PR #126, opened 2025-10-15
feat(ui): add drag-and-drop sorting for provider list

ZyphrZero:
Implement drag-and-drop functionality to allow users to reorder providers with custom sort indices.

Features:
- Install @dnd-kit libraries for drag-and-drop support
- Add sortIndex field to Provider type (frontend & backend)
- Implement SortableProviderItem component with drag handle
- Add update_providers_sort_order Tauri command
- Sync tray menu order with provider list sorting
- Add i18n support for drag-related UI text

Technical details:
- Use @dnd-kit/core and @dnd-kit/sortable for smooth drag interactions
- Disable animations for immediate response after drop
- Update tray menu immediately after reordering
- Sort priority: sortIndex → createdAt → name

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #179, opened 2025-11-07
fix(ui): sync tray menu order after drag-and-drop sorting

ZyphrZero:
The drag-and-drop sorting feature introduced in PR #126 (9eb991d) was updating the provider sort order in the backend and frontend, but failed to update the tray menu to reflect the new order.

Changes:
- Add updateTrayMenu() call after successful sort order update
- Ensures tray menu items are immediately reordered to match the UI

This fixes the issue where dragging providers in the main window would not update their order in the system tray menu.

Fixes: 9eb991d (feat(ui): add drag-and-drop sorting for provider list)


farion1231:
Thanks for your contribution!

PR #202, opened 2025-11-11
feat(gemini): add Gemini provider integration

yovinchen:
- Add gemini_config.rs module for .env file parsing
- Extend AppType enum to support Gemini
- Implement GeminiConfigEditor and GeminiFormFields components
- Add GeminiIcon with standardized 1024x1024 viewBox
- Add Gemini provider presets configuration
- Update i18n translations for Gemini support
- Extend ProviderService and McpService for Gemini

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "cc-switch",
-  "version": "3.6.1",
+  "version": "3.6.2",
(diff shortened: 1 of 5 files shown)


farion1231:
Thanks for your contribution!

PR #210, opened 2025-11-12
fix(gemini): relax validation when adding providers

yovinchen:
Allow users to create Gemini provider configurations without API key and fill it in later. Split validation into two modes:

- validate_gemini_settings: Basic structure check (used when adding)
- validate_gemini_settings_strict: Full validation (used when switching)

This fixes the error 'Gemini config missing required field: GEMINI_API_KEY' when trying to add Gemini providers from presets like PackyCode or Google.

Changes:
- Add validate_gemini_settings_strict for switching validation
- Update write_gemini_live to use strict validation
- Add comprehensive tests for both validation modes


farion1231:

PR #225, opened 2025-11-14
fix(prompt): correct live file backfill priority in enable flow

yovinchen:
Fix backfill logic in prompt enable workflow:
- Prioritize backfilling live file content to currently enabled prompt (prevent data loss)
- Create backup only when no enabled prompt exists and content is new (avoid duplicate backups)
- Implement staged persistence (save after backfill + save after enable)
- Add explicit logging for backfill/backup operations

Also simplify string formatting in prompt_files.rs with inline format strings.


farion1231:
Many thanks for your contribution!

PR #259, opened 2025-11-19
docs: Add installation instructions for Arch Linux

wugeer:
Add installation instructions for Arch Linux via `paru`


farion1231:

PR #276, opened 2025-11-22
fix(codex): use http_headers instead of headers in MCP config

Lutra-Fs:
## Summary
- Fixed Codex MCP configuration to use `http_headers` field name instead of `headers`
- Codex CLI expects `http_headers` per their official documentation
- Maintains backward compatibility by reading from both field names

## Changes
- Updated `import_from_codex()` to read from both `http_headers` (correct) and `headers` (legacy) with priority to `http_headers`
- Updated `json_server_to_toml_table()` to write `http_headers` instead of `headers`
- Updated `core_fields` lists to use `http_headers` for HTTP/SSE type servers

## Context
This follows the same pattern as the recent Gemini fix (commit c5335ce) where provider-specific format requirements are handled during config conversion.

## Related Issues
Fixes the issue where MCP servers with HTTP headers don't work in Codex CLI because CC Switch was using incorrect field name.
https://github.com/openai/codex/blob/main/docs/config.md#mcp_servers

### DIFF
--- a/src-tauri/src/mcp.rs
+++ b/src-tauri/src/mcp.rs
@@ -449,7 +449,7 @@ pub fn import_from_codex(config: &mut MultiAppConfig) -> Result<usize, AppError>
             // 核心字段（需要手动处理的字段）
             let core_fields = match typ {
                 "stdio" => vec!["type", "command", "args", "env", "cwd"],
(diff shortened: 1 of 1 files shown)


farion1231:

PR #299, opened 2025-11-26
fix(skill): use full key for deduplication to allow same-name skills …

yovinchen:
…from different repos


farion1231:

PR #427, opened 2025-12-21
feat(failover): add auto-failover master switch with proxy integration

yovinchen:
- Add persistent auto_failover_enabled setting in database
- Add get/set_auto_failover_enabled commands
- Provider router respects master switch state
- Proxy shutdown automatically disables failover
- Enabling failover auto-starts proxy server
- Optimistic updates for failover queue toggle


farion1231:
感谢您的贡献！

PR #500, opened 2025-12-29
fix: use local timezone and robust DST handling in usage stats

yovinchen:
- Change from UTC to local timezone for daily/hourly trends
- Use SQLite 'localtime' modifier for date grouping
- Replace single().unwrap() with earliest().unwrap_or_else() to handle DST transition edge cases gracefully


farion1231 (approved):
感谢您的贡献！

PR #535, opened 2026-01-06
Fix Gemini base URL alias handling

### LINKED ISSUE #238: 希望 本地代理（无需重启 cc 切换）+协议转换+连接 cli（用 gemini-cli或者千问-cli）
如题

godnight10061:
## Summary
Fixes #238 by treating `GEMINI_BASE_URL` as an alias for `GOOGLE_GEMINI_BASE_URL` end-to-end (proxy takeover/cleanup, provider parsing, deeplinks, and UI), so Gemini CLI base URL works regardless of env var name.

## Changes
- Backend (Tauri/Rust)
  - Proxy takeover writes both keys; cleanup removes both when pointing at local proxy
  - Gemini base-url extraction accepts both keys (adapters, provider service, usage helpers, auth detection)
  - Universal provider export includes both keys
  - Deeplink import accepts either key
- Frontend (React/TS)
  - Provider editor defaults/presets keep both keys in sync
  - Display / endpoint collection falls back to either key
- Tests
  - MSW: add handler for `check_env_conflicts`
  - Integration: increase App integration test timeout to reduce flakiness
  - Rust: add unit test ensuring `GEMINI_BASE_URL` is accepted

## Why
Some parts of the app/docs use `GEMINI_BASE_URL` while others used `GOOGLE_GEMINI_BASE_URL`, causing inconsistent behavior (especially around proxy takeover and config import). Supporting both removes that footgun.

## Tests
- `pnpm test:unit`
- `pnpm typecheck`
- `cd src-tauri && cargo test --no-run`

## Notes
- Backwards compatible.
- If both keys are present, `GOOGLE_GEMINI_BASE_URL` takes precedence (we write both to the same value from the UI and takeover flow).


godnight10061:
Thanks for the feedback!

PR #547, opened 2026-01-08
docs: add download options for each system version in 3.9.0 release note

su-fen:
This PR updates the release notes for version 3.9.0 by adding the download file names corresponding to each system version. These file names were previously missing from the release notes.

Changes:
- Added the file names for each system version.
- Clarified the system compatibility by listing the correct download file names for each platform.


farion1231 (approved):

PR #580, opened 2026-01-10
fix(gemini): convert timeout params to Gemini CLI format

cxyfer:
Fixes issue reported at https://linux.do/t/topic/1214498/45

- Fix MCP timeout parameter compatibility for Gemini CLI
  - Claude Code/Codex uses `startup_timeout_sec` and `tool_timeout_sec`
  - Gemini CLI only supports single `timeout` param in milliseconds
  - These params were passed directly, causing "Unrecognized key" errors

- Add timeout conversion logic in `set_mcp_servers_map`
  - Collect all timeout values (sec and ms variants)
  - Take max value and convert to milliseconds
  - Remove original fields and insert Gemini-compatible `timeout`

- Use default timeout when not configured
  - Default: 60000ms (max of startup=10s, tool=60s)

### DIFF
--- a/src-tauri/src/gemini_mcp.rs
+++ b/src-tauri/src/gemini_mcp.rs
@@ -134,6 +134,33 @@ pub fn set_mcp_servers_map(
         obj.remove("homepage");
         obj.remove("docs");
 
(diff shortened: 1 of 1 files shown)


farion1231 (approved):

Issue #453, opened 2025-12-24
[claude code skills] Windows requires 'cmd /c' wrapper to execute npx


word911:
[version] 
CC-Switch-v3.8.2-Windows-Portable.zip

[env]
Claude Code 版本 v2.0.73
windows 版本 Windows 10 专业工作站版 22H2

[Question] 
windows 环境下，Claude code 中MCP配置警告。具体如下：

在claude code下执行命令：
/doctor

返回显示：
MCP Config Diagnostics

 For help configuring MCP servers, see: https://code.claude.com/docs/en/mcp

 [Contains warnings] User config (available in all your projects)
 Location: C:\Users\nobody\.claude.json
  └ [Warning] [context7] mcpServers.context7: Windows requires 'cmd /c' wrapper to execute npx
  └ [Warning] [memory] mcpServers.memory: Windows requires 'cmd /c' wrapper to execute npx   
  └ [Warning] [sequential-thinking] mcpServers.sequential-thinking: Windows requires 'cmd /c' wrapper to execute npx

烦请在下个版本中更新。谢谢！

farion1231:
感谢反馈，我研究一下！

PR #592, opened 2026-01-11
fix(mcp): skip cmd /c wrapper for WSL target paths

cxyfer:
Fixes unintended behavior introduced in v3.9.0/v3.9.1 from #453 (also reported in #557)

## Background

- #453 reported that Claude Code `/doctor` shows warnings about missing `cmd /c` wrapper for npx/npm commands on Windows
- v3.9.1 added automatic `cmd /c` wrapping for MCP export to fix this warning
- However, when the Claude config directory is set to a WSL network path (e.g., `\wsl$\Ubuntu\home\user\.claude`), the `cmd /c` wrapper breaks MCP server connections because WSL runs Linux

## Problem

```bash
$ claude mcp list
context7-wrap: cmd /c npx -y @upstash/context7-mcp - ✗ Failed to connect
context7-no-wrap: npx -y @upstash/context7-mcp - ✓ Connected
```

## Solution

- Add `is_wsl_path()` to detect WSL network paths (`\wsl$\` and `\wsl.localhost\`)
- Skip `wrap_command_for_windows()` when target config path is a WSL path
- Add debug logging when WSL path is detected

## Changes

- Add `is_wsl_path()` function using `std::path::Component::Prefix` for robust UNC path detection
- Modify `set_mcp_servers_map()` to check target path before wrapping
- Add comprehensive tests for various WSL distributions (Ubuntu, Debian, Arch, Fedora, etc.)

## Limitations

- Only detects direct UNC paths; mapped drive letters (e.g., `Z:` → `\wsl$\...`) cannot be detected

### DIFF
--- a/src-tauri/src/claude_mcp.rs
+++ b/src-tauri/src/claude_mcp.rs
@@ -67,19 +67,21 @@ fn wrap_command_for_windows(_obj: &mut Map<String, Value>) {
 
 /// 检测路径是否为 WSL 网络路径（如 \\wsl$\Ubuntu\... 或 \\wsl.localhost\Ubuntu\...）
 /// WSL 环境运行的是 Linux，不需要 cmd /c 包装
+/// 注意：仅检测直接 UNC 路径，映射磁盘符（如 Z: -> \\wsl$\...）无法检测
 #[cfg(windows)]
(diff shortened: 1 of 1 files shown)


farion1231 (approved):

PR #610, opened 2026-01-12
fix(provider): persist endpoint auto-select state

yovinchen:
- Add endpointAutoSelect field to ProviderMeta for persistence
- Lift autoSelect state from EndpointSpeedTest to ProviderForm
- Save auto-select preference when provider is saved
- Restore preference when editing existing provider


YangYongAn:
我有一个 pr，请核查

https://github.com/farion1231/cc-switch/pull/609

PR #671, opened 2026-01-17
fix(docs): change 'ArchLinux 用户' to 'ArchLinux Users' in README

Shuimo03:
change 'ArchLinux 用户' to 'ArchLinux Users' in README"


farion1231 (approved):

PR #868, opened 2026-02-01
fix(skills): handle Windows path separator in installed status matching

stmoonar:
Use regex to split directory path by both / and \ to correctly extract the install name on Windows, fixing the issue where installed skills were not showing as installed in the discovery page filter.


farion1231 (approved):

PR #1005, opened 2026-02-11
feat(proxy): fix thinking rectifiers and resolve clippy warnings

yovinchen:
- add thinking budget rectifier flow with single retry on anthropic budget errors

- align thinking signature rectification behavior with adaptive-safe handling

- expose requestThinkingBudget in settings/ui/i18n and default rectifier config to disabled

- fix clippy warnings in model_mapper format args and RectifierConfig default derive


yovinchen:
fix  #991

PR #1054, opened 2026-02-16
fix(proxy): NVIDIA Harmony/OpenAI Chat streaming and routing

jnorthrup:
## Summary
- Add `reasoning_content` field support in streaming parser for NVIDIA responses
- Add `#[serde(flatten)] HashMap` to handle NVIDIA's extra JSON fields (logprobs, token_ids, etc.)
- Fix routing: `apiFormat='openai_chat'` forces `/v1/chat/completions`, never `/v1/responses`
- Add `harmonySupport` provider meta flag to control Harmony API usage
- Fix non-streaming response: use `reasoning` field when `content` is null/empty
- Add debug logging for endpoint routing decisions

## Problem
NVIDIA API returns model output in the `reasoning` field instead of `content` field, causing empty responses. Additionally, NVIDIA's streaming responses include extra fields (`logprobs`, `token_ids`, `prompt_token_ids`) that caused JSON deserialization failures.

## Solution
1. Extended `Delta` struct to accept `reasoning_content` and unknown fields via `#[serde(flatten)]`
2. Updated response transform to check both `content` and `reasoning` fields
3. Fixed endpoint routing to respect `apiFormat` setting and not auto-redirect to Harmony API

## Test plan
- Verified NVIDIA streaming via `curl` to proxy
- Verified Claude Code client works with NVIDIA provider
- Model mapping correctly routes `claude-haiku-4-5` → `openai/gpt-oss-120b`


jnorthrup:
what is your vision for a model proxy with NO MODEL PICKER ? this is ancient goodwill on my part to reuse someone else's precedent model picker that sat an equally long time before i saw it

farion1231:
> what is your vision for a model proxy with NO MODEL PICKER ? this is ancient goodwill on my part to reuse someone else's precedent model picker that sat an equally long time before i saw it

The concern here is not the feature itself, but the size of the PR.
This change set is far too large to review safely. Please split it into smaller, focused PRs.

jnorthrup:
@farion1231 i wouldn't touch an ancient PR like this, there was a window of opportunity where presumably you reviewed it and passed on it, and the other PR from the guy before.  with a month+ since having to write the features i like,  there are merges adding conflicts out of sync while this main branch in my tree was growing past the reasonable share point.

I'm not motivated to submit significant PR's like a model picker the entire codebase has been built in a brittle way that i refactored completely in my own tree.  i merged your changes last week and it broke my tree, sorry i can't give this much time

PR #1061, opened 2026-02-16
Add API `models/`  enumeration  

jnorthrup:
image

## Summary
- add OpenAI-compatible model API enumeration command with caching support for provider model lists
- keep model inputs fully editable while showing dropdown suggestions/refresh when a model list is available
- preserve model names as raw strings (including `/`) instead of inferring provider from model text
- add DB schema compatibility by supporting migration from version 5 to 6

## Validation
- `cargo test --manifest-path src-tauri/Cargo.toml schema_migration_upgrades_v5_to_v6_compat -- --nocapture`


farion1231:

PR #1222, opened 2026-03-02
fix: align outline button text tone with usage refresh control

Sube-py:


farion1231 (commented):
This visual change affects a fairly large area. Could you provide some screenshots so we can better understand the impact?

Sube-py:
> This visual change affects a fairly large area. Could you provide some screenshots so we can better understand the impact?

My bad, In dark mode:

image

When the button hovers:
image


After repair:
image

PR #1334, opened 2026-03-06
Fix unclickable Linux window controls on GNOME/Wayland

iron6909:
## Summary
- add a Linux-specific Tauri config that uses the system title bar
- disable custom Tauri drag regions on Linux headers and full-screen panels
- note the fix in the changelog

## Problem
On Ubuntu 24.04 / GNOME 46 / Wayland, the main window's maximize and close buttons can appear but not receive clicks.

## Root cause
Linux was still inheriting the shared window setup while the frontend also marked the custom header as a Tauri drag region. In GNOME/Wayland this can leave the native controls underneath an overlapping custom title/header area.

## Fix
- add `src-tauri/tauri.linux.conf.json` with `titleBarStyle: "Visible"`
- gate custom drag-region attributes behind `!isLinux()` in:
  - `src/App.tsx`
  - `src/components/common/FullScreenPanel.tsx`

## Validation
- `corepack pnpm test:unit`
- `corepack pnpm build:renderer`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- **Linux Window Controls**: Use the system title bar on Linux and disable custom drag regions there so GNOME/Wayland window buttons stay clickable
+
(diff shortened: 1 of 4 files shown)


farion1231:
@codex

PR #1339, opened 2026-03-07
Preserve Codex MCP config across provider writes

smile7up:
## Summary
- preserve Codex mcp_servers and projects when writing provider snapshots back to ~/.codex/config.toml
- preserve the same tables when proxy hot-switch updates the Codex live backup
- add regression coverage for config merging and proxy hot-switch backup preservation

## Root cause
cc-switch was writing Codex provider config as a full replacement config.toml, and proxy hot-switch backup updates reused provider settings directly. That dropped user-managed sections like mcp_servers, so Codex MCP entries disappeared after provider switches or proxy restore flows.

## Testing
- cargo test merge_codex_live_config_preserves_mcp_servers_and_projects
- cargo test switch_proxy_target_preserves_codex_mcp_servers_in_live_backup
- cargo test switch_proxy_target_updates_live_backup_when_taken_over

### DIFF
--- a/src-tauri/src/codex_config.rs
+++ b/src-tauri/src/codex_config.rs
@@ -138,12 +138,16 @@ pub fn merge_codex_live_config(
         return Ok(provider_doc.to_string());
     }
 
-    let existing_doc = existing_live_text.parse::<DocumentMut>().map_err(|e| {
-        AppError::Config(format!(
-            "解析现有 Codex config.toml 失败 ({}): {e}",
-            config_path.display()
-        ))
-    })?;
+    let existing_doc = match existing_live_text.parse::<DocumentMut>() {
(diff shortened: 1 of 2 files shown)


farion1231:
@codex

smile7up:
Addressed.\n\nI changed Codex live-table preservation to be best-effort: if the existing live config.toml (or the backup config used during proxy hot-switch) is malformed, cc-switch now logs a warning and proceeds with the new provider config instead of failing the write/switch path.\n\nAdded regression coverage for both cases:\n- merge_codex_live_config_falls_back_when_existing_live_is_malformed\n- switch_proxy_target_ignores_malformed_codex_live_backup_when_merging\n\nRe-ran the earlier preservation tests as well.

PR #1350, opened 2026-03-08
fix: support JSON-with-comments for OpenCode/Claude config and keep provider save on live sync failure

crmmc:
## Background

  When `opencode.json` or Claude live config contains comments, strict JSON parsing fails.
  This could block provider-related save flows (including non-OpenCode providers) because live sync paths also touch OpenCode/Claude config parsing.

  ## What changed

  1. Enabled comment-tolerant JSON parsing for OpenCode and Claude live config reads.
  2. Changed provider add/update flow to preserve CC Switch internal save (DB) even if live sync fails.
     - Live sync / MCP sync failures are now returned as warnings instead of hard-failing the whole save.
  3. Added frontend warning toasts for these non-fatal sync failures, so users still see clear error feedback.
  4. Updated related API typing and test mocks for the new warning payload.

  ## Why this fixes the issue

  - Commented config files are now parsed correctly, avoiding unnecessary parse failures.
  - Save behavior is split into:
    - **internal config persistence** (must succeed)
    - **live sync** (best-effort with explicit warnings)

  So users no longer lose save capability in CC Switch due to external live config parse/sync errors.

  ## Validation

  - `pnpm typecheck`
  - `pnpm format:check`
  - `pnpm test:unit` (24 files, 167 tests)

  All passed.

  Thanks for the review.

### DIFF
--- a/src-tauri/src/commands/provider.rs
+++ b/src-tauri/src/commands/provider.rs
@@ -5,7 +5,8 @@ use crate::app_config::AppType;
 use crate::error::AppError;
 use crate::provider::Provider;
 use crate::services::{
-    EndpointLatency, ProviderService, ProviderSortUpdate, SpeedtestService, SwitchResult,
+    EndpointLatency, ProviderMutationResult, ProviderService, ProviderSortUpdate, SpeedtestService,
+    SwitchResult,
 };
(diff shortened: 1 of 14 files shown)


farion1231:
@codex

PR #1506, opened 2026-03-15
fix: restore Claude session resume in Ghostty

canyonsehun:
## Summary
- replace Ghostty session launch argument construction with a dedicated helper
- stop passing resume commands through Ghostty command execution paths that break on macOS
- open Ghostty with the target working directory and feed the resume command through `--input=raw:` instead
- add unit coverage for Ghostty argument generation and escaping

## Root cause
Ghostty resume launches were building command execution arguments incorrectly on macOS, which led to failures like:
- `login: claude: No such file or directory`
- Ghostty prompts to allow executing `/bin/zsh`, followed by a blank shell launch

The previous implementation also dropped `cwd` when building the Ghostty launch command.

## Testing
- `cargo test terminal::tests`
- manually verified Ghostty launches with Claude resume commands on macOS using the rebuilt app

### DIFF
--- a/src-tauri/src/session_manager/terminal/mod.rs
+++ b/src-tauri/src/session_manager/terminal/mod.rs
@@ -77,39 +77,10 @@ end tell"#
 }
 
 fn launch_ghostty(command: &str, cwd: Option<&str>) -> Result<(), String> {
-    // Ghostty usage: open -na Ghostty --args +work-dir=... -e shell -c command
-
-    // Using `open` to launch.
-    let mut args = vec!["-na", "Ghostty", "--args"];
-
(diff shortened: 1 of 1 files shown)


farion1231:
此问题已在 v3.12.3 中修复。修复内容：FIX: Ghostty Session Restore（#1506），修复了 macOS 上 Ghostty 恢复 Claude 会话时构建命令参数错误的问题，之前会出现「login: claude: No such file or directory」或空 shell 启动等问题。

请更新到最新版本 v3.12.3 验证。如问题仍然存在，欢迎重新打开此 issue。

Issue #1544, opened 2026-03-17
中转代理bug


rendanPatto:
[2026-03-18][00:48:14][INFO][cc_switch_lib::proxy::forwarder] [Claude] >>> 请求 URL: https://api-ai.gitcode.com/v1/chat/completions (model=zai-org/GLM-5)
[2026-03-18][00:48:26][INFO][cc_switch_lib::proxy::forwarder] [Claude] >>> 请求 URL: https://api-ai.gitcode.com/v1/chat/completions (model=zai-org/GLM-5)
[2026-03-18][00:48:36][ERROR][cc_switch_lib::proxy::handlers] [Claude] 解析上游响应失败: expected value at line 1 column 1, body: data:{"id":"chatcmpl-b6752d331a81ee2d","object":"chat.completion","created":1773766110,"model":"glm-5","choices":[{"index":0,"message":{"role":"assistant","content":"Hi! 😊 How can I help you with your project today?","refusal":null,"annotations":null,"audio":null,"function_call":null,"tool_calls":[],"reasoning":"The user is just saying \"hi\" variations repeatedly. This seems like they're just greeting or testing. I should respond simply and ask if they need help with something."},"logprobs":null,"finish_reason":"stop","stop_reason":154827,"token_ids":null}],"service_tier":null,"system_fingerprint":null,"usage":{"prompt_tokens":20008,"total_tokens":20058,"completion_tokens":50,"prompt_tokens_details":null},"prompt_logprobs":null,"prompt_token_ids":null,"kv_transfer_params":null}

yovinchen:
https://docs.gitcode.com/docs/apis/post-api-v-5-chat-completions/ 官方文档

其中 chat 接口具体为 https://api.gitcode.com/api/v5/chat/completions

关于完整URL 的支持能力将在下一个版本中更新

测试 Qwen/Qwen3-Coder-480B-A35B-Instruct 模型可用 未找到免费 glm-5模型测试

<img width="767" height="563" alt="Image" src="https://github.com/user-attachments/assets/c7311c10-5911-4582-8b6e-4fcb541780dd" />

JasonSun9527:
same issue

yovinchen:
请发一下您的配置信息

> same issue

PR #1561, opened 2026-03-19
feat(proxy): add full URL mode and refactor endpoint rewriting

yovinchen:
- Added a Claude full URL mode, allowing vendor addresses to be used directly as complete upstream interfaces.

- Unified the Claude endpoint/path/query rewrite logic into the proxy forwarder.

- Removed beta queries during Anthropic -> OpenAI/Copilot rewrites to avoid incompatible parameter pass-through.

- Streaming health checks now support both full URL and Responses API.

- Added a full URL toggle on the frontend, preventing switching of providers dependent on the proxy when the proxy is not running.


yovinchen:
fix #1544

PR #1632, opened 2026-03-23
fix(tray): left-click tray icon opens main window directly on Windows

MarkSunDev:
## Problem

On Windows, left-clicking the system tray icon shows the context menu — same behavior as right-click. To open the main window, users must click the tray icon and then click "Open main window" from the menu, which is an unnecessary extra step.

## Solution

- Set `show_menu_on_left_click(false)` to disable menu popup on left-click
- Handle `MouseButton::Left` + `MouseButtonState::Up` in `on_tray_icon_event` to show and focus the main window directly
- Right-click behavior (context menu) remains unchanged

## Testing

- Left-click tray icon → main window shows and receives focus
- Right-click tray icon → context menu opens as before
- "Open main window" menu item still works correctly
- macOS and Linux behavior unaffected (`set_skip_taskbar` is `#[cfg(target_os = "windows")]`)

### DIFF
--- a/src-tauri/src/lib.rs
+++ b/src-tauri/src/lib.rs
@@ -632,16 +632,31 @@ pub fn run() {
 
             // 构建托盘
             let mut tray_builder = TrayIconBuilder::with_id("main")
-                .on_tray_icon_event(|_tray, event| match event {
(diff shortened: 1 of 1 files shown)


farion1231:
@codex

PR #1652, opened 2026-03-24
Fix malformed Codex common config detection

hikaruczl:
## Summary
- normalize malformed Codex TOML before common-config subset checks and removal
- lift snippet root scalar keys back to the TOML root when they were accidentally parsed under the trailing table
- add regression tests for malformed Codex common config appended after [notice.model_migrations]

## Testing
- pnpm exec vitest run tests/utils/providerConfigUtils.codex.test.ts
- pnpm typecheck


farion1231:
@codex review

PR #1683, opened 2026-03-26
Feat: support installing skills from Git URLs and support GitLab repositories in repo manager

KT-lcz:
Allow skills to be installed directly from HTTPS Git repositories instead of requiring ZIP archives, while reusing the shared install pipeline and adding selector, symlink, and rollback safeguards.

Allow GitLab skill repositories to be added and discovered from the repo manager while keeping GitHub on the existing ZIP flow. Route GitLab discovery and installs through the hardened git clone staging pipeline with server-side repo validation.


farion1231:
@codex review

PR #1720, opened 2026-03-27
fix(proxy): strip hop-by-hop headers from upstream response

xyz3282836:
When forwarding upstream responses to downstream clients, hop-by-hop headers (e.g. connection, transfer-encoding, keep-alive) were passed through as-is. This caused axum/hyper to reject the response with "user sent unexpected header" and drop the connection, resulting in "stream disconnected" / "Empty reply from server" on the client side.

Fix: strip all RFC 2616 §13.5.1 hop-by-hop headers (including any extra headers listed in the Connection header value) before building the downstream response, in both streaming and non-streaming paths.

### DIFF
--- a/src-tauri/src/proxy/response_processor.rs
+++ b/src-tauri/src/proxy/response_processor.rs
@@ -13,7 +13,7 @@ use super::{
 use axum::response::{IntoResponse, Response};
 use bytes::Bytes;
 use futures::stream::{Stream, StreamExt};
-use reqwest::header::HeaderMap;
+use reqwest::header::{HeaderMap, HeaderName};
 use serde_json::Value;
 use std::{
     sync::{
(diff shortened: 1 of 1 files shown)


farion1231 (commented):
Thanks for your contribution. I think the direction of this PR is good, and the underlying issue is worth fixing, because forwarding hop-by-hop response headers downstream can indeed trigger axum/hyper errors like `user sent unexpected header`, which may surface as `stream disconnected` or `Empty reply from server`.

That said, could you please take a look at the following issues before merging?

1. This PR currently fixes the generic passthrough path in `process_response`, but it does not cover the Claude/OpenRouter transform path in `handle_claude_transform`. In that path, the non-streaming branch still forwards most upstream response headers, and the streaming branch still explicitly sets `Connection: keep-alive`, so the same class of issue may still happen there.

2. The new stripped header set still seems incomplete. In particular, `proxy-connection` is not removed, and hyper also treats it as an illegal connection-specific header in HTTP/2. That means responses coming through certain proxies or gateways could still reproduce the same downstream failure.

I think this PR is a good start, but at the moment it looks more like a partial fix than a complete one. It would be great to cover the transform path as well and add some tests for streaming, non-streaming, and transformed responses so the behavior is fully locked down. I'd like to hear your opinion.

farion1231:
@codex review

PR #1721, opened 2026-03-27
feat: persist session titles and restore codex session speed

Leopard-S:
## Summary
- persist custom session titles in the session manager and keep the original title available for display/search
- restore Codex session responsiveness by skipping heavy tool output rows when loading message history
- add backend/frontend tests for title override persistence and session rename UX

## Details
- stores session title overrides in SQLite via a new `session_overrides` table
- applies overrides across Claude, Codex, Gemini, OpenClaw, and OpenCode session metadata
- shows renamed session titles in the UI while still displaying the original title on the right side
- keeps search matching against the original title after rename
- avoids long-session slowdown in Codex by only loading normal `message` items into the message list

## Verification
- `pnpm test:unit -- tests/components/SessionManagerPage.test.tsx`
- `pnpm typecheck`

## Notes
- `cargo test` still reports existing unrelated failures in `openclaw_config` / `session_manager::providers::openclaw`
- at least one of those failures reproduces on clean `upstream/main`, so this PR does not introduce that baseline red state

### DIFF
--- a/src-tauri/src/session_manager/mod.rs
+++ b/src-tauri/src/session_manager/mod.rs
@@ -101,6 +101,30 @@ pub fn rename_session(
         .map_err(|e| format!("Failed to save session title override: {e}"))
 }
 
+fn original_session_title(session: &SessionMeta) -> String {
+    session
+        .title
+        .clone()
+        .or_else(|| {
+            session.project_dir.as_deref().and_then(|project_dir| {
+                let trimmed = project_dir.trim();
+                if trimmed.is_empty() {
+                    return None;
+                }
+                let normalized = trimmed.trim_end_matches(['/', '\\']);
+                if normalized.is_empty() {
+                    return Some(trimmed.to_string());
+                }
+                normalized
+                    .split(['/', '\\'])
+                    .next_back()
+                    .filter(|segment| !segment.is_empty())
+                    .map(|segment| segment.to_string())
+            })
+        })
+        .unwrap_or_else(|| session.session_id.chars().take(8).collect())
+}
+
 fn apply_title_overrides(
     sessions: &mut [SessionMeta],
     overrides: Vec<crate::database::SessionTitleOverride>,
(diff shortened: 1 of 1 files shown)


Leopard-S:
Addressed in 7c41fd27.

Changes made:
- Added a fallback path when loading session_overrides fails, so the session list still loads from disk.
- Removed the permissive v7 -> v6 normalization branch; newer schema versions are rejected strictly again.
- Cleared persisted session title overrides after successful session deletion, including the sqlite: OpenCode path.

Verification:
- cargo test --manifest-path src-tauri/Cargo.toml schema_migration_rejects_future_version -- --nocapture
- cargo test --manifest-path src-tauri/Cargo.toml schema_migration_rejects_v7_even_if_session_overrides_exists -- --nocapture
- cargo test --manifest-path src-tauri/Cargo.toml load_session_title_overrides_falls_back_on_query_error -- --nocapture
- cargo test --manifest-path src-tauri/Cargo.toml delete_session_clears_title_override_metadata -- --nocapture
- pnpm typecheck

Please take another look when you have time.

Leopard-S:
Addressed in 52cb00aa.

Changes made:
- Fixed the rename cache restore path so clearing a custom session title now restores the real fallback name (`originalTitle`, or the derived `projectDir` basename / session id fallback when there was no explicit title).
- Updated the test MSW state to mirror the same restore behavior, so the fallback-name path is covered in the frontend integration test.
- Adjusted Codex session message loading to keep the `function_call` tool name as a lightweight `tool` message while still skipping `function_call_output`.

Verification:
- `pnpm test:unit -- tests/components/SessionManagerPage.test.tsx`
- `pnpm typecheck`
- `cargo test --manifest-path src-tauri/Cargo.toml load_messages_keeps_function_call_name_but_skips_output -- --nocapture`

Leopard-S:
Addressed in 80d620f5.

Changes made:
- Preserved the original display title when applying persisted session title overrides, even when the original name comes from the fallback display path (`title -> projectDir basename -> session id prefix`).
- Added a regression test for fallback-derived original titles after reload.
- Added a small edge-case test to keep root project directories consistent with the existing frontend fallback behavior.

Verification:
- `cargo test --manifest-path src-tauri/Cargo.toml applies_custom_title_and_preserves_fallback_display_title -- --nocapture`
- `cargo test --manifest-path src-tauri/Cargo.toml preserves_root_project_dir_as_original_title -- --nocapture`
- `pnpm build` (release exe + Windows bundles were produced successfully; the command still exits non-zero at the updater signing step because `TAURI_SIGNING_PRIVATE_KEY` is not configured in this environment)

PR #1725, opened 2026-03-27
fix: show fallback message when import config error is empty

### LINKED ISSUE #1715: 没有相应模型配置时，点击【导入当前配置】弹出的错误提示文案为空
Image

bluzername:
## Problem

When clicking "Import Current Config" and the import fails (for example no model configuration available), the error notification popup shows with empty text. User sees the toast but has no idea what went wrong.

Screenshot from issue shows the empty error message clearly.

## Root Cause

In `ProviderList.tsx` line 243, the `onError` handler does `toast.error(error.message)`. But sometimes the Tauri invoke returns an error where `error.message` is empty string. The toast then renders with no visible text.

## Fix

Added fallback using `||` operator:

```typescript
toast.error(error.message || t("provider.importCurrentFailed"));
```

When `error.message` is empty, it shows the localized "Failed to import current configuration" text instead.

Also added the `importCurrentFailed` i18n key to all 3 locale files:
- English: "Failed to import current configuration"
- Chinese: "导入当前配置失败"
- Japanese: "現在の設定のインポートに失敗しました"

## Files changed

- `src/components/providers/ProviderList.tsx` - added fallback in onError handler
- `src/i18n/locales/en.json` - added importCurrentFailed key
- `src/i18n/locales/zh.json` - added importCurrentFailed key  
- `src/i18n/locales/ja.json` - added importCurrentFailed key

Closes #1715

### DIFF
--- a/src/components/providers/ProviderList.tsx
+++ b/src/components/providers/ProviderList.tsx
@@ -240,7 +240,7 @@ export function ProviderList({
       }
     },
     onError: (error: Error) => {
-      toast.error(error.message);
+      toast.error(error.message || t("provider.importCurrentFailed"));
     },
   });
 
--- a/src/i18n/locales/en.json
+++ b/src/i18n/locales/en.json
@@ -148,7 +148,8 @@
         "oauthHint": "Google official uses OAuth personal authentication, no need to fill in API Key. The browser will automatically open for login on first use.",
         "apiKeyPlaceholder": "Enter Gemini API Key"
       }
-    }
+    },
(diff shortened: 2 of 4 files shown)


farion1231 (commented):
Thank you for your contribution! Could you please check whether it would be better to use the following approach?

ProviderList.tsx:243 — the project’s standard error extraction function should be used

- toast.error(error.message || t("provider.importCurrentFailed"));
+ toast.error(extractErrorMessage(error) || t("provider.importCurrentFailed"));

farion1231:
@codex review

PR #1729, opened 2026-03-28
chore: switch package manager from pnpm to bun

lispking:
https://bun.com/
image


farion1231 (commented):
why?

lispking:
The main motivation is performance and simpler tooling. Bun is npm-compatible, but installs dependencies much faster in practice, and it can also run scripts directly, so the overall dev setup becomes a bit more streamlined. 

My thinking was: if the project does not rely on pnpm-specific behavior, using Bun could improve contributor onboarding and reduce install time locally and in CI. 

That said, I understand package-manager choice is also about consistency and maintainability, so if pnpm is the preferred convention here, I’m okay with keeping it.

farion1231:
> The main motivation is performance and simpler tooling. Bun is npm-compatible, but installs dependencies much faster in practice, and it can also run scripts directly, so the overall dev setup becomes a bit more streamlined.
> 
> My thinking was: if the project does not rely on pnpm-specific behavior, using Bun could improve contributor onboarding and reduce install time locally and in CI.
> 
> That said, I understand package-manager choice is also about consistency and maintainability, so if pnpm is the preferred convention here, I’m okay with keeping it.

Thanks for your contribution! This is a significant change, so I’d like to take some time to evaluate it carefully before making a decision.

PR #1747, opened 2026-03-29
fix(provider): show persistent config highlight for additive-mode providers

yovinchen:
## Summary
- Provider cards in additive-mode apps (opencode, openclaw) now show a persistent blue highlight when the provider is in the live config, instead of only on hover
- This makes it easy to scan which configurations are enabled at a glance without hovering each card

Closes https://github.com/farion1231/cc-switch/issues/1692

## Test plan
- [x] Verify opencode/openclaw provider cards show blue border when in config (without hover)
- [x] Verify non-additive apps (Claude/Codex/Gemini) retain existing behavior
- [x] Verify OMO cards are unaffected
- [x] Verify proxy-takeover green highlight still works correctly


yovinchen:
@codex review

PR #1769, opened 2026-03-30
fix: align newapi usage query with token endpoint

### LINKED ISSUE #1768: New API usage query default points to the wrong endpoint
- [x] I searched existing issues and this is not a duplicate

## Environment

| Field           | Value                    |
| --------------- | ------------------------ |
| App version     | 3.12.3                   |
| Harness         | OpenCode                 |
| Model           | GPT-5.4(newapi) / openai-gpt-5.4 |
| OS + shell      | macOS + zsh              |

## What happened?

CC Switch's built-in `New API` usage-query default was guiding users toward `/api/user/self` and exposing `Access Token` plus `User ID` as the main input model, even though the more appropriate default for this usage-query path is a token-usage query based on `Base URL + API Key`.

As a result, users opening the built-in `New API` usage-query flow were given a default configuration that was not the best fit for the expected setup.

## Steps to reproduce

1. Open a provider's usage-query modal in CC Switch.
2. Select the built-in `New API` template.
3. Observe the default endpoint, credential fields, and help example.

## Expected behavior

The built-in `New API` default should guide users toward the token-usage flow:

- `GET {{baseUrl}}/api/usage/token`
- `Authorization: Bearer {{apiKey}}`
- `User-Agent: cc-switch/1.0`
- `Base URL + API Key` as the default input fields

## Actual behavior

Data fetch does not succeed.

## Evidence

- targeted inspection showed the built-in template text still referenced the older flow before the fix
- the New API token-usage path matched the intended "Base URL + API Key" usage pattern better than the old default
- after the fix, targeted regression tests were added for template text and New API field behavior

LanternCX:
## What problem are you trying to solve?

CC Switch's built-in `New API` usage-query flow was using an unsuitable default. It pointed users to `/api/user/self` and exposed `Access Token` plus `User ID` as the main input model, even though the more appropriate default for this flow is a token-usage query based on `Base URL + API Key`.

This made the built-in path harder to use than it needed to be and created a mismatch between the expected setup and the default guidance shown in the product.

## What does this PR change?

This PR updates the built-in `New API` usage-query default to use `GET {{baseUrl}}/api/usage/token` with `Bearer {{apiKey}}`, aligns the New API modal fields and help example with that flow, refreshes the Chinese usage-query guide, and adds regression tests for the template and field behavior.

## Is this change appropriate for the core library?

Yes.

This is a correction to shared built-in product behavior, not a project-specific customization. Anyone using the existing `New API` usage-query flow in CC Switch can benefit from a more appropriate default.

## What alternatives did you consider?

1. Keep the old `/api/user/self` default and add a second New API template.
   This would preserve the old flow, but it would also keep the less suitable default in place and force users to pick between overlapping options.

2. Partially update the old template while keeping the old credential model.
   That would mix the new endpoint with the old input expectations and make the built-in flow even more confusing.

3. Make the conversion value configurable in the same PR.
   That may be reasonable later, but it would widen this change from "fix the default flow" into a broader product change. I kept this PR focused.

## Does this PR contain multiple unrelated changes?

No.

The code, UI, tests, and documentation changes all serve the same single goal: correcting the built-in `New API` usage-query default flow and keeping all user-facing entry points consistent with it.

## Existing PRs

- [x] I have reviewed open and closed PRs for duplicates or prior art
- Related PRs: none found

## Environment tested

| Harness  | Model           | Model version/ID |
| -------- | --------------- | ---------------- |
| OpenCode | GPT-5.4(newapi) | openai/gpt-5.4   |

## Evaluation

- What was the initial prompt used to start the session that led to this change?
  - fix issue #1768
- How many verification runs did you execute after making the change?
  - Two targeted test runs after fixing the local environment, plus a production build attempt and a human review.
- How did outcomes change compared to before the change?
  - Before: the built-in `New API` flow still pointed to the old endpoint and old credential model.
  - After: the built-in template, help example, modal fields, and Chinese guide all point to the token-usage endpoint with API Key auth, and the targeted regression tests pass.

## Verification

- [x] Targeted regression tests were added for the New API template and field behavior
- [x] This change was tested beyond the happy path

Coverage included:

- checking that the template no longer contains `/api/user/self`
- checking that the template no longer contains `New-Api-User`, `{{accessToken}}`, or `{{userId}}`
- checking that `total_available = 0` remains a valid result via presence-based validation
- checking that legacy saved `accessToken` / `userId` configurations still open in New API mode

## Human review

- [x] A human has reviewed the complete proposed diff before submission

Fix #1768

### DIFF
--- a/docs/user-manual/zh/2-providers/2.5-usage-query.md
+++ b/docs/user-manual/zh/2-providers/2.5-usage-query.md
@@ -5,6 +5,7 @@
 用量查询功能允许你配置自定义脚本，实时查询供应商的剩余额度、已用量等信息。
 
 **使用场景**：
+
 - 查看 API 账户剩余余额
 - 监控套餐使用情况
 - 多套餐额度汇总显示
@@ -37,18 +38,18 @@ CC Switch 提供三种预设模板：
     url: "{{baseUrl}}/user/balance",
     method: "GET",
     headers: {
-      "Authorization": "Bearer {{apiKey}}",
-      "User-Agent": "cc-switch/1.0"
-    }
+      Authorization: "Bearer {{apiKey}}",
+      "User-Agent": "cc-switch/1.0",
+    },
   },
-  extractor: function(response) {
+  extractor: function (response) {
     return {
       isValid: response.is_active || true,
       remaining: response.balance,
-      unit: "USD"
+      unit: "USD",
     };
-  }
-})
+  },
+});
 ```
 
 **配置参数**：
@@ -64,38 +65,38 @@ CC Switch 提供三种预设模板：
 ```javascript
 ({
   request: {
-    url: "{{baseUrl}}/api/user/self",
+    url: "{{baseUrl}}/api/usage/token",
     method: "GET",
     headers: {
-      "Content-Type": "application/json",
-      "Authorization": "Bearer {{accessToken}}",
-      "New-Api-User": "{{userId}}"
+      Authorization: "Bearer {{apiKey}}",
+      "User-Agent": "cc-switch/1.0",
     },
   },
   extractor: function (response) {
-    if (response.success && response.data) {
-      return {
-        planName: response.data.group || "默认套餐",
-        remaining: response.data.quota / 500000,
-        used: response.data.used_quota / 500000,
-        total: (response.data.quota + response.data.used_quota) / 500000,
-        unit: "USD",
-      };
-    }
+    const hasTotalAvailable = response?.data?.total_available !== undefined;
     return {
-      isValid: false,
-      invalidMessage: response.message || "查询失败"
+      isValid: response.code === true && hasTotalAvailable,
+      remaining: hasTotalAvailable
+        ? response.data.total_available / 500000
+        : undefined,
+      unit: "USD",
     };
   },
-})
+});
 ```
 
 **配置参数**：
 | 参数 | 说明 |
 |------|------|
 | Base URL | New API 服务地址 |
-| Access Token | 访问令牌 |
-| User ID | 用户 ID |
+| API Key | 用于认证的密钥 |
+
+**结果说明**：
+
+- 只有当 `response.code === true` 且 `response.data.total_available` 存在时，才视为有效返回
+- 当余额为 `0` 时，依然视为查询成功，只是表示当前可用余额为 0
+- `500000` 是这套内置预设使用的默认换算值，不代表所有服务都使用同一规则
+- 如果你的服务返回格式或换算方式不同，请改用「自定义模板」
 
 ## 通用配置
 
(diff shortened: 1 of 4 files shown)


farion1231:
@codex review

farion1231 (commented):
Do you think this change would work for all New API providers? From what I’ve seen, some services still rely on Access Token and User ID for usage queries.

LanternCX:
@farion1231 

I re-checked the docs and the implementation. Here are the points I could confirm:

1. `/api/usage/token` is explicitly present in the current official `newapi` docs, and its meaning is token usage lookup.
2. `/api/user/self` also still exists and belongs to the user-profile side of the API.
3. `API Token / API Key` and the user-level `Access Token` used by the old flow are not the same thing: `API Token / API Key` is a token-level credential used to call the API directly, while `Access Token` is a user-level credential and usually needs to be used together with `New-Api-User` / `userId`.

That means if a service can expose the `newapi` API at all, it will usually expose `API Token / API Key`; but it may not expose `Access Token`, and it may not expose `userId` to end users. In many cases, this happens because the provider does not allow users to register on New API directly.

So if we need to choose a default usage-query template, there are really two directions:

1. Keep the `/api/user/self` path.
   This path only works as a default if the service provider is willing to expose user-level `Access Token` and `userId` to end users.

2. Switch to the `/api/usage/token` path.
   This path is more token-usage-oriented, and its assumptions are closer to what many users actually have in practice: `Base URL + API Key`.

The new endpoint is officially supported. Whether older-version compatibility should still dominate here depends on your judgment about the current provider landscape and compatibility goal.

If you think the default should prioritize compatibility with more existing provider differences, then this PR can be closed.

If you think the default should prioritize the more common user-facing setup, which is `Base URL + API Key`, then this PR is reasonable.

farion1231:
Thanks a lot for the detailed explanation, this is very helpful.

I’ll reach out to the New API author to get their perspective on this
before making a final decision.

Really appreciate your insights here.

Cao Xin ***@***.***> 于2026年4月2日周四 10:36写道：

farion1231:
Thanks for the thorough investigation and the well-written PR, @LanternCX. Your analysis of the difference between user-level and token-level credentials was really helpful.

After further testing, I looked into both `/api/usage/token` and the OpenAI-compatible `/v1/dashboard/billing/subscription` endpoint as alternatives. The issue I found is that these token-level endpoints return **per-key quota** rather than **account-level balance** — and for keys with unlimited quota, `hard_limit_usd` returns `100000000`, which isn't useful to display.

Most users want to see their overall account balance, which only `/api/user/self` can provide. While it does require Access Token + User ID (which not all providers expose), it gives the right level of information for the use case.

So I've decided to keep the current `/api/user/self` approach for now. I really appreciate the effort you put into this — your explanation in the comments was spot-on and helped me understand the trade-offs much better. If the ecosystem shifts toward token-level endpoints being more widely useful, we can revisit this.

LanternCX:
@farion1231 

That makes perfect sense. I indeed hadn't considered the issue with the return value for unlimited quotas. 

For users holding an Access Token, the API Tokens they create are typically set to an unlimited quota. 

Therefore, changing from `/api/user/self` to `/api/usage/token` would actually cause unnecessary configuration hassle for those users, which is definitely not the expected behavior.

I will go ahead and close this PR. Thanks again for the discussion!

PR #1775, opened 2026-03-30
Respect env_key when reading and writing Codex auth

ghost:
This makes Codex auth follow the configured `env_key` instead of assuming every provider uses `OPENAI_API_KEY`.

Changes in this PR:
- read auth from the key named by `env_key` in `config.toml`
- write API key edits back to that same field
- keep the frontend helpers and Rust-side extraction logic in sync
- add focused tests for `env_key`-driven auth handling

This mainly fixes Azure-style Codex providers where the transport config is fine, but the saved auth field name is not.


farion1231:

PR #1803, opened 2026-03-31
Fix New API usage query default endpoint (#1768)

### LINKED ISSUE #1768: New API usage query default points to the wrong endpoint

MavenRain:
## Summary
- Change the built-in New API usage-query template endpoint from `/api/user/self` to `/api/usage/token`
- Switch authentication from `Access Token` + `User ID` headers to `Bearer {{apiKey}}` + `User-Agent`
- Update the UI credential fields to show `API Key` + `Base URL` instead of `Base URL` + `Access Token` + `User ID`
- Update documentation (en, zh, ja) to match the new template

## Test plan
- [ ] Open a provider's usage-query modal, select the New API template, and verify the script shows `/api/usage/token` with `apiKey`
- [ ] Verify the credential fields show API Key and Base URL (no Access Token or User ID)
- [ ] Test the usage query against a New API service to confirm data fetches successfully

Closes #1768

### DIFF
--- a/docs/user-manual/en/2-providers/2.5-usage-query.md
+++ b/docs/user-manual/en/2-providers/2.5-usage-query.md
@@ -64,12 +64,12 @@ Designed specifically for New API-type proxy services:
 ```javascript
 ({
   request: {
-    url: "{{baseUrl}}/api/user/self",
+    url: "{{baseUrl}}/api/usage/token",
     method: "GET",
     headers: {
       "Content-Type": "application/json",
-      "Authorization": "Bearer {{accessToken}}",
-      "New-Api-User": "{{userId}}"
+      "Authorization": "Bearer {{apiKey}}",
+      "User-Agent": "cc-switch/1.0"
     },
   },
   extractor: function (response) {
@@ -93,9 +93,8 @@ Designed specifically for New API-type proxy services:
 **Configuration parameters**:
 | Parameter | Description |
 |-----------|-------------|
+| API Key | Authentication key for the New API service |
 | Base URL | New API service URL |
-| Access Token | Access token |
-| User ID | User ID |
 
 ## General Configuration
 
--- a/docs/user-manual/ja/2-providers/2.5-usage-query.md
+++ b/docs/user-manual/ja/2-providers/2.5-usage-query.md
@@ -64,12 +64,12 @@ New API タイプの中継サービス専用に設計されています：
 ```javascript
 ({
   request: {
-    url: "{{baseUrl}}/api/user/self",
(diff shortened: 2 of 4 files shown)


farion1231:
@codex review

farion1231:
Thanks for the contribution, @MavenRain!

After testing the `/api/usage/token` endpoint across several providers, I found that it returns per-key quota rather than account-level balance. For keys with unlimited quota, it returns a very large number (`100000000`) which doesn't provide useful information to users.

Since most users want to see their overall account balance, I've decided to keep the current `/api/user/self` approach. See #1769 for the full discussion.

Appreciate your time on this!

MavenRain:
> Thanks for the contribution, @MavenRain!
> 
> After testing the `/api/usage/token` endpoint across several providers, I found that it returns per-key quota rather than account-level balance. For keys with unlimited quota, it returns a very large number (`100000000`) which doesn't provide useful information to users.
> 
> Since most users want to see their overall account balance, I've decided to keep the current `/api/user/self` approach. See #1769 for the full discussion.
> 
> Appreciate your time on this!

Yeah, no worries . . . it gave me a chance to learn a little bit about cc-switch, so I am thankful for that opportunity!

PR #1810, opened 2026-03-31
feat(proxy): add tool schema rectifier for non-OpenAI provider compatibility

wejack639:
## Problem

When using Codex CLI (v0.117+) with CC Switch proxying to DashScope/Bailian, requests fail with:

```
InvalidParameter: The parameters, when provided as a dict, must confirm to a valid openai-compatible JSON schema
```

## Root Cause

Codex CLI sends ~188 tools via the OpenAI Responses API. Two patterns are incompatible with DashScope and similar providers:

1. **Non-function tool types** (e.g. `web_search`, `code_interpreter`) - these are OpenAI-specific built-in tools that third-party providers don't recognize
2. **Missing `required` field** in tool `parameters` - DashScope requires this field to be present even when empty (`[]`), but ~48 of Codex's tools omit it (all-optional-param tools)

## Solution

Add a **Tool Schema Rectifier** that runs in the proxy `forward()` pipeline (after `filter_private_params_with_whitelist`):

- Filter out tools where `type != "function"`
- Patch `"required": []` into any tool parameters missing the field

This follows the same pattern as existing rectifiers (`thinking_rectifier`, `thinking_budget_rectifier`).

## Testing

- 9 unit tests covering: no tools, empty tools, web_search removal, code_interpreter removal, missing required patch, existing required skip, combined rectification, no-params tools, non-object params
- All tests pass: `cargo test -- tool_schema_rectifier`
- Manually verified: Codex CLI `codex exec 'say hello'` works through CC Switch -> DashScope after this fix

## Files Changed

| File | Change |
|---|---|
| `src-tauri/src/proxy/tool_schema_rectifier.rs` | New - rectifier implementation + 9 tests |
| `src-tauri/src/proxy/mod.rs` | Register new module |
| `src-tauri/src/proxy/forwarder.rs` | Call rectifier in forward() pipeline |

### DIFF
--- a/src-tauri/src/proxy/forwarder.rs
+++ b/src-tauri/src/proxy/forwarder.rs
@@ -825,18 +825,21 @@ impl RequestForwarder {
         // 默认使用空白名单，过滤所有 _ 前缀字段
         let mut filtered_body = filter_private_params_with_whitelist(request_body, &[]);
 
-        // Tool Schema 整流：修复非 OpenAI provider 的 tool 兼容性
-        // - 移除非 function 类型的 tool（如 web_search）
-        // - 为缺少 required 字段的 parameters 补充空数组
-        let tool_rectify_result =
-            super::tool_schema_rectifier::rectify_tool_schema(&mut filtered_body);
-        if tool_rectify_result.applied {
-            let tag = adapter.name();
-            log::info!(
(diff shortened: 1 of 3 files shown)


farion1231:
@codex review

PR #1829, opened 2026-04-01
ci: add dependabot.yml config

bengbengbalabalabeng:
## Changes

Introduces `Dependabot` to help automate dependency updates and improve supply chain security.

After merging this PR, you don't need to wait for the scheduled cycle. You can directly go to the repository panel's `Insights` -> `Dependency graph` -> `Dependabot` and click `Check for updates` to immediately trigger a scan test.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,40 @@
+version: 2
+updates:
+  - package-ecosystem: "npm"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+    labels:
+      - "dependencies"
+      - "frontend"
+    commit-message:
+      prefix: "chore(deps)"
+    groups:
+      frontend-deps:
+        patterns:
+          - "*"
+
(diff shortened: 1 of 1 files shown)


farion1231 (commented):
Recently, supply chain attacks have been becoming more frequent, so keeping dependency versions pinned is a relatively safer choice. What do you think?

bengbengbalabalabeng:
> Recently, supply chain attacks have been becoming more frequent, so keeping dependency versions pinned is a relatively safer choice. What do you think?

I agree with your point. At the same time, the versions detected by Dependabot can serve as a useful reference for evaluating dependency upgrades. This allows developers to stay informed without actively tracking version changes, since Dependabot handles the monitoring for us. We can decide whether to adopt the latest version on a case‑by‑case basis, and if a particular update is not suitable, we can simply ignore it using `@dependabot ignore current dependency update`.

farion1231 (approved):

PR #1871, opened 2026-04-03
feat(usage): add speed column to request logs

SunneeYang:
## Summary
- add a per-request speed column in the request log table
- compute speed from output tokens and generation duration fallback logic
- add `tokensPerSecond` locale entries for zh/en/ja

## Test plan
- [x] `pnpm typecheck`
- [x] confirm only 4 files changed vs `main`


farion1231:
@codex review

PR #1916, opened 2026-04-06
fix(claude): preserve hooks and includeCoAuthoredBy

### LINKED ISSUE #1907: [Bug] hooks in ~/.claude/settings.json get randomly deleted when switching nodes
## Bug Description

When using cc-switch to switch between nodes, the "hooks" section in `~/.claude/settings.json` is sometimes automatically deleted. This does not happen every time, but occasionally the hooks configuration disappears, requiring the user to reconfigure hooks from scratch.

## Steps to Reproduce

1. Configure some hooks in `~/.claude/settings.json`
2. Use cc-switch to switch between different nodes
3. Observe that sometimes the "hooks" section in `settings.json` is missing/deleted

## Expected Behavior

Switching nodes should not modify or delete the hooks configuration in `~/.claude/settings.json`.

## Additional Context

This is an intermittent issue - not every switch operation causes the problem, but it happens frequently enough to be disruptive to the user workflow.

zerone0x:
## Summary / 概述

When syncing Claude providers, preserve existing `hooks` and `includeCoAuthoredBy` values from the current `~/.claude/settings.json` if the provider config does not define them. This prevents those fields from being wiped during node/provider switches.

## Related Issue / 关联 Issue

Fixes #1907

## Screenshots / 截图

| N/A | N/A |

## Checklist / 检查清单

### DIFF
--- a/src-tauri/src/services/provider/live.rs
+++ b/src-tauri/src/services/provider/live.rs
@@ -666,7 +666,24 @@ pub(crate) fn write_live_snapshot(app_type: &AppType, provider: &Provider) -> Re
     match app_type {
         AppType::Claude => {
             let path = get_claude_settings_path();
-            let settings = sanitize_claude_settings_for_live(&provider.settings_config);
+            let mut settings = sanitize_claude_settings_for_live(&provider.settings_config);
+
+            if path.exists() {
+                if let Ok(existing) = read_json_file::<Value>(&path) {
+                    if let (Some(existing_obj), Some(settings_obj)) =
+                        (existing.as_object(), settings.as_object_mut())
+                    {
+                        for key in ["hooks", "includeCoAuthoredBy"] {
+                            if !settings_obj.contains_key(key) {
+                                if let Some(value) = existing_obj.get(key) {
+                                    settings_obj.insert(key.to_string(), value.clone());
+                                }
+                            }
+                        }
+                    }
+                }
+            }
+
             write_json_file(&path, &settings)?;
         }
         AppType::Codex => {
--- a/src-tauri/tests/provider_service.rs
+++ b/src-tauri/tests/provider_service.rs
@@ -594,7 +594,11 @@ fn provider_service_switch_claude_updates_live_and_state() {
         },
         "workspace": {
             "path": "/tmp/workspace"
-        }
+        },
+        "hooks": {
+            "before": ["echo hello"]
+        },
+        "includeCoAuthoredBy": false
     });
     std::fs::write(
         &settings_path,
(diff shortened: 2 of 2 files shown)


farion1231:

farion1231:
Hi @zerone0x, thanks for taking the time to look into #1907 and for the clean, well-scoped PR. Unfortunately I don't think this direction is the right fix — let me explain the reasoning so we're on the same page.

**Why a hardcoded field whitelist isn't the fix we want**

cc-switch treats each provider as a **complete, self-contained snapshot** of `settings.json`. Switching intentionally replaces the whole file, not just `env`. For fields that should follow the user *across* providers (like `hooks`, `includeCoAuthoredBy`, and similar), we already have a dedicated mechanism: the **Shared Config Snippet**, which is stored separately and deep-merged into whichever provider gets written to live. `includeCoAuthoredBy: false` is in fact the default content of that snippet (`DEFAULT_COMMON_CONFIG_SNIPPET` in `src/components/providers/forms/hooks/useCommonConfigSnippet.ts`).

Layering a hardcoded `["hooks", "includeCoAuthoredBy"]` fallback on top of that creates several problems:

1. **Overlaps with the Shared Config layer** — these two fields would end up handled by two mechanisms at once. A user who deliberately customizes `includeCoAuthoredBy` via the shared snippet and then sets it differently on one provider would silently get the old value back from the live file.
2. **Breaks the "full snapshot" semantics** — a provider can no longer be reasoned about in isolation; its effective on-disk config becomes a function of whatever happens to be in the live file at write time.
3. **Removes the ability to deliberately drop a field** — if a user *intentionally* removes `hooks` from one provider ("this account should not run any hooks"), the new logic silently restores it from live. There's no way to express "I actually want this field gone".
4. **Doesn't scale** — the same ask will keep coming back for other fields: `permissions`, `apiKeyHelper`, `model`, `statusLine`, etc. A hardcoded whitelist can't keep up, and each addition blurs the ownership of that field further.
5. **Earlier field-merge experiments already showed this to be fragile** — we've tried variations of field-level preservation before and hit problems with nested arrays (`hooks` itself is one), "user removed vs user never set" ambiguity, and multi-layer merge conflicts. The current "full snapshot + explicit shared config" model was the deliberate outcome of that experience.

**The intended fix for #1907**

The recommended path is already documented in the [README FAQ](https://github.com/farion1231/cc-switch#faq) ("My plugin configuration disappeared after switching providers"): extract `hooks` / `includeCoAuthoredBy` into the Shared Config Snippet, and make sure "Write Shared Config" is enabled on each Claude provider. I posted a detailed walkthrough on #1907 — feel free to take a look there.

**If you'd still like to contribute in this area**

A couple of directions that *would* be a good fit, if you're interested:

- **Improve FAQ discoverability** — the current FAQ entry is phrased around "plugin configuration" and users with `hooks` / `includeCoAuthoredBy` issues may not connect the dots. A wording update (or an additional FAQ entry that calls out these fields by name) would directly help future issues like #1907.

Either of those I'd be happy to review. Thanks again for digging in — the investigation and test you added were solid, it's just that the approach conflicts with the design direction we want to stick with.

PR #1922, opened 2026-04-06
fix(provider): prevent config JSON blinking when editing with proxy running

timwuhaotian:
## Summary

Fixes an issue where the config JSON editor would blink/flash constantly when editing a provider while the proxy is running.

## Root Cause

When the proxy is running, the providers query refetches every 10 seconds ([queries.ts:64](src/lib/query/queries.ts#L64)). This triggered a cascade of state updates:

1. Provider object reference changes on refetch
2. `initialData` recomputes (due to `provider?.meta` dependency)
3. `defaultValues` recomputes → `form.reset()` was called
4. `settingsConfig` changes → `useCommonConfigSnippet` triggers state updates → **blinking**

## Fix

- **ProviderForm.tsx**: Skip `form.reset()` in edit mode - the form should only initialize once when the dialog opens, not on every provider data change
- **EditProviderDialog.tsx**: Remove `provider?.meta` from `initialData` dependencies to stabilize memoization

## Test Plan

1. Start the app with proxy running
2. Open edit dialog for a provider
3. Observe the config JSON - it should remain stable without blinking
4. Wait 10+ seconds (through multiple refetch cycles)
5. Confirm config still stable

## Checklist

- [x] `pnpm typecheck` passes
- [ ] No Rust code changed
- [ ] No user-facing text changed (no i18n updates needed)

### DIFF
--- a/src/components/providers/EditProviderDialog.tsx
+++ b/src/components/providers/EditProviderDialog.tsx
@@ -139,6 +139,8 @@ export function EditProviderDialog({
   }, [liveSettings, provider?.settingsConfig]); // 只依赖 settingsConfig，不依赖整个 provider
 
   // 固定 initialData，防止 provider 对象更新时重置表单
+  // 注意：不依赖 provider?.meta，避免 providers refetch 时触发重新计算导致闪烁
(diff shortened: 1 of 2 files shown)


yovinchen:
@codex review

PR #1936, opened 2026-04-07
feat: add MiniMax-M2.7-highspeed provider presets

octo-patch:
## Summary

- Add **MiniMax Highspeed** (CN) preset using `MiniMax-M2.7-highspeed` model via `api.minimaxi.com/anthropic`
- Add **MiniMax en Highspeed** (international) preset using `MiniMax-M2.7-highspeed` model via `api.minimax.io/anthropic`
- Add unit tests covering all four MiniMax presets (M2.7 and M2.7-highspeed, CN and international variants)

`MiniMax-M2.7-highspeed` delivers the same performance as MiniMax-M2.7 but is faster and more agile — useful for latency-sensitive workflows.

## Changes

- `src/config/claudeProviderPresets.ts` — added two highspeed presets after the existing MiniMax M2.7 presets
- `tests/config/claudeProviderPresets.test.ts` — added 12 unit tests for all MiniMax preset variants

## Test plan

- [x] `pnpm test:unit -- tests/config/claudeProviderPresets.test.ts` — all 23 tests pass
- [x] No regressions in existing MiniMax presets

API documentation: https://platform.minimax.io/docs/api-reference/text-anthropic-api


farion1231 (commented):
Thanks for the contribution. However, after comparing all fields, the only difference between MiniMax Highspeed and the existing MiniMax preset is the ANTHROPIC_MODEL string  suffix — the endpoint, auth, category, theme, and promotion key are identical. Users can achieve the same result by editing the ANTHROPIC_MODEL field after creating a MiniMax provider. It doesn’t seem like an additional preset is necessary.

PR #1942, opened 2026-04-07
[codex] Normalize fragmented OpenAI chat system prompts

yovinchen:
## Summary
- normalize fragmented `system` prompts into a single leading OpenAI chat `system` message
- preserve the existing non-system message conversion flow
- add a regression test covering fragmented `system` prompt normalization

## Why
Some strict OpenAI-compatible chat backends, including the Nvidia/Qwen path discussed in #1881, reject requests when Claude-side prompt fragments are forwarded as multiple `system` messages. Reordering alone is not sufficient for those providers; the prompt needs to be collapsed into one leading `system` message.

## Validation
- `cargo test proxy::providers::transform --manifest-path src-tauri/Cargo.toml`

Closes #1881

PR #1946, opened 2026-04-08
[codex] Preserve cache_control when merging system prompts

### LINKED ISSUE #1945: [BUG] Claude Code fails with "$.input_tokens" error and disconnects when using local LLM / custom CLAUDE_BASE_URL
Describe the bug
When pointing Claude Code to a local LLM server (via llama.cpp, Ollama, or LiteLLM) using the CLAUDE_BASE_URL environment variable, the client frequently crashes or disconnects. The most common error is a JavaScript evaluation error: Cannot read property 'input_tokens' of undefined (or similar $.input_tokens path error).

Even when protocol-matching proxies are used to mimic the Anthropic /v1/messages format, the client often terminates the connection (cancel task) immediately after the prompt is processed but before/during the streaming of the response.

To Reproduce
Start a local OpenAI-compatible server (e.g., llama.cpp on port 8080).

Use a proxy (like LiteLLM or a custom Python script) to map the local API to Anthropic's message format.

Set environment variable: :CLAUDE_BASE_URL="http://127.0.0.1:8092/v1"

Run claude.

Enter any prompt (e.g., "hi").

Observe the crash or "Connection disconnected" message in the terminal.

Expected behavior
Claude Code should be resilient to slight variations in the usage object within the streaming chunks, or at least provide a meaningful error message instead of a hard crash/disconnect when input_tokens are not immediately available in the first chunk.

Screenshots / Logs
Proxy Side Logs (Python/FastAPI):

Plaintext
# The client sends a request, but then cancels it
srv stop: cancel task, id_task = 2006
slot release: id 0 | task 2006 | stop processing: n_tokens = 18, truncated = 0
Client Side Error:

Plaintext
Error evaluating $.input_tokens: undefined
# OR
Connection disconnected.
Environment Context
OS: Windows 11 (PowerShell)

Claude Code Version: Latest as of April 2026

Local Model: Qwen 3.5 Coder (via llama.cpp)

Proxy: LiteLLM / Custom FastAPI Proxy

Additional context
It seems Claude Code strictly expects a message_start event with a valid usage object containing input_tokens at the very beginning of the stream. If the local model takes time to process the prompt (high TTFT), or if the final message_delta chunk is missing the usage stats, the client-side parser fails.

This makes it nearly impossible to use Claude Code with local engineering workflows, which is a major limitation for developers in air-gapped or privacy-sensitive environments.

## Related Issues

After posting this issue, I found several related issues in this repository:

### #1370 - "llama.cpp部署的模型无法使用" (OPEN)
- **Similarity**: Also reports issues with llama.cpp deployed models and cc-switch compatibility
- **Status**: Still OPEN, indicating the problem persists
- **Key info**: User reports 404 errors with llama.cpp + Qwen3-Coder-30B, while curl tests work fine

### #1678 - "Claude Code Auto Mode Classifier Compatibility Issue" (OPEN)
- **Similarity**: Claude Code compatibility issues with local proxies
- **Status**: Still OPEN, indicating broader compatibility problems

### #759 - "增加token使用量返回" (OPEN)
- **Similarity**: Related to token usage tracking and input_tokens/output_tokens statistics

## Additional Context

This issue appears to be a specific manifestation of a broader compatibility problem between Claude Code and local LLM servers. The core issue seems to be that Claude Code has strict expectations about the API response format, particularly regarding the `usage` object and `input_tokens` field.

The fact that multiple users are reporting similar issues suggests this is a significant limitation for developers wanting to use Claude Code with local/private LLM workflows.

yovinchen:
## Summary
- preserve message-level `cache_control` when fragmented `system` prompts are merged into a single leading OpenAI chat `system` message
- intentionally drop `cache_control` only when merged `system` fragments carry conflicting cache hints
- add regression coverage for both the preserved and conflicting merge cases

## Why
PR #1942 fixed strict OpenAI-compatible backends that reject fragmented `system` prompts, but the merge path rebuilt the combined `system` message with only `role` and `content`. That dropped `cache_control` from the original `system` blocks and could disable prompt-prefix caching for cache-aware upstreams.

This follow-up keeps the strict-backend compatibility fix while preserving cache hints whenever they can be inherited unambiguously.

## Validation
- `cargo test proxy::providers::transform --manifest-path src-tauri/Cargo.toml`

Closes #1942
Closes #1945


yovinchen:
@codex review

yovinchen:
@codex review

farion1231 (approved):

PR #1976, opened 2026-04-09
Fix Ghostty session restore launch path

### LINKED ISSUE #1975: Ghostty opens blank terminal instead of launching Claude Code
## Summary
When `cc switch` is configured to use Ghostty as the preferred terminal, clicking to open a terminal opens a blank Ghostty window instead of launching directly into a Claude Code session.

The same flow works correctly with Terminal.app, which opens a terminal already running Claude Code.

## Steps to reproduce
1. Set the preferred terminal in `cc switch` to Ghostty
2. Use the UI action to open a terminal / open Claude Code from `cc switch`
3. Observe that Ghostty opens a normal blank shell window

## Expected behavior
Ghostty should behave like Terminal.app and open directly into a Claude Code terminal session, rather than only opening an empty terminal window.

## Actual behavior
Ghostty opens, but Claude Code is not started automatically.

## Additional context
- Terminal.app works as expected on the same machine
- Ghostty is listed as a supported/selectable terminal option in `cc switch`
- This suggests terminal detection works, but the Ghostty launch/integration path may not be correctly passing through the command that should start Claude Code

## Environment
- macOS
- Preferred terminal: Ghostty
- Comparison terminal: Terminal.app

If helpful, I can test potential fixes or provide more environment details.

Suda202:
## Summary
- launch Ghostty sessions through the user's shell instead of injecting raw terminal input
- reuse the existing shell command builder so `cwd` handling matches other terminal integrations
- remove the Ghostty raw-input helpers and replace the tests with shell-command assertions

## Test plan
- [ ] Set preferred terminal to Ghostty in cc-switch
- [ ] Open a Claude session terminal from the session manager
- [ ] Verify Ghostty opens directly into the resumed Claude Code session
- [ ] Confirm Terminal.app behavior is unchanged

Closes #1975

### DIFF
--- a/src-tauri/src/session_manager/terminal/mod.rs
+++ b/src-tauri/src/session_manager/terminal/mod.rs
@@ -78,19 +78,29 @@ end tell"#
 }
 
 fn launch_ghostty(command: &str, cwd: Option<&str>) -> Result<(), String> {
-    let full_command = build_shell_command(command, cwd);
     let shell = std::env::var("SHELL").unwrap_or_else(|_| "/bin/zsh".to_string());
 
+    let mut args = vec![
+        "-na".to_string(),
+        "Ghostty".to_string(),
+        "--args".to_string(),
+        "--quit-after-last-window-closed=true".to_string(),
+    ];
+
+    if let Some(dir) = cwd {
+        if !dir.trim().is_empty() {
+            args.push(format!("--working-directory={dir}"));
+        }
+    }
+
+    args.push("-e".to_string());
+    args.push(shell);
+    args.push("-l".to_string());
+    args.push("-c".to_string());
+    args.push(command.to_string());
+
     let status = Command::new("open")
-        .arg("-na")
-        .arg("Ghostty")
-        .arg("--args")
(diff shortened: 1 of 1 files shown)


yovinchen:
@codex review

Suda202:
The CI failures are not caused by the logic in this PR — they're rebase conflicts with changes that landed on main after this PR was opened.

Specifically:
- **Backend**: `tests/skill_sync.rs` fails to compile because the `InstalledSkill` struct on main now has two new required fields (`content_hash` and `updated_at`) that the test initializers in this PR don't include (E0063).
- **Frontend**: `UnifiedSkillsPanel.test.tsx` failure also appears to be a main-branch compatibility issue.

The fix in this PR (switching Ghostty from raw terminal input injection to the shell command builder) is the right approach — I can confirm that even with standalone Ghostty installed (no cmux), cc-switch opened a blank terminal window. This PR addresses the root cause.

Could a maintainer help rebase this onto the latest main and update the test fixtures? Happy to provide more context if needed.

farion1231 (approved):

PR #1978, opened 2026-04-09
fix(tray): use an app-specific tray id

### LINKED ISSUE #1977: Use an app-specific Linux tray id to avoid tray-icon collisions with other Tauri apps
On Linux, `cc-switch` currently builds its tray icon with `TrayIconBuilder::with_id("main")`.

That becomes a problem when another Tauri app does the same thing. The `tray-icon` crate uses the tray id to derive both the AppIndicator identifier and the temporary PNG path under `$XDG_RUNTIME_DIR/tray-icon`, so two apps using the same id end up sharing the same runtime resources.

Concrete example:
- `cc-switch` uses `with_id("main")`
- `clash-verge-rev` also uses `with_id("main")`
- `cc-switch` starts first and exposes `/run/user/$UID/tray-icon/tray-icon-main-0.png`
- Starting `clash-verge-rev` creates the same `tray-icon-main-0.png`, then updates its icon to `tray-icon-main-1.png` / `tray-icon-main-3.png`
- During that update, `tray-icon` deletes the previously active `tray-icon-main-0.png`
- GNOME AppIndicator then tries to refresh `cc-switch` using the stale path and the icon disappears or falls back to an error/placeholder icon

This is not specific to Clash Verge Rev; that app is just an easy reproducer because it also uses Tauri + `tray-icon` and also uses `"main"` as the tray id.

Suggested fix:
- use an app-specific tray id such as `"cc-switch"`
- update internal `tray_by_id("main")` call sites to use the same app-specific tray id constant

I have a small PR ready that does exactly that.

lidaxian121:
## Summary
- replace the shared tray id `"main"` with the app-specific id `"cc-switch"`
- update internal `tray_by_id(...)` lookups to use the same constant
- add a small unit test that guards against regressing back to `"main"`

## Why
On Linux, `tray-icon` derives both the AppIndicator identifier and the temporary PNG filename from the tray id. Using a generic id like `"main"` makes `cc-switch` collide with other Tauri apps that do the same thing.

A concrete reproducer is `clash-verge-rev`, which also uses `with_id("main")`:
- `cc-switch` starts first and exposes `/run/user/$UID/tray-icon/tray-icon-main-0.png`
- starting `clash-verge-rev` creates the same `tray-icon-main-0.png`
- when `clash-verge-rev` updates its tray icon, `tray-icon` deletes the previously active `tray-icon-main-0.png` and moves on to `tray-icon-main-1.png` / `tray-icon-main-3.png`
- GNOME AppIndicator then refreshes `cc-switch` against a stale path and the icon disappears or falls back to a broken placeholder

Using an app-specific tray id avoids the collision entirely.

Closes #1977.

## Verification
- `cargo test --lib tray_id_is_unique_to_app --manifest-path src-tauri/Cargo.toml`

### DIFF
--- a/src-tauri/src/commands/failover.rs
+++ b/src-tauri/src/commands/failover.rs
@@ -162,7 +162,7 @@ pub async fn set_auto_failover_enabled(
 
     // 刷新托盘菜单，确保状态同步
     if let Ok(new_menu) = crate::tray::create_tray_menu(&app, &state) {
-        if let Some(tray) = app.tray_by_id("main") {
+        if let Some(tray) = app.tray_by_id(crate::tray::TRAY_ID) {
             let _ = tray.set_menu(Some(new_menu));
         }
     }
--- a/src-tauri/src/lib.rs
+++ b/src-tauri/src/lib.rs
@@ -168,7 +168,7 @@ async fn update_tray_menu(
 ) -> Result<bool, String> {
     match tray::create_tray_menu(&app, state.inner()) {
         Ok(new_menu) => {
-            if let Some(tray) = app.tray_by_id("main") {
+            if let Some(tray) = app.tray_by_id(tray::TRAY_ID) {
                 tray.set_menu(Some(new_menu))
                     .map_err(|e| format!("更新托盘菜单失败: {e}"))?;
                 return Ok(true);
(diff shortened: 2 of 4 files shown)


yovinchen:
@codex review

farion1231 (approved):

PR #1984, opened 2026-04-10
Align Thinking fallback with main-model-only Claude mappings

yovinchen:
## Summary
- align Claude Thinking model hydration with the existing Haiku-style fallback behavior
- show `ANTHROPIC_MODEL` in the Thinking field when `ANTHROPIC_REASONING_MODEL` is unset
- keep the fallback read-only so saving a main model does not silently rewrite reasoning config

## Testing
- `pnpm typecheck`
- `pnpm test:unit` *(fails due to an unrelated existing mock issue in `tests/components/UnifiedSkillsPanel.test.tsx` for `useCheckSkillUpdates`)*


yovinchen:
@codex review

farion1231 (approved):

PR #1985, opened 2026-04-10
[codex] restore auth tab localization in settings

yovinchen:
## What changed
- added the missing `settings.tabAuth` locale entry in `zh`, `en`, and `ja`
- kept the existing settings tab rendering logic unchanged

## Why
The settings page already reads the auth tab label from i18n, but the locale bundles did not define `settings.tabAuth`. That caused the auth tab to fall back to the hardcoded Chinese default instead of showing the proper localized label.

## Impact
- the auth tab now shows the expected localized label in Chinese, English, and Japanese
- no component behavior or layout logic changed

## Root cause
The auth tab was wired to a translation key that was never added to the locale files.

## Validation
- `pnpm format:check`
- `pnpm typecheck`

### DIFF
--- a/src/i18n/locales/en.json
+++ b/src/i18n/locales/en.json
@@ -243,6 +243,7 @@
     "title": "Settings",
     "general": "General",
     "tabGeneral": "General",
+    "tabAuth": "Auth",
     "tabAdvanced": "Advanced",
     "tabProxy": "Proxy",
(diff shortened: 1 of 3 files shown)


yovinchen:
@codex review

farion1231 (approved):

PR #2022, opened 2026-04-12
feat(opencode): add dedicated UI for model token limits (limit.context & limit.output)

wwminger:
## Summary

Previously, there was no way to edit OpenCode model token limits in the UI. The `limit` field was excluded from extra fields as a known/reserved key (`OPENCODE_KNOWN_MODEL_KEYS`), but had no dedicated input. Users could only set limits via presets.

This PR adds a **Token Limits** section with number inputs for `limit.context` (context window) and `limit.output` (max output), placed between Model Properties and SDK Options in the expanded model details.

## Changes

- `OpenCodeFormFields.tsx`: Added `handleModelLimitContextChange` and `handleModelLimitOutputChange` handlers + Token Limits UI section
- `src/i18n/locales/en.json`, `zh.json`, `ja.json`: Added 6 i18n keys (tokenLimits, limitContext, limitOutput, limitContextPlaceholder, limitOutputPlaceholder, tokenLimitsHint)

## Testing

- `pnpm typecheck` passed
- Manual verification: Token Limits section appears in expanded model details with working number inputs

## Screenshots

The Token Limits section appears when expanding a model in the OpenCode provider form, allowing users to set context window and max output token limits directly.

### DIFF
--- a/src/components/providers/forms/OpenCodeFormFields.tsx
+++ b/src/components/providers/forms/OpenCodeFormFields.tsx
@@ -466,6 +466,53 @@ export function OpenCodeFormFields({
     });
   };
 
+  // Model limit handlers (limit.context and limit.output)
+  const handleModelLimitContextChange = (modelKey: string, value: string) => {
+    const model = models[modelKey];
+    const numValue = value.trim() ? Number(value) : undefined;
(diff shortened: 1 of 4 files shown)


farion1231:
@codex review

PR #2023, opened 2026-04-12
fix(opencode): use json5 parser for trailing comma tolerance

wwminger:
## Summary

OpenCode CLI writes `opencode.json` with trailing commas (valid JSONC), but CC Switch parsed it with `serde_json` (strict JSON), causing errors like:

> JSON 解析错误: /home/wm/.config/opencode/opencode.json: trailing comma at line 35 column 3

Switch to `json5::from_str` which accepts both JSON and JSONC. The `json5` crate is already a project dependency (used by `openclaw_config.rs` for the same reason).

## Changes

- `src-tauri/src/opencode_config.rs`: Changed `serde_json::from_str(&content)` to `json5::from_str(&content)`, error type from `AppError::json()` to `AppError::Config()`

## Testing

- `cargo check` passed
- `cargo clippy` — 0 warnings
- `cargo test` — all 743+ tests passed

### DIFF
--- a/src-tauri/src/opencode_config.rs
+++ b/src-tauri/src/opencode_config.rs
@@ -61,7 +61,8 @@ pub fn read_opencode_config() -> Result<Value, AppError> {
     }
 
     let content = std::fs::read_to_string(&path).map_err(|e| AppError::io(&path, e))?;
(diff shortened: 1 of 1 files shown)


farion1231:
@codex review

farion1231 (approved):

PR #2031, opened 2026-04-12
Fix usage script variable replacement for custom providers

West-Pavilion:
This fixes a pretty confusing usage-query bug in custom templates.

If a provider is using the custom usage template, the UI shows `{{apiKey}}` and `{{baseUrl}}` as supported variables. In practice, though, they could fail to resolve unless the raw values were pasted directly into the script.

## What changed

- make usage-script test execution fall back to the provider's saved credentials when the usage form fields are left empty
- extend provider credential extraction so Codex / OpenAI-style providers can resolve `OPENAI_API_KEY` and `base_url` from their stored config
- accept lowercase aliases like `{{apikey}}` and `{{baseurl}}` in addition to the existing variable names
- add focused tests for Codex credential fallback and template replacement behavior

## Verification

- `cargo test codex --lib`
- `cargo test test_build_script_with_vars_supports_legacy_lowercase_aliases --lib`

This should make the actual behavior line up with what the modal tells users.


farion1231:
@codex review

AkimioJR:
I can still reproduce this issue on v3.16.2 with this setup:

- App/provider type: Codex provider
- Provider setup: custom configuration
- Usage query template: Custom
- Script variables used: `{{baseUrl}}` / `{{apiKey}}`

The usage query test/query does not correctly inject the current Codex provider credentials from the saved provider config, so the request fails unless the values are hardcoded. This seems to match the issue described in this PR, especially the fallback from saved provider credentials for custom usage templates.

I also checked the PRs merged after the v3.16.2 release and did not find a merged fix covering the Codex + custom usage template path yet.

Could you confirm whether this PR covers the following path?

`Codex provider -> Custom config -> Usage Query -> Custom template -> {{apiKey}} / {{baseUrl}}`

If it does, I think this PR would still be useful to merge/rebase. I’m happy to help verify it against the v3.16.2 reproduction case.

PR #2034, opened 2026-04-13
Add Anthropic web-search bridge for Responses providers

linzj:
## Summary
- bridge Anthropic web_search tools to OpenAI Responses web_search_call requests
- round-trip web search history/results across request, non-streaming response, and SSE streaming paths
- preserve web search usage accounting and reject unsupported constraints instead of widening behavior

## Scope
This is not limited to `codex_oauth`. It affects every Claude provider configured with `api_format = "openai_responses"`, because both request/response transforms and streaming handlers are selected by API format.

## Notes
- forced Anthropic `tool_choice` for `web_search` falls back to `auto`
- empty `allowed_domains` are omitted instead of being sent as an empty Responses filter array
- unsupported `blocked_domains` and `max_uses` now fail fast instead of being silently widened or ignored
- Responses requests only include supported web search fields (`web_search_call.action.sources`)
- SSE streaming now emits `server_tool_use` for non-query web search actions and deduplicates added/done tool-use emission

### DIFF
--- a/src-tauri/src/proxy/providers/streaming_responses.rs
+++ b/src-tauri/src/proxy/providers/streaming_responses.rs
@@ -62,6 +62,157 @@ fn tool_item_key_from_event(data: &Value) -> Option<String> {
     None
 }
 
+#[inline]
+fn output_item_from_event(data: &Value) -> &Value {
+    data.get("item").unwrap_or(data)
+}
+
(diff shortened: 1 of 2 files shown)


farion1231:
@codex review

farion1231:
@codex review

PR #2060, opened 2026-04-13
fix(proxy): strip hop-by-hop response headers per RFC 7230

yovinchen:
## Summary

- Add `strip_hop_by_hop_response_headers` to remove hop-by-hop headers from proxy responses per RFC 2616 / RFC 7230
- Applied across all three response paths: SSE passthrough, streaming (`handle_streaming`), and non-streaming (`handle_non_streaming`)
- Also strips extension headers listed in the `Connection` header value
- Remove hardcoded `Connection: keep-alive` from stream_check Claude requests to align with other providers

## Changes

- `src-tauri/src/proxy/response_processor.rs` — add `strip_hop_by_hop_response_headers`, call it in streaming/non-streaming handlers, add unit tests
- `src-tauri/src/proxy/handlers.rs` — remove hardcoded `Connection: keep-alive` from SSE path, call strip function when rebuilding responses
- `src-tauri/src/services/stream_check.rs` — remove `Connection: keep-alive` from Claude health check requests

## Test plan

- [ ] `cargo test` passes including two new unit tests
- [ ] Proxy responses no longer contain `Connection`, `Transfer-Encoding`, `Keep-Alive` or other hop-by-hop headers
- [ ] Business headers (`Content-Type`, `Content-Length`, etc.) are preserved correctly


farion1231:
@codex review

farion1231 (approved):

PR #2101, opened 2026-04-15
fix(skills): sync imported skills to app directories after import

yaoguohh:
import_from_apps() saves skills to the database but never calls sync_to_app_dir(), so skills appear as "installed" in the UI while symlinks are missing from the target app directories.
Added sync_to_app_dir() calls after db.save_skill(), matching the pattern used by install() and toggle_app().


farion1231 (approved):

PR #2116, opened 2026-04-15
feat(gemini): support OAuth preview models and release builds

OXOOOOX:
Align Gemini defaults with the working preview model names and add the official CLI install/login flow so OAuth providers can be configured end-to-end. Keep updater artifacts disabled for local Tauri builds and enable them only in release builds so local packaging succeeds without signing keys.

## Related Issue / 关联 Issue

Fixes #

## Screenshots / 截图

## Checklist / 检查清单


OXOOOOX (commented):
6

PR #2154, opened 2026-04-17
Add StepFun CN and EN presets

hengm3467:
## Summary
- add separate StepFun CN and StepFun en presets
- map Claude to Anthropic Step Plan endpoints and OpenCode/OpenClaw to OpenAI-compatible Step Plan endpoints
- replace the StepFun icon with the new round gradient logo

## Verification
- pnpm typecheck
- pnpm build:renderer


hengm3467:
Close as the new / clean PR is #2155

PR #2179, opened 2026-04-19
feat(ui): add quick-set button for model mapping fields

lispking:
Add a "一键设置" (Quick Set) button in the model mapping section to simplify provider configuration. When users enter a model name in any of the five model fields, they can now click this button to populate all fields with that same value.

This addresses the UX friction of manually filling all five model mapping fields (主模型, 推理模型, Haiku, Sonnet, Opus) when the provider uses the same model name across all request types.

Implementation:
- Add Wand2 icon import from lucide-react
- Insert quick-set button alongside existing fetch-models button
- Logic picks first non-empty model value and applies to all fields
- Show success toast after applying
- Disabled state when all model fields are empty
- Add i18n strings for zh, en, ja locales

Relates to user feedback about tedious model configuration workflow.

image

### DIFF
--- a/src/components/providers/forms/ClaudeFormFields.tsx
+++ b/src/components/providers/forms/ClaudeFormFields.tsx
@@ -24,7 +24,7 @@ import {
   DropdownMenuSeparator,
   DropdownMenuTrigger,
 } from "@/components/ui/dropdown-menu";
-import { ChevronDown, ChevronRight, Download, Loader2 } from "lucide-react";
+import { ChevronDown, ChevronRight, Download, Loader2, Wand2 } from "lucide-react";
 import EndpointSpeedTest from "./EndpointSpeedTest";
 import { ApiKeySection, EndpointField, ModelInputWithFetch } from "./shared";
 import { CopilotAuthSection } from "./CopilotAuthSection";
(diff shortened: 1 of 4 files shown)


farion1231 (commented):
As far as I know, almost all providers return multiple models. Is this feature really necessary?

lispking:
> As far as I know, almost all providers return multiple models. Is this feature really necessary?

It’s definitely necessary, and this button doesn’t affect the original usage.

lispking:
Otherwise, when sharing a single model, you have to copy it in five different places, which makes the experience really poor.

farion1231:
But for now, the range of cases where this feature is useful is still
fairly limited. What do you think about adding another option as well:
copying the main model’s fields to the other model mappings?

King ***@***.***> 于2026年4月19日周日 20:51写道：

lispking:
The current setup is to fill in the main model, then click the one-click apply button to sync it to the other input fields.

This scenario is highly applicable in many proxy (relay) service setups, and a significant proportion of users rely on such services.

For relay/proxy services, usually only a few models work well—but they can occasionally fail. When that happens, you have to duplicate the relay configuration for different models and tweak them repeatedly. The amount of work scales with how many models the relay provides, which makes the setup really frustrating.

farion1231 (approved):

PR #2187, opened 2026-04-19
feat(deeplink): allow overriding OpenCode `npm` package via URL param

oaker-io:
## Summary

- Adds an `npm` parameter to the `ccswitch://` provider deeplink so callers can pick which `@ai-sdk/*` package the OpenCode provider should use.
- Fixes the case where a relay exposes both OpenAI-compatible and Gemini-native endpoints: today a Gemini import lands as `@ai-sdk/openai-compatible`, and the user has to open the provider config and hand-edit `npm` to `@ai-sdk/google` before chat works.
- Default stays `@ai-sdk/openai-compatible`, so existing deeplinks are unaffected.

## What changed

- `DeepLinkImportRequest` gains `npm: Option<String>` (camelCase `npm` in the URL).
- `parse_deeplink_url` reads `npm` from the query string for the `provider` resource.
- `merge_additive_config` (OpenCode / OpenClaw JSON config pass-through) now carries `npm` from the Base64 config when the URL param is absent.
- `build_opencode_settings` uses the override and falls back to `@ai-sdk/openai-compatible`.
- Docs: all three locales of `5.3-deeplink.md` get a new row for `npm`.

## Context

Example deeplink from a relay that routes Gemini under `/v1beta`:

```
ccswitch://v1/import?resource=provider&app=opencode&name=Relay
  &endpoint=https%3A%2F%2Frelay.example.com%2Fv1beta
  &apiKey=sk-...
  &model=gemini-3-flash
  &npm=%40ai-sdk%2Fgoogle
```

Without this change the imported provider is usable for discovery but fails on the first chat because the `openai-compatible` SDK can't speak Gemini-native. With the change, one click lands a working config.

## Tests

- `test_parse_opencode_deeplink_with_npm_param` — URL param survives parsing.
- `test_build_opencode_provider_defaults_npm_when_absent` — default unchanged.
- `test_build_opencode_provider_honors_npm_override` — override wins.

All 24 `deeplink::tests::*` pass locally (`cargo test deeplink`). `cargo fmt --check` and `cargo clippy -- -D warnings` clean.

## Test plan

- [x] `cargo fmt --check` passes
- [x] `cargo clippy --manifest-path src-tauri/Cargo.toml -- -D warnings` passes
- [x] `cargo test --manifest-path src-tauri/Cargo.toml` passes (24 deeplink tests + 2 integration tests)
- [ ] Manual: trigger an OpenCode deeplink with `npm=@ai-sdk/google` and confirm the provider config in CC Switch shows the overridden package

### DIFF
--- a/docs/user-manual/en/5-faq/5.3-deeplink.md
+++ b/docs/user-manual/en/5-faq/5.3-deeplink.md
@@ -53,6 +53,7 @@ ccswitch://v1/import?resource={type}&app={app}&name={name}&...
 | `haikuModel` | No | Haiku model (Claude only) |
 | `sonnetModel` | No | Sonnet model (Claude only) |
 | `opusModel` | No | Opus model (Claude only) |
+| `npm` | No | npm package for OpenCode provider (default `@ai-sdk/openai-compatible`; e.g. `@ai-sdk/google` for Gemini-native routing) |
 | `notes` | No | Notes |
 | `icon` | No | Icon |
 | `config` | No | Base64-encoded configuration content |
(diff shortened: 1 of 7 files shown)


farion1231:
@codex review

PR #2211, opened 2026-04-21
fix(skills): prevent duplicate imports when import button is double-clicked

### LINKED ISSUE #2139: Bug: Skills Import Count Multiplied When Button Clicked Multiple Times
### Self Checks / 自检

### CC Switch Version / 版本号

v3.13.0

### Operating System / 操作系统

Windows

### Related App / 涉及应用

Claude Code

### Steps to Reproduce / 重现步骤

  Description

  When opening the Skills management panel and clicking the "Import Existing" button, if the user quickly clicks the "Confirm" button twice (or multiple times when the interface lags due to importing many
  skills), the displayed count of "Installed Claude:" becomes incorrect - it shows a multiple of the actual count based on the number of clicks.

  Root Cause

  1. The import confirmation button in ImportSkillsDialog (line 841) is not disabled during mutation execution, allowing users to click multiple times
  2. The cache update in useImportSkillsFromApps hook (lines 207-215) directly appends to the array without deduplication logic, causing the same skills to be added multiple times to the cache

  Steps to Reproduce

  1. Open the Skills management panel
  2. Click "Import Existing" button
  3. In the import dialog, quickly click the "Import" button multiple times (or click repeatedly when importing many skills causes UI lag)
  4. After import completes, observe the "Installed Claude:" count

  

### Expected Behavior / 期望行为

Expected Behavior

  - Import button should be disabled during the import process
  - Skills count should display the correct number, not multiplied by click count

 

### Actual Behavior / 实际行为

 Actual Behavior

  - Button remains clickable during import
  - Count shows a multiplied number (e.g., if clicked 3 times and imported 5 skills, count shows 15)

  Affected Files

  - src/components/skills/UnifiedSkillsPanel.tsx - ImportSkillsDialog component
  - src/hooks/useSkills.ts - useImportSkillsFromApps hook

 

### Additional Context / 补充信息

 Suggested Fix

  1. Pass importMutation.isPending state to ImportSkillsDialog and disable the button when importing
  2. Add deduplication logic to the cache update in useImportSkillsFromApps.onSuccess

  Environment

  - CC Switch version: 3.13.0
  - Platform: All (Windows/macOS/Linux)

TuYv:
## Summary / 概述

When a user taps the "Import" confirm button multiple times in the Skills import dialog — either deliberately or because the UI lags while importing many skills — the installed-skills count balloons to a multiple of the real count.

Two related defects combined to produce this:

1. **The confirm button stayed clickable during a pending mutation.** It only checked `selected.size === 0`, so every extra tap fired another `importFromApps` mutation.
2. **`useImportSkillsFromApps` appended results to the installed cache without deduping.** Re-firing the mutation stacked the same skills into the list.

### Changes

- `ImportSkillsDialog` gains an `isImporting` prop; the Import and Cancel buttons are disabled while the mutation is pending. This follows the `isRestoring` / `isDeleting` pattern already used by `RestoreSkillsDialog` in the same file.
- Cache-merge logic is extracted as a pure function `mergeImportedSkills(existing, imported)` in a new `src/hooks/useSkills.helpers.ts` so it can be unit-tested without React / React Query setup. It dedupes by `InstalledSkill.id` (newer entries replace stale ones) and short-circuits to the original reference when `imported` is empty so React Query does not notify subscribers about a no-op update.
- Unit tests cover empty cache, re-import of the same id, stale replacement, empty-payload reference stability, and unrelated appends.

### Scope

Only `useImportSkillsFromApps` is touched here. `useInstallSkill` and `useInstallSkillsFromZip` also perform naive cache appends but are not reachable from a dialog double-click in the same way, and changing them is out of scope for this fix — happy to open a follow-up if desired.

## Related Issue / 关联 Issue

Fixes #2139

## Screenshots / 截图

N/A — the behaviour change is functional. Reproduction steps and root cause are documented in the linked issue; after the fix the confirm button goes disabled on first click and the installed count matches the imported count regardless of how many times the button is tapped.

## Checklist / 检查清单

- [x] \`pnpm typecheck\` passes
- [x] \`pnpm format:check\` passes for the files touched in this PR (three pre-existing warnings in \`src/components/providers/forms/hooks/*\` are unchanged and not in scope)
- [ ] \`cargo clippy\` — N/A, no Rust changes
- [x] i18n — no new user-facing strings

### DIFF
--- a/src/components/skills/UnifiedSkillsPanel.tsx
+++ b/src/components/skills/UnifiedSkillsPanel.tsx
@@ -447,6 +447,7 @@ const UnifiedSkillsPanel = React.forwardRef<
       {importDialogOpen && unmanagedSkills && (
         <ImportSkillsDialog
           skills={unmanagedSkills}
+          isImporting={importMutation.isPending}
           onImport={handleImport}
           onClose={() => setImportDialogOpen(false)}
         />
@@ -593,6 +594,7 @@ interface ImportSkillsDialogProps {
     foundIn: string[];
     path: string;
   }>;
+  isImporting: boolean;
   onImport: (imports: ImportSkillSelection[]) => void;
   onClose: () => void;
 }
@@ -717,6 +719,7 @@ const RestoreSkillsDialog: React.FC<RestoreSkillsDialogProps> = ({
 
 const ImportSkillsDialog: React.FC<ImportSkillsDialogProps> = ({
   skills,
+  isImporting,
   onImport,
   onClose,
 }) => {
@@ -835,10 +838,13 @@ const ImportSkillsDialog: React.FC<ImportSkillsDialogProps> = ({
           </div>
 
           <div className="flex justify-end gap-3">
-            <Button variant="outline" onClick={onClose}>
+            <Button variant="outline" onClick={onClose} disabled={isImporting}>
               {t("common.cancel")}
             </Button>
-            <Button onClick={handleImport} disabled={selected.size === 0}>
+            <Button
+              onClick={handleImport}
+              disabled={selected.size === 0 || isImporting}
+            >
               {t("skills.importSelected", { count: selected.size })}
             </Button>
           </div>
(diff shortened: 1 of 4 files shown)


farion1231 (approved):

PR #2218, opened 2026-04-21
[codex] Stabilize Codex OAuth cache routing

### LINKED ISSUE #2217: Stabilize Codex OAuth prompt cache routing
## Problem

Claude Code requests routed through the Codex OAuth provider can miss OpenAI/Codex prompt cache more often than expected. The proxy already sends a `prompt_cache_key`, but it historically falls back to provider-level identity and does not send the Codex-like session headers that help the backend route related turns consistently.

Real traffic also exposed two runtime blockers while testing this path:

- `rustls` can panic when both crypto providers are available and no process default is installed.
- Codex OAuth forces Responses API streaming, but the Claude transform handler can still try to parse the upstream SSE body as JSON when the original client request was non-streaming.

## Expected behavior

For Codex OAuth-backed Claude proxy requests:

- Prefer stable client-provided session identity for prompt cache routing.
- Avoid using generated per-request UUIDs as upstream cache keys.
- Send Codex-like session headers when a real client session id exists.
- Handle upstream SSE Responses payloads even when the original Claude request did not set `stream: true`.
- Avoid runtime TLS provider panics.

## Observed impact from local verification

On local proxy logs around the fix window, Codex provider hit rate improved in the sampled workload from roughly 17% before to 86% after, and cache-read/input ratio improved from roughly 2.3% to 18.2%. This is not a strict A/B test, but it confirms the fixed path is actively producing `cache_read_tokens`.

majiayu000:
## What Changed

- Use a client-provided session id as the Codex OAuth `prompt_cache_key` when routing Claude requests through OpenAI Responses.
- Keep explicit `promptCacheKey` overrides authoritative, and fall back to provider id when the proxy only generated a per-request UUID.
- Add Codex-like session headers (`session_id`, `x-client-request-id`, `x-codex-window-id`) for Codex OAuth requests when a real client session exists.
- Handle upstream Responses SSE even if the original Claude request was non-streaming.
- Install the rustls ring crypto provider during app setup to avoid runtime TLS provider panics in the proxy path.

## Why

Claude-to-Codex OAuth proxy traffic was not consistently routing related turns to the same prompt-cache identity. It could still hit cache through backend prefix caching, but local traffic showed unstable hit rates. Official Codex uses conversation identity and session headers for Responses API routing, so this PR aligns CC Switch's Codex OAuth proxy path more closely with that behavior.

The same local validation uncovered two blockers: rustls could panic without an explicit crypto provider, and forced-streaming Codex OAuth responses could be parsed as JSON when the incoming Claude request was non-streaming.

## Impact

For Codex OAuth-backed Claude providers, related requests with a real client session id should be more likely to reuse prompt-cache prefixes. Generated proxy UUIDs are deliberately excluded because they would fragment cache identity.

Local observed cache metrics around this change improved in the sampled workload from roughly 17% hit rate to 86%, with cache-read/input ratio increasing from roughly 2.3% to 18.2%. This is not a strict A/B test, but confirms the fixed path emits and receives active cache signals.

## Validation

- `cargo test codex_oauth`
- Local Tauri dev app started successfully
- `curl http://127.0.0.1:15721/health`
- Local proxy logs showed `cache_read_tokens` after restart

## Remaining Risk

- This still does not implement `x-codex-turn-state` capture/replay.
- Full `cargo test` was not run cleanly while the local CC Switch dev app owned port `15721`; the relevant filtered test suite passed.

Fixes #2217

### DIFF
--- a/src-tauri/src/lib.rs
+++ b/src-tauri/src/lib.rs
@@ -273,6 +273,8 @@ pub fn run() {
         .plugin(tauri_plugin_opener::init())
         .plugin(tauri_plugin_store::Builder::new().build())
         .setup(|app| {
+            let _ = rustls::crypto::ring::default_provider().install_default();
+
             // 预先刷新 Store 覆盖配置，确保后续路径读取正确（日志/数据库等）
             app_store::refresh_app_config_dir_override(app.handle());
             panic_hook::init_app_config_dir(crate::config::get_app_config_dir());
--- a/src-tauri/src/provider.rs
+++ b/src-tauri/src/provider.rs
@@ -258,7 +258,8 @@ pub struct ProviderMeta {
     pub is_full_url: Option<bool>,
     /// Prompt cache key for OpenAI Responses-compatible endpoints.
     /// When set, injected into converted Responses requests to improve cache hit rate.
-    /// If not set, provider ID is used automatically during Claude -> Responses conversion.
+    /// If not set, Codex OAuth uses the current session ID; other Claude -> Responses
+    /// conversions fall back to provider ID.
     #[serde(rename = "promptCacheKey", skip_serializing_if = "Option::is_none")]
     pub prompt_cache_key: Option<String>,
     /// 累加模式应用中，该 provider 是否已写入 live config。
--- a/src-tauri/src/proxy/forwarder.rs
+++ b/src-tauri/src/proxy/forwarder.rs
@@ -55,6 +55,8 @@ pub struct RequestForwarder {
     current_provider_id_at_start: String,
     /// 代理会话 ID（用于 Gemini Native shadow replay）
     session_id: String,
+    /// Session ID 是否由客户端提供；生成值不能作为上游缓存身份。
+    session_client_provided: bool,
     /// 整流器配置
     rectifier_config: RectifierConfig,
     /// 优化器配置
@@ -77,6 +79,7 @@ impl RequestForwarder {
         app_handle: Option<tauri::AppHandle>,
         current_provider_id_at_start: String,
         session_id: String,
+        session_client_provided: bool,
         _streaming_first_byte_timeout: u64,
         _streaming_idle_timeout: u64,
         rectifier_config: RectifierConfig,
@@ -92,6 +95,7 @@ impl RequestForwarder {
             app_handle,
             current_provider_id_at_start,
             session_id,
+            session_client_provided,
             rectifier_config,
             optimizer_config,
             copilot_optimizer_config,
@@ -966,7 +970,8 @@ impl RequestForwarder {
                     mapped_body,
                     provider,
                     api_format,
-                    Some(&self.session_id),
+                    self.session_client_provided
+                        .then_some(self.session_id.as_str()),
                     Some(self.gemini_shadow.as_ref()),
                 )?
             } else {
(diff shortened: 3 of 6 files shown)


majiayu000:
image

image
这是改动前后的缓存使用对比 
顺便这次修复了一个 sse的422的bug

farion1231:
Hi @majiayu000 — heads up that I force-pushed this branch to rebase onto `main` and resolve a conflict with #2235, which landed separately. Quick summary of what changed and why; happy to revert any of it if you disagree.

**Context**

- #2235 already fixes the same 422 you were targeting (non-streaming Claude request → forced Codex OAuth Responses SSE being parsed as JSON), but via a different strategy: it adds `should_use_claude_transform_streaming` and routes the response through the existing streaming transformer so the client gets `text/event-stream`.
- That approach has a useful extra property we want to keep: the generic `response.is_sse()` guard also protects any other provider that unexpectedly returns SSE, not just `codex_oauth`.
- Your aggregator (`responses_sse_to_response_value`) is the more correct fix for the Anthropic non-streaming contract — non-streaming clients should still receive JSON, not SSE — but it shouldn't replace #2235's generic SSE guard.

**What the rebase does**

Dropped (superseded by #2235):
- `should_use_stream_transform` (functionally identical to `should_use_claude_transform_streaming` on main)
- The `if is_stream && upstream_is_stream` branch rework — main's single `if use_streaming` branch is kept

Kept and tightened:
- `responses_sse_to_response_value` — but triggered by a **narrow override** (`!is_stream && is_codex_oauth && api_format == "openai_responses"`), not by a broad `api_format == "openai_responses"` check. This prevents accidental regressions for non-Codex Responses providers (Azure OpenAI Responses, etc.) whose upstreams genuinely return non-stream JSON.
- I rewrote the aggregator to use `proxy::sse::take_sse_block` / `strip_sse_field` instead of a hand-rolled `split("\n\n")`. The original would have silently produced `TransformError` against any upstream that uses the standard `\r\n\r\n` SSE delimiter (i.e. most real HTTPS upstreams); `take_sse_block` supports both `\n\n` and `\r\n\r\n` and is already the convention elsewhere in `proxy::`.
- Added regression tests for CRLF delimiters, `response.failed`, and the missing `response.completed` defensive branch.

Fully kept (the core contribution of this PR):
- `prompt_cache_key` sourced from the client session id for `codex_oauth`
- `session_client_provided` guard so generated proxy UUIDs never leak as cache identity
- `session_id` / `x-client-request-id` / `x-codex-window-id` headers
- `rustls::crypto::ring::default_provider().install_default()`
- `provider.rs` comment update

**Final branch layout**

Two commits on top of `main`:
1. `472d10e5 Stabilize Codex OAuth cache routing` — your original work, authorship preserved
2. `374c5c6d feat(proxy): aggregate forced Codex OAuth SSE into JSON for non-streaming clients` — the narrow override layered on top of #2235

CI:
- `cargo fmt` — clean
- `cargo test --lib` — 975 passed
- `cargo clippy --lib -- -D warnings` — no new warnings from this PR (there is one pre-existing `manual_ok_err` warning on `session_manager/providers/gemini.rs:38` that exists on `main` unrelated to this change)

Let me know if you'd prefer a different shape for any of this, or if you want me to move the rebase commit back under your authorship.

farion1231 (approved):
感谢您的贡献！

PR #2227, opened 2026-04-22
fix: use TOML parser instead of regex for Codex model extraction (#2222)

nmsn:
Fixes #2222

Regex only matched `model=...` on first line, breaking configs with comments or multiline TOML. Use the existing toml crate to parse properly.


farion1231 (approved):

PR #2360, opened 2026-04-26
fix(env-checker): stop false-positives from inline-prefix vars in shell functions

### LINKED ISSUE #516: 环境变量冲突 检测有逻辑问题
### 现象 
**每次启动 都会弹窗提示 有检测到系统环境变量冲突 发现2个环境变量可能会覆盖您的配置**

Image

### 问题描述

软件运行时检测到 3 个环境变量冲突，但这些变量实际上是自己写的本地 shell 函数内部的局部变量，不应该被标记为冲突：

- `local v_gemini_old=xxx` (来源: `~/.zshrc:264`)
- `local v_gemini_remote=xxx` (来源: `~/.zshrc:275`)

### 问题原因

`src-tauri/src/services/env_checker.rs` 中的 `check_shell_configs` 函数匹配逻辑过于宽松（第 97-122 行）：

```rust
// 只要行包含 = 且变量名包含关键词就会被匹配
if trimmed.starts_with("export ")
    || (!trimmed.starts_with('#') && trimmed.contains('='))
{
    // ...
    if keywords.iter().any(|k| var_name.to_uppercase().contains(k)) {
        // 标记为冲突
    }
}
```
当前使用 contains 模糊匹配，导致任何变量名中包含 GEMINI、ANTHROPIC、OPENAI 的变量都会被误报为冲突。

## 建议修复，因为误判

因为我的`.zshrc` 中包含 自己定义的其他变量，例如 `local v_gemini_remote` 其实只是 本地 func 的内部变量而已

---

## 下面是AI的分析建议，不太懂内部代码，所以仅供参考

应该改为精确匹配官方环境变量名，而不是模糊包含关键词：

```rust
fn get_keywords_for_app(app: &str) -> Vec<&str> {
    match app.to_lowercase().as_str() {
        "claude" => vec!["ANTHROPIC_API_KEY"],
        "codex" => vec!["OPENAI_API_KEY"],
        "gemini" => vec!["GEMINI_API_KEY", "GOOGLE_API_KEY"],
        _ => vec![],
    }
}
```

然后用精确匹配或前缀匹配代替 contains：

```rust
// 精确匹配
if keywords.iter().any(|k| var_name.to_uppercase() == *k)

// 或前缀匹配
if keywords.iter().any(|k| var_name.to_uppercase().starts_with(k))
```

marovole:
## Problem

The env-conflict scanner in `src-tauri/src/services/env_checker.rs` produces aggressive false positives. The detector uses substring matching (`name.to_uppercase().contains(keyword)`) and treats every non-comment `VAR=value` line as a global export, which incorrectly flags two very common patterns:

**1. Variable names that merely *contain* a keyword**

```bash
export LOG_ANTHROPIC_TIMING=1   # not the API key
local v_gemini_remote=...       # local helper
```

**2. Inline command-prefix assignments inside shell functions** — the user-reported pattern from #516:

```bash
cc() {
  ANTHROPIC_AUTH_TOKEN="..." \
  ANTHROPIC_BASE_URL="..." \
  claude --dangerously-skip-permissions "$@"
}
```

POSIX semantics make those assignments visible **only** to the spawned `claude` child. They don't pollute the calling shell, don't leak into siblings, and cannot override what cc-switch configures. A user with five such helper functions sees the conflict dialog report **26 phantom conflicts on every launch**.

Verified empirically — the user's clean login shell shows zero leaked vars:

```bash
$ env -i HOME="$HOME" PATH="/usr/bin:/bin" zsh -i -c \
    'env | grep -iE "anthropic|claude_code"'
# (empty)
```

Closes #516.

## Fix

`src-tauri/src/services/env_checker.rs`:

- **Exact (case-insensitive) name matching** against a whitelist of variables that actually drive each managed CLI (Anthropic API/auth token/base URL/model selectors for `claude`, `OPENAI_*` for `codex`, `GEMINI_API_KEY` / `GOOGLE_API_KEY` / `GOOGLE_GEMINI_API_KEY` for `gemini`). No more `contains`.
- **Brace-depth tracking** while scanning rc files: any line that lies inside a `{ ... }` block is skipped, which is exactly where inline command prefixes live.
- **`export`-only matching** at the top level. A bare `VAR=value` outside `export` is either an inline prefix on a continued line or an unexported shell var; neither leaks into descendant processes.

A naive char-counter for `{`/`}` is good enough for typical rc files; braces inside strings/comments cause harmless over-skipping rather than false positives.

## Tests

Four new regression tests cover the exact patterns from the bug report, plus the property we still want to preserve (real exports must still be flagged):

- `test_skips_inline_command_prefix_inside_function` — the user-reported pattern
- `test_flags_top_level_export` — must still flag genuine `export ANTHROPIC_API_KEY=...`
- `test_does_not_flag_unrelated_var_with_keyword_substring` — `LOG_ANTHROPIC_REQUESTS`, `MY_OPENAI_PROXY_LOG`, `local v_gemini_remote`
- `test_does_not_flag_bare_top_level_assignment` — `VAR=...` without `export`

Tests share `$HOME` so they're annotated with `#[serial_test::serial]` (the crate is already in dev-dependencies).

```text
running 5 tests
test services::env_checker::tests::test_get_keywords_returns_exact_var_names ... ok
test services::env_checker::tests::test_does_not_flag_unrelated_var_with_keyword_substring ... ok
test services::env_checker::tests::test_flags_top_level_export ... ok
test services::env_checker::tests::test_does_not_flag_bare_top_level_assignment ... ok
test services::env_checker::tests::test_skips_inline_command_prefix_inside_function ... ok

test result: ok. 5 passed; 0 failed
```

## Behavior change worth a second look

Previously a bare top-level `ANTHROPIC_API_KEY="..."` (no `export`) was flagged. Now it isn't, because POSIX shells don't propagate non-exported vars to children. If you'd rather keep flagging it for the rare `set -a` (allexport) case, happy to gate that on detecting `set -a` earlier in the file — let me know.

### DIFF
--- a/src-tauri/src/services/env_checker.rs
+++ b/src-tauri/src/services/env_checker.rs
@@ -31,12 +31,30 @@ pub fn check_env_conflicts(app: &str) -> Result<Vec<EnvConflict>, String> {
     Ok(conflicts)
 }
 
-/// Get relevant keywords for each app
-fn get_keywords_for_app(app: &str) -> Vec<&str> {
+/// Get the list of environment variable names that, when present in the
+/// user's shell or process environment, can override what cc-switch
+/// configures. Names must match exactly (case-insensitive); substring
+/// matching produced false positives for unrelated vars whose names merely
+/// contained "ANTHROPIC", "OPENAI", or "GEMINI" (e.g. local helper vars
+/// such as `LOG_ANTHROPIC_*`, `v_gemini_remote`).
+fn get_keywords_for_app(app: &str) -> Vec<&'static str> {
     match app.to_lowercase().as_str() {
-        "claude" => vec!["ANTHROPIC"],
-        "codex" => vec!["OPENAI"],
-        "gemini" => vec!["GEMINI", "GOOGLE_GEMINI"],
+        "claude" => vec![
+            "ANTHROPIC_API_KEY",
+            "ANTHROPIC_AUTH_TOKEN",
+            "ANTHROPIC_BASE_URL",
+            "ANTHROPIC_MODEL",
+            "ANTHROPIC_DEFAULT_OPUS_MODEL",
+            "ANTHROPIC_DEFAULT_SONNET_MODEL",
+            "ANTHROPIC_DEFAULT_HAIKU_MODEL",
+            "ANTHROPIC_REASONING_MODEL",
+        ],
+        "codex" => vec!["OPENAI_API_KEY", "OPENAI_BASE_URL"],
+        "gemini" => vec![
+            "GEMINI_API_KEY",
+            "GOOGLE_API_KEY",
+            "GOOGLE_GEMINI_API_KEY",
+        ],
         _ => vec![],
     }
 }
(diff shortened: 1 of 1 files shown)


farion1231:
@codex review

farion1231:
@claude review

PR #2361, opened 2026-04-26
fix(auto-launch): clean up legacy macOS login items left by pre-#462 builds

marovole:
## Problem

#462 fixed `auto_launch.rs` to register the `.app` bundle path on macOS so login items don't open via Terminal.app. **However, users who had auto-launch enabled before that fix still have a broken login item** that points at the inner Unix executable:

```
/Applications/CC Switch.app/Contents/MacOS/cc-switch
```

That path makes LaunchServices treat the entry as a *Unix Executable File*, so every login spawns a Terminal window full of cc-switch's startup logs. The legacy entry also uses the old display name `cc-switch` (lowercase, hyphen) while current code uses `CC Switch` — meaning:

- `is_auto_launch_enabled()` returns `false` (the in-app toggle shows "off"), and
- `disable_auto_launch()` can't see the legacy item to remove it,

so the broken state is **invisible from inside the app** and persists indefinitely. Reproduced on cc-switch 3.14.1 upgraded from a pre-#462 install:

```bash
$ osascript -e 'tell application "System Events" to get the path of login item "cc-switch"'
/Applications/CC Switch.app/Contents/MacOS/cc-switch
$ osascript -e 'tell application "System Events" to get the kind of login item "cc-switch"'
Unix Executable File
```

## Fix

Add a one-shot self-heal called from the Tauri `setup` hook:

`src-tauri/src/auto_launch.rs` — new `migrate_legacy_macos_login_items()`:
1. Runs an AppleScript that looks for login items named **`cc-switch`** *or* **`CC Switch`** whose `path` contains `/Contents/MacOS/`. That path shape is unambiguously the buggy one — no `.app` bundle ever has its bundle path inside `Contents/MacOS/`.
2. Deletes the matching entries.
3. If anything was removed, calls `enable_auto_launch()` to re-register with the proper `.app` bundle path so the user keeps their original auto-launch preference.

The AppleScript uses `try` blocks so a missing item is not an error. On non-macOS targets the function is a no-op.

`src-tauri/src/lib.rs` — invoke once from `setup`, right after the log plugin is configured so the migration log line is captured.

The migration only fires when a buggy entry is actually found (returns early otherwise), and it never touches login items that don't match the cc-switch display names.

## Tests

```text
running 5 tests
test auto_launch::tests::test_migrate_legacy_login_items_compiles_on_macos ... ok
test auto_launch::tests::test_get_macos_app_bundle_path_not_in_bundle ... ok
test auto_launch::tests::test_get_macos_app_bundle_path_dev_build ... ok
test auto_launch::tests::test_get_macos_app_bundle_path_valid ... ok
test auto_launch::tests::test_get_macos_app_bundle_path_with_spaces ... ok

test result: ok. 5 passed; 0 failed
```

The AppleScript path itself is intentionally not exercised in `cargo test` because it would mutate the developer's actual login items; the test ensures the function stays compileable and reachable.

## Manual verification plan for reviewer

On a macOS host that has the buggy state (or simulated via `osascript -e 'tell application "System Events" to make login item at end with properties {name:"cc-switch", path:"/Applications/CC Switch.app/Contents/MacOS/cc-switch", hidden:false}'`):

1. Launch the dev build → log shows `✓ Removed legacy macOS login item …` and `✓ Re-registered auto-launch using the .app bundle path`.
2. Verify `osascript -e 'tell application "System Events" to get the path of every login item whose name is "CC Switch"'` now returns the `.app` bundle path.
3. Re-launch → migration is silent (no false repeat).

## Note on UX choice

The migration runs silently rather than prompting the user — the buggy state is unambiguously wrong and the heal preserves the user's intent (`enable_auto_launch` after delete). Happy to swap in a one-time toast/notification if you'd prefer something more visible.

### DIFF
--- a/src-tauri/src/auto_launch.rs
+++ b/src-tauri/src/auto_launch.rs
@@ -68,6 +68,88 @@ pub fn is_auto_launch_enabled() -> Result<bool, AppError> {
         .map_err(|e| AppError::Message(format!("检查开机自启状态失败: {e}")))
 }
 
+/// 迁移由旧版本写入的 macOS 登录项。
+///
+/// PR #462 之前的构建版本将 `.app/Contents/MacOS/<exe>` 这一可执行文件路径
+/// 直接注册为登录项；这条路径在 LaunchServices 看来是 "Unix Executable
+/// File"，被打开时会附带一个终端窗口（用户能看到 cc-switch 的全部启动日志）。
+/// 老版本注册时使用的显示名也是 `cc-switch`（小写连字符），而当前代码使用
+/// `CC Switch`，因此 [`is_auto_launch_enabled`] 永远查不到这一条旧记录，
+/// 它会一直驻留在系统登录项里。
+///
+/// 此函数在启动期调用一次：
+/// 1. 通过 AppleScript 查找名为 `cc-switch` 或 `CC Switch`、且路径包含
+///    `/Contents/MacOS/` 的登录项 —— 这种路径形态一定是错的；
+/// 2. 删除命中的登录项；
+/// 3. 若确实删除了任何条目，立即用正确的 `.app` bundle 路径重新启用开机
+///    自启，保留用户原本的意图。
+///
+/// 在非 macOS 平台上是空操作。
+#[cfg(target_os = "macos")]
+pub fn migrate_legacy_macos_login_items() {
+    use std::process::Command;
+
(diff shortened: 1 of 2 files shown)


farion1231:
@codex review

farion1231:
@claude review

PR #2363, opened 2026-04-26
fix(terminal): silence ShellExecute dialog when launching Windows Terminal

JWM0203:
## Summary

Fixes a long-standing bug where picking **Windows Terminal** as the preferred terminal causes a modal dialog `Windows 找不到文件 'wt'` to pop up on every launch, even though the existing fallback to `cmd` is supposed to handle the failure silently.

### Root cause

`launch_windows_terminal` and the Windows branch of `launch_terminal_running` both invoke:

```rust
run_windows_start_command(&["wt", "cmd", "/K", &bat_path], "Windows Terminal")
```

which expands to `cmd /C start wt cmd /K <bat>`. When `start` cannot resolve `wt`, it routes through `ShellExecute`, which **first pops a modal "Windows cannot find" dialog and only then returns failure**. By the time the existing `result.is_err()` cmd-fallback fires, the user has already seen the popup.

This commonly hits users who run cc-switch elevated: an admin-elevated process inherits the SYSTEM PATH, which lacks the per-user `%LOCALAPPDATA%\Microsoft\WindowsApps` directory where `wt.exe` lives — so even though `wt` is installed, `start wt` can't find it.

### Fix

Resolve a concrete path to `wt.exe` *before* invoking `start`:

1. Try `where wt` (covers the normal case).
2. Fall back to an explicit probe of `%LOCALAPPDATA%\Microsoft\WindowsApps\wt.exe` (covers the elevated-launch scenario).
3. If neither resolves, return `Err(...)` so the existing cmd fallback path kicks in cleanly — no modal dialog.

When `wt.exe` is found, pass the **full resolved path** to `start` (with an empty title argument: `start "" "<full path>" cmd /K <bat>`). `ShellExecute` against a concrete path doesn't show the "cannot find" dialog.

The change is minimal and contained: one new helper `resolve_windows_terminal_path()` plus a two-line replacement at each of the two existing Windows launch sites.

## Test plan

- [x] Built locally with `pnpm tauri build --no-bundle` on Windows 11 Pro 23H2 (MSVC 14.44, rustc 1.95.0)
- [x] Reproduced the bug with the released v3.14.1 binary: dialog shows on every Provider launch when wt is the preferred terminal
- [x] Verified the patched binary launches Windows Terminal cleanly with no dialog under both normal and elevated launch contexts
- [x] Verified `cmd` and `PowerShell` selections continue to work unchanged
- [x] No new warnings introduced; existing warnings in this file are pre-existing

### DIFF
--- a/src-tauri/src/commands/misc.rs
+++ b/src-tauri/src/commands/misc.rs
@@ -1180,6 +1180,49 @@ fn which_command(cmd: &str) -> bool {
         .unwrap_or(false)
 }
 
+/// Resolve the full path to `wt.exe` (Windows Terminal). Returns None if not found.
+///
+/// We must resolve a concrete path *before* invoking `cmd /C start wt …`. If
+/// `start` can't locate `wt` it routes through ShellExecute, which pops a modal
+/// "Windows 找不到文件 'wt'" dialog and only *then* returns failure — the existing
+/// fallback to cmd happens after the dialog is already on screen.
+///
+/// `where wt` covers the normal case. The explicit `LOCALAPPDATA\Microsoft\
+/// WindowsApps\wt.exe` probe covers the common elevated-launch scenario, where
+/// the SYSTEM PATH inherited by an admin-elevated cc-switch lacks the user's
+/// per-user WindowsApps directory.
+#[cfg(target_os = "windows")]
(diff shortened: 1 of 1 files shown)


farion1231:
@codex review

farion1231:
@claude review

PR #2371, opened 2026-04-26
Fix duplicate Codex session usage in aggregate stats

xandersbell:
## Summary

This PR applies a minimal, low-intrusion fix for duplicate Codex usage appearing in aggregate statistics.

I encountered an issue where the macOS app usage dashboard showed a pseudo provider named `Codex (Session)` alongside the actual manually imported Codex provider. In many cases, both rows had the same request/token/cost values, while `Codex (Session)` had no latency / first-token timing data.

After tracing the code, the likely cause is that the same Codex CLI request can be recorded from two sources:

- the real proxy request log, under the actual provider id
- the Codex session log import, under `_codex_session`

Because Codex does not currently have a shared stable request id between these two sources, both records can be counted in aggregate stats. Codex session usage is also synced periodically, so the matching window needs to tolerate some timestamp skew between proxy logging and session-log import.

## What changed

To respect the existing interface and minimize the patch size, this PR does not change:

- public Tauri command interfaces
- frontend data types
- database schema
- session import behavior
- raw request log visibility

Instead, it only adjusts aggregate usage queries to exclude a `codex_session` row when a matching real Codex proxy row exists within a small time window.

The duplicate check currently requires:

- `app_type = codex`
- session row provider id is `_codex_session`
- matching real proxy row exists
- same status code
- same model
- same input/output/cache token counts
- timestamps within 90 seconds

Raw request logs still show both records for inspection.

## Why this is intentionally limited

This is a patch-level fix, not a redesign of the usage data model.

A more fundamental fix would likely require the project owner to decide how Codex proxy logs and Codex session logs should share identity, deduplicate at import time, or model multiple usage sources consistently through rollups. That may involve schema changes or broader product decisions, so I kept this PR narrowly scoped.

Known limitation: historical data already rolled up into `usage_daily_rollups` cannot be precisely deduplicated by this patch because the rollup table does not retain enough source/detail fields.

## Tests

Ran:

```bash
cargo fmt --manifest-path src-tauri/Cargo.toml --check
cargo test --manifest-path src-tauri/Cargo.toml services::usage_stats --lib
```

Result:

```text
11 passed; 0 failed
```

Thanks for maintaining the project.

### DIFF
--- a/src-tauri/src/services/usage_stats.rs
+++ b/src-tauri/src/services/usage_stats.rs
@@ -130,6 +130,55 @@ fn provider_name_coalesce(log_alias: &str, provider_alias: &str) -> String {
     )
 }
 
+const CODEX_SESSION_DEDUP_WINDOW_SECS: i64 = 90;
+
+fn sql_col(alias: &str, column: &str) -> String {
+    if alias.is_empty() {
+        column.to_string()
+    } else {
+        format!("{alias}.{column}")
+    }
+}
+
+/// Exclude Codex session rows from aggregate statistics when the same request
+/// is already present as a real proxy log. Raw request logs still show both
+/// sources so the original data remains inspectable.
+fn codex_session_duplicate_exclusion(log_alias: &str) -> String {
+    let app_type = sql_col(log_alias, "app_type");
+    let provider_id = sql_col(log_alias, "provider_id");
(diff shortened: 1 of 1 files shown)


farion1231:
@codex review

farion1231:
@claude review

PR #2432, opened 2026-04-28
feat(terminal): add Warp as a preferred terminal option

alexey-a-abramov:
## Summary

Adds Warp Terminal as a selectable preferred-terminal target on macOS and Linux, alongside Terminal.app, iTerm2, Alacritty, Kitty, Ghostty, WezTerm and Kaku.

Selecting **Warp** in *Settings → Preferred Terminal* and clicking the terminal button (or *Resume in terminal* in the Session Manager) will activate Warp, open a new tab, and run the cc-switch launcher command in it.

## Why a custom launcher

Warp does not implement Terminal.app's `do script` AppleScript verb, so the standard cc-switch launcher pattern (write a temp `.sh`, then `tell application "X" to do script`) does not work as-is. The launcher therefore:

1. Activates Warp via AppleScript.
2. Opens a new tab with Cmd+T (System Events).
3. Sends keystrokes via System Events to type `bash '/tmp/cc_switch_launcher_<pid>.sh'` and Enter.

This requires the user to grant cc-switch the **Accessibility** permission (System Settings → Privacy & Security → Accessibility) the first time. If Warp fails to launch for any reason, cc-switch falls back to Terminal.app like it already does for every other preferred terminal.

On Linux, `warp-terminal` is added to the default terminal list with the `-e` argument convention used by the other CLI terminals.

## Files touched

- `src/components/settings/TerminalSettings.tsx` — `warp` option in the macOS dropdown, `warp-terminal` option in the Linux dropdown.
- `src/i18n/locales/{en,zh,ja}.json` — *Warp* label under `settings.terminal.options.macos.warp` and `...linux.warp`.
- `src-tauri/src/commands/misc.rs` — new `launch_macos_warp` helper, `warp` branches added to `launch_macos_terminal` and `launch_terminal_running`, `warp-terminal` added to both Linux fallback lists.
- `src-tauri/src/session_manager/terminal/mod.rs` — `launch_warp` function and `warp` branch in the `launch_terminal` dispatcher used by the *Resume in terminal* flow.

## Verification

- `pnpm typecheck` — passes.
- `cargo check` — passes.
- `pnpm tauri build` — builds `CC Switch.app` successfully (the optional updater signing step is skipped because `TAURI_SIGNING_PRIVATE_KEY` is not set, but that is unrelated to this change).
- Manually verified end-to-end on macOS: selecting Warp opens a new Warp tab and runs the cc-switch launcher script automatically.

## How to test

1. Build and run.
2. *Settings → Preferred Terminal* → **Warp**.
3. Click any *Open in Terminal* button on a provider, or *Resume in terminal* in the Session Manager.
4. First run only: macOS will prompt to grant cc-switch Accessibility permission. After approving, click again — Warp will activate, open a fresh tab and run the bash launcher.

[Conversation link](https://app.warp.dev/conversation/c9d93262-d946-4e99-9234-89e18273d1af)

Co-Authored-By: Oz <oz-agent@warp.dev>

### DIFF
--- a/src-tauri/src/commands/misc.rs
+++ b/src-tauri/src/commands/misc.rs
@@ -944,6 +944,7 @@ exec bash --norc --noprofile
     // Note: Kitty doesn't need the -e flag, others do
     let result = match terminal {
         "iterm2" => launch_macos_iterm2(&script_file),
+        "warp" => launch_macos_warp(&script_file),
         "alacritty" => launch_macos_open_app("Alacritty", &script_file, true),
         "kitty" => launch_macos_open_app("kitty", &script_file, false),
         "ghostty" => launch_macos_open_app("Ghostty", &script_file, true),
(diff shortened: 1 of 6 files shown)


farion1231:
@codex review

farion1231 (commented):
When I tested this before, Warp did not support launching with an additional command. Did it work correctly in your testing?

farion1231 (commented):
Thanks for the contribution! While reviewing this, I noticed [#2466](https://github.com/farion1231/cc-switch/issues/2466) was just merged and covers the same feature on macOS using Warp's warp://action/new_tab URI scheme. The URI-based approach avoids needing System Events / Accessibility permission and the activate-then-keystroke timing race, so we're going to land on that one.

The Linux portion in this PR (warp-terminal -e bash <script>) also has a latent bug: the existing Linux launcher uses Command::spawn() and treats successful fork as success, so if warp-terminal doesn't accept -e (which appears to be the case), the user sees no error but the command never runs. We'd want to verify Warp Linux's actual launch contract before adding it to the fallback list.

Closing this in favor of [#2466](https://github.com/farion1231/cc-switch/issues/2466). Thanks again for taking the time!

PR #2469, opened 2026-04-29
fix(config): sort JSON keys alphabetically for deterministic output

fuleinist:
## Summary
Fixes non-deterministic JSON key ordering in settings.json files that caused large spurious git diffs when switching configs.

## Problem
When cc-switch writes to settings.json (for Claude Code, Gemini, etc.), the JSON keys could appear in different orders after modifications. This is because serde_json's IndexMap preserves insertion order, but deep merge operations and re-serialization don't guarantee consistent key ordering across writes.

## Solution
Added a `sort_json_keys()` function that recursively sorts all JSON object keys alphabetically before writing. This ensures deterministic output regardless of the order keys were inserted during config building.

## Changes
- Added `sort_json_keys()` helper function in `config.rs`
- Modified `write_json_file()` to sort keys before serialization
- Preserves nested array order (only object keys are sorted)
- Minimal, surgical change touching only the JSON serialization path

## Testing
The fix was verified by:
1. Code compiles successfully
2. Existing unit tests pass
3. Manual verification: switching configs multiple times produces identical settings.json output (same key order)

### DIFF
--- a/src-tauri/src/config.rs
+++ b/src-tauri/src/config.rs
@@ -186,8 +186,8 @@ pub fn write_json_file<T: Serialize>(path: &Path, data: &T) -> Result<(), AppErr
 
     let value = serde_json::to_value(data).map_err(|e| AppError::JsonSerialize { source: e })?;
     let sorted_value = sort_json_keys(&value);
-    let json =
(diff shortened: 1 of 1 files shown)


farion1231:
@codex review

farion1231 (approved):

PR #2551, opened 2026-05-03
[codex] Extract VS Code request for session titles

Xvvln:
## Summary

- Extract the real user request from Codex VS Code session messages that start with `# Context from my IDE setup:`.
- Keep existing title filtering for `# AGENTS.md` and `<environment_context>` injected messages.
- Add regression tests for VS Code IDE context messages with and without a `## My request for Codex:` section.

## Root Cause

Codex sessions launched from VS Code can store the IDE context and the real prompt in the same `role=user` message. The session manager used the first user message as the title, so many Codex sessions were displayed as the same `# Context from my IDE setup:` prefix instead of the actual request.

## Validation

- `git diff --check`
- `pnpm typecheck`
- `pnpm format:check`


farion1231:
@codex review
@claude review

Xvvln:
@farion1231 
image
this is a real scene

PR #2573, opened 2026-05-04
fix(settings): use app config dir for app settings

keta1:
## Summary

- Store AppSettings in the app config directory instead of hard-coding ~/.cc-switch/settings.json
- Update the settings-path helper to return a concrete app-config path
- Keep the documentation comment aligned with the actual storage location

## Migration Note

The current implementation changes where local AppSettings are read and written. Users who previously used a custom CC Switch config directory may temporarily lose their local settings because the app will start reading `<app_config_dir>/settings.json` instead of the old `~/.cc-switch/settings.json` location.

Those users need to manually migrate their existing `settings.json` into the active app config directory.

### DIFF
--- a/src-tauri/src/settings.rs
+++ b/src-tauri/src/settings.rs
@@ -169,7 +169,7 @@ impl WebDavSyncSettings {
 
 /// 应用设置结构
 ///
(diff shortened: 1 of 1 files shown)


farion1231:
@codex review
@claude review

Issue #2495, opened 2026-04-30
I’m using Claude Code with multiple providers/models for different tasks, and I often need to keep multiple terminal sessions open at the same time.


amumu3792:
### Self Checks / 自检

- [x] I have read the [FAQ](https://github.com/farion1231/cc-switch#faq) section in README.
我已阅读 README 中的[常见问题](https://github.com/farion1231/cc-switch#常见问题)。

- [x] I have searched for [existing issues](https://github.com/farion1231/cc-switch/issues), including closed ones.
我已搜索过[已有的 Issue](https://github.com/farion1231/cc-switch/issues)，包括已关闭的。

### Related App / 涉及应用

Claude Code

### Problem or Motivation / 问题或动机

I’m using Claude Code with multiple providers/models for different tasks, and I often need to keep multiple terminal sessions open at the same time.

My actual workflow is:
- Session A: official Claude subscription
- Session B: OpenRouter with inclusionai/ling-2.6-1t
- Session C: NVIDIA or another provider

The problem is that cc-switch writes provider-related settings into Claude Code’s global user settings/env, so switching one provider affects all running and future sessions. This makes it impossible to keep multiple concurrent Claude Code sessions isolated from each other.

What I need is not just global switching, but session-level or profile-level isolation.

Feature requests:
1. Support multiple named profiles, e.g. `official`, `openrouter-ling`, `nvidia`
2. Allow launching Claude Code with a specific profile without modifying the global shared config
3. Keep provider settings isolated per launched session
4. Avoid mutating `~/.claude/settings.json` globally for provider/model/token fields
5. Optionally support separate config directories per profile

Example desired usage:
- `cc-switch run official`
- `cc-switch run openrouter-ling`
- `cc-switch run nvidia`

This would allow multiple Claude Code terminals to coexist without config contamination.

### Contribution / 参与贡献

- [ ] I am interested in contributing to this feature.
我有兴趣参与开发此功能。

ZHYXNJD:
Hi, I also encountered the same problem, and I am interested in contributing to this feature.

samzong:
+1

ShuBuShao:
这个功能使用场景挺多的 +1

PR #2574, opened 2026-05-05
feat(provider): add configurable terminal startup command

### LINKED ISSUE #2575: Support configurable terminal startup command for Open Terminal
### Self Checks / 自检

### Related App / 涉及应用

All supported apps / 通用功能

Currently, the "Open Terminal" action uses a fixed startup command for each app.

This works for the default workflow, but it is too limited for users who want CC Switch to inject the selected provider environment and then run a custom terminal/session command.

Common examples:

1. Some users manage terminal sessions with Zellij or Tmux, and do not want CC Switch to directly launch the app command. They want CC Switch to open the selected terminal, inject provider environment variables, and then attach/create a session:

```bash
zellij attach --create {sessionName}
tmux new -A -s {sessionName}
```

2. Some Claude Code users need to append startup flags, for example:

```bash
claude --settings {settingsFile} --dangerously-skip-permissions
```

3. Some users launch apps through wrapper commands or aliases:

```bash
hapi claude --settings {settingsFile}
```

At the moment, these workflows require manually re-running commands after the terminal opens, which makes the "Open Terminal" action less useful for session-managed or wrapper-based workflows.

Related issues:

- #2495
- #2508
- #1548

### Proposed Solution / 建议方案

Add an optional provider-level "Terminal startup command" template.

When empty, CC Switch keeps the existing default behavior.

When configured, the existing "Open Terminal" flow should:

1. Use the selected terminal app as configured today, such as Terminal.app, iTerm2, Warp, etc.
2. Apply the selected provider configuration and inject provider environment variables
3. Run the configured command template instead of the built-in default startup command

This keeps terminal selection and startup command customization separate:

- terminal preference controls where the command runs
- terminal startup command controls what runs after the provider environment is loaded

Useful template variables could include:

- `{settingsFile}`: temporary provider config/settings file path
- `{providerId}`: selected provider id
- `{sessionName}`: generated safe session name
- `{app}`: current app type
- `{cwd}`: selected working directory

### Additional Context / 补充信息

Example commands:

```bash
zellij attach --create {sessionName}
tmux new -A -s {sessionName}
claude --settings {settingsFile} --dangerously-skip-permissions
hapi claude --settings {settingsFile}
```

This should be backward compatible: if the command template is empty, each app should continue using its current default startup command.

### Contribution / 参与贡献

我有兴趣参与开发此功能。

suharvest:
## Summary

- Add a provider-level terminal startup command template for the existing "Open Terminal" action
- Keep current app-specific defaults when the template is empty
- Support reusable variables such as `{settingsFile}`, `{providerId}`, `{sessionName}`, `{app}`, and `{cwd}`
- Add provider form UI, tooltip documentation, and i18n strings for English, Chinese, and Japanese

## Motivation

The current "Open Terminal" action uses a fixed command per app. This is limiting for workflows where users want CC Switch to inject the selected provider environment, but then run a custom terminal/session command.

Examples:

```bash
zellij attach --create {sessionName}
tmux new -A -s {sessionName}
claude --settings {settingsFile} --dangerously-skip-permissions
hapi claude --settings {settingsFile}
```

This keeps terminal app selection independent from the startup command: users can still choose Terminal.app, iTerm2, Warp, etc., while customizing what runs inside that terminal after the provider environment is loaded.

Related issues:

- Closes #2575
- #2495
- #2508
- #1548

## Testing

- `pnpm typecheck`
- `cargo test --manifest-path src-tauri/Cargo.toml commands::misc`


farion1231:
@codex review
@claude review

suharvest:
Addressed the automated review feedback in ba751fde. The terminal launcher now skips invalid environment variable names before emitting shell exports / Windows set commands, and the cleanup trap now uses a cleanup function instead of nested shell quotes.\n\nRe-tested:\n- cargo fmt --check --manifest-path src-tauri/Cargo.toml\n- cargo test --manifest-path src-tauri/Cargo.toml commands::misc

suharvest:
刚刚 rebase 到最新的 main 解决冲突，原因是 main 在此 PR 之后又合并了若干改动（其中 #2416 universal-provider duplicate action 等同样改动了 `ProviderForm.tsx`）。

冲突点和解决方式：

1. `ProviderForm.tsx` `ProviderFormProps` 接口附近 — main 把 `interface` 改成了 `export interface`，本 PR 在该接口前新增了 `getDefaultTerminalCommand` 辅助函数。两边保留：保留 main 的 `export`，同时保留本 PR 的辅助函数。
2. `ProviderForm.tsx` `nextMeta` 字段构造处 — main 新增 `claudeDesktopMode: undefined`，本 PR 新增 `terminalCommandTemplate`。两个字段并存即可。

其余文件（`misc.rs`、`provider.rs`、i18n locales、`types.ts`）由 git 自动合并完成，无人工冲突。

已本地 `tsc --noEmit` 通过。

farion1231:
@codex review

suharvest:
Pushed b8c2461b addressing the latest Codex review:

**P1 — `default_provider_terminal_command` non-exhaustive for `ClaudeDesktop`**: confirmed and fixed. The rebase onto `main` pulled in `AppType::ClaudeDesktop` (from #2416), and the match arm was missing — `cargo check` actually fails on the pre-fix tree. Added a `ClaudeDesktop => ""` arm: Claude Desktop is a GUI app rather than a CLI, so the launcher now just exports the env vars and drops the user at a prompt. Test extended to cover the new arm.

**P2 — Windows env values containing `"`**: confirmed and fixed. There is no reliable escape for a literal `"` inside `set "K=V"`, so I switched `build_windows_set_commands` to the unquoted `set K=V` form. `escape_windows_batch_value` already handles the metacharacters that matter (`& | < > ( ) % ^`), and `"` becomes a literal. Added a regression test (`build_windows_set_commands_preserves_double_quotes_in_value`) and updated the two existing tests to match the new output shape.

**P1 — env var name sanitization**: I believe this one is a false positive on the re-review — the protection is already present in the reviewed commit (`3adda82576`). `build_shell_export_commands` (misc.rs:1406) and `build_windows_set_commands` (misc.rs:1523) both filter via `is_valid_env_var_name` (misc.rs:1534), which enforces the POSIX `[A-Za-z_][A-Za-z0-9_]*` rule and rejects exactly the `FOO;rm -rf ~` style key the bot cited. `git blame` puts those lines in `3adda82576` because the earlier `ba751fde` fix got rolled forward when I rebased. There is also an existing regression test `build_shell_export_commands_skips_invalid_env_names` (misc.rs:2021) covering that vector, plus the Windows-side equivalent.

Verified locally:
- `cargo check --manifest-path src-tauri/Cargo.toml --all-targets`
- `cargo test --manifest-path src-tauri/Cargo.toml --lib` (relevant `commands::misc::tests::*` all pass)

suharvest:
Pushed d2703f03 to close one more Windows-side hardening gap I noticed while re-reading the diff after the previous fix.

**CR/LF injection through `escape_windows_batch_value`**: any value that flows into the generated `.bat` (`set K=V` env values, `{providerId}` template substitutions, `set "CCSWITCH_PROVIDER_ID=..."`, `echo` line) passes through `escape_windows_batch_value`, which caret-escapes `& | < > ( ) % ^` but lets `\r` / `\n` through. An embedded newline ends the current batch line, so the remainder of the value gets parsed as a fresh batch command. This vector was opened by the new templating path on Windows (the env/template lines didn't exist before this PR), so it's in scope for the same hardening pass.

There is no caret escape that survives a literal newline inside a `.bat` file, so the fix is to strip `\r` / `\n` at the top of `escape_windows_batch_value` before the caret pass. None of the call sites (env values, provider id, settings file path, cwd, app name) have a legitimate use for newlines.

Added two regression tests: `escape_windows_batch_value_strips_crlf` and `build_windows_set_commands_strips_crlf_in_value`.

While checking that, I also noticed the launcher temp files (`cc_switch_launcher_<pid>.sh`, `claude_<provider>_<pid>.json`, `cc_switch_claude_<pid>.bat`) use PID-based predictable names + `std::fs::write` rather than O_EXCL, which leaves a small symlink-race window on shared `temp_dir`. That code pre-exists this PR (the writes blame to 2025-12 / 2026-01 commits, not the templating commit), so I think it's better as a separate issue rather than expanding this PR's scope — happy to file one if you'd like.

Verified locally:
- `cargo fmt --manifest-path src-tauri/Cargo.toml`
- `cargo clippy --manifest-path src-tauri/Cargo.toml --all-targets -- -D warnings`
- `cargo test --manifest-path src-tauri/Cargo.toml --lib commands::misc::tests` (23 passed)

farion1231:
@codex review

farion1231:
@codex review
@claude review

suharvest:
Pushed c98eeb4a addressing the remaining Codex P2 (cascading token replacement in \`render_command_template\`).

**Root cause**: the previous implementation folded `str::replace` over each `(key, value)` pair, so a value substituted for an earlier token could itself contain `{anotherKey}` and get re-expanded on a later iteration. In the current code path none of the five values can legally contain `{...}` (they are UUIDs, an `AppType` enum string, or `temp_dir()`-rooted paths), so the bug was not reachable — but the invariant is fragile and would break the moment a new placeholder fed user-controlled text.

**Fix**: switched to a single-pass scanner that walks the template once, recognizes `{key}` against the replacements table only when it appears in the original template, and emits the value verbatim. A substituted value containing `{...}` is therefore preserved as literal text.

Regression tests:
- `render_command_template_does_not_cascade_replacements` — verifies that a value of `{b}` substituted for `{a}` does not get further expanded when `{b}` is also a placeholder.
- `render_command_template_leaves_unknown_placeholders_intact` — confirms unknown `{...}` tokens pass through unchanged (matches prior behavior).

Verified locally:
- \`cargo fmt --manifest-path src-tauri/Cargo.toml\`
- \`cargo clippy --manifest-path src-tauri/Cargo.toml --all-targets -- -D warnings\`
- \`cargo test --manifest-path src-tauri/Cargo.toml --lib commands::misc::tests\`

Issue #2535, opened 2026-05-02
[Bug] Live import creates duplicate id="default" provider for Codex on every startup


JiangHe12:
### Self Checks

- [x] I have read the [FAQ](https://github.com/farion1231/cc-switch#faq) section in README.
- [x] I have searched for [existing issues](https://github.com/farion1231/cc-switch/issues), including closed ones.

### CC Switch Version

3.14.1

### Operating System

Windows

### Related App

Codex

### Steps to Reproduce

1. Install CC Switch v3.14.1, configure a Codex provider (using ChatGPT OAuth authentication)
2. Verify that the `providers` table in `cc-switch.db` already has an entry with `id="codex-official"`
3. Close CC Switch
4. Restart CC Switch
5. Check the log file `cc-switch.log` — it shows: `✓ Imported live config for codex as default provider`
6. Open `cc-switch.db` with SQLite and run: `SELECT id, name, is_current, category FROM providers WHERE app_type='codex'`

**This happens on every startup.**

### Expected Behavior

On startup, if the `providers` table already has entries with `app_type='codex'` (whether `codex-official` or any custom provider), the live import should be skipped — just like Claude and Gemini, which show `○ claude already has providers; live import skipped`.

It should NOT create a duplicate entry with `id="default"`, `name="default"`, `category="custom"` on every startup.

### Actual Behavior

Every time CC Switch starts, even though the `providers` table already contains `id="codex-official"` (with `category="official"`, created by the `official_providers_seeded` mechanism), the live import check still considers the codex group as having "no providers", triggers the import, and creates a duplicate entry with `id="default"`.

**Database result:**

| id | name | is_current | category |
|---|---|---|---|
| codex-official | OpenAI Official | 0 | official |
| default | default | **1** | custom |

Both records have identical auth tokens. The `default` entry has `created_at = NULL`, indicating it was not created through the normal INSERT flow.

**Comparison with Claude and Gemini:**
- Claude: log shows `○ claude already has providers; live import skipped` ✅
- Gemini: log shows `○ gemini already has providers; live import skipped` ✅
- Codex: log shows `✓ Imported live config for codex as default provider` ❌

**Root cause hypothesis:** The live import's "has providers" check excludes the `codex-official` entry (which has `category="official"`), causing the check to return false. Claude and Gemini have user-added non-official entries, so their checks pass.

### Additional Context

**Environment:**
- CC Switch version: 3.14.1
- OS: Windows 11 Home China (10.0.26200)
[...]

PR #2590, opened 2026-05-06
Fix Codex startup live import duplication

DhruvShankpal:
Fix Issue #2535
## Summary

Fixes the startup live-import path so CC Switch does not recreate 'id="default"' for Codex when the app already has provider entries such as 'codex-official'.

Also keeps local current-provider settings in sync when a live import does happen.

## What changed

- Added a stricter startup-only provider existence check:
  - skip live import if the app already has any provider row
  - this includes official seeded providers like 'codex-official'
- Updated live import to also write the local current-provider setting after setting DB 'is_current'

## Why

Before this change, startup import only skipped when a non-official provider existed. That meant Codex could still auto-import live config even when 'codex-official' was already present, creating a 'default' provider and causing DB/settings current-provider drift.

## Result

- Existing Codex installs with 'codex-official' no longer create 'default' on every startup
- Fresh installs can still import live config once when no providers exist
- DB 'is_current' and local 'currentProvider' settings stay aligned after import

## Notes
I could not run automated tests in the current shell because 'cargo' and 'pnpm' were not available in the environment.
I am relatively new to Rust and used the assistance of codex to help solve the problem so advise supervision before merge.

### DIFF
--- a/src-tauri/src/database/dao/providers.rs
+++ b/src-tauri/src/database/dao/providers.rs
@@ -535,6 +535,22 @@ impl Database {
         Ok(ids)
     }
 
+    /// 判断指定 app 下是否已存在任意 provider。
+    ///
+    /// 启动阶段的 live import 需要使用这个更严格的判断：
+    /// 只要该 app 已经有任何 provider（包括官方 seed），就不应再自动导入 `default`。
+    pub fn has_any_provider_for_app(&self, app_type: &str) -> Result<bool, AppError> {
+        let conn = lock_conn!(self.conn);
+        let exists: bool = conn
+            .query_row(
+                "SELECT EXISTS(SELECT 1 FROM providers WHERE app_type = ?1)",
+                params![app_type],
+                |row| row.get(0),
+            )
+            .map_err(|e| AppError::Database(e.to_string()))?;
+        Ok(exists)
+    }
+
     /// 判断指定 app 下是否存在非官方种子的供应商。
     ///
     /// 比 `get_all_providers` 轻量得多：只读 id 列、无 endpoint 子查询、首条命中即返回。
--- a/src-tauri/src/lib.rs
+++ b/src-tauri/src/lib.rs
@@ -494,6 +494,18 @@ pub fn run() {
             for app_type in
                 crate::app_config::AppType::all().filter(|t| !t.is_additive_mode())
             {
+                if app_state
+                    .db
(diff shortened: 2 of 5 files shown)


JiangHe12:
Thanks for the PR! The approach looks correct — the root cause is that `has_non_official_seed_provider` treats `codex-official` as "no providers", so adding a broader check at startup makes sense.

A few things:

1. Could you verify manually that a fresh install (empty DB) still correctly imports live config on first run?
2. Could you check that restarting after the fix doesn't create new `default` entries?
3. The third change (syncing `set_current_provider` to settings) is a nice touch but could be a separate commit for clarity — up to the maintainer.

Overall the fix is clean and addresses the issue well. 👍

DhruvShankpal:
Fix: Prevent duplicate Codex default provider on restart & add startup import tests

Hey @JiangHe12 I fixed the issue where restarting the app could create duplicate Codex default provider entries. The startup logic now checks for any existing provider before importing the live config, ensuring the import only happens on a true fresh install.

What’s included:

Refactored startup guard to check for any provider, not just non-official ones.

Added tests to verify:
Fresh install imports Codex live config as default provider.
Restart with existing provider skips import and does not create duplicates.

Test results:
Both new tests pass locally:

`codex_startup_import_fresh_install_imports_once_and_syncs_current_setting`
`codex_startup_import_skips_when_only_official_seed_exists`

Notes:
I locally installed the dependencies and the tests passed ok.
I synced set_current_provider to settings after import , let me know if you’d like the settings sync change split into a separate commit. You previously said it was optional so I didn't do that here, but i can explicitly change if you want.

DhruvShankpal:
Also @JiangHe12  I'm curious what the appropriate etiquette is when improving on my initial pull request. Is it better to just commit changes to same branch(like i did above) or create a new branch?

JiangHe12:
The refactored approach with `should_import_default_config_on_startup` is cleaner, and the two tests cover exactly what I was concerned about — looks good to me.

For your etiquette question: committing to the same branch is the standard practice. It automatically updates this PR, keeps all discussion in one place, and the maintainer can review the full diff. Creating a new branch would mean a new PR and split context. So what you did is correct.

farion1231:
@codex review

DhruvShankpal:
all thats left is maintainer merging right? any steps from me?

farion1231 (approved):

PR #2645, opened 2026-05-08
fix: rewrite response model field back to request model

GinBan1:
## Summary

When the proxy maps models (e.g. `claude-sonnet-4-6` → `mimo-v2.5-pro`), the upstream API response contains the mapped model name. This causes Claude Code's `/context` command to display the wrong model name, since it reads from local config rather than the API response.

This PR rewrites the `model` field in the response back to the original request model, so the client always sees the model name it configured. The proxy's model mapping remains fully functional for routing — only the response metadata is adjusted.

### `response_processor.rs`
- **Non-streaming passthrough**: After decompression, parse JSON response, rewrite `model` field to `ctx.request_model`, re-serialize, and strip `Content-Length` header
- **Streaming passthrough**: Intercept SSE `message_start` events and rewrite `model` field in the `message` object before yielding bytes to client
- Updated `create_logged_passthrough_stream` signature to accept `request_model: Option<String>`

### `handlers.rs`
- **Transform mode**: After OpenAI/Gemini → Anthropic conversion, rewrite `model` field before serialization
- Updated stream creation call to pass `request_model`

## Motivation

Users who switch between providers (e.g. from Anthropic to Xiaomi MiMo) via cc-switch want Claude Code's `/context` to display the correct model name. Currently, `/context` always shows the original Claude model name (e.g. "Sonnet 4.6") because Claude Code reads from its local config, not from the API response. With this fix, the response `model` field matches what Claude Code expects, so the display is always correct.

## Test plan

- [ ] Non-streaming request with model mapping: verify response `model` matches request model
- [ ] Streaming request with model mapping: verify SSE `message_start` event `model` matches request model
- [ ] Request without model mapping: verify no-op (model already matches)
- [ ] Non-streaming transform mode: verify converted response `model` matches request model

### DIFF
--- a/src-tauri/src/proxy/handlers.rs
+++ b/src-tauri/src/proxy/handlers.rs
@@ -241,6 +241,7 @@ async fn handle_claude_transform(
             "Claude/OpenRouter",
             Some(usage_collector),
             timeout_config,
+            Some(ctx.request_model.clone()),
         );
 
         let mut headers = axum::http::HeaderMap::new();
@@ -327,6 +328,10 @@ async fn handle_claude_transform(
             }
         });
     }
+
+    // 将转换后的响应 model 字段改写回请求原始模型名
+    let mut anthropic_response = anthropic_response;
+    anthropic_response["model"] = json!(ctx.request_model);
 
     // 构建响应
     let mut builder = axum::response::Response::builder().status(status);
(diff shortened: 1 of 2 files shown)


farion1231:
@codex review
@claude review

PR #2749, opened 2026-05-12
fix(skills): fall back to junctions for Windows skill links

TheLonelyDevil9:
## What?

Adds a Windows fallback for skill sync when directory symlink creation fails. CC Switch now treats skill sync links as managed links: Unix symlinks, Windows directory symlinks, or Windows junctions.

## Why?

Non-admin Windows sessions without Developer Mode can fail directory symlink creation with an administrator-privilege error. That breaks live skill sync across CLI harnesses and can remove an existing link during toggle/re-sync flows.

## How?

- Keeps Unix symlink behavior unchanged.
- Tries `std::os::windows::fs::symlink_dir` first on Windows.
- Falls back to an NTFS junction using `CreateFileW` and `DeviceIoControl(FSCTL_SET_REPARSE_POINT)`, avoiding `cmd.exe`/`mklink` shell parsing.
- Detects Windows directory reparse points as managed skill links.
- Removes managed links without recursively deleting the target directory.
- Normalizes Windows link targets with `\??\` / `\\?\` prefixes before SSOT comparisons.
- Updates skill sync tests to validate live managed-link behavior and orphan cleanup through the public sync path.

## Testing?

- `git diff --check`
- Manual Windows validation: junctions work in the affected non-admin session where symlink creation fails.
- Not run locally: `cargo fmt`, `cargo test`, or Rust build checks because Rust tooling is not available in this shell.

## Screenshots (optional)

Not applicable.

## Anything Else?

No matching open upstream PR was found for Windows skill symlink-to-junction fallback.

### DIFF
--- a/src-tauri/src/services/skill.rs
+++ b/src-tauri/src/services/skill.rs
@@ -25,10 +25,10 @@ use crate::error::format_skill_error;
 #[derive(Debug, Clone, Copy, PartialEq, Eq, Serialize, Deserialize, Default)]
 #[serde(rename_all = "lowercase")]
 pub enum SyncMethod {
-    /// 自动选择：优先 symlink，失败时回退到 copy
+    /// 自动选择：优先托管链接，失败时回退到 copy
     #[default]
     Auto,
-    /// 符号链接（推荐，节省磁盘空间）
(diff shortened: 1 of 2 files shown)


farion1231:
@codex review

PR #2760, opened 2026-05-13
fix(hermes): refresh model.base_url on provider switch

### LINKED ISSUE #2757: [Bug] Hermes model.base_url not updated when switching providers (HTTP 404 on auxiliary services)
### Self Checks / 自检

### CC Switch Version / 版本号

v3.14.1

### Operating System / 操作系统

macOS

### Related App / 涉及应用

Other / 其他

### Steps to Reproduce / 重现步骤

                                                                                               
     1. Open CC Switch → Switch Hermes provider to DeepSeek                                                             
        - Verify ~/.hermes/config.yaml shows base_url: https://api.deepseek.com ✓                                       
     2. Open CC Switch → Switch Hermes provider to Kimi For Coding                                                      
     3. Check ~/.hermes/config.yaml                                                                                     
    

### Expected Behavior / 期望行为

yaml                                                                                                               
     model:                                                                                                             
       default: kimi-for-coding                                                                                         
       provider: kimi                                                                                                   
       base_url: https://api.kimi.com/coding   

### Actual Behavior / 实际行为

yaml                                                                                                               
     model:                                                                                                             
       default: kimi-for-coding                                                                                         
       provider: kimi                                                                                                   
       base_url: https://api.deepseek.com   # ← Still DeepSeek's URL!                                                   
                                                                                                             

### Additional Context / 补充信息

Bug Report: Hermes model.base_url Not Updated When Switching Providers                                             
     Environment                                                                                                        
     - CC Switch Version: 3.14.1                                                                                        
     - OS: macOS 15.x (Sequoia)                                                                                         
     - Hermes Config Path: ~/.hermes/config.yaml                                                                        
     Problem Description                                                                                                
     When switching between AI providers in CC Switch's Hermes integration, the model.base_url field in config.yaml     
     is not correctly updated to match the newly selected provider. Instead, it retains the previous provider's         
     base URL, causing all API requests (including auxiliary services like title generation) to fail with HTTP 404      
     errors.                                                                                                            
     Steps to Reproduce                                                                                                 
     1. Open CC Switch → Switch Hermes provider to DeepSeek                                                             
        - Verify ~/.hermes/config.yaml shows base_url: https://api.deepseek.com ✓                                       
     2. Open CC Switch → Switch Hermes provider to Kimi For Coding                                                      
     3. Check ~/.hermes/config.yaml                                                                                     
     Expected Result                                                                                                    
     yaml                                                                                                               
     model:                                                                                                             
       default: kimi-for-coding                                                                                         
       provider: kimi                                                                                                   
       base_url: https://api.kimi.com/coding                                                                            
                                                                                                                        
     Actual Result                                                                                                      
     yaml                                                                                                               
     model:                                                                                                             
       default: kimi-for-coding                                                                                         
       provider: kimi                                                                                                   
       base_url: https://api.deepseek.com   # ← Still DeepSeek's URL!                                                   
                                                                                                                        
     Impact                                                                                                             
     - Primary: Hermes cannot connect to the correct API endpoint                                                       
     - Secondary: Auxiliary services (title generation, session search, etc.) return HTTP 404                           
     - Workaround required: Manual edit of config.yaml after every provider switch                                      
     Evidence from CC Switch Database                                                                                   
                                                                                                                        
     The provider configuration in CC Switch's own SQLite database (~/.cc-switch/cc-switch.db) is correct:              
                                                                                                                        
     sql                                                                                                                
     SELECT id, name, settings_config FROM providers WHERE app_type = 'hermes' AND id = 'kimi';                         
                                                                                                                        
     Returns:                                                                                                           
     json                                                                                                               
     {                                                                                                                  
       "name": "kimi_coding",                                                                                           
       "base_url": "https://api.kimi.com/coding",                                                                       
       "api_mode": "anthropic_messages",                                                                                
       "models": [{"id":"kimi-for-coding","name":"Kimi For Coding","context_length":256000}]                            
     }                                                                                                                  
                                                                                                                        
     This proves the bug is in the config file generation logic, not the provider storage.                              
     Root Cause Analysis                                                                                                
                                                                                                                        
     From backup files in ~/.cc-switch/backups/hermes/:                                                                 
                                                                                                                        
     yaml                                                                                                               
     hermes_20260513_115741.yaml (generated by CC Switch)                                                               
     model:                                                                                                             
       default: kimi-for-coding      # ← Correct (from provider config)                                                 
       provider: kimi                # ← Correct (from provider config)                                                 
       base_url: https://api.deepseek.com   # ← WRONG (stale/default value)                                             
                                                                                                                        
     The base_url field appears to be:                                                                                  
     1. Either hardcoded to a default and not overridden during provider switch                                         
     2. Or cached from the previous provider and not refreshed                                                          
     Suggested Fix                                                                                                      
                                                                                                                        
     In src/hermes_config.rs (or equivalent config generation module), ensure that when generating the model:           
     section:                                                                                                           
                                                                                                                        
     rust                                                                                                               
     // Pseudocode                                                                                                      
     let provider_config = db.get_provider(app_type, provider_id);                                                      
     config.model = ModelConfig {                                                                                       
         default: provider_config.default_model,                                                                        
         provider: provider_config.id,                                                                                  
         base_url: provider_config.base_url,  // ← Must come from provider config!                                      
     };                                                                                                                 
                                                                                                                        
     Currently it seems base_url is either:                                                                             
     - Read from a stale cache                                                                                          
     - Using a hardcoded fallback                                                                                       
     - Not updated during the provider switch transaction                                                               
     Workaround (for users)                                                                                             
                                                                                                                        
     Until fixed, users can:                                                                                            
     1. Manually edit ~/.hermes/config.yaml after each switch                                                           
     2. Use an auto-fix script that runs on shell startup                                                               
                                                                                                                        
     Labels: bug, hermes, provider-switch                                                                               
     Priority: High (breaks core functionality)

lihan3238:
## Summary

Fixes #2757. `apply_switch_defaults` in `src-tauri/src/hermes_config.rs` used `..current` struct-update to preserve user customizations (`context_length`, `max_tokens`, `extra`). But this also silently preserved `base_url` across provider switches — so when the user switched from DeepSeek to Kimi For Coding, the `model.base_url` field in `~/.hermes/config.yaml` kept pointing at DeepSeek. Auxiliary services (title generation, session search) then 404'd against the wrong endpoint.

The issue reporter (@KaisvenZ) provided a complete root-cause analysis including DB-vs-yaml diff evidence and pseudocode for the fix.

## Change

`src-tauri/src/hermes_config.rs` `apply_switch_defaults`:

```diff
+    let new_base_url = settings_config
+        .get("base_url")
+        .and_then(|v| v.as_str())
+        .map(|s| s.trim().to_string())
+        .filter(|s| !s.is_empty());
+
     let current = get_model_config()?.unwrap_or_default();
     let merged = HermesModelConfig {
         default: first_model_id.or(current.default.clone()),
         provider: Some(provider_id.to_string()),
+        base_url: new_base_url.or(current.base_url.clone()),
         ..current
     };
```

Use the new provider's `settings_config.base_url` when set, falling back to the existing value when the new provider's settings omits `base_url` (so a user override survives a switch to a provider that does not declare its own).

Docstring updated to spell out the new precedence.

## Tests

Added two regression tests:

- `apply_switch_defaults_updates_base_url_to_new_provider` — direct reproduction of #2757 (DeepSeek → Kimi switch with both providers declaring their own `base_url`)
- `apply_switch_defaults_ignores_blank_base_url_in_settings` — guards the fallback path so a whitespace-only `base_url` in settings does not blank out an existing override

The four pre-existing `apply_switch_defaults` tests continue to pass unchanged. None of them asserted `base_url` with a non-empty `settings.base_url` value — they covered:

- (`sets_default_and_provider`) settings *had* `base_url` but didn't assert on it → still passes; new behavior just makes the assertion possible
- (`preserves_user_context_length`) settings *lacked* `base_url`, asserts user override is preserved → still works (falls back to current via `.or()`)
- (`updates_provider_even_without_models`) settings *had* `base_url` but didn't assert on it → still passes
- (`keeps_old_default_when_first_model_id_is_blank`) settings lacked `base_url` → still works

## Behavior change note

This change refreshes `model.base_url` whenever the new provider declares one, which is a user-visible behavior change. The previous code would have kept a manually-edited `model.base_url` even across a provider switch. Two views are reasonable:

- **Issue reporter's view (this PR)**: provider switching is a "use this provider's settings" gesture; provider-specific fields including `base_url` should refresh.
- **Alternative view**: user-set overrides should always win, even across provider switches.

Happy to defer if you prefer the alternative — the simplest revert is to drop the `base_url:` line in the struct literal and remove the new test. I went with the issue-reporter's view because the breakage they describe (404s on auxiliary services) is consistent and affects real users, and because the previous "always preserve" behavior was implicit in the `..current` rather than explicitly designed.

## Verification

- `cargo fmt --check` passes locally
- Could not run `cargo test` locally (Tauri build needs `pkg-config` + `libssl-dev` system packages I don't have sudo for); diff is small and identical in shape to surrounding tested code, relying on the backend CI job to verify.

### DIFF
--- a/src-tauri/src/hermes_config.rs
+++ b/src-tauri/src/hermes_config.rs
@@ -795,8 +795,15 @@ pub fn set_model_config(model: &HermesModelConfig) -> Result<HermesWriteOutcome,
 /// still have a runnable configuration (Hermes will surface a clear error
 /// if the default no longer belongs to the active provider).
 ///
-/// Existing fields in `model:` (`context_length` / `max_tokens` / `base_url`
-/// / `extra`) are preserved via struct-update.
+/// `model.base_url` is overwritten when the new provider's `settings_config`
+/// declares a `base_url` — provider switching MUST refresh the upstream
+/// endpoint, otherwise auxiliary services (title generation, etc.) hit the
+/// previous provider and return 404. When the new provider has no
+/// `base_url`, the existing value is preserved so a user override survives.
+///
+/// `context_length` / `max_tokens` / `extra` are always preserved via
+/// struct-update so user-set customizations through the Model panel survive
+/// the switch.
 pub fn apply_switch_defaults(
     provider_id: &str,
     settings_config: &serde_json::Value,
@@ -810,10 +817,17 @@ pub fn apply_switch_defaults(
         .map(|s| s.trim().to_string())
         .filter(|s| !s.is_empty());
 
+    let new_base_url = settings_config
+        .get("base_url")
+        .and_then(|v| v.as_str())
+        .map(|s| s.trim().to_string())
+        .filter(|s| !s.is_empty());
+
     let current = get_model_config()?.unwrap_or_default();
     let merged = HermesModelConfig {
         default: first_model_id.or(current.default.clone()),
         provider: Some(provider_id.to_string()),
+        base_url: new_base_url.or(current.base_url.clone()),
         ..current
     };
     set_model_config(&merged)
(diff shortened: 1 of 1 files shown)


farion1231:
@codex review

MicroMilo:
Also issue https://github.com/farion1231/cc-switch/pull/2760.

PR #2782, opened 2026-05-14
fix(ui): suppress default browser context menu

kaisar945:
Prevent the WebView/browser context menu from appearing on regular app surfaces, which can expose browser-like actions such as Back, Forward, Stop, and Reload in the desktop UI.

Keep native context menus enabled for editable controls and CodeMirror editors so copy, paste, and text editing actions remain available.

## Related Issue / 关联 Issue

Fixes #

## Screenshots / 截图

## Checklist / 检查清单

### DIFF
--- a/src/main.tsx
+++ b/src/main.tsx
@@ -26,6 +26,24 @@ try {
   // 忽略平台检测失败
 }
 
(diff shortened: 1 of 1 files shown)


farion1231:
@codex review

PR #2784, opened 2026-05-14
fix(skills): install correct skill from skills.sh search results

zhaomoran:
## Problem

When searching skills.sh and multiple results share the same `directory` name (e.g. `agent-browser` from different repos), clicking install on any card always installed the **first** matching skill.

**Root cause:** `SkillCard` passed `skill.directory` to `onInstall`, and `handleInstall` used `Array.find(s => s.directory === directory)` — which returns the first match regardless of which card was clicked.

## Fix

Use the unique `key` field (`directory:repoOwner:repoName`) as the identifier passed from `SkillCard` to `onInstall`/`onUninstall`, and match on `key` in `handleInstall`.

**Changed files:**
- `src/components/skills/SkillCard.tsx` — `onInstall`/`onUninstall` props now receive `key` instead of `directory`
- `src/components/skills/SkillsPage.tsx` — `handleInstall` matches by `s.key === key`

## Test

Added `tests/components/SkillsPageInstall.test.tsx`: renders two skills.sh results with the same `directory` but different repos, clicks the second card's install button, and asserts the correct skill is passed to the install mutation.

### DIFF
--- a/src/components/skills/SkillCard.tsx
+++ b/src/components/skills/SkillCard.tsx
@@ -18,8 +18,8 @@ type SkillCardSkill = DiscoverableSkill & { installed: boolean };
 
 interface SkillCardProps {
   skill: SkillCardSkill;
-  onInstall: (directory: string) => Promise<void>;
-  onUninstall: (directory: string) => Promise<void>;
+  onInstall: (key: string) => Promise<void>;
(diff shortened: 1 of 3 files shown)


farion1231:
@codex review

farion1231 (approved):

PR #2786, opened 2026-05-14
fix(proxy): normalize OpenAI Chat tool payloads

ringoshiina:
## Summary
- normalize missing/null Anthropic tool `input_schema` into an OpenAI Chat function parameters object
- strip Anthropic-only `cache_control` fields from OpenAI Chat system messages, text content parts, and tools
- collapse single text content parts back to a plain string after stripping `cache_control`

## Why
Strict OpenAI-compatible chat backends reject Anthropic-only fields such as `cache_control`, and function/tool schemas must be JSON Schema objects. Claude Code can send server tools with a null schema and cache-control metadata, which can otherwise produce provider-side 400 validation errors.

## Tests
- `cargo test --lib proxy::providers::transform`

### DIFF
--- a/src-tauri/src/proxy/providers/transform.rs
+++ b/src-tauri/src/proxy/providers/transform.rs
@@ -491,8 +491,33 @@ fn convert_message_to_openai(
 }
 
 /// 清理 JSON schema（移除不支持的 format）
-pub fn clean_schema(mut schema: Value) -> Value {
+pub fn clean_schema(schema: Value) -> Value {
+    clean_schema_inner(schema, true)
+}
+
+fn clean_schema_inner(mut schema: Value, require_object_schema: bool) -> Value {
+    if !schema.is_object() {
+        return if require_object_schema {
+            json!({
+                "type": "object",
+                "properties": {}
+            })
+        } else {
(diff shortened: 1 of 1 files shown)


farion1231:
@codex review

farion1231 (commented):
Unconditionally removing `cache_control` would fix the 400 errors on strict OpenAI Chat backends, but it may break support for Anthropic-style prompt caching in OpenAI-compatible gateway scenarios such as OpenRouter, Envoy AI Gateway, and LiteLLM.

Do you think this issue could exist?

ringoshiina:
Yes, I think that issue could exist.\n\nI updated the PR to avoid that regression: it now preserves the existing `cache_control` behavior for OpenAI-compatible gateways and only keeps the safer schema normalization change. In other words, missing/null Anthropic tool `input_schema` values are normalized to a valid OpenAI function parameters object, but Anthropic-style prompt caching metadata is no longer stripped in this PR.\n\nI also rebased the branch onto the latest `main` and ran:\n\n`cargo test --lib proxy::providers::transform`\n\nResult: 140 passed.

PR #2787, opened 2026-05-14
fix(proxy): detect wrapped reasoning providers

ringoshiina:
## Summary
- preserve `reasoning_content` for OpenAI Chat providers when compatible identifiers appear in the provider name
- inspect provider env/default-model settings for `deepseek`, `kimi`, or `moonshot` identifiers
- cover wrapped aggregator configurations where the request model remains Claude-style and the base URL does not reveal the underlying reasoning model

## Why
The existing detection handles direct DeepSeek/Kimi/Moonshot models and base URLs. Aggregator-style providers can hide the actual reasoning backend in provider names or model mapping env vars, so tool-call follow-up requests may still need `reasoning_content` even when the request model is Claude-style.

## Tests
- `cargo test --lib proxy::providers::claude`


farion1231:
@codex review

PR #2801, opened 2026-05-15
fix(terminal): Ghostty opens clean window instead of cloning existing tabs

### LINKED ISSUE #2798: fix: `open -a` in launch_macos_open_app ignores --args when terminal is already running
## Bug Description

When using "Open Terminal" (打开终端) for a provider, if the target terminal (Ghostty, Alacritty, WezTerm, Kaku) is **already running**, the command silently does nothing — no new tab/window is created and `claude` is not launched.

The terminal gets focus but `--args` are ignored.

## Root Cause

`launch_macos_open_app` in `src-tauri/src/commands/misc.rs` uses:

```rust
cmd.arg("-a").arg(app_name).arg("--args");
```

On macOS, `open -a <App> --args ...` only passes `--args` when the app is **launched for the first time**. If the app is already running, `open -a` simply activates the existing instance and **discards** `--args`.

## Expected Behavior

A new terminal window should open and execute `claude --settings <config>`, regardless of whether the terminal is already running.

## Fix

Change `open -a` to `open -na` (the `-n` flag forces a new instance):

```diff
-    cmd.arg("-a").arg(app_name).arg("--args");
+    cmd.arg("-na").arg(app_name).arg("--args");
```

Note: the similar function in `session_manager/terminal/mod.rs` (used for session resume) already correctly uses `open -na` for Ghostty and other terminals.

## Affected Terminals

All terminals going through `launch_macos_open_app`:
- Ghostty
- Alacritty
- Kitty
- WezTerm
- Kaku

## Environment

- macOS
- Ghostty terminal (already running)
- CC Switch latest (commit 0d095555)

luw2007:
## Summary

- Add dedicated `launch_macos_ghostty` with `--quit-after-last-window-closed=true` so Ghostty opens a single clean window running claude, instead of cloning all existing tabs
- Change `launch_macos_open_app` from `open -a` to `open -na` so other terminals (Alacritty/Kitty/WezTerm/Kaku) correctly open a new window when already running

## Root Cause

- `open -a Ghostty --args ...`: when Ghostty is already running, macOS silently ignores `--args` — no new window, no claude
- `open -na Ghostty --args ...`: forces a new instance but clones all existing tabs into it
- Fix: use Ghostty-specific args (`--quit-after-last-window-closed=true -e bash <script>`) consistent with the existing `launch_ghostty` in `session_manager/terminal/mod.rs`

## Test Plan

- [x] Ghostty already running → click "Open Terminal" → single clean window with claude launches
- [x] `cargo check` passes

Closes #2798

### DIFF
--- a/src-tauri/src/commands/misc.rs
+++ b/src-tauri/src/commands/misc.rs
@@ -950,7 +950,7 @@ exec bash --norc --noprofile
         "warp" => launch_macos_warp(&script_file),
         "alacritty" => launch_macos_open_app("Alacritty", &script_file, true),
         "kitty" => launch_macos_open_app("kitty", &script_file, false),
-        "ghostty" => launch_macos_open_app("Ghostty", &script_file, true),
+        "ghostty" => launch_macos_ghostty(&script_file),
         "wezterm" => launch_macos_open_app("WezTerm", &script_file, true),
         "kaku" => launch_macos_open_app("Kaku", &script_file, true),
         _ => launch_macos_terminal_app(&script_file), // "terminal" or default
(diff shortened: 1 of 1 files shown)


farion1231:
@codex review

farion1231 (approved):

PR #2817, opened 2026-05-15
fix(proxy): convert upstream errors to Anthropic format in transform path

leeyang1990:
## Summary

When cc-switch proxies requests with `needs_transform=true` (e.g., OpenAI-format providers used through Claude Code), upstream errors are forwarded verbatim in OpenAI format. Claude Code cannot parse these non-Anthropic error responses, causing it to silently retry indefinitely — resulting in tool execution hangs.

## Changes

- Added `map_proxy_error_to_anthropic()` in `src-tauri/src/proxy/error.rs` to convert upstream error responses into Anthropic's expected `{"type":"error","error":{"type":"...","message":"..."}}` format
- Added `extract_openai_error_message()` helper to parse OpenAI-style error bodies
- Applied the conversion in `src-tauri/src/proxy/handlers.rs` when `needs_transform=true` and the forwarder returns an error
- Added 7 unit tests covering various error format scenarios

## Test Plan

- [x] `cargo test` — all 7 new tests pass
- [x] Built release binary and live-tested with Claude Code + LiteLLM proxy
- [x] Verified: tool_choice conversion works (existing fix by author)
- [x] Verified: upstream errors now properly surfaced to Claude Code instead of hanging
- [x] Verified: streaming responses unaffected

### DIFF
--- a/src-tauri/src/proxy/error.rs
+++ b/src-tauri/src/proxy/error.rs
@@ -204,3 +204,218 @@ pub fn categorize_error(error: &reqwest::Error) -> ErrorCategory {
         ErrorCategory::Retryable
     }
 }
+
+/// Convert a ProxyError into the Anthropic error format when the response
+/// is destined for a Claude Code client that went through a format-transform
+/// path (openai_chat / openai_responses / gemini_native).
+///
+/// Claude Code expects errors in Anthropic shape:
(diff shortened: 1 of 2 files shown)


farion1231:
@codex review

PR #2822, opened 2026-05-16
[codex] Fix Codex CLI detection for mise-managed Node installs

iambinlin:
## Summary

Fix false-negative local Codex CLI detection in CC Switch when Codex is installed through a mise-managed Node runtime.

In this setup, `codex --version` works in the user's terminal, but CC Switch may show `not installed or not executable` in Settings -> About -> Local Environment Check.

image

## Problem

Tauri GUI apps often start with a reduced PATH compared with an interactive terminal session. When direct execution of `codex --version` fails, CC Switch falls back to scanning common CLI install locations.

Before this change, the fallback scan covered locations such as:

- `~/.local/bin`
- `~/.npm-global/bin`
- `~/n/bin`
- Volta
- fnm
- nvm

However, it did not cover mise-managed Node installs, such as:

- `~/.local/share/mise/shims`
- `~/.local/share/mise/installs/node/<version>/bin`

As a result, Codex could be installed and usable from the terminal while still being reported as unavailable by CC Switch.

## Changes

- Add mise Node search path support to the local CLI fallback scanner:
  - `~/.local/share/mise/shims`
  - `~/.local/share/mise/installs/node/*/bin`
- Reuse the existing path de-duplication helper.
- Add a unit test covering mise shims and installed Node bin directories.

## Validation

Ran:

```bash
cargo fmt
cargo test commands::misc::tests
```

Result:

- 14 passed
- 0 failed

Manual verification:

- Codex is installed under a mise-managed Node directory.
- After the fix, CC Switch detects the local Codex version correctly in the About page environment check.

## Risk

Low.

This change only extends fallback CLI search paths. It does not change:

- version parsing
- remote latest-version checks
- frontend rendering
- provider switching
- config read/write behavior
- WSL detection behavior

## Scope

Affects local tool version detection fallback logic on non-Windows platforms.

Expected impact:

- Codex installed via mise is now detected correctly.
- Existing Claude, Gemini, OpenCode, nvm, fnm, Volta, and standard PATH detection behavior remains unchanged.

### DIFF
--- a/src-tauri/src/commands/misc.rs
+++ b/src-tauri/src/commands/misc.rs
@@ -560,6 +560,27 @@ fn tool_executable_candidates(tool: &str, dir: &Path) -> Vec<std::path::PathBuf>
     }
 }
 
+fn extend_mise_node_search_paths(paths: &mut Vec<std::path::PathBuf>, home: &Path) {
+    if home.as_os_str().is_empty() {
+        return;
+    }
+
+    let mise_base = home.join(".local/share/mise");
+    push_unique_path(paths, mise_base.join("shims"));
+
+    let node_installs = mise_base.join("installs").join("node");
+    if node_installs.exists() {
+        if let Ok(entries) = std::fs::read_dir(&node_installs) {
+            for entry in entries.flatten() {
+                let bin_path = entry.path().join("bin");
+                if bin_path.exists() {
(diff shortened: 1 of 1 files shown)


farion1231:
@codex review

farion1231 (approved):

PR #2831, opened 2026-05-16
fix: respect default profile in Windows Terminal launch

### LINKED ISSUE #2830: Windows Terminal option ignores default profile and always opens cmd on Windows
### Self Checks / 自检

### CC Switch Version / 版本号

3.15.0

### Operating System / 操作系统

Windows

### Related App / 涉及应用

Claude Code

### Steps to Reproduce / 重现步骤

Steps to Reproduce
Open Windows Terminal settings.
Set the default profile to PowerShell 7.x.
Open CC Switch.
Go to Settings -> Preferred Terminal.
Select Windows Terminal.
Click the terminal button from a provider card.

### Expected Behavior / 期望行为

Expected Behavior
CC Switch should open Windows Terminal using the user's default profile, which in this case is PowerShell 7.x.

Actual Behavior
CC Switch opens Windows Terminal, but the tab is cmd instead of the default PowerShell profile.

Sometimes the launched terminal also shows:

'/d' is not recognized as an internal or external command,
operable program or batch file.
Source-level Cause
It looks like the Windows Terminal branch is hardcoded to launch cmd:

src-tauri/src/commands/misc.rs
Current code:

"wt" => run_windows_start_command(&["wt", "cmd", "/K", &bat_path], "Windows Terminal"),
This appears in at least two places in the current codebase.

Because wt is explicitly followed by cmd, Windows Terminal never gets a chance to use its default profile.

### Actual Behavior / 实际行为

Actual Behavior
CC Switch opens Windows Terminal, but the tab is cmd instead of the default PowerShell profile.

Sometimes the launched terminal also shows:

'/d' is not recognized as an internal or external command,
operable program or batch file.

### Summary

When `Preferred Terminal` is set to `Windows Terminal`, CC Switch opens **Windows Terminal with a `cmd` tab**, instead of respecting the user's default Windows Terminal profile.

On my machine, Windows Terminal's default profile is **PowerShell 7.6.1**, but clicking the terminal button in CC Switch still opens `cmd`.

In some cases, the launched terminal also shows an error like:

```text
'/d' is not recognized as an internal or external command,
operable program or batch file.

### PR DESCRIPTION
**Pull request by lihan3238**
## Summary

- Removes hardcoded `"cmd"` from the `wt` (Windows Terminal) launch arguments so that Windows Terminal uses the user's configured default profile instead of always opening cmd.exe.
- The batch file is still passed as the command to execute — Windows Terminal will invoke it through whatever shell the default profile specifies (PowerShell, cmd, etc.).
- Affected 2 locations in `src-tauri/src/commands/misc.rs` (lines 1300, 1545).

Closes #2830

## Before

```rust
"wt" => run_windows_start_command(&["wt", "cmd", "/K", &bat_path], "Windows Terminal"),
```

This forced Windows Terminal to always open a `cmd` tab, ignoring the user's default profile setting.

## After

```rust
"wt" => run_windows_start_command(&["wt", &bat_path], "Windows Terminal"),
```

Windows Terminal now opens using the default profile and executes the setup batch file through that profile's shell.

## Test Plan

- [x] Code change is minimal and syntactically correct (2 array element removals)
- [ ] Manual test on Windows: set default profile to PowerShell 7, click terminal button, verify PowerShell opens instead of cmd
- [ ] Manual test on Windows: with cmd as default profile, verify existing behavior is preserved

### DIFF
--- a/src-tauri/src/commands/misc.rs
+++ b/src-tauri/src/commands/misc.rs
@@ -1297,7 +1297,29 @@ del \"%~f0\" >nul 2>&1
             &["powershell", "-NoExit", "-Command", &ps_cmd],
             "PowerShell",
         ),
-        "wt" => run_windows_start_command(&["wt", &bat_path], "Windows Terminal"),
+        "wt" => {
+            // Run claude directly in the default WT profile, bypassing the
+            // CMD-specific bat file so the profile shell is preserved.
+            let config_str = config_file.to_string_lossy().to_string();
+            let mut wt_args = vec!["wt".to_string(), "new-tab".to_string()];
+            if let Some(dir) = cwd {
+                wt_args.push("-d".to_string());
+                wt_args.push(dir.to_string_lossy().to_string());
+            }
+            wt_args.push("--".to_string());
+            wt_args.push("claude".to_string());
+            wt_args.push("--settings".to_string());
+            wt_args.push(config_str);
+            let arg_refs: Vec<&str> = wt_args.iter().map(|s| s.as_str()).collect();
+            let cfg = config_file.to_path_buf();
+            let bat = bat_file.to_path_buf();
+            std::thread::spawn(move || {
+                std::thread::sleep(std::time::Duration::from_secs(5));
+                let _ = std::fs::remove_file(&cfg);
+                let _ = std::fs::remove_file(&bat);
+            });
+            run_windows_start_command(&arg_refs, "Windows Terminal")
+        }
         _ => run_windows_start_command(&["cmd", "/K", &bat_path], "cmd"), // "cmd" or default
     };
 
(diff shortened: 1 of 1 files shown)


farion1231:
@codex review

lihan3238:
Good catch. Updated in 8c74923c with two improvements:

1. For the claude launch path (launch_windows_terminal): now bypasses the CMD-specific bat file entirely and runs claude directly via wt new-tab -- claude --settings, using -d for cwd. The default profile shell is fully preserved.
2. For the generic command path (launch_terminal_running): uses explicit wt new-tab -- syntax to ensure WT uses the default profile instead of matching a CMD profile.

Temp file cleanup is handled via a delayed background thread since the bat file self-delete is no longer in play for the WT case.

h1557066969-afk:
👍

PR #2839, opened 2026-05-16
Fix Linux desktop integration and GNOME window startup

xiaoyueyoqwq:
## Summary

This PR fixes several Linux desktop integration and GNOME/Wayland window startup issues observed on ZorinOS 18.1 with the native GTK/Wayland backend.

### Desktop integration

- Enables the GTK application id so Linux desktops can associate the runtime window with `com.ccswitch.desktop`.
- Sets the main window title to `CC Switch` instead of an empty title.
- Adds a deb/rpm desktop template with `StartupWMClass=cc-switch`.
- Changes the Linux bundle category to `DeveloperTool`, producing development-oriented desktop categories.
- Updates the Flatpak desktop entry to use `Development;IDE;` and `StartupWMClass=cc-switch`.

### GNOME/Wayland window startup

- Keeps the native GTK/GNOME titlebar instead of forcing app-managed window controls.
- Removes the previous Linux resize nudge that resized the window by 1px and then restored it. On GNOME/Wayland this could inflate the startup window and pollute saved window state.
- After Tauri shows the window, refreshes GTK presentation with `gtk_window.show_all()` and `gtk_window.present()`. This fixes the native titlebar button hit-testing issue where minimize/maximize/close could be inert until the user manually maximized/restored the window.
- Keeps a delayed focus retry to avoid first-click focus loss.

### Linux window size memory

- On Linux, the Tauri window-state plugin now persists only maximized state.
- Adds separate Linux-only normal window size memory.
- Does not persist or restore window position, matching Wayland expectations.
- Ignores startup resize noise and skips saving maximized/minimized/hidden window sizes.
- Clamps restored size to the current monitor work area.
- Saves the latest normal size before close-to-tray hiding and app exit.

## Verification

- `cargo check` passes in `src-tauri`.
- `pnpm tauri build --bundles deb` produced a deb package successfully:
  - `src-tauri/target/release/bundle/deb/CC Switch_3.15.0_amd64.deb`
  - The command exits non-zero only at updater signing because the local environment has a public updater key but no `TAURI_SIGNING_PRIVATE_KEY`.
- The generated deb was manually installed and tested on ZorinOS 18.1 / GNOME Wayland:
  - app appears in the expected development app category
  - window identity is registered correctly
  - cold-start window size is normal
  - native GNOME titlebar buttons work immediately
  - Linux normal window size memory works without restoring position

## Notes

- Runtime diagnostics showed a Wayland backend (`WaylandWindowHandle`), so X11-only diagnostics such as `xprop` and `xwininfo` are not useful for this issue.
- A client-side/app-managed titlebar workaround was tested and rejected because it changed native GNOME behavior and regressed window dragging.

### DIFF
--- a/flatpak/com.ccswitch.desktop.desktop
+++ b/flatpak/com.ccswitch.desktop.desktop
@@ -5,5 +5,6 @@ Comment=All-in-One Assistant for Claude Code, Codex & Gemini CLI
 Exec=cc-switch
 Icon=com.ccswitch.desktop
 Terminal=false
-Categories=Utility;Development;
+Categories=Development;IDE;
 StartupNotify=true
+StartupWMClass=cc-switch
--- a/src-tauri/Cargo.lock
+++ b/src-tauri/Cargo.lock
@@ -749,6 +749,7 @@ dependencies = [
  "dirs 5.0.1",
  "flate2",
  "futures",
+ "gtk",
  "http",
  "http-body",
  "http-body-util",
--- a/src-tauri/Cargo.toml
+++ b/src-tauri/Cargo.toml
@@ -84,6 +84,7 @@ json-five = "0.3.1"
 tauri-plugin-single-instance = "2"
 
 [target.'cfg(target_os = "linux")'.dependencies]
+gtk = "0.18.2"
 webkit2gtk = { version = "2.0.1", features = ["v2_16"] }
 
 [target.'cfg(target_os = "windows")'.dependencies]
--- a/src-tauri/desktop/cc-switch.desktop.hbs
+++ b/src-tauri/desktop/cc-switch.desktop.hbs
@@ -1,0 +1,10 @@
+[Desktop Entry]
+Type=Application
+Name={{name}}
+Comment={{comment}}
+Exec={{exec}}
+Icon={{icon}}
+Terminal=false
+Categories={{categories}}
+StartupNotify=true
+StartupWMClass=cc-switch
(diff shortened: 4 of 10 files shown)


xiaoyueyoqwq:
Linux tray restore note:

During testing on ZorinOS 18.1 / GNOME Wayland, the initial startup titlebar-button issue and the tray-restore titlebar-button issue looked similar, but they turned out to be different paths.

Cold startup was fixed by showing the Tauri window first, then refreshing the native GTK window with `gtk_window.show_all()` + `gtk_window.present()` and a delayed focus/present retry. That handles the first realization/mapping of the GTK/Wayland window.

The remaining tray-restore issue happened after the main window had already been realized once and was later hidden to tray. Logs showed that after tray restore the GTK window, content child, and titlebar were all visible/mapped/realized/sensitive/focused, but the native GNOME titlebar buttons could still be inert until the window received a real window-manager configure event. Manually double-clicking the titlebar to maximize/restore reliably fixed it.

We tested several restore-side workarounds:
- Repeating `show_all()` / `present()` / focus did not fix tray restore.
- A forced maximize/unmaximize pulse fixed it, but was visually disruptive and not acceptable UX.
- Same-size GTK resize / queue-resize did not fix it and could reintroduce window-size growth.
- A 1px resize pulse did not fix it either.

The least disruptive workaround we found is Linux-specific: when close-to-tray is enabled, use `minimize()` instead of `hide()` for the main window. This avoids the hide/show remap path that leaves GNOME/Wayland titlebar hit-testing stale. The tradeoff is that desktop environments with Windows-like taskbars, such as ZorinOS, may show CC Switch as a minimized/folded taskbar item instead of a fully hidden tray-only window. Given the native-titlebar regression, this is currently the safer Linux behavior.

If anyone got a better idea welcome to contribution code, thanks! :)

farion1231:
@codex review

xiaoyueyoqwq:
经过基于GNOME50的验证该问题在v3.20.0已经得到解决

PR #2852, opened 2026-05-17
feat(proxy): Codex Chat Compat mode - DeepSeek and other Chat Completions providers

OmMaBaMiHong:
## Summary

Adds a **Chat Compat mode** that enables cc-switch to act as a proxy for Codex CLI when the upstream provider only supports **Chat Completions API** (e.g. DeepSeek, Qwen, GLM, Moonshot, etc).

## How it works

When `chat_compat: true` is set in the provider config, two bidirectional adapters are activated:

1. **Request path** (`responses_to_chat.rs`): Converts Codex CLI's Responses API format → Chat Completions format before forwarding
2. **Response path** (`chat_to_responses.rs`): Converts upstream Chat Completions response → Responses API SSE events for Codex CLI

## DeepSeek-specific fixes included

- Forces non-streaming for DeepSeek to simplify response handling
- Prevents infinite retry loop when DeepSeek returns errors
- Reorders interleaved messages and handles dangling `tool_calls`
- Matches `tool_calls` to `outputs` to prevent "insufficient" error
- Safe UTF-8 char boundary truncation to prevent panics on multi-byte strings

## Usage

Add `chat_compat: true` to your DeepSeek (or other Chat Completions-only) provider config in cc-switch, then use it as a proxy for Codex CLI.


suyunzzz:
在哪里开启，我现在是使用一个python脚本做的转发

OmMaBaMiHong:
image

farion1231:
@codex review

Ken-Leo:
The reason why /compact or autocompact fails for the 3rd-party model API compaction calling seems to be that Codex tries to send users' session content back to its models for training or other uses. The right way is that the contexts should be streamed to their real upstream providers. Appreciate it! @OmMaBaMiHong I also wonder if this fix also fits the Codex Desktop app?

Yuxin-Qiao:
Can confirm this compact endpoint issue also affects MiniMax (and likely all Chat API providers — GLM, Qwen, etc.) when used as Codex providers through Chat Compat mode.

Observed on 3.16.2: `/v1/responses/compact` gets routed through `handle_codex_chat_to_responses_transform()`, forwarded as a normal chat request to MiniMax's Chat API, which of course returns a normal response with zero compaction. Codex believes compaction succeeded while the context window keeps growing until it hits the model limit.

The compact rewrite fix in this PR would resolve it for all Chat Compat providers, not just DeepSeek. Hope this gets merged soon — happy to help test with MiniMax.

PR #2887, opened 2026-05-18
[codex] Sync Codex env key for desktop app

Suda202:
## Summary
- Sync the active Codex provider `env_key` into the macOS launch environment after successful live config writes.
- Unset stale launch environment keys when switching away from a custom provider or when the configured key has no auth value.
- Keep non-macOS behavior unchanged.

## Why
Codex Desktop launched from Finder or Dock reads the macOS launch environment, not the shell environment. When CC Switch writes a custom Codex provider config with an `env_key`, the desktop app can fail to see the matching API key unless that key is also synced into launchd.

## Validation
- `cargo test codex_launch_env_actions`
- `cargo test provider_service_switch_codex`
- `git diff --check`

### DIFF
--- a/src-tauri/src/codex_config.rs
+++ b/src-tauri/src/codex_config.rs
@@ -9,6 +9,8 @@ use crate::error::AppError;
 use serde_json::Value;
 use std::fs;
 use std::path::Path;
+#[cfg(target_os = "macos")]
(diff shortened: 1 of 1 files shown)


farion1231:
@codex review

PR #2928, opened 2026-05-19
fix(deeplink): preserve custom env fields when importing Claude provi…

### LINKED ISSUE #2927: [Bug] Deep link 导入后，config 中的自定义 env 字段（如 ANTHROPIC_CUSTOM_HEADERS）未被保存
### Self Checks / 自检

### CC Switch Version / 版本号

3.15.0

### Operating System / 操作系统

macOS

### Related App / 涉及应用

Claude Code

### Steps to Reproduce / 重现步骤

1. 构造如下 Deep Link：
```
ccswitch://v1/import?resource=provider&app=claude&name=My+Provider&endpoint=https://your-proxy.example.com&apiKey=sk-xxx&model=custom-model&config=<base64({"env":{"ANTHROPIC_CUSTOM_HEADERS":"Cookie: session=abc","ANTHROPIC_DEFAULT_HAIKU_MODEL":"fast-model","API_TIMEOUT_MS":"3000000"}})>&configFormat=json
```

2. 打开该 Deep Link，CC Switch 弹出导入预览弹窗。
3. 3. 预览弹窗中可以看到 `ANTHROPIC_CUSTOM_HEADERS` 等所有 env 字段。
4. 4. 点击『确认导入』。
5. 5. 在 Provider 编辑器中查看刚导入的 Provider 配置。

### Expected Behavior / 期望行为

预览弹窗中展示的所有 env 字段，在用户确认导入后应全部持久化到数据库。预览内容与实际写入结果应保持一致。

建议在 `src-tauri/src/deeplink/provider.rs` 的导入逻辑中，将 `config` 参数解码后的完整 `env` 块与 query 参数映射的标准字段**合并写入**，而非丢弃非标准字段。

### Actual Behavior / 实际行为

`ANTHROPIC_CUSTOM_HEADERS` 及其他通过 `config` 传入的自定义字段**均未出现**在已保存的 Provider 配置中，只有标准字段被写入：

```json
{
  "env": {
    "ANTHROPIC_AUTH_TOKEN": "sk-xxx",
    "ANTHROPIC_BASE_URL": "https://your-proxy.example.com",
    "ANTHROPIC_MODEL": "custom-model"
  }
}
```

`ANTHROPIC_CUSTOM_HEADERS`、`ANTHROPIC_DEFAULT_HAIKU_MODEL`、`API_TIMEOUT_MS` 等均丢失。

### Additional Context / 补充信息

v3.11.0 曾出现过类似的数据丢失问题：Provider 切换时非白名单自定义字段被静默丢弃，该问题已在 v3.12.0 回退修复。此 Issue 描述的是 Deep Link 导入场景下的同类问题——`config` 参数中的额外 env 字段在预览中可见，但实际未被保存。

已在 v3.14.1 和 v3.15.0 两个版本复现。

相关源码位置：`src-tauri/src/deeplink/provider.rs`

doutuifei:
## Summary

Fixes #2927. When importing a Claude provider via deep link, the preview dialog correctly shows every env field carried in the `config` payload — but on confirmation `build_claude_settings` only persisted the standard fields mapped from URL params (`apiKey`, `endpoint`, `model`, `haiku/sonnet/opusModel`). Any non-standard env keys such as `ANTHROPIC_CUSTOM_HEADERS`, `API_TIMEOUT_MS`, or `CLAUDE_CODE_DISABLE_EXPERIMENTAL_BETAS` were silently dropped, making the preview and the actual write diverge.

## Approach

Start the env map from the inline `config` payload (best-effort decode via the new `extract_claude_config_env` helper), then overlay the URL-param-derived standard fields on top. The net effect is that URL params remain authoritative for standard Claude fields (unchanged), custom env keys present in `config` now survive the import (the fix), and deep links without a `config` payload behave exactly as before (backward compatible).

No `pub` API surface is changed — `parse_and_merge_config` (used by the preview command in `commands/deeplink.rs`) keeps its current signature. The decode happens inside `build_claude_settings` via a private helper that returns `None` on any failure, falling back gracefully to URL-param-only behavior.

## Scope

This PR fixes the Claude path only (`build_claude_settings`). Codex / Gemini / OpenCode / OpenClaw / Hermes have similar patterns but different config shapes (TOML auth, npm/options/models, snake_case/camelCase YAML), so they need targeted fixes rather than a blanket change. Happy to follow up in separate PRs if the maintainer wants them addressed.

## Tests

Adds two regression tests in `src-tauri/src/deeplink/tests.rs`. `test_build_claude_provider_preserves_custom_env_fields` confirms `ANTHROPIC_CUSTOM_HEADERS`, `API_TIMEOUT_MS`, `CLAUDE_CODE_DISABLE_EXPERIMENTAL_BETAS` survive import, and that URL params override same-named config fields (e.g. URL `haikuModel` wins over `ANTHROPIC_DEFAULT_HAIKU_MODEL` in config). `test_build_claude_provider_without_config_unchanged` guards the backward-compatible path: a deep link without `config` produces exactly the same env shape as before (`ANTHROPIC_AUTH_TOKEN` + `ANTHROPIC_BASE_URL` only, nothing else).

Results:

```
cargo test --lib deeplink   → 27 passed, 0 failed
cargo test --lib            → 1184 passed, 0 failed
cargo clippy --lib --tests -- -D warnings   → clean
cargo fmt --check           → clean
```

## Manual verification

Built locally with `pnpm tauri dev` on macOS, ran a deep link carrying `ANTHROPIC_CUSTOM_HEADERS` plus the other custom env fields, and confirmed in the provider editor that all fields shown in the preview are persisted after clicking Import. Verified on cc-switch v3.15.0 (same behavior reproduces on v3.14.1).

## Notes for review

Re-decoding the base64 `config` payload inside `build_claude_settings` is intentional: `parse_and_merge_config` is a `pub` function used independently for preview, so threading the decoded value through every `build_*` function would be invasive. Decode cost is negligible on this one-shot import path.

`extract_claude_config_env` uses silent fallback (`?` on every fallible op) because `parse_and_merge_config` has already validated the payload upstream — if anything fails here it's safer to fall back to URL-param-only behavior than to surface a new error class on a path that previously succeeded.

### DIFF
--- a/src-tauri/src/deeplink/provider.rs
+++ b/src-tauri/src/deeplink/provider.rs
@@ -244,8 +244,23 @@ fn build_provider_meta(request: &DeepLinkImportRequest) -> Result<Option<Provide
 }
 
 /// Build Claude settings configuration
+///
+/// Merges env from the inline config (if any) with the standard fields from URL params.
+/// URL params take priority — they overwrite same-named fields from the config.
+/// Non-standard env fields (e.g. `ANTHROPIC_CUSTOM_HEADERS`, `API_TIMEOUT_MS`,
+/// `CLAUDE_CODE_DISABLE_EXPERIMENTAL_BETAS`, ...) are preserved as-is so that
+/// providers requiring extra environment variables work after deeplink import.
 fn build_claude_settings(request: &DeepLinkImportRequest) -> serde_json::Value {
-    let mut env = serde_json::Map::new();
+    // Start from the full env block in the inline config (if present), so any
+    // custom env vars the user passed via `config=<base64-json>` survive the
+    // import. Falling back to an empty map keeps the previous behavior for
+    // deeplinks that don't carry a config field.
+    let mut env = extract_claude_config_env(request).unwrap_or_default();
+
+    // Now overwrite / fill in the standard fields from URL params. URL params
+    // are authoritative because they're what the deeplink builder put on the
+    // wire — for Claude these are: ANTHROPIC_AUTH_TOKEN, ANTHROPIC_BASE_URL,
+    // ANTHROPIC_MODEL, and the haiku/sonnet/opus model aliases.
     env.insert(
         "ANTHROPIC_AUTH_TOKEN".to_string(),
         json!(request.api_key.clone().unwrap_or_default()),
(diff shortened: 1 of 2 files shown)


farion1231:
@codex review

farion1231 (approved):

PR #2990, opened 2026-05-20
feat: add MiMo reasoning_content support for Claude Code proxy

zhangyapu1:
## Summary

MiMo requires `reasoning_content` to be passed back in multi-turn agent conversations with tool calls (see [MiMo docs](https://platform.xiaomimimo.com/docs/zh-CN/usage-guide/passing-back-reasoning_content)). Without it, the API returns 400.

cc-switch already has full `thinking` → `reasoning_content` conversion logic in `transform.rs`, gated by `preserve_reasoning_content` in `claude.rs`. However, the whitelist `is_reasoning_content_compatible_identifier` only includes `moonshot`, `kimi`, and `deepseek` — MiMo is missing.

### Changes

**`claude.rs`** — Add `mimo` and `xiaomimimo` to the `is_reasoning_content_compatible_identifier` whitelist so `preserve_reasoning_content = true` when proxying to MiMo.

**`transform.rs`** — Handle `redacted_thinking` blocks: Claude Code encrypts historical thinking into `redacted_thinking` across turns. Without handling, these blocks are silently dropped, leaving `reasoning_content` empty and triggering MiMo's 400. When `preserve_reasoning_content` is on, inject a `[redacted thinking]` placeholder.

## Test plan

- [ ] Configure cc-switch with a MiMo provider (`apiFormat: "openai_chat"`, base_url containing `xiaomimimo`)
- [ ] Use Claude Code through cc-switch to have a multi-turn conversation with tool calls
- [ ] Verify no 400 errors from MiMo on follow-up turns
- [ ] Verify `reasoning_content` is present in the converted request body (check cc-switch logs)

### DIFF
--- a/src-tauri/src/claude_desktop_config.rs
+++ b/src-tauri/src/claude_desktop_config.rs
@@ -21,6 +21,8 @@ const CONFIG_LIBRARY_DIR: &str = "configLibrary";
 const GATEWAY_TOKEN_SETTING_KEY: &str = "claude_desktop_gateway_token";
 const CLAUDE_DESKTOP_PROXY_PREFIX: &str = "/claude-desktop";
 const DEFAULT_CREATED_AT: &str = "2024-01-01T00:00:00Z";
+const MIMO_REDACTED_THINKING_PLACEHOLDER: &str = "[redacted thinking]";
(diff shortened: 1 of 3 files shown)


farion1231:
@codex review

farion1231 (approved):

PR #3027, opened 2026-05-21
feat(usage): real-time stats refresh + fix codex sync panic on non-ASCII model names

in30mn1a:
## Summary

This PR addresses a long-standing issue where the usage statistics dashboard only refreshed after restarting cc-switch. Diagnosing this surfaced two root causes — both fixed here.

## Problem 1: Codex session sync panics on non-ASCII model names

`normalize_codex_model` in `services/session_usage_codex.rs` slices the last 11 bytes of the model name to detect an ISO date suffix:

```rust
let suffix = &name[name.len() - 11..];
```

For a model name containing multi-byte UTF-8 characters (real-world example from my config: `【官】glm-5.1`), `name.len() - 11` lands inside a multi-byte char, panicking the tokio worker:

```
thread 'tokio-rt-worker' panicked at services/session_usage_codex.rs:79:27:
start byte index 5 is not a char boundary; it is inside '官' (bytes 3..6) of `【官】glm-5.1`
```

Once the background sync task panics, **no Codex session logs are imported** until the app is restarted (where `Database::init` → `rollup_and_prune` happens to flush pending data). This is the actual reason stats appear to "only refresh on restart" for users who don't route through the cc-switch proxy.

**Fix**: Guard the slice with `is_char_boundary` + `is_ascii` (a valid date suffix is always ASCII, so non-ASCII names can safely bypass the slicing).

## Problem 2: Dashboard doesn't reflect new logs in real time

Even after fixing Problem 1, the `UsageDashboard` only polls every 30s and disables `refetchIntervalInBackground`, so freshly-imported data is invisible until the next poll or window refocus.

**Fix**: Push-based refresh.

- New `usage_events` module emits a `usage-log-recorded` event whenever `proxy_request_logs` actually gains a new row. Coverage:
  - Proxy `log_request` (streaming + non-streaming, success + error)
  - Claude / Codex / Gemini session sync (only when a row is truly inserted, so `INSERT OR IGNORE` skips don't trigger empty refreshes)
  - Startup `rollup_and_prune`
- 200ms debounce inside `notify_log_recorded` collapses bursts. A Codex sync importing 3000+ entries produces ~2 emits, not 3000.
- Global `OnceLock<AppHandle>` lets call sites that don't already hold an `AppHandle` (e.g. `UsageLogger`) notify without signature churn.
- Frontend `useUsageEventBridge` listens for the event and invalidates `usageKeys.all`. Mounted only on `UsageDashboard` so the listener is unsubscribed automatically when navigating away.

## Verification

- ✅ `cargo check` passes (existing 25 dead-code warnings in `commands/misc.rs` are pre-existing and unrelated)
- ✅ `tsc --noEmit` passes
- ✅ Manually verified end-to-end: with my actual config containing the panic-triggering model name, Codex sync imported 3145 entries successfully, produced 2 debounced emits (both `emit usage-log-recorded 成功` in logs), and the dashboard updated within ~200ms.

## Files

| Type | File |
|------|------|
| 🐛 Fix | `src-tauri/src/services/session_usage_codex.rs` (the panic) |
| ✨ New | `src-tauri/src/usage_events.rs` (event module) |
| ✨ New | `src/hooks/useUsageEventBridge.ts` (frontend listener) |
| 🔧 Edit | `src-tauri/src/lib.rs` (init in `setup`) |
| 🔧 Edit | `src-tauri/src/proxy/usage/logger.rs` (notify on `log_request`) |
| 🔧 Edit | `src-tauri/src/services/session_usage{,_codex,_gemini}.rs` (notify on insert) |
| 🔧 Edit | `src-tauri/src/database/dao/usage_rollup.rs` (notify on prune) |
| 🔧 Edit | `src/components/usage/UsageDashboard.tsx` (mount the bridge hook) |

## Behaviour notes

- The session-sync notify paths only fire when a row is actually inserted, so dedup-skipped writes don't cause unnecessary frontend refreshes.
- Gemini's `INSERT … ON CONFLICT … DO UPDATE` path reuses the existing `conn.changes() > 0` check.
- AppHandle is injected after the log plugin is registered, so the init log line is captured in `cc-switch.log`.

### DIFF
--- a/src-tauri/src/database/dao/usage_rollup.rs
+++ b/src-tauri/src/database/dao/usage_rollup.rs
@@ -89,6 +89,9 @@ impl Database {
                     log::info!(
                         "Rolled up and pruned {deleted} proxy_request_logs (retain={retain_days}d)"
                     );
+                    // 归档触发了表结构变化，前端 30 天前的统计可能跟着变，
+                    // 通知一次让 UsageDashboard 重拉数据
+                    crate::usage_events::notify_log_recorded();
                 }
                 Ok(deleted)
             }
--- a/src-tauri/src/lib.rs
+++ b/src-tauri/src/lib.rs
@@ -32,6 +32,7 @@ mod settings;
 mod store;
 
 mod tray;
+mod usage_events;
 mod usage_script;
 
 pub use app_config::{AppType, InstalledSkill, McpApps, McpServer, MultiAppConfig, SkillApps};
@@ -334,6 +335,11 @@ pub fn run() {
                         .build(),
                 )?;
             }
+
+            // 注入 AppHandle 给 usage_events，让无 AppHandle 持有的写日志路径
+            // 也能向前端推送 `usage-log-recorded`。
+            // 放在日志系统初始化之后，确保 init 的日志能正常输出。
+            usage_events::init(app.handle().clone());
 
             // 初始化数据库
             let app_config_dir = crate::config::get_app_config_dir();
--- a/src-tauri/src/proxy/usage/logger.rs
+++ b/src-tauri/src/proxy/usage/logger.rs
@@ -102,6 +102,9 @@ impl<'a> UsageLogger<'a> {
         )
         .map_err(|e| AppError::Database(format!("记录请求日志失败: {e}")))?;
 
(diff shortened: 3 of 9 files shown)


farion1231:
@codex review

farion1231 (approved):

PR #3047, opened 2026-05-22
fix: harden Codex chat tool conversion

mittyok:
## Problem

When Codex App is configured to use an OpenAI-compatible Chat Completions upstream, CC Switch converts Codex `Responses API` requests into Chat Completions requests.

Some Codex workflows, including generating commit messages, rely on tool calls. In those flows the Responses input may contain `function_call` / `function_call_output` items that are not a perfectly valid Chat Completions tool-message block after a naive conversion.

This can produce invalid Chat Completions message sequences such as:

- an `assistant.tool_calls` message without matching following `tool` messages
- a standalone `role=tool` message without a preceding assistant tool call
- a delayed/interleaved tool output after a normal user/assistant message

Many OpenAI-compatible Chat Completions providers reject those sequences with 400-level errors, which makes Codex features such as commit-message generation fail.

## Root Cause

The previous Responses → Chat conversion treated tool call items too independently:

- `function_call` could be converted into `assistant.tool_calls` even when no valid output followed.
- `function_call_output` could become a `role=tool` message even when it was orphaned or separated from its matching call.
- The converter did not fully enforce the Chat Completions invariant that every `assistant.tool_calls` block must be immediately followed by corresponding `tool` messages.

Responses API input is more flexible as a sequence of typed items, but Chat Completions has stricter adjacency and pairing rules for tool calls.

## Solution

This PR hardens the Codex Responses → Chat Completions conversion by converting tool items as contiguous tool blocks instead of independent messages.

Changes include:

- scan consecutive `function_call` items as a candidate tool block
- only emit `assistant.tool_calls` for calls that have immediately following matching `function_call_output` items
- drop orphan `function_call` items instead of emitting invalid assistant tool calls
- drop orphan or interleaved `function_call_output` items instead of emitting standalone `role=tool` messages
- support both `call_id` and `id` when matching tool outputs
- lower the Codex conversion summary log from `warn` to `debug` to avoid noisy production logs

The goal is to prefer a valid, non-400 Chat Completions request over preserving invalid tool fragments that the upstream cannot accept.

## Validation

Ran targeted Rust tests:

```bash
cargo test transform_codex_chat --lib
```

Result:

- 10 passed
- 0 failed

New coverage includes:

- orphan function calls are dropped
- partial tool calls only keep paired calls
- interleaved tool output after a normal message is dropped
- orphan tool output is dropped
- tool output matching supports `id` fallback

## Risk

Low to medium.

This only affects Codex Responses → Chat Completions conversion. The change intentionally drops invalid/orphaned tool fragments that would otherwise cause Chat Completions upstreams to reject the request. In rare malformed histories, this may lose some tool context, but it avoids hard request failure and keeps the generated Chat message sequence protocol-valid.

### DIFF
--- a/package.json
+++ b/package.json
@@ -91,5 +91,6 @@
     "sonner": "^2.0.7",
     "tailwind-merge": "^3.3.1",
     "zod": "^4.1.12"
-  }
+  },
+  "packageManager": "pnpm@10.33.0+sha1.46a0bef28aad690765997ab6da6d5475bdd4148e"
 }
--- a/src-tauri/src/proxy/forwarder.rs
+++ b/src-tauri/src/proxy/forwarder.rs
@@ -1148,7 +1148,7 @@ impl RequestForwarder {
         let request_body = if codex_responses_to_chat {
             let input_items = mapped_body.get("input").and_then(|v| v.as_array());
             if let Some(items) = input_items {
-                log::warn!(
+                log::debug!(
                     "[codex->chat] converting {} input items: {}",
                     items.len(),
                     items.iter().enumerate().map(|(i, item)| {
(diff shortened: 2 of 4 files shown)


farion1231:
@codex review

PR #3060, opened 2026-05-22
fix: terminate stale Claude CLI processes after provider switch

### LINKED ISSUE #3057: Bug: Switching back to Claude Official still uses previous provider in active sessions
## Describe the bug

After switching from **Claude Official** (OAuth) to a third-party provider (e.g. DeepSeek), then switching back to **Claude Official**, the active Claude Code session continues to use the previous provider instead of the official Anthropic endpoint.

## Steps to reproduce

1. Start with Claude Official (OAuth) as the active provider
2. In CC Switch, switch to DeepSeek (or any third-party provider)
3. Confirm DeepSeek is working in the current Claude Code session
4. In CC Switch, switch back to Claude Official
5. CC Switch UI shows "使用中" (In Use) for Claude Official
6. Continue using the **same** Claude Code session — requests still go to DeepSeek

## Expected behavior

After switching back to Claude Official, the running Claude Code session should use the official Anthropic endpoint.

## Actual behavior

The running session still routes requests to the previously active provider (DeepSeek), even though `~/.claude/settings.json` has been correctly updated to `{"env": {}}` by CC Switch.

## Root cause analysis

CC Switch's `switch_normal` correctly overwrites `~/.claude/settings.json` with the new provider's config. The file on disk is updated properly.

However, Claude Code reads `settings.json` at **process startup** and injects the `env` section as process-level environment variables. Once the process is running, changes to `settings.json` are not picked up — the old env vars (e.g. `ANTHROPIC_BASE_URL=https://api.deepseek.com/anthropic`) remain in the process's memory for the lifetime of that session.

CC Switch's `switch_normal` function (in `src-tauri/src/services/provider/mod.rs`) ends with an MCP sync and returns — there is **no logic to kill, restart, or notify running Claude Code processes** after the switch.

## Suggested fix

After writing the new live config, CC Switch should kill any running `claude` processes (or send `SIGHUP`) so they restart and pick up the updated `settings.json`. This would make provider switches take effect immediately without requiring users to manually restart their terminal sessions.

## Workaround

Close the current Claude Code terminal session and open a new one after switching providers. The new process will read the updated `settings.json` and use the correct provider.

## Environment

- CC Switch version: 2026.5.19 (via macOS app bundle)
- macOS: Darwin 25.4.0
- Providers tested: Claude Official (OAuth) ↔ DeepSeek

eDramas:
## Problem

Closes #3057

After switching providers in CC Switch, the active Claude Code session continues to use the previous provider. **Closing the terminal window and reopening it does not help.**

## Deep-dive: why closing the terminal doesn't work

Claude Code uses a persistent background daemon that maintains a pool of pre-warmed `--bg-spare` worker processes. The daemon keeps running even after all terminal windows are closed.

**Each spare process reads `settings.json` at spawn time** and bakes the `env` section (e.g. `ANTHROPIC_BASE_URL`) into its own process environment. After that, no amount of file changes affects the already-spawned spare.

When a spare is claimed (assigned to a new session), the daemon immediately spawns a **replacement spare** that reads the *current* `settings.json`. So:

1. While DeepSeek is active → newly spawned spares pick up `ANTHROPIC_BASE_URL=https://api.deepseek.com/…`
2. CC Switch switches back to Claude Official → `settings.json` is correctly updated to `{"env":{}}`
3. **But the daemon's spare pool still contains idle spares that were spawned while DeepSeek was active**
4. A new session may be assigned one of those stale spares → still routes to DeepSeek

## How idle spares are distinguished from active sessions

The daemon's `spare/` directory contains socket files:

| State | Files present |
|---|---|
| Idle spare (not yet claimed) | `<id>.claim.sock` **and** `<id>.pty.sock` |
| Active session (already claimed) | only `<id>.pty.sock` (claim.sock removed by daemon) |

This makes it safe to target **only** processes whose `claim.sock` still exists — active sessions are completely unaffected.

## Fix

After `write_live_with_common_config` completes for `AppType::Claude`:

1. Scan `/tmp/cc-daemon-*/*/spare/` for `*.claim.sock` files
2. For each: run `pkill -TERM -f "--bg-spare <full-sock-path>"` — targets exactly the idle spare that owns that socket
3. The daemon detects the exited spares and re-spawns fresh ones that read the updated `settings.json`

Result: provider switch takes effect for all future sessions **without** closing any windows or killing the daemon.

Failures are non-fatal: a `"claude_spare_refresh_failed"` warning is added to `SwitchResult.warnings` and logged; the switch itself succeeds regardless.

## What was tried and rejected

An earlier attempt used `pkill -TERM -x claude` (exact process name match). This was **rejected** because the name `claude` matches not just spare processes but also `claude agents` (the UI), the daemon itself, and active sessions — causing the app window to close and ongoing work to be lost.

## Testing

- `cargo check` passes (rustc 1.95.0, darwin arm64)
- Logic verified against live process list and `~/.claude/daemon/roster.json` structure
- Idle spare identification confirmed by inspecting `/tmp/cc-daemon-501/<hash>/spare/` socket files on a live system
- Active sessions (those without `claim.sock`) are not targeted by the pkill pattern

### DIFF
--- a/src-tauri/src/services/provider/mod.rs
+++ b/src-tauri/src/services/provider/mod.rs
@@ -50,10 +50,6 @@ pub struct ProviderService;
 #[serde(rename_all = "camelCase")]
 pub struct SwitchResult {
     pub warnings: Vec<String>,
-    /// True when the app type reads settings.json at process startup (env vars are
-    /// baked into the process environment). The frontend should surface a prompt
-    /// asking the user to restart their active sessions.
-    pub requires_session_restart: bool,
 }
 
 #[cfg(test)]
@@ -1619,11 +1615,18 @@ impl ProviderService {
         // Sync MCP
         McpService::sync_all_enabled(state)?;
 
-        // Claude Code reads settings.json at process startup and bakes the env
-        // section into the process environment. Changing the file has no effect
-        // on already-running sessions. Signal the frontend to prompt the user.
+        // Idle spare processes were spawned before the provider switch and still
+        // hold the old provider's env vars (e.g. ANTHROPIC_BASE_URL). Kill only
+        // those idle spares so the daemon re-spawns fresh ones that read the
+        // updated settings.json. Active sessions (whose claim.sock is already
+        // taken) are unaffected.
         if matches!(app_type, AppType::Claude) {
-            result.requires_session_restart = true;
+            if let Err(e) = kill_idle_claude_spare_processes() {
+                log::warn!("Could not refresh Claude spare pool after provider switch: {e}");
+                result
+                    .warnings
+                    .push("claude_spare_refresh_failed".to_string());
+            }
         }
 
         Ok(result)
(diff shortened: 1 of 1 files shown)


farion1231:
@codex review

eDramas:
Good catch — fixed in 78d35442.

Added `--` before the pattern so `pkill` treats `--bg-spare …` as data rather than an option flag on both BSD (macOS) and procps (Linux):

\`\`\`rust
.args(["-TERM", "-f", "--", &format!("--bg-spare {sock_path}")])
\`\`\`

PR #3097, opened 2026-05-24
fix: improve usage detail long field display

michaelx1993:
## Summary
- make request detail dialogs wider with internal scrolling so long fields do not break layout
- render long values as compact expandable previews with copy actions
- remove native table tooltip for model labels to avoid huge hover popups

## Test
- ./node_modules/.bin/tsc --noEmit


farion1231:
@codex review

michaelx1993:
Closing this PR because the long-field display issue belongs to the sub3 Prompt request logs UI and has been fixed/released there instead.

PR #3129, opened 2026-05-25
[codex] Fix custom usage script summaries

### LINKED ISSUE #3128: Custom usage script test can show undefined and saved query can reject HTTP LAN base URL
## Problem

Custom usage scripts can pass the modal test but still fail or render poorly in the regular provider usage path.

Two cases reproduce this:

1. A custom extractor returns `used` and `total` but omits `remaining`, which is documented and typed as optional. The test success toast still formats only `remaining`, so it can display output like:

```text
Test successful![Coco OpenRouter] Remaining: undefined %
```

2. For a Claude provider whose `ANTHROPIC_BASE_URL` is a non-localhost HTTP relay, for example:

```text
http://10.37.192.156:8090/anthropic
```

and a custom usage script that directly calls a separate HTTP quota endpoint:

```js
({
  request: {
    url: "http://10.37.192.156:18344/user/balance",
    method: "GET",
    headers: {
      Authorization: "Bearer <provider api key>",
      "User-Agent": "cc-switch/1.0",
    },
  },
  extractor: function (response) {
    return {
      planName: "Coco OpenRouter",
      used: response.used_percent,
      total: 100,
      unit: "%",
    };
  },
});
```

The modal test path can succeed, but the saved provider usage query path can fail before the custom request runs because it falls back to the provider base URL and validates that HTTP base URL even in `templateType: "custom"` mode.

## Expected behavior

- Test success summaries should not render `undefined` when optional quota fields are omitted.
- Custom usage scripts should be able to call their explicit request URL without an unrelated provider base URL fallback causing HTTPS validation failure.
- Docs should match the extractor schema: all return fields are optional and `extra` is a string.

## Notes

This was found with a local/LAN relay and a separate LAN quota endpoint. The API key is the existing provider API key; no separate usage key is required.

hanhan3344:
## Summary

Fixes #3128.

This PR fixes two custom usage-query edge cases:

- custom usage script summaries now handle optional fields correctly, so `used`/`total` can display a usage percentage when `remaining` is omitted instead of rendering `undefined`
- `templateType: "custom"` usage scripts no longer validate an unrelated provider `base_url` fallback before executing an explicit request URL
- usage-query docs now match the actual extractor schema: all fields are optional and `extra` is a string

## Root Cause

The modal success formatter assumed every successful result had `remaining`, even though `UsageData.remaining` is optional and the in-app extractor help says all fields are optional.

For saved provider usage queries, `query_usage` can fall back to the provider base URL. That fallback was validated before custom script execution, so a custom script calling a separate explicit HTTP LAN quota endpoint could be blocked by the provider's non-localhost HTTP relay URL even though custom request URL validation intentionally allows this mode.

## Validation

- `pnpm exec vitest run tests/utils/usageDisplay.test.ts`
- `pnpm typecheck`
- `cargo test --manifest-path src-tauri/Cargo.toml usage_script`
- Full `pnpm test:unit` also passed locally: 44 files, 244 tests


farion1231:
@codex review

farion1231 (approved):

PR #3140, opened 2026-05-25
refactor: replace JSON deep copy with deepClone helper and extract useTauriEvent hook

ChongBiaoZhang:
Replace production and test `JSON.parse(JSON.stringify())` deep-copy patterns with a shared `deepClone()` helper. The helper uses native `structuredClone()` when available and falls back to recursive cloning for older WebViews that are still within the app's macOS 12 support range, so `tsconfig` can stay on ES2020.

Extract a `useTauriEvent` hook to eliminate repeated Tauri event listener boilerplate (`useEffect` + active/disposed flag + async `listen`) across App.tsx and useUsageCacheBridge.ts. The hook handles async registration, race-condition guards, and cleanup automatically.

Additional cleanup in this PR:

- Preserve the previous `universal-provider-synced` listener ordering: invalidate provider queries first, then update the tray menu.
- Stop touching the pre-existing `cc-switch-main/` dead-copy directory; it can be removed in a separate cleanup PR.

### DIFF
--- a/cc-switch-main/src/config/universalProviderPresets.ts
+++ b/cc-switch-main/src/config/universalProviderPresets.ts
@@ -137,7 +137,7 @@ export function createUniversalProviderFromPreset(
     apps: { ...preset.defaultApps },
     baseUrl,
     apiKey,
-    models: structuredClone(preset.defaultModels),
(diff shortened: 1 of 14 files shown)


farion1231:
@codex review

farion1231 (approved):

PR #3164, opened 2026-05-26
fix(proxy): robust tool call arguments parsing for long-context scenarios

china-powers:
## Background

When using cc-switch as a proxy between Claude Code and non-standard OpenAI-compatible upstreams (e.g., DeepSeek V4 Pro via opencode), prolonged sessions with 100K+ context would cause ALL subsequent tool calls to fail with "required parameter missing".

Root cause: Silent fallback to empty object `{}` when the `arguments` field of `tool_calls` could not be parsed as a JSON string.

### 1. Response direction (`openai_to_anthropic`)

The `arguments` field parsing was a single fragile chain:
```rust
func.get("arguments").and_then(|a| a.as_str()).unwrap_or("{}");
serde_json::from_str(args_str).unwrap_or(json!({}));
```

Now explicitly handles all upstream variants:
- **String** with valid JSON → parse normally
- **String** empty → log warning, fallback to `{}`
- **Object** → serialize directly (non-standard upstreams return objects instead of strings)
- **Null** → log warning, fallback to `{}`
- **Missing** → log warning, fallback to `{}`
- **Malformed JSON string** → log error with truncated payload, fallback to `{}`

Each path has a unique log level (warn/error) for debuggability.

### 2. Request direction (`anthropic_to_openai`)

Coerce `Value::Null` in `tool_use.input` to `json!({})` before serialization. Previously `canonical_json_string(&Value::Null)` would produce the literal string `"null"` in the `arguments` field, which is invalid for the OpenAI contract.

### 3. Tests

Added 12 new tests covering all edge cases:
- `test_openai_to_anthropic_simple_text`
- `test_openai_to_anthropic_tool_call_string_args` (happy path)
- `test_openai_to_anthropic_tool_call_object_args` (non-standard)
- `test_openai_to_anthropic_tool_call_null_args`
- `test_openai_to_anthropic_tool_call_empty_string_args`
- `test_openai_to_anthropic_tool_call_missing_args`
- `test_openai_to_anthropic_multiple_tool_calls` (mixed types)
- `test_openai_to_anthropic_content_refusal`
- `test_openai_to_anthropic_reasoning_content`
- `test_anthropic_to_openai_tool_use_null_input`

### DIFF
--- a/.github/workflows/build-windows.yml
+++ b/.github/workflows/build-windows.yml
@@ -26,18 +26,13 @@ jobs:
       - name: Patch tauri config to disable updater signing
         shell: pwsh
         run: |
-          $config = Get-Content src-tauri/tauri.conf.json -Raw | ConvertFrom-Json
-          $config.bundle.createUpdaterArtifacts = $false
-          $config | ConvertTo-Json -Depth 10 | Set-Content src-tauri/tauri.conf.json
+          (Get-Content src-tauri/tauri.conf.json) -replace '"createUpdaterArtifacts": true', '"createUpdaterArtifacts": false' | Set-Content src-tauri/tauri.conf.json
 
       - name: Install frontend deps
         run: pnpm install --frozen-lockfile
 
       - name: Build Tauri app
-        run: pnpm tauri build
-        env:
-          TAURI_SIGNING_PRIVATE_KEY: ''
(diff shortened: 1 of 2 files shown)


farion1231:
@codex review

PR #3175, opened 2026-05-26
feat(provider-preset): add sorting tabs

Orchardxyz:
## Summary

Add sorting tabs to the Provider Preset selector so users can switch between the existing default order and an alphabetical view without changing the underlying preset definitions.

This keeps the current behavior as the default, adds a faster way to find providers when the list is long, and remembers the selected sort mode across repeated add-provider flows.

## Screenshots

| Before | After |
| image | image |

## Checklist / 检查清单


farion1231:
@codex review

PR #3182, opened 2026-05-26
Add timed lightweight mode controls

wzxnb2333:
﻿## Summary

This adds a timed lightweight-mode workflow for users who keep the app in the tray most of the time.

- Adds settings for entering lightweight mode after the main window has been closed, with a default 20-minute delay and custom minute input.
- Adds startup handling so auto-launch can start directly in lightweight mode when enabled.
- Keeps the tray menu state in sync while a delayed lightweight-mode transition is pending, including canceling the pending transition when the main window is reopened.
- Adds localized settings copy for English, Simplified Chinese, and Japanese.

## Verification

- cargo fmt --check
- cargo test auto_lightweight --lib
- cargo check --lib
- .\\node_modules\\.bin\\prettier.CMD --check "src/**/*.{js,jsx,ts,tsx,css,json}"
- .\\node_modules\\.bin\\tsc.CMD --noEmit --pretty false
- .\\node_modules\\.bin\\vite.CMD build
- cargo build --release


wzxnb2333:
Closing this while I rework the window startup behavior.

PR #3188, opened 2026-05-26
Change tray left click to open main window

fangfuzha:
This pull request updates the tray icon event handling logic to improve user experience and platform integration. The main changes involve distinguishing between mouse hover and click events on the tray icon, and enhancing the behavior when the tray icon is clicked, especially with the left mouse button.

**Tray event handling improvements:**

* Split the handling of `TrayIconEvent::Enter` (mouse hover) and `TrayIconEvent::Click` (mouse click), so each event can trigger specific actions independently.
* On left mouse button click, the main window is brought to the foreground (unminimized, shown, focused), with additional platform-specific fixes for Windows, Linux, and macOS. If the main window does not exist and the app is in lightweight mode, it attempts to exit lightweight mode and recreate the window.

**User interface behavior:**

* Changed `.show_menu_on_left_click` from `true` to `false`, so left-clicking the tray icon no longer opens the tray menu, aligning with the new behavior of focusing the main window instead.


farion1231:
@codex review

fangfuzha:
@claude `app` is moved into the `async move` task at lines 812–814, but it’s used again afterwards (line 818). This should fail to compile due to use-after-move. Fix by cloning into a separate variable for the spawned task (e.g., `let app_for_refresh = app.clone();` and move that into the async block), keeping `app` available for the window logic.
can you fix it！

fangfuzha:
Look here

fangfuzha:
@farion1231 need review

PR #3189, opened 2026-05-26
fix(codex-session): aggregate token_count events per turn to prevent log explosion

DeliciousBuding:
## Summary

Each Codex streaming API call emits ~33 `token_count` events during its lifetime. The session log sync created one `proxy_request_logs` row per event, inflating the record count ~33x and total token estimates ~38x (since `total_token_usage` counts full context windows, not API-billed tokens).

### Before (per-event insertion)

A session with 745 `turn_context` events produced 24,932 `proxy_request_logs` rows. Observed on a production instance:

| Model | codex_session rows | Actual API calls | Inflation |
|-------|-------------------|-----------------|-----------|
| gpt-5.5 | 560,193 | 24,742 | 23x |
| gpt-5.4 | 52,422 | 10 | 5,242x |
| **DB size** | **558 MB** | | |

### After (per-turn flush)

Deltas are accumulated within each turn boundary and flushed once when the next `turn_context`, `session_meta`, or EOF is reached. One record per API call.

## Changes

- Added `pending_delta`, `pending_model`, `pending_timestamp` to `FileParseState`
- Added `flush_pending()` method that emits one aggregated record per turn
- Flush triggers on `turn_context`, `session_meta`, and end-of-file
- `event_msg` token_count: accumulate into `pending_delta` instead of inserting immediately

## Test Plan

- All 26 existing `session_usage*` tests pass
- `cargo check` clean (only pre-existing warnings)
- Manual verification: cross-referenced cc-switch `codex_session` counts against NewAPI database

### DIFF
--- a/src-tauri/src/services/session_usage_codex.rs
+++ b/src-tauri/src/services/session_usage_codex.rs
@@ -66,21 +66,40 @@ struct FileParseState {
 
 impl FileParseState {
     /// 将当前 turn 的累积 delta flush 为一条记录（如果非零）
+    ///
+    /// `should_insert` 为 false 时只推进 event_index 不写库，用于增量同步
+    /// 的 replay 阶段保持 request_id 序列确定性。
     fn flush_pending(
         &mut self,
         db: &Database,
         imported: &mut u32,
         skipped: &mut u32,
+        should_insert: bool,
     ) {
-        let delta = match self.pending_delta.take() {
+        let delta = match self.pending_delta.as_ref() {
             Some(d) if !d.is_zero() => d,
             _ => return,
+        };
+
+        // 始终推进 event_index，保证增量同步 resume 时 request_id 不碰撞
+        self.event_index += 1;
+
+        if !should_insert {
+            // replay 阶段：只推进索引，实际插入已在历史同步中完成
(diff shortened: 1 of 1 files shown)


farion1231:
@codex review

DeliciousBuding (commented):
Addressed all three review findings in d5cad89:

1. **event_index determinism (P1)**: `flush_pending` now takes `should_insert` flag. During incremental sync replay (`line_offset <= last_offset`), turn boundaries still accumulate and call `flush_pending` with `should_insert=false` to advance `event_index` in lockstep. New data flushes with `should_insert=true`.

2. **Data loss on insert failure**: `pending_delta/model/timestamp` are only cleared after a successful DB insert. On error, the accumulator is preserved for the next sync retry.

3. **Timestamp None overwrite**: `pending_timestamp` is only updated when the parsed timestamp is `Some(_)`, so an earlier captured value is not erased by a later `None`.

All 26 `session_usage*` tests continue to pass.

MAXEUR5:
我也有这个问题，我是在好几个电脑之间来回备份导入导出，结果发现proxy_request_logs越来越大了

PR #3253, opened 2026-05-28
fix(model-fetch): support Bailian Token Plan model list endpoint

coderTom1226:
## Summary

- Fix model list fetch failure for Alibaba Bailian **Token Plan** gateway (`token-plan.*.maas.aliyuncs.com`)
- Prepend `/compatible-mode/v1/models` to candidate URLs (verified HTTP 200 with OpenAI-compatible response)
- Treat Token Plan 400/parse failures as soft failures and continue trying other candidates
- Fall back to curated model IDs from Token Plan documentation when all endpoints fail

## Problem

When Base URL is e.g. `https://token-plan.cn-beijing.maas.aliyuncs.com/apps/anthropic`, CC Switch tries:
- `.../apps/anthropic/v1/models` → 404
- `.../v1/models` → 404
- `.../models` → 404

UI shows: "未找到可用的模型列表端点".

The correct listing endpoint is `.../compatible-mode/v1/models`, which returns a valid models list.

## Related

- Complements #2815 (generic candidate retry logic); this PR adds Token Plan–specific endpoint discovery and fallback.

## Test plan

- [x] Added unit test `test_candidates_token_plan_prepends_compatible_mode_models`
- [ ] `cargo test model_fetch` (CI)
- [ ] Manual: configure Bailian Token Plan provider in CC Switch, click "获取模型列表" → should return model list instead of error
- [ ] Regression: DeepSeek / Zhipu / SiliconFlow model fetch still works

### DIFF
--- a/src-tauri/src/services/model_fetch.rs
+++ b/src-tauri/src/services/model_fetch.rs
@@ -7,6 +7,7 @@
 use reqwest::StatusCode;
 use serde::{Deserialize, Serialize};
 use std::time::Duration;
+use url::Url;
 
 /// 获取到的模型信息
 #[derive(Debug, Clone, Serialize, Deserialize)]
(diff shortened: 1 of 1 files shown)


coderTom1226:
Superseded by a revised PR that removes hardcoded fallback model IDs and relies solely on the /compatible-mode/v1/models API endpoint.

PR #3256, opened 2026-05-28
fix(model-fetch): fetch Bailian Token Plan models via compatible-mode endpoint

coderTom1226:
## Summary

- Fix model list fetch for Alibaba Bailian **Token Plan** gateway (`token-plan.*.maas.aliyuncs.com`)
- Prepend `/compatible-mode/v1/models` to candidate URLs (returns OpenAI-compatible model list)
- Treat Token Plan 400/parse/empty responses as soft failures so other candidates can be tried
- **No hardcoded model IDs** — models are fetched dynamically from the API only

## Problem

When Base URL is e.g. `https://token-plan.cn-beijing.maas.aliyuncs.com/apps/anthropic`, CC Switch tries:
- `.../apps/anthropic/v1/models` → 404
- `.../v1/models` → 404
- `.../models` → 404

UI shows: "未找到可用的模型列表端点".

The correct listing endpoint is `.../compatible-mode/v1/models`, which returns a valid models list.

## Related

- Supersedes #3253 (removed hardcoded fallback model list per review feedback)
- Complements #2815 (generic candidate retry logic)

## Test plan

- [ ] Added unit test `test_candidates_token_plan_prepends_compatible_mode_models`
- [x] `cargo test model_fetch` (CI)
- [ ] Manual: configure Bailian Token Plan provider, click "获取模型列表" → should return API model list
- [ ] Regression: DeepSeek / Zhipu / SiliconFlow model fetch still works

### DIFF
--- a/src-tauri/src/services/model_fetch.rs
+++ b/src-tauri/src/services/model_fetch.rs
@@ -50,36 +50,11 @@ const KNOWN_COMPAT_SUFFIXES: &[&str] = &[
 
 /// 是否为阿里云百炼 Token Plan（`*.maas.aliyuncs.com`，且 host 含 `token-plan.`）。
 ///
-/// 该网关通常不把 Anthropic Base URL 映射到可用的 `GET /v1/models`，甚至会返回 400 JSON，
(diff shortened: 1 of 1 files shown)


farion1231:
@codex review

PR #3346, opened 2026-05-30
feat(hermes): add discover_models toggle to provider form

w1ndcn:
## What does this PR do?

Adds a "Discover models from API" toggle switch to the Hermes provider configuration form. When disabled, the provider writes `discover_models: false` to Hermes's `config.yaml`, so Hermes only uses the manually configured model list and does not probe the `/models` endpoint.

## Why is this needed?

Hermes's live `/models` discovery currently replaces the user's entire configured model list (see NousResearch/hermes-agent#35145). While that bug is being fixed upstream, users also need a UI-level control to opt out of live discovery entirely for providers where they only want specific models.

CC Switch had no way to set `discover_models: false` — the only option was to manually edit `config.yaml`.

## Changes

| File | Change |
|------|--------|
| `useHermesFormState.ts` | Add `hermesDiscoverModels` state + `handleHermesDiscoverModelsChange` handler |
| `HermesFormFields.tsx` | Add `ToggleRow` switch UI (Search icon, between API key and model list) |
| `ProviderForm.tsx` | Pass `discoverModels` / `onDiscoverModelsChange` props to `HermesFormFields` |
| `hermesProviderPresets.ts` | Add `discover_models?: boolean` to `HermesProviderSettingsConfig` interface |
| `zh.json`, `en.json`, `ja.json`, `zh-TW.json` | Add i18n translations for the toggle label and description |

## Design decisions

- The toggle writes an explicit `true` / `false` value rather than deleting the key on re-enable. This prevents the Rust forward-compat merge from resurrecting a stale `false` from the on-disk YAML.
- The toggle defaults to `true` (Hermes default behavior). Only `discover_models: false` is explicitly written when the user opts out.
- The implementation follows the exact same pattern as `rate_limit_delay` (same hook structure, same props passing, same ToggleRow component).

## Screenshots

image

## Checklist

- [x] I've tested my changes
- [x] I've added i18n translations for all
- [x] My changes follow the existing code patterns in this project
- [x] I've added type declarations for new fields

### DIFF
--- a/src/components/providers/forms/HermesFormFields.tsx
+++ b/src/components/providers/forms/HermesFormFields.tsx
@@ -30,6 +30,7 @@ import {
   ChevronDown,
   ChevronRight,
   Loader2,
+  Search,
 } from "lucide-react";
 import {
   DropdownMenu,
@@ -39,6 +40,7 @@ import {
   DropdownMenuSeparator,
   DropdownMenuTrigger,
 } from "@/components/ui/dropdown-menu";
+import { ToggleRow } from "@/components/ui/toggle-row";
 import { ApiKeySection } from "./shared";
 import {
   fetchModelsForConfig,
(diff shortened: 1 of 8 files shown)


farion1231:
@codex review

PR #3360, opened 2026-05-30
fix(codex): always update model catalog JSON on provider switch

Postroggy:
## Summary

- Codex `cc-switch-model-catalog.json` and `model_catalog_json` in `config.toml` were only regenerated when the proxy was active (takeover) or a live backup existed. Switching providers while the proxy was idle left the catalog stale — requiring a full CC Switch restart to pick up the new provider's models.
- Root cause: the catalog update was gated inside the `should_sync_backup` block in `hot_switch_provider`.
- Fix: after the `should_sync_backup` block, unconditionally call `write_codex_provider_live_with_catalog` for Codex provider switches when the proxy is not currently active (`!live_taken_over`). This mirrors what `live.rs` already does during a normal provider apply.

## Test plan

- [ ] New test `hot_switch_codex_provider_updates_model_catalog_json_without_proxy_takeover`: seeds two providers (each with `modelCatalog`), switches without proxy takeover, asserts catalog JSON file is created and contains the new provider's model, and that `config.toml` now includes `model_catalog_json`.
- [ ] All existing `hot_switch_codex_*` tests still pass.
- [ ] Full test suite: 1369 passed, 0 new failures (2 pre-existing port-binding flaky tests unaffected).

### DIFF
--- a/src-tauri/src/services/proxy.rs
+++ b/src-tauri/src/services/proxy.rs
@@ -1860,23 +1860,25 @@ impl ProxyService {
             }
         }
 
-        // When proxy is not active, catalog update is skipped by the block above.
-        // Always regenerate the Codex model catalog JSON on provider switch so that
-        // `/model` reflects the new provider's models without requiring a restart.
-        if !live_taken_over && matches!(app_type_enum, AppType::Codex) {
-            let auth = provider.settings_config.get("auth");
-            let config_str = provider
-                .settings_config
-                .get("config")
-                .and_then(|v| v.as_str());
-            if let Some(auth) = auth {
-                let _ = crate::codex_config::write_codex_provider_live_with_catalog(
-                    &provider.settings_config,
-                    provider.category.as_deref(),
-                    auth,
-                    config_str,
-                );
-            }
+        if has_backup && !live_taken_over && matches!(app_type_enum, AppType::Codex) {
+            let effective_settings = build_effective_settings_with_common_config(
+                self.db.as_ref(),
+                &AppType::Codex,
+                &provider,
+            )
+            .map_err(|e| format!("构建 Codex 有效配置失败: {e}"))?;
+            let auth = effective_settings
+                .get("auth")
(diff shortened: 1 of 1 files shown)


AideYu:
I can reproduce a related Codex model catalog issue on CC Switch 3.16.0 on macOS.

The PR description matches one part of the behavior: switching Codex providers while the proxy is idle can leave the live `cc-switch-model-catalog.json` / `model_catalog_json` stale.

There seems to be an additional persistence case worth covering:

1. Create or edit a Codex provider, for example `DeepSeek Direct`, with `settings_config.modelCatalog.models` containing:
   - `deepseek-v4-flash`
   - `deepseek-v4-pro`
2. Save the provider and verify the model mapping appears in the UI.
3. Switch from that provider back to an imported OpenAI/Gmail Codex account.
4. Re-open the DeepSeek provider editor.

Observed result:

- The model mapping rows disappear from the provider editor.
- The provider row in SQLite loses `settings_config.modelCatalog`.
- The provider's `settings_config.config` can also lose the `model_catalog_json = ...` line.

Expected result:

- Switching away from a Codex provider should not mutate or drop that provider's saved `modelCatalog`.
- The live catalog/config should be refreshed for the active provider, but inactive provider snapshots should keep their model mappings.

Local workaround that confirms the failure mode:

- I added a SQLite trigger that restores `settings_config.modelCatalog` and `model_catalog_json` for that specific provider if CC Switch writes an update without them.
- After that, a `DeepSeek Direct -> Gmail` switch no longer deletes the saved model mapping, and the Gmail/OpenAI live model list no longer includes the DeepSeek models.

So #3360 looks directionally correct for the live catalog refresh path, but it may also need a test/assertion that provider switching does not rewrite an inactive Codex provider's saved `modelCatalog` to null/missing.

farion1231:
@codex review

chofuhoyu:
This is important and should solve my problem in https://github.com/farion1231/cc-switch/issues/3356#issuecomment-4582255173

farion1231 (approved):

PR #3369, opened 2026-05-30
feat(usage): display output tokens per second (TPS) in RequestLogTable

juexe:
This PR adds a calculation and display of output tokens per second (TPS) inside the output tokens cell of the RequestLogTable if latency and output token count are valid.

### DIFF
--- a/src/components/usage/RequestDetailPanel.tsx
+++ b/src/components/usage/RequestDetailPanel.tsx
@@ -7,6 +7,7 @@ import {
 } from "@/components/ui/dialog";
 import { useRequestDetail } from "@/lib/query/usage";
 import { getFreshInputTokens, isUnpricedUsage } from "@/types/usage";
+import { formatOutputTokensPerSecond } from "./format";
 
 interface RequestDetailPanelProps {
   requestId: string;
@@ -56,6 +57,7 @@ export function RequestDetailPanel({
   const freshInput = getFreshInputTokens(request);
   const isCacheInclusive = request.inputTokens !== freshInput;
   const unpriced = isUnpricedUsage(request);
+  const outputTps = formatOutputTokensPerSecond(request);
 
   return (
     <Dialog open onOpenChange={onClose}>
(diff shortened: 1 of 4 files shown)


farion1231:
@codex review

juexe:
@codex continue review?

juexe:
@farion1231 Please proceed with the review.

578223592:
> @farion1231 Please proceed with the review.

i like this feature

Musdkar:
is this feat still being considered？

farion1231:
Merged, thank you @juexe! Nice first contribution, and thanks for your patience while this sat in the queue. It also closes #5936, which asked for exactly this column.

What I checked on my side:
- Merge tree against current main (611 commits past your base): `pnpm typecheck`, the full unit suite (1111 tests) and prettier are all clean; the request-log table refactor since your base merged without conflict.
- Replayed the formula over my local request logs: on ~300 Claude streaming rows the generation-window version averages ~58 tok/s versus ~29 tok/s with output ÷ total latency, so switching to `latencyMs - firstTokenMs` after the Codex review comment was the right call.
- Traced where `firstTokenMs` is recorded on every proxied streaming path (Anthropic, Chat, Responses and Gemini upstreams, passthrough and converted): the first collected event always sits at the head of the stream, so the window is sound for all managed apps.

Follow-ups on our side, not on this PR:
- The backend never writes `duration_ms` today. We will record first-content-delta to last-content-delta there, and your `durationMs` branch will pick it up automatically.
- A floor guard for relays that buffer the whole response and flush it at the end (one row out of ~240 on my machine showed a few thousand tok/s).

It will ship in the next release.

yzxq1:
你好，请问这个请求合并到最新的版本CC Switch v3.20.3 [Latest](https://github.com/farion1231/cc-switch/releases/latest)了吗？
https://github.com/farion1231/cc-switch/commit/d695a2d77fd9081eafd3e9eedcbf2a97b3410928

PR #3376, opened 2026-05-30
fix(codex): preserve model catalog on provider update

AideYu:
## Summary
- Preserve a Codex provider's saved `settings_config.modelCatalog` when an update payload omits that CC Switch private field.
- Preserve an existing `model_catalog_json = ...` line in the provider's saved `config` when the update payload omits it.
- Add a regression test covering the DeepSeek Direct-style case where a provider with two model mappings is updated by a payload that only contains `auth` and `config`.

## Why
A user can save a Codex provider with model mappings, then later switch away or update from a stale edit form that does not include `modelCatalog`. The backend currently treats the omitted private field as a deletion and overwrites `settings_config`, so reopening the provider shows an empty model mapping table.

Provider switching should refresh the active live catalog/config, but it should not erase inactive providers' saved model mappings.

## Test plan
- `git diff --check` passes locally.
- Could not run `cargo test update_codex_provider_preserves_existing_model_catalog_when_update_omits_it --lib` locally because this machine does not have `cargo` installed (`zsh: command not found: cargo`).

### DIFF
--- a/src-tauri/src/services/provider/mod.rs
+++ b/src-tauri/src/services/provider/mod.rs
@@ -349,6 +349,77 @@ mod tests {
     }
 
     #[test]
+    fn update_codex_provider_preserves_existing_model_catalog_when_update_omits_it() {
+        with_test_home(|state, _home| {
+            let existing = Provider::with_id(
+                "deepseek-direct".into(),
+                "DeepSeek Direct".into(),
+                json!({
+                    "auth": { "OPENAI_API_KEY": "test-key" },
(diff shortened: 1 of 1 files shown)


farion1231:
@codex review

farion1231 (commented):
Thanks for the contribution and the regression test! This was actually fixed on `main` a few hours after you branched (commit b248ae51, "Fix Codex model catalog being wiped by live-config backfill"), covering both the switch-away and stale edit-form cases. Closing as superseded — but please keep them coming. 🙏

PR #3407, opened 2026-05-31
ci: switch ubuntu-22.04 to ubuntu-latest to reduce runner queue time

trdong19:
## Problem
Ubuntu 22.04 runners were queuing for 8+ minutes, causing build delays.

## Solution
Switched all  references to  in both workflow files:
-  (4 occurrences)
-  (1 occurrence)

This gives access to a larger GitHub-hosted runner pool and should significantly reduce queue times.


farion1231:
@codex review

PR #3448, opened 2026-05-31
fix(proxy): attribute streamed usage model

cxyfer:
## Summary

- Fix routing-mode usage attribution for streamed responses.
- Prefer the model reported in the streamed usage payload when available.
- Keep the original request model available separately for fallback/pricing behavior.

## Background

In routing mode, streamed usage logging could record the request/default Claude model as the attributed model even when the transformed stream usage payload already included the actual upstream model.

For example, usage statistics could be attributed to a default model name such as `claude-opus-4-8` instead of the provider model that actually handled the request.

The fix keeps the existing stream event filtering and extraction behavior unchanged, and only adjusts the usage logging path to prefer `usage.model` when it is present.

260601_013350_cc-switch

## Changes

- Use `usage.model` as the logged attribution model when the streamed usage payload includes it.
- Fall back to the original request model when the streamed usage payload has no model.
- Continue passing the original request model to `log_usage` separately so fallback/pricing logic remains intact.
- Keep the change scoped to proxy stream usage/model attribution.

## Test plan

- [x] Verify routing-mode usage statistics with DeepSeek OpenAI Chat Completions.
- [x] Verify routing-mode usage statistics with sub2api OpenAI Responses.
- [x] Verify routing-mode usage statistics with sub2api OpenAI Chat Completions.

## Screenshots

260601_013619_cc-switch

### DIFF
--- a/src-tauri/src/proxy/handler_config.rs
+++ b/src-tauri/src/proxy/handler_config.rs
@@ -44,11 +44,13 @@ pub fn claude_stream_usage_event_filter(data: &str) -> bool {
 }
 
 fn openai_stream_usage_event_filter(data: &str) -> bool {
-    data.contains("\"usage\"")
+    data.contains("\"usage\"") || data.contains("\"model\"")
 }
 
 pub fn codex_stream_usage_event_filter(data: &str) -> bool {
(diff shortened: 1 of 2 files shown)


cxyfer:
@codex review

PR #3539, opened 2026-06-02
fix: quote path in Windows autostart registry entry

Ersiter:
## Problem

auto-launch v0.5.0 writes the executable path to `HKCU\Software\Microsoft\Windows\CurrentVersion\Run` **without wrapping it in double quotes**:

```
CC Switch : D:\Users\ersit\AppData\Local\Programs\CC Switch\cc-switch.exe
```

When the path contains spaces (like `CC Switch`), Windows interprets the space as an argument delimiter and tries to execute `D:\...\Programs\CC` — which doesn't exist. The app silently fails to start on boot, even though Task Manager shows it as "Enabled".

## Root Cause

The bug is in the `auto-launch` crate ([source](https://github.com/zzzgydi/auto-launch)), not in CC Switch's code:

```rust
// auto-launch/src/windows.rs
root_key.create(AL_REGKEY)?.set_string(
    &self.app_name,
    format!("{} {}", &self.app_path, &self.args.join(" ")),
    //     ^ no quoting — breaks on paths with spaces
)?;
```

## Fix

Bypass the `auto-launch` crate on Windows and write the registry entry directly via `winreg`, with the path properly quoted:

```
"D:\Users\ersit\AppData\Local\Programs\CC Switch\cc-switch.exe"
```

Also manages `StartupApproved\Run` to keep Task Manager's Startup tab consistent with the actual auto-start state.

- macOS/Linux: unchanged, continue using `auto-launch` crate
- Windows: direct `winreg` calls with quoted paths
- `is_enabled()` strips quotes when comparing, so it handles both old (unquoted) and new (quoted) registry entries transparently

### DIFF
--- a/src-tauri/src/auto_launch.rs
+++ b/src-tauri/src/auto_launch.rs
@@ -1,4 +1,107 @@
 use crate::error::AppError;
+
+// ── Windows: 手动管理注册表，修复 auto-launch crate 不加引号的 bug ──────────
+// auto-launch v0.5.0 写入 HKCU\...\Run 时没有给路径加双引号，
+// 导致路径含空格的应用（如 "CC Switch"）开机无法自启。
+// 此处直接使用 winreg 写入带引号的路径，并同步管理
+// StartupApproved\Run 以保持与任务管理器"启动"标签页一致。
+
+#[cfg(target_os = "windows")]
+mod platform {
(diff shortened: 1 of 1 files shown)


farion1231:
@codex review

PR #3602, opened 2026-06-02
fix(proxy): preserve user's Claude model version in takeover mode

### LINKED ISSUE #3600: Proxy takeover always overwrites Claude model version to hardcoded constants
## Problem

When proxy takeover is enabled, switching providers always overwrites the `ANTHROPIC_DEFAULT_*_MODEL` fields in Claude Code's `settings.json` with hardcoded constants (`claude-opus-4-8`, `claude-sonnet-4-6`, `claude-haiku-4-5`), regardless of what the user has configured in their provider settings.

**Example:** A user configures their provider with `ANTHROPIC_DEFAULT_OPUS_MODEL = "claude-opus-4-6"` because their upstream API only supports Opus 4.6. After switching providers, `settings.json` is overwritten to:

```json
{
  "ANTHROPIC_DEFAULT_OPUS_MODEL": "claude-opus-4-8",
  "ANTHROPIC_DEFAULT_OPUS_MODEL_NAME": "claude-opus-4-6",
  "ANTHROPIC_DEFAULT_SONNET_MODEL": "claude-sonnet-4-6",
  "ANTHROPIC_DEFAULT_SONNET_MODEL_NAME": "claude-opus-4-6",
  "ANTHROPIC_DEFAULT_HAIKU_MODEL": "claude-haiku-4-5",
  "ANTHROPIC_DEFAULT_HAIKU_MODEL_NAME": "claude-opus-4-6"
}
```

The `MODEL_NAME` fields correctly reflect the user's choice, but the `MODEL` fields are always forced to the hardcoded constants from `CLAUDE_TAKEOVER_*_MODEL`.

## Impact

- Claude Code sees the wrong model version (e.g. thinks it's running Opus 4.8 when the upstream actually serves Opus 4.6), which affects system prompts and internal behavior.
- The model mapper correctly maps the request at the proxy layer, but providers that validate model IDs strictly may reject `claude-opus-4-8` if they only support `claude-opus-4-6`.

## Root Cause

In `src-tauri/src/services/proxy.rs`, `push_claude_takeover_role_fields()` unconditionally uses the hardcoded constant for the `MODEL` field (line 256):

```rust
let mut client_model = takeover_model.to_string(); // always "claude-opus-4-8"
```

This ignores the user's chosen Claude model version even when the upstream model is a valid Claude model of the same role family.

## Proposed Fix

When the upstream model is a Claude model of the same role family (e.g. `claude-opus-4-6` for the opus slot) AND uses the dash-separated format that Claude Code accepts, use it directly instead of the hardcoded constant. Fall back to the constant for non-Claude models (e.g. `MiniMax-M2`, `gpt-5.4`).

This preserves:
- The user's chosen Claude model version in `settings.json`
- Correct Claude Code internal behavior (system prompts match actual model)
- Proxy model mapping (still routes via role keyword matching)
- Backward compatibility for non-Claude upstream models

## Environment

- CC Switch version: 3.16.1
- OS: macOS
- Related app: Claude Code

vincenthcui:
## Summary

- When proxy takeover is active, `push_claude_takeover_role_fields` now preserves the user's chosen Claude model version (e.g. `claude-opus-4-6`) in `settings.json` instead of always overwriting with the hardcoded constant (`claude-opus-4-8`).
- Falls back to the hardcoded constant for non-Claude upstream models (e.g. `MiniMax-M2`, `gpt-5.4`) and dot-separated formats (e.g. `claude-opus-4.6` used by Copilot).

## Problem

When a user configures `ANTHROPIC_DEFAULT_OPUS_MODEL = "claude-opus-4-6"` in their provider, the proxy takeover unconditionally overwrites it to `claude-opus-4-8` in `settings.json`. This causes:

1. Claude Code thinks it's running Opus 4.8 (wrong system prompts / internal behavior)
2. Providers that only support specific Claude versions may reject the model ID

## How it works

In `push_claude_takeover_role_fields`, before writing the `MODEL` field:

1. Strip `[1M]` marker and provider prefix (e.g. `anthropic/`) from the upstream model
2. Check if the base model is a Claude model of the **same role family** (contains `"claude-"` prefix + role keyword like `"opus"`)
3. Check it uses **dash-separated format** (`claude-opus-4-6`, not `claude-opus-4.6`) — Claude Code only accepts this format
4. If all conditions match → use the upstream model directly; otherwise → use the hardcoded constant

## Test plan

- [x] All 3 `managed_account_claude_takeover_*` tests pass (Copilot `claude-haiku-4.5` correctly falls back to constant due to dot format)
- [x] All 4 `hot_switch_*` tests pass
- [x] All 12 `model_mapper` tests pass
- [x] `normal_claude_takeover_*` test passes

Fixes #3600

### DIFF
--- a/src-tauri/src/claude_desktop_config.rs
+++ b/src-tauri/src/claude_desktop_config.rs
@@ -743,7 +743,7 @@ fn is_compatible_opus_route_alias(route_id: &str, requested: &str) -> bool {
 /// 按角色关键词(opus / haiku / sonnet)归类一个 Claude 模型名/route_id。
 /// 仅在命中明确角色词时返回 Some,未知模型返回 None(不回落,保持精确报错语义)。
 /// 与前端 `routeRoleFromId` 同序(opus → haiku → sonnet)。
-fn claude_role_keyword(model: &str) -> Option<&'static str> {
+pub(crate) fn claude_role_keyword(model: &str) -> Option<&'static str> {
     let normalized = model.to_ascii_lowercase();
     if normalized.contains("opus") {
         Some("opus")
--- a/src-tauri/src/services/proxy.rs
+++ b/src-tauri/src/services/proxy.rs
@@ -257,23 +257,18 @@ impl ProxyService {
         // AND uses the dash-separated ID format that Claude Code accepts
         // (e.g. claude-opus-4-6, not claude-opus-4.6), preserve the user's
         // chosen version instead of forcing the hardcoded constant.
+        //
+        // Reuse the canonical helpers in `claude_desktop_config` so we don't
+        // accept invalid IDs like `claude-opus-` while still rejecting the
+        // dot-separated format used by Copilot (e.g. `claude-haiku-4.5`).
         let stripped = Self::strip_claude_one_m_marker(upstream_model);
         let base = stripped.rsplit('/').next().unwrap_or(&stripped);
-        let base_lower = base.to_lowercase();
-        let role = if takeover_model.contains("haiku") {
-            "haiku"
-        } else if takeover_model.contains("opus") {
-            "opus"
-        } else if takeover_model.contains("sonnet") {
-            "sonnet"
-        } else {
-            ""
-        };
-        let is_dash_format = !base.contains('.');
-        let mut client_model = if !role.is_empty()
-            && is_dash_format
-            && base_lower.starts_with("claude-")
-            && base_lower.contains(role)
+        let base_role = crate::claude_desktop_config::claude_role_keyword(base);
+        let takeover_role = crate::claude_desktop_config::claude_role_keyword(takeover_model);
+        let mut client_model = if !base.contains('.')
+            && base_role.is_some()
+            && base_role == takeover_role
+            && crate::claude_desktop_config::is_claude_safe_model_id(base)
         {
             base.to_string()
         } else {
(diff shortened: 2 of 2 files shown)


farion1231:
@codex review

DawnElixir:
Please help merge this fix since we meet this issue as well.

farion1231 (commented):
Thank you for your contribution! The main requested change is in push_claude_takeover_role_fields. The PR currently hand-rolls this validation:

base_lower.starts_with("claude-") && base_lower.contains(role)
This is weaker than the existing canonical helper, is_claude_safe_model_id. It would preserve invalid IDs such as claude-opus- and claude-3-5-sonnet-20241022, while the existing helper rejects them. I suggest reusing the canonical validation while keeping the PR’s intended dot-format rejection:

!base.contains('.')
    && is_claude_safe_model_id(base)
    && claude_role_keyword(base) == claude_role_keyword(takeover_model)
claude_role_keyword may need to become pub(crate) or be moved to a shared helper.

A secondary issue is missing test coverage. The new happy path is not asserted today. Please add tests for:

ANTHROPIC_DEFAULT_OPUS_MODEL = "claude-opus-4-6" preserves "claude-opus-4-6"
claude-opus-4-6[1M] preserves "claude-opus-4-6[1M]"
optionally, claude-opus- falls back to "claude-opus-4-8"
One minor consistency issue: MODEL strips the anthropic/ prefix, but the fallback MODEL_NAME does not. This can produce MODEL=claude-opus-4-6 and MODEL_NAME=anthropic/claude-opus-4-6. Low impact, but worth aligning.

vincenthcui:
Thanks for the feedback, @farion1231! All four review points have been addressed in supplementary commit `7e0e10e0` on top of `738a0085`.

Concrete changes:

1. **Canonical helper reuse** — replaced the hand-rolled `base_lower.starts_with("claude-") && base_lower.contains(role)` check with the canonical `claude_role_keyword` + `is_claude_safe_model_id` pair, both from `claude_desktop_config`. `claude_role_keyword` was made `pub(crate)` so the proxy module can reuse it. The hand-rolled role classification and string-prefix check are gone.

2. **Consistency fix** — the `anthropic/` provider prefix is now also stripped from the fallback `MODEL_NAME` value, so `MODEL` and `MODEL_NAME` agree when the upstream value includes the prefix.

3. **Three new regression tests** covering the happy path you asked for:
   - `claude-opus-4-6` → `ANTHROPIC_DEFAULT_OPUS_MODEL = "claude-opus-4-6"`
   - `claude-opus-4-6[1M]` → `MODEL = "claude-opus-4-6[1M]"`, `MODEL_NAME = "claude-opus-4-6"`
   - `claude-opus-` → falls back to `claude-opus-4-8` (degenerate ID, rejected by `is_claude_safe_model_id`)

Test results on the full PR branch: `managed_account_claude_takeover_*` 6/6, `normal_claude_takeover_*` 1/1, `hot_switch_*` 4/4, `proxy::model_mapper` 12/12 all pass. Behaviour-preserving for non-Claude upstream models (e.g. `MiniMax-M2.7`, `gpt-5.4`) — they still fall back to the hardcoded constant.

@DawnElixir mentioned the same issue earlier; happy to address any other concerns if you have them, otherwise this should be ready to merge when you have a moment. 🙏

Issue #3688, opened 2026-06-04
[Bug] Proxy toggle stuck: Live remains on http://127.0.0.1:15721 when stop_with_restore restores a corrupted backup


YongmaoLuo:
## Summary
After any history in which `stop_with_restore()` failed to restore the user's original Live (app crash, settings.json unwritable, partial-write, etc.), the proxy toggle becomes a no-op: every subsequent `start` → `stop` cycle leaves `~/.claude/settings.json` (and Codex/Gemini equivalents) pointing at `http://127.0.0.1:15721` instead of the provider's real URL. The user has to manually fix the file.

## Repro
1. Get CC Switch into a state where `proxy_live_backup.original_config` is itself a proxy placeholder (e.g. `{"env":{"ANTHROPIC_AUTH_TOKEN":"PROXY_MANAGED","ANTHROPIC_BASE_URL":"http://127.0.0.1:15721"}}`). The simplest way to reach this is:
   - Call `start_with_takeover` (Live goes to placeholder, backup saves the *real* config)
   - Kill the process before the rest of the takeover completes (or otherwise end up with Live still showing the placeholder from a prior takeover)
   - Call `start_with_takeover` again — it now re-reads Live and **saves the proxy placeholder as the "original Live"** in the backup row, overwriting the previous good backup
2. Now any `stop_with_restore` reads the corrupted backup and writes it back to Live → Live stays at the proxy URL forever.
3. The user toggles the proxy off, Claude Code's settings.json still points at the dead local port, the model stops working.

## Root Cause
Both `backup_live_configs()` (line ~1086) and `backup_live_config_strict()` (line ~1122) unconditionally read Live and save it to the backup row, without checking whether Live is already a proxy placeholder. In the corrupted state described above, this overwrites a good backup with the proxy config.

`restore_live_config_for_app_with_fallback_inner()` (line ~1462) blindly writes the parsed backup back to Live if one exists. c3d810a added the SSOT fallback for the *missing-backup* case, but not for the *backup-is-itself-corrupt* case.

## Proposed Fix
- In both backup paths: detect "Live is already a proxy placeholder" and skip the save (preserves any existing good backup).
- In the restore path: detect "parsed backup is itself a proxy placeholder" and fall through to the existing SSOT path.

This is a defense-in-depth change on top of c3d810a — it doesn't change any happy-path behavior, it just makes the system self-heal from a corrupted backup state.

## Related Upstream Issues
[...]

Issue #1238, opened 2026-03-02
s 401 Unauthorized: Incorrect API key provided: PROXY_MA*AGED.


Realtyxxx:
<img width="1430" height="568" alt="Image" src="https://github.com/user-attachments/assets/16642a94-5e0d-46a6-8843-911b670338b1" />

farion1231:
这种情况是 CCSWITCH 开启代理之后被意外关闭了，导致请求地址没有正确回填。再打开一次，就会恢复正确

Realtyxxx:
这个回填的请求地址是什么？，重复流程没有生效

Realtyxxx:
<img width="1249" height="829" alt="Image" src="https://github.com/user-attachments/assets/8a53454c-ef2e-499f-be4d-e7d3f3a5463d" />
@farion1231

imBlanker:
[...]

PR #3689, opened 2026-06-04
fix(proxy): skip backup/restore when Live is already a proxy placeholder

YongmaoLuo:
## Summary

Defense-in-depth fix on top of #3688 (and the c3d810a follow-up). When the system reaches a corrupted state where `Live` (e.g. `~/.claude/settings.json`) is already a proxy placeholder *and* the backup row is also a proxy placeholder, the proxy toggle becomes a no-op — every `start` → `stop` cycle leaves Live pinned at `http://127.0.0.1:15721`.

This PR adds placeholder detection to both sides of the backup/restore pipeline so the system self-heals instead of locking into the bad state.

## What changes

1. **`backup_live_configs()` and `backup_live_config_strict()`** — detect that Live is already a proxy placeholder and skip the save, preserving any existing good backup. Without this, re-takeover in a corrupted state would silently overwrite a good backup with the proxy config itself.

2. **`restore_live_config_for_app_with_fallback_inner()`** — if the parsed backup is itself a proxy placeholder, fall through to the SSOT (current provider DB) path that c3d810a added. Without this, restore would write the corrupted backup back to Live and re-pin the user at the proxy URL.

3. **New helper `live_has_proxy_placeholder_for_app(app_type, config)`** — single dispatch to the existing `is_claude_live_taken_over` / `codex_live_has_proxy_placeholder` / `is_gemini_live_taken_over` helpers, so the check stays in lockstep with the per-app detection.

## Why this is safe

- **No happy-path change.** Both checks only fire when the data is already in an obviously-bad state (Live or backup contains `PROXY_MANAGED` / `http://127.0.0.1:15721`).
- **Existing data preserved.** Backup side: if there's a good backup already, the skip preserves it. Restore side: if the backup is corrupt, the existing SSOT path is reached via the existing fallthrough — the backup row itself is not touched.
- **Existing tests still pass.** 37 proxy tests in `services::proxy::tests` continue to pass; 3 new tests cover the new behavior.

## Tests added

- `restore_falls_through_to_ssot_when_backup_is_proxy_placeholder` — corrupted backup → restore rebuilds Live from provider DB, not from backup.
- `backup_skips_when_live_is_already_proxy_placeholder` — Live is already a placeholder → `backup_live_config_strict` skips, existing good backup untouched.
- `bulk_backup_skips_all_when_live_is_proxy_placeholder` — all three apps (Claude/Codex/Gemini) Live=placeholder → `backup_live_configs` skips every save.

## Related issues / commits

- c3d810a — `fix(proxy): tighten takeover detection and use fallback restore on disable` (the SSOT-fallback foundation this builds on)
- #1118, #1540, #1198, #1238 — same symptom class, closed in v3.12.2 with the "Proxy Takeover Restore Flow reworked" fix
- #3688 — this issue (filed in tandem with the PR)

## Diff

```diff
+    /// 判断给定的 Live/备份配置是否已被代理接管（包含占位符）
+    fn live_has_proxy_placeholder_for_app(app_type: &AppType, config: &Value) -> bool {
+        match app_type {
+            AppType::Claude => Self::is_claude_live_taken_over(config),
+            AppType::Codex => Self::codex_live_has_proxy_placeholder(config),
+            AppType::Gemini => Self::is_gemini_live_taken_over(config),
+            _ => false,
+        }
+    }

 async fn backup_live_configs(&self) -> Result<(), String> {
     if let Ok(config) = self.read_claude_live() {
+        if Self::live_has_proxy_placeholder_for_app(&AppType::Claude, &config) {
+            log::warn!("claude Live 已被代理接管，不备份...");
+        } else {
+            // existing save path
+        }
     }
     // ... same for codex, gemini
 }

 async fn backup_live_config_strict(&self, app_type: &AppType) -> Result<(), String> {
     // ... existing read ...
+    if Self::live_has_proxy_placeholder_for_app(app_type, &config) {
+        log::warn!("{app_type_str} Live 已被代理接管，不备份...");
+        return Ok(());
+    }
     // ... existing save ...
 }

 async fn restore_live_config_for_app_with_fallback_inner(&self, app_type: &AppType) -> ... {
     if let Some(backup) = backup {
         let config: Value = serde_json::from_str(&backup.original_config)?;
+        if Self::live_has_proxy_placeholder_for_app(app_type, &config) {
+            log::warn!("{app_type_str} 备份本身已是代理占位符...");
+        } else {
+            self.write_live_config_for_app(app_type, &config)?;
+            return Ok(());
+        }
     }
     // ... fall through to SSOT path
 }
```

### DIFF
--- a/src-tauri/src/services/proxy.rs
+++ b/src-tauri/src/services/proxy.rs
@@ -1086,32 +1086,48 @@ impl ProxyService {
     async fn backup_live_configs(&self) -> Result<(), String> {
         // Claude
         if let Ok(config) = self.read_claude_live() {
-            let json_str = serde_json::to_string(&config)
-                .map_err(|e| format!("序列化 Claude 配置失败: {e}"))?;
-            self.db
-                .save_live_backup("claude", &json_str)
-                .await
-                .map_err(|e| format!("备份 Claude 配置失败: {e}"))?;
+            // 跳过已被代理接管的 Live：避免把代理占位符当作"原始 Live"存进备份槽。
+            // 否则下次 start_with_takeover 在异常历史状态下（Live 已是占位符）再次
+            // 调用本函数，会用代理配置覆盖一个原本正常的备份；之后 stop 恢复时
+            // 即便走到备份路径也会把代理占位符再写回 Live，永久卡在 127.0.0.1:15721。
+            if Self::live_has_proxy_placeholder_for_app(&AppType::Claude, &config) {
+                log::warn!("claude Live 已被代理接管，不备份（避免把代理配置固化进备份槽）；下次 stop 会从 SSOT 重建 Live");
+            } else {
+                let json_str = serde_json::to_string(&config)
+                    .map_err(|e| format!("序列化 Claude 配置失败: {e}"))?;
+                self.db
+                    .save_live_backup("claude", &json_str)
+                    .await
+                    .map_err(|e| format!("备份 Claude 配置失败: {e}"))?;
+            }
         }
 
         // Codex
         if let Ok(config) = self.read_codex_live() {
-            let json_str = serde_json::to_string(&config)
-                .map_err(|e| format!("序列化 Codex 配置失败: {e}"))?;
-            self.db
-                .save_live_backup("codex", &json_str)
-                .await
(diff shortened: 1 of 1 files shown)


farion1231:
@codex review

farion1231 (approved):

PR #3771, opened 2026-06-05
feat: add image context model for Claude proxy

Pswzy:
## Summary
- Add an optional image processing model for Claude and Claude Desktop proxy providers.
- When a request contains images, the proxy first asks the configured image model to extract image context, then removes image blocks and injects the extracted text back into the original message order before forwarding to the main mapped model.
- Keep existing unsupported-image fallback behavior when no image model is configured.

## Details
- Adds ProviderMeta.imageModel and matching frontend provider form fields with i18n strings.
- Supports Anthropic image blocks plus OpenAI-style input_image/image_url data URL blocks and nested image MIME metadata.
- Avoids recursive image preprocessing/model mapping for the image-analysis request.
- Uses an English internal analysis protocol while preserving parsing compatibility for Chinese image-analysis section labels.
- Automatically enables Claude Code proxy takeover when the active provider has an image model, and restores it when switching to providers without one.

## Tests
- pnpm typecheck
- pnpm test:unit
- cargo +stable test image_context::tests
- cargo +stable test media_sanitizer::tests
- cargo +stable test provider_meta_serializes_image_model


farion1231:
@codex review

Pswzy:
Updated the PR to address the automated review and CI formatting failure.

Changes in the follow-up commit:
- Guard the Claude proxy takeover disable path so it only runs when the edited provider is the current provider, matching the enable path.
- Added a focused `useProviderActions` test covering inactive text-only provider edits while Claude proxy takeover is active.
- Ran Prettier on the files flagged by CI.

Verified locally:
- `pnpm format:check`
- `pnpm typecheck`
- `pnpm exec vitest run tests/hooks/useProviderActions.test.tsx`
- `pnpm test:unit`

makoMakoGo:
> Updated the PR to address the automated review and CI formatting failure.\n\nChanges in the follow-up commit:\n- Guard the Claude proxy takeover disable path so it only runs when the edited provider is the current provider, matching the enable path.\n- Added a focused useProviderActions test covering inactive text-only provider edits while Claude proxy takeover is active.\n- Ran Prettier on the files flagged by CI.\n\nVerified locally:\n- pnpm format:check\n- pnpm typecheck\n- pnpm exec vitest run tests/hooks/useProviderActions.test.tsx\n- pnpm test:unit

Thanks for the pull request. When using AI to create pull requests and comments, please conduct a self-check first. Line breaks and escape characters were not properly handled.

Pswzy:
> > Updated the PR to address the automated review and CI formatting failure.\n\nChanges in the follow-up commit:\n- Guard the Claude proxy takeover disable path so it only runs when the edited provider is the current provider, matching the enable path.\n- Added a focused useProviderActions test covering inactive text-only provider edits while Claude proxy takeover is active.\n- Ran Prettier on the files flagged by CI.\n\nVerified locally:\n- pnpm format:check\n- pnpm typecheck\n- pnpm exec vitest run tests/hooks/useProviderActions.test.tsx\n- pnpm test:unit
> 
> Thanks for the pull request. When using AI to create pull requests and comments, please conduct a self-check first. Line breaks and escape characters were not properly handled.

Sorry about that, and thanks for pointing it out.

The previous comment was posted with escaped newline characters by mistake. I have edited it and fixed the formatting. I’ll double-check PR comments more carefully before posting them in the future.

Pswzy:
> > > Updated the PR to address the automated review and CI formatting failure.\n\nChanges in the follow-up commit:\n- Guard the Claude proxy takeover disable path so it only runs when the edited provider is the current provider, matching the enable path.\n- Added a focused useProviderActions test covering inactive text-only provider edits while Claude proxy takeover is active.\n- Ran Prettier on the files flagged by CI.\n\nVerified locally:\n- pnpm format:check\n- pnpm typecheck\n- pnpm exec vitest run tests/hooks/useProviderActions.test.tsx\n- pnpm test:unit
> > 
> > 
> > Thanks for the pull request. When using AI to create pull requests and comments, please conduct a self-check first. Line breaks and escape characters were not properly handled.
> 
> Sorry about that, and thanks for pointing it out.
> 
> The previous comment was posted with escaped newline characters by mistake. I have edited it and fixed the formatting. I’ll double-check PR comments more carefully before posting them in the future.

Pswzy:
I’ve rebased the branch onto the latest main and resolved the merge conflicts.

I also re-ran the relevant checks locally:
- pnpm format:check
- pnpm typecheck
- pnpm test:unit
- cargo test for the image context and media sanitizer modules

Everything is passing locally now. Thanks!

Pswzy:
@farion1231  @makoMakoGo 
All merge conflicts have been resolved, and both frontend and backend checks are passing now.

Could you please review and merge this PR when you have time? Thanks!

farion1231 (commented):
Thank you very much for your contribution.

However, this project already has an image sanitizer to handle models that do not support images. Would this functionality overlap with the existing implementation?

Pswzy:
> Thank you very much for your contribution.
> 
> However, this project already has an image sanitizer to handle models that do not support images. Would this functionality overlap with the existing implementation?

Thanks for pointing this out. The two features address related but different cases.

The existing image sanitizer removes unsupported image blocks and replaces them
with an `[Unsupported Image]` marker, allowing text-only models to continue
without an upstream error, but the image information is lost.

This feature is opt-in and uses a configured vision model/provider to extract
the image content first. It then replaces the image blocks with textual context
before sending the final request to the main model.

The image-context processing runs before the existing sanitizer. After a
successful extraction, no image blocks remain, so the sanitizer becomes a
no-op. When no image model is configured, the existing sanitizer behavior is
unchanged.

There is some shared image-block detection logic, but the runtime behaviors are
complementary rather than duplicates.

Pswzy:
@farion1231 

The merge conflict has been resolved and the branch has been updated with the latest `main`.

Could you please approve the pending workflow and review the PR again when you have a chance? Thank you!

farion1231 (commented):
Thank you for your contribution, but to implement this feature, the PR adds too much complexity, and I feel the tradeoff may not be worth it.

This PR still has several issues:

**Automatically disabling Claude takeover can cause regressions**

`imageModel` is treated as the only reason to use takeover. When switching to or editing the currently active Claude provider without an `imageModel`, proxy takeover is automatically disabled. This breaks user-enabled takeover, as well as scenarios that depend on the proxy, such as OpenAI Chat/Responses, Copilot, full URL mode, failover, and usage tracking.

I suggest removing the automatic disable behavior and only keeping the automatic enable behavior for providers with `imageModel`.

**Claude’s ProxyToggle can disappear**

The Claude branch in `App.tsx` is missing `settingsData?.enableLocalProxy`. When local proxy is enabled globally, but the current provider is not under takeover and has no `imageModel`, the UI no longer provides an entry point to re-enable takeover.

I suggest adding this condition back.

**Vision sub-request failures break the main request**

Currently, if image recognition fails, the whole request returns an error. This also breaks requests that could previously have succeeded through the existing media sanitizer fallback.

I suggest making vision failures fail open: log the failure, then fall back to the existing image sanitizer / unsupported marker behavior.

**Additional vision cost is invisible**

The vision sub-request is consumed internally and does not go through the existing response processor, so it is not included in usage tracking. Users will incur additional token costs, but those costs will not appear in the usage panel.

I suggest adding usage logging for the vision request, or at least clearly documenting this in the UI/docs.

**Repeated image recognition can amplify cost**

Image context processing runs inside each `forward()` attempt. Failover, rectifier retries, and repeated conversation history across turns can all trigger repeated calls to the vision model.

I suggest adding image-hash-based caching, or moving preprocessing outside the retry/failover loop so it only runs once.

Pswzy:
> Thank you for your contribution, but to implement this feature, the PR adds too much complexity, and I feel the tradeoff may not be worth it.
> 
> This PR still has several issues:
> 
> **Automatically disabling Claude takeover can cause regressions**
> 
> `imageModel` is treated as the only reason to use takeover. When switching to or editing the currently active Claude provider without an `imageModel`, proxy takeover is automatically disabled. This breaks user-enabled takeover, as well as scenarios that depend on the proxy, such as OpenAI Chat/Responses, Copilot, full URL mode, failover, and usage tracking.
> 
> I suggest removing the automatic disable behavior and only keeping the automatic enable behavior for providers with `imageModel`.
> 
> **Claude’s ProxyToggle can disappear**
> 
> The Claude branch in `App.tsx` is missing `settingsData?.enableLocalProxy`. When local proxy is enabled globally, but the current provider is not under takeover and has no `imageModel`, the UI no longer provides an entry point to re-enable takeover.
> 
> I suggest adding this condition back.
> 
> **Vision sub-request failures break the main request**
> 
> Currently, if image recognition fails, the whole request returns an error. This also breaks requests that could previously have succeeded through the existing media sanitizer fallback.
> 
> I suggest making vision failures fail open: log the failure, then fall back to the existing image sanitizer / unsupported marker behavior.
> 
> **Additional vision cost is invisible**
> 
> The vision sub-request is consumed internally and does not go through the existing response processor, so it is not included in usage tracking. Users will incur additional token costs, but those costs will not appear in the usage panel.
> 
> I suggest adding usage logging for the vision request, or at least clearly documenting this in the UI/docs.
> 
> **Repeated image recognition can amplify cost**
> 
> Image context processing runs inside each `forward()` attempt. Failover, rectifier retries, and repeated conversation history across turns can all trigger repeated calls to the vision model.
> 
> I suggest adding image-hash-based caching, or moving preprocessing outside the retry/failover loop so it only runs once.

Thanks for the detailed review. I’ve pushed an update addressing the feedback:

1. Removed the automatic Claude takeover disable behavior. Providers with imageModel can still auto-enable takeover, but existing user-enabled takeover is no longer turned off just because the active provider has no imageModel.
2. Restored the Claude ProxyToggle visibility condition so the takeover entry point remains available when local proxy is enabled globally.
3. Vision sub-request failures now fail open by logging the failure and replacing image blocks with the existing unsupported-image marker before continuing the main request.
4. Vision sub-request usage is now recorded with real latency instead of 0ms.
5. Added image-context caching. Full-request caching is preserved, and repeated identical single-image content can reuse cached analysis to avoid repeated vision calls, while multi-image requests with different images still go through vision analysis to preserve relationship context.

Pswzy:
@farion1231 
The latest changes have been pushed and the full local CI flow has passed.

Could you please approve the pending workflow and review the PR when you have time?

Pswzy:
@farion1231  Thanks again for taking a look.

I think this feature is quite practical for real usage. It helps solve a common issue where the main model cannot actually read images but may still guess based on surrounding context, which can lead to misleading answers instead of a clear fallback.

All checks are passing now. When you have time, could you please review this PR again?

Pswzy:
@farion1231 
Friendly reminder when you have time: all checks are passing now, and this feature helps avoid misleading answers when image content cannot be recognized. Could you please review this PR again?

farion1231 (commented):
Thanks for the continued iterations on this — real care clearly went into addressing
the earlier review: removing the auto-disable path, restoring the ProxyToggle
condition, failing open on vision errors, adding usage logging, and the two-tier
cache are all genuine improvements. We also agree with your core architectural
choice: preprocessing (textualizing images once) is functionally the right shape,
since whole-request routing to a vision model would permanently degrade the session
— images stay in the transcript, so every later turn would keep routing away from
the main model.

After a full review, however, we've decided not to merge the feature in its current
form. The blockers are structural rather than implementation bugs:

**1. Non-deterministic injection destabilizes the main conversation's prompt cache.**
The injected text is LLM-generated, and the analysis cache is in-memory (fully
cleared once it reaches 256 entries). Any CC Switch restart or cache eviction
mid-session re-analyzes the images and produces different text — which changes the
message prefix and invalidates the upstream prompt cache for the rest of the
conversation. This project has been burned badly by
semantically-equivalent-but-byte-different request mutations before (see the #3775
revert and the #4297 cache-hit regression), so byte-stability of anything injected
into the working path is a hard requirement. Fixing this properly means persisting
analysis results (e.g. in SQLite), which adds yet another layer to an already large
change.

**2. Cost amplification in the analysis request.**
`build_ordered_analysis_content` replays the entire conversation text (all turns,
all roles, tool-result contents), and `copy_request_options` copies the original
`system` — Claude Code's system prompt alone is tens of thousands of tokens. Every
*new* image triggers a full-context vision call, so screenshot-heavy sessions grow
roughly quadratically in cost. The caching only helps repeated images, not new ones.

**3. The fail-open path is not equivalent to the existing fallback.**
`replace_images_after_image_context_failure` strips images unconditionally,
bypassing both the rectifier gates (`enabled && request_media_fallback`) and the
text-only heuristic. A user who configured `imageModel` for cost reasons while
using a vision-capable main model would silently lose images on a transient vision
failure. The form hint ("leave empty to keep the existing fallback") implies a
symmetry that doesn't hold.

**4. Smaller issues**, listed for completeness: `inject_content_blocks` drops
`cache_control` from replaced blocks (the sanitizer's marker replacement preserves
it); a takeover-enable failure inside the switch flow surfaces as "switch failed"
even though the switch itself succeeded; `claudeProxyTakeoverDisabled` is now a
dead i18n key in all four locales; the per-image cache only activates when all
images in a request are identical.

### The parts of this PR we'd gladly take

Your diagnosis of the *availability* gaps is correct and valuable — but those
belong in the existing rectifier, not a new pipeline. We'd welcome these as small
standalone PRs:

1. **Chinese error keywords in `is_unsupported_image_error`** — a real gap: CN
   gateways return Chinese error bodies, so the reactive image retry never triggers
   today (see #5025). Please tighten the keyword combinations somewhat, since this
   runs on the working path for all users.
2. **OpenAI-style image block detection** (`input_image` / data-URL / nested MIME)
   in the media sanitizer — also a genuine fix.
3. While at it, `known_text_only_model` is missing newer models such as `glm-5.2`;
   additions there directly close reported 400s.

These three changes close the actual breakage reported in issues at a tiny fraction
of the risk surface.

If you want to keep pursuing the vision-preprocessing feature itself, the
preconditions we'd need are: (a) a persisted analysis cache so injected text is
byte-stable across restarts, (b) the analysis request trimmed to the image-bearing
user message rather than full history + system, and (c) the fail-open path routed
through the existing rectifier gating. We recognize that is meaningful extra work —
which is exactly the complexity concern — but stating it explicitly seems fairer
than a vague "too complex".

Thanks again for the effort here; the split-out PRs above would be immediately
mergeable improvements.

farion1231:
One correction to my earlier comment: I attributed #5025 to Chinese error bodies, but after checking the user's logs, the actual error is English: `400 Model only support text input`. It slips past the reactive retry for two independent reasons: the message never mentions "image" (so the `mentions_image` gate fails before the hints even run), and the hint list has `"only supports text"` while the gateway writes `"only support text"` (no "s"). So the Chinese keyword addition, while welcome, would not have closed #5025.

We've since fixed that specific gap on our side: `glm-5.2` added to `known_text_only_model` (exact match, to avoid stripping images from a future `glm-5.2v` multimodal variant), plus a self-evident phrase check for `only support(s) text` that bypasses the image-mention gate. It will land on main shortly.

That narrows the split-out PRs we'd still welcome from this work to two pieces:

1. **Chinese error keywords** (tightened) — as defense-in-depth for gateways that do return Chinese error bodies;
2. **OpenAI-style image block detection** (`input_image` / data-URL / nested MIME) in the media sanitizer.

PR #3773, opened 2026-06-05
feat(provider): open Claude provider-scoped terminal

Amiyadesi:
﻿## Summary / ??

This PR keeps the provider terminal work intentionally small: Claude Code only.

When a user opens a terminal from a Claude provider card, cc-switch starts a new terminal session with that provider's env and a temporary Claude `--settings` file. It does not switch the live/global Claude config, and it does not touch other terminals that are already open in the same project.

I reviewed the isolation path again before updating this PR and found two real issues. First, repeated launches were using pid-based temp filenames, so two provider terminals opened from the same cc-switch process could reuse the same settings/script file. The updated launcher uses a per-launch suffix, so provider A, provider B, and a normal main-env terminal can sit next to each other without sharing temp files or process env.

Second, a parent cc-switch process can already have Claude env vars such as `ANTHROPIC_AUTH_TOKEN` set. The launcher now clears known Claude auth/base env keys before exporting the selected provider values on bash, cmd, and PowerShell. That keeps presets that use `ANTHROPIC_API_KEY` from being accidentally routed through an inherited auth token.

PowerShell now gets its own launcher script instead of going through `.bat`, so the provider env remains available in the PowerShell session after Claude exits. cmd/wt keep the batch path; macOS/Linux export the env before running Claude and then `exec bash` so the shell keeps the provider env. Failed launches clean up temp settings/script files.

## Related Issue / ?? Issue

No linked issue.

## Screenshots / ??

Not captured. This changes terminal launch behavior rather than a visible UI surface.

## Checklist / ????

- [x] Claude Code only; non-Claude apps are rejected at the command boundary
- [x] Provider env is scoped to the spawned terminal session, not the main app process or live config
- [x] Inherited Claude auth/base env vars are cleared before provider env is exported
- [x] Multiple provider terminals get separate temp settings/scripts
- [x] PowerShell, cmd/wt, macOS, and Linux launchers keep env available after Claude exits
- [x] `.omx/` is not part of the PR

## Verification / ??

- `rustup run stable cargo fmt --manifest-path src-tauri/Cargo.toml --check`
- `rustup run stable cargo test --manifest-path src-tauri/Cargo.toml --lib commands::misc::tests::extract_claude_terminal_env_vars_uses_api_key_field_without_auth_token -- --nocapture`
- `rustup run stable cargo test --manifest-path src-tauri/Cargo.toml --lib commands::misc::tests::build_shell_env_exports_quotes_values_and_skips_unsafe_keys -- --nocapture`
- `rustup run stable cargo test --manifest-path src-tauri/Cargo.toml --lib commands::misc::tests::build_windows_env_commands_escapes_values_and_skips_unsafe_keys -- --nocapture`
- `rustup run stable cargo test --manifest-path src-tauri/Cargo.toml --lib commands::misc::tests::build_powershell_env_commands_quotes_values_and_skips_unsafe_keys -- --nocapture`
- `rustup run stable cargo clippy --manifest-path src-tauri/Cargo.toml --lib --no-deps`
- `git diff --check`

Earlier verification kept from this branch:

- `rustup run stable cargo test --manifest-path src-tauri/Cargo.toml extract_claude_terminal_env_vars --lib`
- `rustup run stable cargo test --manifest-path src-tauri/Cargo.toml next_provider_terminal_launch_suffix --lib`
- `rustup run stable cargo test --manifest-path src-tauri/Cargo.toml build_windows_powershell_script --lib`
- `rustup run stable cargo test --manifest-path src-tauri/Cargo.toml write_claude_config_writes_env_object --lib`
- `rustup run stable cargo test --manifest-path src-tauri/Cargo.toml sanitize_temp_file_component_replaces_path_chars --lib`
- `node .\node_modules\typescript\bin\tsc --noEmit`
- `node .\node_modules\prettier\bin\prettier.cjs --check src/**/*.{js,jsx,ts,tsx,css,json}`

Notes:

- Clippy finishes with two existing warnings outside this change: `src/settings.rs` has an unused import, and `src/commands/misc.rs:645` has a needless `return`.
- I also checked `rustup run stable cargo test --manifest-path src-tauri/Cargo.toml --lib`. On this Windows checkout it is not green because two unrelated tests fail on raw backslashes in temp paths: `services::proxy::tests::codex_restore_from_backup_preserves_model_catalog_pointer` and `session_manager::providers::openclaw::tests::delete_session_updates_index_and_removes_jsonl`.

### DIFF
--- a/src-tauri/src/commands/misc.rs
+++ b/src-tauri/src/commands/misc.rs
@@ -2,12 +2,14 @@
 
 use crate::app_config::AppType;
 use crate::init_status::{InitErrorPayload, SkillsMigrationPayload};
+use crate::provider::Provider;
 use crate::services::ProviderService;
 use once_cell::sync::Lazy;
 use regex::Regex;
 use std::collections::HashMap;
 use std::path::{Path, PathBuf};
 use std::str::FromStr;
+use std::sync::atomic::{AtomicU64, Ordering};
 use tauri::AppHandle;
 use tauri::State;
 use tauri_plugin_opener::OpenerExt;
@@ -17,6 +19,8 @@ use std::os::windows::process::CommandExt;
 
 #[cfg(target_os = "windows")]
 const CREATE_NO_WINDOW: u32 = 0x08000000;
+
+static NEXT_PROVIDER_TERMINAL_LAUNCH_ID: AtomicU64 = AtomicU64::new(1);
 
 /// 打开外部链接
 #[tauri::command]
@@ -2393,6 +2397,10 @@ pub async fn open_provider_terminal(
     cwd: Option<String>,
 ) -> Result<bool, String> {
     let app_type = AppType::from_str(&app).map_err(|e| e.to_string())?;
+    if app_type != AppType::Claude {
+        return Err("Provider terminal currently supports Claude Code only".to_string());
+    }
+
     let launch_cwd = resolve_launch_cwd(cwd)?;
 
     // 获取提供商配置
@@ -2403,9 +2411,7 @@ pub async fn open_provider_terminal(
         .get(&providerId)
         .ok_or_else(|| format!("提供商 {providerId} 不存在"))?;
 
-    // 从提供商配置中提取环境变量
-    let config = &provider.settings_config;
-    let env_vars = extract_env_vars_from_config(config, &app_type);
+    let env_vars = extract_claude_terminal_env_vars(provider);
 
     // 根据平台启动终端，传入提供商ID用于生成唯一的配置文件名
     launch_terminal_with_env(env_vars, &providerId, launch_cwd.as_deref())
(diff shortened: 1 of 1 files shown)


farion1231:
@codex review

farion1231 (commented):
Thank you for your contribution. Please check if this issue exists:

[P2] src-tauri/src/commands/misc.rs:3162, 3177, 3233 on PR head 8e9c78c8: the launcher exports provider env vars, but never clears inherited Claude auth vars. For presets using apiKeyField: "ANTHROPIC_API_KEY" such as PatewayAI, a cc-switch process launched with ANTHROPIC_AUTH_TOKEN already set will spawn a terminal containing both the old inherited token and the provider’s ANTHROPIC_API_KEY. Since the app itself treats ANTHROPIC_AUTH_TOKEN as higher priority, this can route the provider-scoped terminal through the wrong credential and breaks the isolation promise. I’d unset known Claude auth/base env keys before exporting provider values on bash, cmd, and PowerShell, or explicitly clear the sibling auth key when selecting one.

PR #3827, opened 2026-06-07
Fix Codex editor showing stale auth key for active third-party providers

abkkkbb:
## Summary

  Fixes a Codex provider editor mismatch when editing the currently active third-party provider.

  When CC Switch preserves the Codex official auth file, the live `~/.codex/auth.json` may remain unchanged while the
  active third-party provider token is projected into `config.toml` as `experimental_bearer_token`.

  Previously, the editor initialized the form from live settings directly, so it could show a stale
  `auth.OPENAI_API_KEY` from `auth.json` even though Codex was actually using the correct provider token from
  `config.toml`.

  This change restores the live projected token back into the provider auth shape when initializing the editor.

  ## Changes

  - Extract `experimental_bearer_token` from live Codex `config.toml` when editing the active provider.
  - Restore that token into `auth.OPENAI_API_KEY` for the editor form.
  - Remove the runtime-projected `experimental_bearer_token` from the editable TOML to avoid saving live projection
  details back into provider storage.
  - Keep the existing `modelCatalog` database SSOT behavior.
  - Add a regression test for the stale live auth case.

  ## Why

  For third-party Codex providers, `auth.json` may intentionally stay unchanged to preserve official Codex auth/login
  state. In that mode, the provider token lives in `config.toml` at runtime. The editor should show the effective
  provider token, not the stale preserved `auth.json` key.

### DIFF
--- a/src/components/providers/EditProviderDialog.tsx
+++ b/src/components/providers/EditProviderDialog.tsx
@@ -9,6 +9,10 @@ import {
   type ProviderFormValues,
 } from "@/components/providers/forms/ProviderForm";
 import { openclawApi, providersApi, vscodeApi, type AppId } from "@/lib/api";
+import {
+  extractCodexExperimentalBearerToken,
+  updateCodexExperimentalBearerToken,
+} from "@/utils/providerConfigUtils";
 
 interface EditProviderDialogProps {
   open: boolean;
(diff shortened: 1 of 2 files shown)


farion1231:
@codex review

abkkkbb:
The previous frontend failure was caused by Prettier formatting. I pushed a formatting-only follow-up commit and
  verified locally:

  - `pnpm format:check`
  - `pnpm typecheck`
  - `pnpm vitest run tests/components/EditProviderDialog.test.tsx`

  The latest workflow appears to be waiting for maintainer approval to run.

PR #3879, opened 2026-06-08
Add managed OAuth account selection for providers

SaladDay:
## Summary
- Add managed ChatGPT/Codex OAuth account binding support for Codex official providers.
- Replace inline OAuth account management in provider add/edit forms with a focused account selector plus a Manage Accounts entry point.
- Reuse the auth center in a focused secondary panel, auto-scroll to the relevant account section, and preserve the provider form state when returning.
- Clarify the unmanaged browser-login option for users who do not want to bind a managed account.

## Merge note
Please avoid squash-merging this PR if possible.

The three commits are intentionally separated and each has a meaningful rollback boundary:
1. `Add managed ChatGPT account binding for Codex`
2. `Consolidate managed account entry points`
3. `Clarify unmanaged Codex login option`

Keeping them as separate commits makes targeted rollback easier and keeps the project history clean.

## Verification
- `npx -y pnpm@10.15.1 typecheck`
- `npx -y pnpm@10.15.1 build:renderer`
- `git diff --check`
- Fresh-eyes xhigh review loop until no blocking issues remained


farion1231:
@codex review

farion1231 (commented):
感谢您的贡献，请查看一下是否有以下问题：

托管 Codex official 写入的 auth.json 只有短期 access_token，没有 refresh 能力。token 过期后 Codex CLI 不能自刷新，用户可能需要重新切换/保存 provider 才恢复。proxy 备份释放时也可能还原过期 token。

token 刷新时持有共享 RwLock 并跨网络 await，且请求走 600s timeout。网络卡住时，切换、认证中心账号操作、后续 token 读取都可能被拖住。

前端 status 查询失败会把 accounts 派生成空数组，select effect 会静默清空已绑定账号。保存后会删除 providerType/authBinding，可能把托管绑定变成未绑定坏配置。

switch/update 先提交 current provider / DB，再写 live。托管 live 写入现在包含网络取 token，失败概率更高；失败后会出现 UI/DB 指向 B，但实际 live 仍是 A。

也应修复
backfill 在 marker 不匹配时不剥离 live auth，可能把用户原生 Codex 登录的 access/refresh token 持久化进托管 provider 的 DB 配置。

新增 select 模式的 4 个 i18n key 缺失，英/日/繁体界面会显示中文 fallback。

YUZHEthefool:
Merged the latest `main` into this PR and completed the conflict resolution and timing-safety follow-ups. Merge commit: `ee15ebfc`.

### Conflict resolution

- Resolved all 14 conflicted files while preserving the latest changes from `main`.
- ChatGPT/Codex OAuth and Grok/xAI OAuth remain fully independent features:
  - Codex uses `codex_oauth` and `reauth_required`.
  - xAI uses `xai_oauth` and `requires_reauth`.
  - They only share the Provider form's type-discrimination entry point; authentication state, account validation, and backend token handling remain separate.
- Auth Center keeps independent sections and scroll targets for Copilot, Codex, and xAI.

### Review feedback addressed

1. Managed Codex `auth.json` now contains a complete refreshable token bundle: `access_token`, `id_token` when available, `refresh_token`, and `last_refresh`. It also adopts refresh material rotated by the Codex CLI before rewriting the file.
2. OAuth HTTP requests now use a 30-second timeout. Token refresh is serialized with per-account locks instead of holding an outer shared write lock across network awaits.
3. Failed Codex/Copilot status queries no longer treat an empty account list as authoritative, so an existing Provider binding is not silently cleared.
4. Managed Codex switch/update operations now preflight the token, write Live/backup state first, and commit DB/current last. Failures restore the exact auth/config/catalog/marker snapshot.
5. Managed Codex backfill always strips Live auth and restores the Provider's stored placeholder auth, preventing native Codex access/refresh tokens from being persisted into the Provider database row.
6. Selector and reauthentication strings are present in English, Japanese, Simplified Chinese, and Traditional Chinese.
7. FAST mode remains available in selector mode.

### Additional timing fix

The first managed Codex Provider previously saved its DB row before token preflight and Live writes. A failed preflight could therefore leave an unusable orphan Provider.

The first-provider path now runs under the Codex switch lock in this order:

`token preflight → Live write → Provider DB row → DB current`

Any failure restores Live state and removes or restores the newly written Provider row. Regression tests cover both a missing managed account and a DB current-commit failure.

### Verification

- `cargo fmt --check`
- `cargo check`
- `pnpm typecheck`
- `pnpm build:renderer`
- 8 managed Codex transaction tests passed
- 20 Codex OAuth manager tests passed
- Token-bundle and backfill regression tests passed
- 60 focused frontend tests across 10 files passed, covering OAuth Provider forms, Auth Center, presets, and all four locales
- `git diff --check`

Only existing bundle-size, stale browser-data, and unmatched MSW request warnings remain; there were no failures.

farion1231 (commented):
Thanks — this has come a long way from the June version. The transaction design is solid: the four-file live snapshot with rollback that preserves newer same-account auth held up under every failure path I traced (direct switch, first add, takeover edit, takeover enable), and all six of my earlier comments are genuinely addressed rather than worked around. Adopt-on-preflight plus retry-from-disk on `RefreshTokenInvalid` is the right shape for CLI token rotation. The merge tree against current main passes the full backend suite (including the 8 managed-transaction tests), typecheck, and all frontend unit tests.

That said, the ownership-tracking design has one structural problem I want fixed before merging, plus a credential-cleanup gap in the auth center.

### [P1] Live-auth ownership is claimed by an access-token fingerprint, which the Codex CLI destroys on its first self-refresh

The marker written at [codex_config.rs#L451](https://github.com/farion1231/cc-switch/blob/ee15ebfc9c9664d455f9e5099f56a3d6ce5ecded/src-tauri/src/codex_config.rs#L451) records `sha256(access_token)`. Access tokens live for hours, and any bare `codex` run past the TTL self-refreshes against auth.openai.com (never through the proxy) and rewrites `auth.json` — after which [`codex_auth_matches_recorded_managed_oauth`](https://github.com/farion1231/cc-switch/blob/ee15ebfc9c9664d455f9e5099f56a3d6ce5ecded/src-tauri/src/codex_config.rs#L464) can never match again. From then on:

- Switching to an unbound official provider does a config-only write ([codex_config.rs#L2564-L2573](https://github.com/farion1231/cc-switch/blob/ee15ebfc9c9664d455f9e5099f56a3d6ce5ecded/src-tauri/src/codex_config.rs#L2564-L2573)), the managed cleanup at [provider/mod.rs#L4254-L4257](https://github.com/farion1231/cc-switch/blob/ee15ebfc9c9664d455f9e5099f56a3d6ce5ecded/src-tauri/src/services/provider/mod.rs#L4254-L4257) silently skips on the marker mismatch, and [`clear_stale_codex_live_auth_after_official_switch`](https://github.com/farion1231/cc-switch/blob/ee15ebfc9c9664d455f9e5099f56a3d6ce5ecded/src-tauri/src/codex_config.rs#L939) only removes third-party residue. Result: the UI says "not bound" while `~/.codex/auth.json` still holds a live managed login that keeps consuming that ChatGPT account.
- With the residue in place, the backup-strip and backfill paths copy the real token bundle into the DB (live_backup / provider row), breaking the "real tokens never enter the DB" invariant this PR otherwise maintains carefully.
- The hot-switch path has no managed cleanup at all — neither branch at [services/proxy.rs#L2650-L2692](https://github.com/farion1231/cc-switch/blob/ee15ebfc9c9664d455f9e5099f56a3d6ce5ecded/src-tauri/src/services/proxy.rs#L2650-L2692) has an equivalent of the direct-switch cleanup, so switching from managed account A to an unbound official provider during takeover keeps serving A.

Minimal fix: make ownership content-based. You already have [`codex_live_auth_is_managed_chatgpt_login`](https://github.com/farion1231/cc-switch/blob/ee15ebfc9c9664d455f9e5099f56a3d6ce5ecded/src-tauri/src/codex_config.rs#L524) and use it in the refresh-sync path — use it (account-scoped) as the deletion predicate instead of the fingerprint, and before deleting, adopt the on-disk refresh token back into the manager (`read_codex_live_auth_refresh_for_account` → `adopt_account_refresh_token`) so a CLI-rotated token is never lost. Add the same cleanup call to the hot-switch transaction. One trade-off to make explicitly: content-based matching also claims a *native* `codex login` of the same account — adopt-before-delete makes that lossless, but leave a comment saying so.

### [P1] Removing an account / auth-center logout leaves a fully refreshable login in `~/.codex/auth.json` indefinitely

[`remove_account`](https://github.com/farion1231/cc-switch/blob/ee15ebfc9c9664d455f9e5099f56a3d6ce5ecded/src-tauri/src/proxy/providers/codex_oauth_auth.rs#L928-L959) and [`clear_auth`](https://github.com/farion1231/cc-switch/blob/ee15ebfc9c9664d455f9e5099f56a3d6ce5ecded/src-tauri/src/proxy/providers/codex_oauth_auth.rs#L979-L1017) clear manager state and the manager's own storage file only, and the command layer ([auth.rs#L327-L332](https://github.com/farion1231/cc-switch/blob/ee15ebfc9c9664d455f9e5099f56a3d6ce5ecded/src-tauri/src/commands/auth.rs#L327-L332), [#L393-L395](https://github.com/farion1231/cc-switch/blob/ee15ebfc9c9664d455f9e5099f56a3d6ce5ecded/src-tauri/src/commands/auth.rs#L393-L395)) adds nothing. If the removed account is the currently bound one, the live file keeps a bundle that bare `codex` can refresh forever — and since removal also unbinds the provider row, the next switch-away no longer knows which account to clean, so the residue is permanent. "Log out of all accounts" should mean the machine no longer holds working credentials cc-switch wrote. Fix: on remove/logout, if live auth is a managed login for that account (content check as above), delete `auth.json` and the marker.

### [P2] Upgraded accounts have `token_updated_at_ms == 0`, which disables the adopt gate

At [codex_oauth_auth.rs#L865-L867](https://github.com/farion1231/cc-switch/blob/ee15ebfc9c9664d455f9e5099f56a3d6ce5ecded/src-tauri/src/proxy/providers/codex_oauth_auth.rs#L865-L867), `token_updated_at_ms <= 0` makes `observed_is_newer` unconditionally true, so any stale same-account token on disk replaces the manager's valid refresh token and is persisted. The retry branch then cannot recover (disk holds the same dead token) → hard logout requiring re-login. This branch has no test. Fix: for `<= 0`, allow stamping the timestamp but don't replace token material.

### [P2] The `RefreshTokenInvalid` recovery can be vetoed by the timestamp heuristic

In the retry branch ([codex_oauth_auth.rs#L633-L652](https://github.com/farion1231/cc-switch/blob/ee15ebfc9c9664d455f9e5099f56a3d6ce5ecded/src-tauri/src/proxy/providers/codex_oauth_auth.rs#L633-L652)) the server has just proven the manager's token dead, so a *different* disk token is strictly better — but adoption is still rejected when `last_refresh` is missing or unparseable, on clock skew, or on a same-millisecond stamp. Fix: let the invalid-token retry bypass the `observed_is_newer` gate.

### [P2] `add_provider` / `update_provider` are sync commands that now run a network preflight

[commands/provider.rs#L38-L61](https://github.com/farion1231/cc-switch/blob/ee15ebfc9c9664d455f9e5099f56a3d6ce5ecded/src-tauri/src/commands/provider.rs#L38-L61) — saving a managed binding runs the OAuth preflight (up to the ~30 s network timeout) on the main thread and freezes the window. `switch_provider` ([#L104-L110](https://github.com/farion1231/cc-switch/blob/ee15ebfc9c9664d455f9e5099f56a3d6ce5ecded/src-tauri/src/commands/provider.rs#L104-L110)) already does async + `spawn_blocking`; these two should match.

### [P2] Re-adding the official Codex preset silently overwrites the existing official provider

[mutations.ts#L53-L72](https://github.com/farion1231/cc-switch/blob/ee15ebfc9c9664d455f9e5099f56a3d6ce5ecded/src/lib/query/mutations.ts#L53-L72): after ensure-seed, the `{...officialProvider, ...rest}` update replaces the existing row's name/binding/usage script with form values without any prompt, and the `inserted` flag returned by `ensureCodexOfficialProvider` is discarded. The `clears a prior managed binding` test currently encodes this as expected behavior — I'd rather see a confirm dialog or a merge that preserves the existing row.

### [P2] A `reauth_required` account can be bound and saved

[CodexOAuthSection.tsx#L134-L138](https://github.com/farion1231/cc-switch/blob/ee15ebfc9c9664d455f9e5099f56a3d6ce5ecded/src/components/providers/forms/CodexOAuthSection.tsx#L134-L138) computes the state and the UI warns that re-login is needed, but the save gate only checks the account exists; the Claude Desktop form has the symmetric block ([ClaudeDesktopProviderForm.tsx#L614](https://github.com/farion1231/cc-switch/blob/ee15ebfc9c9664d455f9e5099f56a3d6ce5ecded/src/components/providers/forms/ClaudeDesktopProviderForm.tsx#L614)). The auth.json written for such an account lacks `id_token`. Fix: block save for a reauth-required selection, matching the warning.

### [P2] The semi-takeover restore path writes a managed bundle without recording the marker

[services/proxy.rs#L2667-L2692](https://github.com/farion1231/cc-switch/blob/ee15ebfc9c9664d455f9e5099f56a3d6ce5ecded/src-tauri/src/services/proxy.rs#L2667-L2692) calls `write_codex_provider_live_with_catalog` directly. The marker is recorded by the other three managed write sites ([live.rs#L1234-L1241](https://github.com/farion1231/cc-switch/blob/ee15ebfc9c9664d455f9e5099f56a3d6ce5ecded/src-tauri/src/services/provider/live.rs#L1234-L1241), [proxy.rs#L3123](https://github.com/farion1231/cc-switch/blob/ee15ebfc9c9664d455f9e5099f56a3d6ce5ecded/src-tauri/src/services/proxy.rs#L3123), [codex_config.rs#L630](https://github.com/farion1231/cc-switch/blob/ee15ebfc9c9664d455f9e5099f56a3d6ce5ecded/src-tauri/src/codex_config.rs#L630)) but not here, so a bundle written through this path can never be claimed by cleanup later. Fix: record the marker here too, or fold marker recording into the write helper so no path can forget it.

### nits

- All four locale files carry a duplicate `providerKeyStatusLoading` key inside `providerForm` (e.g. zh.json L1028 and L1180, a merge artifact) — the later one silently wins.
- The lock-order comment in codex_oauth_auth.rs states the reverse of the actual `storage_lock` → `accounts` acquisition order.
- The switch path could copy the compare-before-write that [`sync_codex_managed_oauth_live_auth_after_refresh`](https://github.com/farion1231/cc-switch/blob/ee15ebfc9c9664d455f9e5099f56a3d6ce5ecded/src-tauri/src/codex_config.rs#L597-L633) already does, closing a small adopt→write race.
- Deleting a bound account silently unbinds the provider with no toast.

**To merge**: the two P1s — they are one fix family (content-based ownership as the deletion predicate, adopt-before-delete, and the same cleanup wired into switch-away, hot-switch, and remove/logout). The P2s can land as fast follow-ups if you prefer, though the adopt-gate pair (`token_updated_at_ms` / `RefreshTokenInvalid`) is small and closes real hard-logout paths. Happy to re-review quickly once the ownership fix lands.

YUZHEthefool:
Implemented the requested follow-up in `4b14026b`.

### Credential ownership and lifecycle

- Replaced access-token fingerprint ownership with account-scoped content matching. Marker v2 records the account ID and remains backward-compatible with legacy v1 markers.
- Managed switch/update/takeover paths now adopt Codex CLI-rotated refresh material before overwriting or deleting live auth, then compare the disk generation again immediately before the mutation.
- Added the same outgoing-account cleanup to direct switches, active-takeover hot switches, managed-to-unbound Official transitions, managed-to-third-party transitions, backup stripping/backfill, and semi-takeover restoration.
- Auth Center account removal and logout now delete matching `auth.json` plus the ownership marker, and serialize with provider add/update/switch/hot-switch through the shared Codex switch lock.
- Proxy restart re-projection now takes that same per-app lock and rechecks takeover state after acquiring it, preventing preflighted credentials from being recreated after removal/logout.

### Token-generation safety

- Added explicit reconciliation outcomes for synchronized, adopted, provably older, ambiguous, and unmanaged generations.
- Legacy accounts and missing/equal timestamps remain fail-closed when token material differs; they are no longer assigned a synthetic timestamp that could make a later retry delete the newer CLI generation.
- All managed bundle writes reject unresolved same-account generation conflicts, including updates while a valid access-token cache exists.
- `RefreshTokenInvalid` still bypasses timestamp ambiguity after the server has rejected the manager token, adopts a different disk generation, and retries once.
- Compare-before-write/delete guards and regression coverage now verify that auth/config/catalog/marker state is left unchanged on a generation mismatch.

### Other review follow-ups

- `add_provider` and `update_provider` now use async `spawn_blocking` command wrappers.
- Managed accounts without an `id_token` are rejected before writing live auth.
- Re-adding the fixed Codex Official preset preserves an existing row.
- Provider forms block `reauth_required` Codex accounts, including the default-account fallback.
- Managed account removal now shows a success toast, and duplicate locale keys were removed.

### Verification

- `cargo fmt --check`
- Full Rust suite: 2,380 passed, 4 ignored, plus all integration suites passed
- `pnpm typecheck`
- `pnpm format:check`
- Full frontend suite: 104 files / 707 tests passed
- `git diff --check`
- Windows x64 release executable and NSIS 3.19.1 package built successfully

Could you take another look when convenient?

farion1231:
已合并，感谢 @SaladDay 的完整实现，也特别感谢 @YUZHEthefool 对审查意见的系统性修复。从 marker v2、内容型归属判定、恢复文件事务到多官方账号与身份迁移，几轮迭代下来质量很扎实，失败注入与并发场景的测试覆盖尤其到位。

合并后留有三个不阻断的小项，欢迎有空时以 follow-up 跟进（不跟也没关系，我们会排期）：

1. Device Code 登录与「注销所有」之间有一个亚秒级竞态：`clear_auth` 执行期间已在网络往返中的 `start_device_flow` 归来后仍会登记 pending code，流程存活。建议 manager 加一个 `login_epoch`（clear 时递增；start 在发起网络请求前采样、登记时不符则放弃）。只需要动 clear 路径——remove 单账号语义上不该作废进行中的登录，不用动。
2. 恢复路径的硬链接探针在 `\\wsl.localhost`（9P 不支持硬链接）和 exFAT 上会 fail-closed，这类配置目录上停止接管的恢复会报"文件系统不支持安全恢复"。可以考虑探针失败时降级为直写并输出告警日志。
3. `native-login-exists` 查询在加载期默认按 true 处理，「使用 Codex 当前登录」选项会短暂不可见，可以给查询加载态一个中间处理。

官方卡整体退出自动故障转移的方向已确认接受，发版说明里我们会向用户说明这一行为变化。

PR #3884, opened 2026-06-08
fix: image_url format not detected/sanitized for Codex->Chat path (DeepSeek 400)

YXY666888:
## Problem

When Codex CLI sends a Requests API request containing input_image items, CC Switch converts to Chat Completions format. DeepSeek rejects image_url blocks with HTTP 400.

## Root Cause

Two bugs compounded:

1. media_sanitizer only detected type:image (Anthropic format), missing type:image_url (OpenAI Chat format)
2. apply_media_prevention was only called in the Claude adapter path, missing the Codex->Chat conversion path
3. is_unsupported_image_error() didn't match 'unknown variant' error messages from upstream

## Fix

- media_sanitizer.rs: handle both image and image_url in detection and replacement
- media_sanitizer.rs: add 'unknown variant' to UNSUPPORTED_HINTS
- forwarder.rs: call apply_media_prevention after responses_to_chat_completions_with_reasoning in Codex path

3 new tests added.

### DIFF
--- a/src-tauri/src/proxy/forwarder.rs
+++ b/src-tauri/src/proxy/forwarder.rs
@@ -1335,10 +1335,12 @@ impl RequestForwarder {
             super::providers::apply_codex_chat_upstream_model(provider, &mut mapped_body);
             let reasoning_config =
                 super::providers::resolve_codex_chat_reasoning_config(provider, &mapped_body);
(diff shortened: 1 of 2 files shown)


farion1231:
@codex review

Issue #1678, opened 2026-03-25
Claude Code Auto Mode Classifier Compatibility Issue


xzjncu:
# Claude Code Auto Mode Classifier Compatibility Issue

## Problem Description

When using CC Switch's Anthropic Base URL proxy (e.g., `http://127.0.0.1:15721`), Claude Code's Auto Mode classifier fails with the following error:

```
The auto mode classifier is temporarily unavailable, so Bash cannot be used right now. (dacEnabled=n/a, hasExtraBody=false)
```

This results in:
- ❌ Bash/WebFetch tools being blocked
- ⚠️ Write/Edit tools working intermittently
- ✅ Read/Grep/Glob tools working normally
- ✅ MCP tools working normally (independent permission path)

## Root Cause Analysis

Through systematic debugging, we identified that the local proxy intercepts Claude Code's API requests to `api.anthropic.com`, but the proxy implementation is incompatible with the Auto Mode classifier service's backend connection requirements.

**Key diagnostic insight**:
- `curl` tests show normal TLS handshake and HTTP 405 response (expected)
- Claude Code's classifier still fails with `dacEnabled=n/a`
- This is because `curl` uses system proxy, while Claude Code uses CC Switch's Base URL proxy

**Network path difference**:
| Tool | Network Path |
|------|-------------|
| `curl` | System proxy (Clash Verge HTTP/HTTPS port) |
| Claude Code | CC Switch Base URL (`http://127.0.0.1:15721`) |

## Reproduction Steps

1. Configure CC Switch with Anthropic Base URL: `http://127.0.0.1:15721`
2. Start Claude Code in Auto Mode
3. Attempt to use Bash tool: `echo "test"`
4. Observe error: `dacEnabled=n/a`

## Workaround

**Current solution**: Disable CC Switch's Anthropic Base URL proxy.

After disabling the proxy and restarting Claude Code session, the classifier immediately recovers.

## Suggested Fix

It would be helpful if CC Switch could:
1. Document the incompatibility with Claude Code's Auto Mode classifier
2. Provide a configuration option to bypass the classifier service endpoint
3. Or investigate why the local proxy breaks the classifier's backend connection

## Environment

- Claude Code: v2.1.74
- CC Switch: v3.12.3
- System: macOS Darwin 25.3.0
- Proxy: Clash Verge (TUN mode enabled/disabled - both tested, not related to this issue)

## Additional Notes

- This issue is **NOT** related to Clash Verge TUN mode (verified)
[...]

PR #3901, opened 2026-06-08
fix(proxy): decode zstd / undeclared-compressed non-streaming responses

EatingCarrot:
Some upstream relays (e.g. one-api/new-api behind Cloudflare + Fastly) zstd-compress non-streaming JSON responses but strip the `content-encoding` header. cc-switch's `read_decoded_body` only decompresses when `content-encoding` is present and did not support zstd, so the body was forwarded as raw zstd bytes while still labeled `content-type: application/json`; the Anthropic SDK then fails to parse it.

This specifically breaks Claude Code's auto permission mode: the command safety classifier issues a NON-streaming request, whereas normal chat streams (and the streaming path already forces
`accept-encoding: identity`). So only auto mode hits the compressed-JSON path and reports "<model> is temporarily unavailable, so auto mode cannot determine the safety of <tool> right now".

- Add `zstd = "0.13"` and support `zstd` in `decompress_body`.
- Add `sniff_compression`: detect gzip (1f 8b) / zstd (28 b5 2f fd) frame magic and decompress even when `content-encoding` is absent or wrong. Valid JSON/text never starts with these magics, so no false positives.
- Wire the sniff fallback into `read_decoded_body`, covering both the passthrough (`handle_non_streaming`) and transform (`handle_claude_transform`) non-streaming paths.
- Add unit tests, including the repro: zstd body with no `content-encoding` header is decoded to parseable JSON.

Refs #1678, #1847

## Related Issue / 关联 Issue

Fixes #

## Screenshots / 截图

## Checklist / 检查清单

### DIFF
--- a/src-tauri/Cargo.lock
+++ b/src-tauri/Cargo.lock
@@ -802,6 +802,7 @@ dependencies = [
  "windows-sys 0.61.2",
  "winreg 0.52.0",
  "zip 2.4.2",
+ "zstd",
 ]
 
 [[package]]
--- a/src-tauri/Cargo.toml
+++ b/src-tauri/Cargo.toml
@@ -43,6 +43,7 @@ reqwest = { version = "0.12", features = ["rustls-tls", "json", "stream", "socks
 arboard = "3.6"
 flate2 = "1"
 brotli = "7"
+zstd = "0.13"
 tokio = { version = "1", features = ["macros", "rt-multi-thread", "time", "sync"] }
 futures = "0.3"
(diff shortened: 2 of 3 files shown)


farion1231:
@codex review

chenx-dust:
Partly duplicated: https://github.com/farion1231/cc-switch/pull/3817
不过希望解决的不是同一个问题

Marukoyo:
希望支持 雀氏很多CDN 都加了 HTTP 响应体压缩

PR #3911, opened 2026-06-08
feat(failover): show mapped upstream model in failover queue

ZQ-jhon:
## What

Surface the **actual upstream model** each provider maps to, directly in the failover queue UI.

When a Claude provider points at a non-Claude upstream via an Anthropic-compatible endpoint (e.g. `ANTHROPIC_MODEL=deepseek-chat`), the failover queue previously only showed the provider's display name. Users couldn't tell that "P2" actually downgrades to DeepSeek/GLM/Qwen. This adds a small `→ <model>` badge to each queue item so the cross-model downgrade chain is visible at a glance.

## Why

cc-switch already supports cross-model failover today: `model_mapper.rs` rewrites the inbound model to each provider's configured `ANTHROPIC_MODEL`, and `select_providers` walks the queue. The only gap was discoverability — the queue UI didn't show where each entry really routes. This PR closes that gap **without touching any backend / routing / circuit-breaker logic**.

## Changes

- **`src/utils/providerConfigUtils.ts`** — new `getModelFromConfig(settingsConfig, appType)` helper that resolves a provider's upstream model name (Claude: `ANTHROPIC_MODEL` with SONNET/OPUS/HAIKU fallback; Codex/Gemini: their respective keys). Mirrors the existing `getApiKeyFromConfig` pattern and matches `model_mapper.rs` precedence.
- **`src/components/proxy/FailoverQueueManager.tsx`** — front-end join: fetch providers via `useProvidersQuery`, build a `providerId → model` map, render a `→ <model>` badge per queue item. Pure UI, no new IPC commands.
- **`src/utils/providerConfigUtils.test.ts`** — 6 new unit tests for `getModelFromConfig`, including the cross-model mapping case. Tests assert behavior/invariants, not data snapshots.
- **`docs/`** — a plan doc and a Chinese usage guide for the Anthropic-compatible cross-model failover route (config example + verification steps).

## Testing

- `tsc --noEmit` (project config): clean, no errors.
- `vitest run`: **54 files / 300 tests pass** (3 pre-existing + 6 new in the util test).
- No backend changes, so no Rust test impact.

## Notes

This is intentionally scoped to the **same-protocol** (Anthropic-compatible) case, which is already functional. Cross-*protocol* failover (Anthropic↔OpenAI↔Gemini) would require a proxy-side protocol bridge and is explicitly out of scope here.

### DIFF
--- a/docs/guide-cross-model-failover.md
+++ b/docs/guide-cross-model-failover.md
@@ -1,0 +1,120 @@
+# 跨模型故障转移使用指南（Anthropic 兼容路线）
+
+> 适用版本：cc-switch v3.16+
+> 目标读者：希望「当首选供应商（如 OpenRouter 代理的 Claude）不可用时，
+> 自动降级到另一个模型（如 DeepSeek / GLM / Qwen）」的用户。
+
+cc-switch 的故障转移**不要求队列里的供应商都是同一个模型**——只要它们都是
+同一个应用（如 Claude）下、且都讲 **Anthropic 兼容协议**的供应商，就能混在一个
+故障转移队列里。代理在转发时会按每个供应商配置的 `ANTHROPIC_MODEL` 把请求重写到
+它真正的上游模型。因此「OpenRouter-Claude 挂了 → 自动切到 DeepSeek」是开箱即用的，
+**无需改任何代码，只需正确配置供应商**。
+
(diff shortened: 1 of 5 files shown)


farion1231:
@codex review

PR #3928, opened 2026-06-09
fix(streaming_codex_chat): prevent empty call_id/name from overwritin…

luantu:
Root cause: push_tool_call_delta() unconditionally overwrites state.call_id and state.name even when id_delta/name_delta are empty strings (Some("") instead of None). Some upstream providers (DeepSeek via thirdparty) send empty id/name in late streaming chunks, corrupting previously-set correct values.

Fixes:
1. push_tool_call_delta: skip empty string overwrites
2. finalize_tools "already added" path: add defensive fallback for empty call_id/name before emitting output_item.done

Fixes the [SRV-006] connection closed before message completed symptom caused by Codex rejecting tool calls with empty name.

## Summary / 概述

Fix tool call state corruption when streaming providers send empty id/name deltas in late chunks, which caused Codex to reject tool calls with empty name and disconnect.

## Related Issue / 关联 Issue

N/A

## Screenshots / 截图

N/A - no UI changes.

## Checklist / 检查清单


farion1231:
@codex review

luantu:
其他PR已解决，45555638 Chat API: skip tool calls with missing function names (#4159)

farion1231:
感谢您的贡献！

luantu ***@***.***> 于2026年6月22日周一 17:53写道：

PR #3958, opened 2026-06-09
fix(media-sanitizer): also detect and strip image_url blocks for text-only models

fan166:
## Problem

When Codex sends a request with images through CC Switch to a **non-vision provider** (like DeepSeek), the proxy converts the Responses API format to Chat Completions format via `transform_codex_chat.rs`. This conversion emits image content as `{ type: "image_url" }` blocks (OpenAI Chat format).

However, `media_sanitizer.rs` only checked for `{ type: "image" }` (Anthropic format), so images in Chat-format requests were forwarded to providers that don't support vision, causing:

> `Failed to deserialize the JSON body into the target type: messages[194]: unknown variant `image_url`, expected `text``

## Fix

Updated two functions in `media_sanitizer.rs`:

1. `content_has_image_blocks()` — now also checks for `"image_url"` type
2. `replace_images_in_content()` — now also replaces `"image_url"` type blocks with the `[Unsupported Image]` marker

## Related

This complements the existing Anthropic-format (`{ type: "image" }`) sanitization that was already working correctly for the `transform.rs` (Anthropic→OpenAI) path.

### DIFF
--- a/src-tauri/src/proxy/media_sanitizer.rs
+++ b/src-tauri/src/proxy/media_sanitizer.rs
@@ -113,7 +113,8 @@ fn content_has_image_blocks(content: &Value) -> bool {
     };
 
     blocks.iter().any(|block| {
-        block.get("type").and_then(Value::as_str) == Some("image")
+        let block_type = block.get("type").and_then(Value::as_str);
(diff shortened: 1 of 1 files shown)


farion1231:
@codex review

PR #3973, opened 2026-06-09
[codex] Fix WebDAV saved password placeholder

xwil1:
## Summary

This PR fixes the WebDAV sync settings password field when a password has already been saved.

Previously, the backend correctly redacted the saved WebDAV password before returning settings to the frontend, but the frontend could only see an empty `password` value. As a result, the password field looked blank even when a password was already configured.

This change adds a non-sensitive `passwordConfigured` flag so the frontend can show a masked placeholder (`••••••••`) without receiving or storing the real password.

## What changed

- Add `passwordConfigured` to WebDAV sync settings returned to the frontend.
- Keep the actual WebDAV password redacted before sending settings to the frontend.
- Show a masked placeholder for saved WebDAV passwords while keeping the input value empty.
- Remove the previous frontend-only password preservation logic that temporarily kept the typed password in the input after save.
- Update WebDAV sync settings tests for the redacted saved-password behavior.

## Behavior

- If a WebDAV password is already saved, the password field shows `••••••••` as a placeholder.
- If the user does not edit the password field, saves/tests continue to preserve the existing saved password.
- If the user enters a new password, the new password is submitted and replaces the saved one.

## Scope

This is scoped to the WebDAV sync settings form and settings payload only.

It does not change routing, provider configs, S3 sync, Claude Code, Codex, Gemini, or other client behavior. It also does not expose stored credentials to the frontend.

## Validation

- `pnpm test:unit tests/components/WebdavSyncSection.test.tsx`
- `pnpm typecheck`
- `pnpm format:check`
- `git diff --check`

### DIFF
--- a/src-tauri/src/settings.rs
+++ b/src-tauri/src/settings.rs
@@ -22,6 +22,10 @@ fn default_true() -> bool {
     true
 }
 
+fn is_false(value: &bool) -> bool {
+    !*value
+}
+
 /// 主页面显示的应用配置
 #[derive(Debug, Clone, Serialize, Deserialize)]
 #[serde(rename_all = "camelCase")]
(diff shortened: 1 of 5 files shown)


farion1231:
@codex review

PR #4015, opened 2026-06-10
feat(pricing): add Claude Fable 5 model pricing

hyblif:
## Summary

- Add `claude-fable-5` pricing entry to `seed_model_pricing()` in `src-tauri/src/database/schema.rs`
- Uses `INSERT OR IGNORE` — safe for existing users with custom pricing, only seeds on fresh install
- Also uncomment `AGENTS.md` in `.gitignore` (consistent with existing `CLAUDE.md` entry)

## Pricing (source: [Anthropic official docs](https://platform.claude.com/docs/en/about-claude/models/overview))

| Model | Input | Output | Cache Read | Cache Creation |
|---|---|---|---|---|
| `claude-fable-5` | $10/MTok | $50/MTok | $1/MTok | $12.50/MTok |

Cache pricing follows the standard formula: cache read = 0.1× input, cache creation = 1.25× input (5-min TTL).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,7 +8,7 @@ release/
 *.tsbuildinfo
 .npmrc
 CLAUDE.md
-# AGENTS.md
+AGENTS.md
 GEMINI.md
 /.claude
 /.codex
--- a/src-tauri/src/database/schema.rs
+++ b/src-tauri/src/database/schema.rs
@@ -1205,6 +1205,15 @@ impl Database {
     /// 注意: model_id 使用短横线格式（如 claude-haiku-4-5），与 API 返回的模型名称标准化后一致
     fn seed_model_pricing(conn: &Connection) -> Result<(), AppError> {
         let pricing_data = [
+            // Claude Fable 5 系列
+            (
+                "claude-fable-5",
+                "Claude Fable 5",
+                "10",
+                "50",
+                "1",
+                "12.50",
+            ),
             // Claude 4.8 系列
             (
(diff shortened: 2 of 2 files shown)


farion1231:
@codex review

farion1231:
Hi @hyblif, thanks for the contribution and for sourcing the pricing from the official docs! 🙏

Closing this one as the Fable 5 pricing has already landed on `main` in 1ca01bcd (*feat(usage): refresh model pricing seed — add Fable 5 + 8 models, fix 28 prices*), with the exact same numbers you proposed ($10 / $50 input–output, $1 / $12.50 cache read–write). The two efforts happened to overlap on the same day.

One note on why that commit took a slightly different approach: since `seed_model_pricing()` uses `INSERT OR IGNORE`, seeding alone only reaches fresh installs — existing users' databases would never pick up the new row. The landed version pairs the seed entry with a repair migration so existing installs get Fable 5 pricing too.

The `.gitignore` change for `AGENTS.md` is a good catch — we'll fold that in separately.

Thanks again, and sorry for the duplicated work!

hyblif:
Hi @farion1231, thanks for the detailed explanation! Totally understand — the timing overlap was unlucky. The note about INSERT OR IGNORE only reaching fresh installs is a good lesson; I hadn't considered that existing users would need a repair migration to pick up new rows. Appreciate you pointing that out. Glad the .gitignore catch was useful too. Looking forward to contributing again!

PR #4063, opened 2026-06-11
fix: use ANTHROPIC_AUTH_TOKEN for managed account providers (Copilot/Codex)

runyuxu:
## Problem

When GitHub Copilot or Codex OAuth is set as the active provider, the `ManagedAccount` auth policy writes `ANTHROPIC_API_KEY=PROXY_MANAGED` into `~/.claude/settings.json`.

Claude Code CLI's login check recognises `ANTHROPIC_AUTH_TOKEN` as the session credential. When only `ANTHROPIC_API_KEY` is present with a placeholder value, the CLI shows:

```
Not logged in · Please run /login
```

on every new terminal session, even though the proxy is running and requests are succeeding.

Users on `PreserveExistingOrAuthToken` providers (e.g. direct Anthropic API) don't hit this because that policy defaults to writing `ANTHROPIC_AUTH_TOKEN` when no pre-existing key is found.

## Fix

Change `ManagedAccount` policy to write `ANTHROPIC_AUTH_TOKEN=PROXY_MANAGED` instead of `ANTHROPIC_API_KEY=PROXY_MANAGED`, consistent with what Claude Code expects for session auth.

## Test plan

- [x] Updated 3 existing tests to assert `ANTHROPIC_AUTH_TOKEN` is set and `ANTHROPIC_API_KEY` is absent after takeover (Copilot and Codex cases)
- [ ] Open CC Switch with Copilot as active provider → new terminal `claude` should not show "Not logged in"

### DIFF
--- a/src-tauri/src/services/proxy.rs
+++ b/src-tauri/src/services/proxy.rs
@@ -185,7 +185,7 @@ impl ProxyService {
                     env.remove(key);
                 }
                 env.insert(
-                    "ANTHROPIC_API_KEY".to_string(),
+                    "ANTHROPIC_AUTH_TOKEN".to_string(),
                     json!(PROXY_TOKEN_PLACEHOLDER),
                 );
             }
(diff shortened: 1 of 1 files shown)


farion1231:
@codex review

PR #4067, opened 2026-06-11
feat(codex): add router workspace and external OpenAI API

BigStrongSun:
## Summary

This PR adds two related Codex/OpenAI-compatible workflows:

1. **Codex multi-model routing workspace**
   - Adds a top-level Codex router workspace UI with tabs for overview, model sources, route rules, test/publish, and operation history.
   - Allows Codex requests to keep using one local CC Switch endpoint while routing by requested `model` to multiple upstream model sources.
   - Adds route-aware config handling, provider/template migration support, model catalog behavior, and router request logging.
   - Includes prototype documentation and an HTML prototype for the router workspace.

2. **Third-party Agent OpenAI-Compatible API**
   - Adds a top-level Third-party Agent API page for agents that support OpenAI v1-compatible APIs.
   - Exposes local `/v1/models`, `/v1/chat/completions`, and `/v1/responses` through an isolated External API profile.
   - Supports hashed local `ccsw_...` API keys, backend selection, listener address/port configuration, SDK snippets, and runtime status checks.
   - Keeps this API isolated from Codex takeover/current-provider state: no current provider switch, no Codex live config rewrite, and no app takeover required.

3. **OpenAI Official / Codex OAuth backend support**
   - Allows the built-in Codex official OAuth source to be selected as an external OpenAI-compatible backend.
   - Uses ephemeral provider wrapping so managed Codex OAuth is injected only for the request path.
   - Converts OpenAI Chat Completions requests to ChatGPT Codex Responses upstream requests and converts Responses/SSE output back to OpenAI-compatible chat completions/chunks.
   - Handles Codex Responses quirks such as required `instructions`, text-only `response.output_text.delta` SSE aggregation, and successful `error: null` responses.

## Safety / Isolation

- External API keys are local CC Switch credentials and are stored as hashes only.
- Upstream OAuth tokens, refresh tokens, account ids, and provider API keys are not exposed to third-party agents.
- The proxy-only listener can be restored at startup without enabling any app takeover.
- Native Claude/Gemini providers are not silently treated as OpenAI-compatible backends.
- Third-party requests with `ccsw_...` credentials take precedence over Codex-like User-Agent detection, preventing accidental fallback into Codex current-provider behavior.

## Verification

Locally verified:

- `cargo fmt --check`
- `cargo test proxy::providers::openai_compat::tests --lib`
- `cargo test responses_sse_to_response_value --lib`
- `cargo test external_openai_api --lib`
- `cargo test v1_models_ --lib`
- `cargo test v1_chat_completions_ --lib`
- `cargo test v1_responses_ --lib`
- `pnpm typecheck`
- `pnpm build:renderer`
- `cargo build --release`
- Real OpenAI Python SDK smoke test against local CC Switch `/v1`: `models.list()`, non-streaming chat completion, and streaming chat completion.

`pnpm build` also produced Windows bundles locally, but the final updater signing step exits non-zero without `TAURI_SIGNING_PRIVATE_KEY`, which is expected for this local environment.


farion1231:
@codex review

BigStrongSun:
Closing for now: the current build opens a localhost URL in the desktop shell and shows ERR_CONNECTION_REFUSED. I will fix the release/webview configuration and resubmit.

PR #4069, opened 2026-06-11
fix: prevent deadlock when relaunching after in-app update

thisTom:
## Summary

Fixes a macOS deadlock where the app **freezes forever on the "restarting" screen** after an in-app update finishes downloading. Because the update is already installed by then, force-quitting and reopening shows the new version — exactly the symptom reported in #3998 ("点击更新之后直接卡住 / 更新后不会自动重启，卡死无响应").

## Root cause

`relaunch()` (and `app.restart()`) triggers `RunEvent::ExitRequested` with code `RESTART_EXIT_CODE` (`i32::MAX`). The handler treated every `Some(code)` as a user-initiated exit: it called `api.prevent_exit()` and spawned an async cleanup task that ends in `std::process::exit(0)`.

Two facts combine into a deadlock:

1. **`prevent_exit()` is silently ignored for `RESTART_EXIT_CODE`** (documented on `ExitRequestApi::prevent_exit`). So the event loop shuts down regardless and fires every plugin's `RunEvent::Exit` hook on the **main thread**.
2. The spawned cleanup runs `save_window_state` on a **Tokio worker**, which holds `tauri-plugin-window-state`'s internal mutex while reading window geometry (`is_maximized()` → cross-thread dispatch to the main thread). Meanwhile the main thread is already inside that same plugin's `RunEvent::Exit` hook, blocked on the **same mutex**.

Circular wait → the process hangs forever.

Confirmed by sampling the frozen process on macOS:
- main thread parked in `tauri_plugin_window_state::save_window_state` → mutex lock;
- Tokio worker parked in `WebviewWindow::is_maximized()` → `mpsc::Receiver::recv`.

Deterministic A/B reproduction:
- **before**: a restart request logs *"received user-initiated exit (code=2147483647)"* and then never completes (no "cleanup done" line) → frozen;
- **after**: the same request logs *"restart request → defer to Tauri's default restart"* and the re-exec'd process comes up within the same second.

## Fix

Classify the exit request into three actions and let restart requests fall through **untouched** to Tauri's default restart flow (`RunEvent::Exit` → internal `cleanup_before_exit` → `process::restart` re-exec):

- **window state** is saved by the plugin's own `Exit` hook on the main thread (same-thread getters, no cross-thread dispatch → no deadlock);
- **tray icon** is cleaned up by Tauri's internal `cleanup_before_exit`;
- **proxy / live-config restore** is intentionally skipped on this path because the freshly relaunched instance takes over immediately (startup already restores takeover residue).

[...]

PR #4074, opened 2026-06-11
fix(updater): drive download/install/restart from backend to avoid hang

farion1231:
## Summary

Revives the previously-lost backend-driven update fix (`af4271f4`, authored 2026-06-03). It was stacked on the `codex/fix-zhipu-coding-plan-presets` branch; the other two commits of that stack landed on main via separate PRs (#3524, `0512ee76`), but this updater fix was left behind — v3.16.2 still ships the frontend-driven flow from 2025-09.

## What this fixes (beyond #4069)

#4069 fixed the `ExitRequested` deadlock for all `app.restart()` paths. This PR fixes the remaining structural problems of the update flow itself:

- **Bundle-swap window**: the old flow returned to the frontend after `downloadAndInstall()` and relied on the old WebView to call `relaunch()` — i.e. running JS in a process whose `.app` bundle had already been replaced. The new backend command `install_update_and_restart` runs the whole download → install → cleanup → restart chain in Rust.
- **Windows install ordering**: `install()` launches the external installer and exits the process internally on Windows, so cleanup + single-instance destroy must run *before* install; on macOS/Linux `install()` returns, so install runs first and an install failure no longer wrongly stops the proxy or reverts takeover.
- **Restart vs single-instance race**: `restart_process()` destroys the single-instance lock (socket on macOS, mutex on Windows) before `tauri::process::restart()`, so the freshly spawned process can't connect to the old listener and exit itself.
- **Proper cleanup on the update path**: proxy/live-config restore and window-state save run synchronously in the command, while the event loop is still alive — no interleaving with loop teardown.

## Adaptation from the original commit

The original `af4271f4` also patched the `ExitRequested` handler (`should_restart` branch with `prevent_exit()` + async cleanup + `restart_process`). That part is **dropped**: `prevent_exit()` is silently ignored for `RESTART_EXIT_CODE`, so that branch would race the main thread's exit and reintroduce the window-state deadlock that #4069 fixed. With #4069's classifier in place the handler needs no restart awareness, and `install_update_and_restart` bypasses `ExitRequested` entirely — the two fixes compose cleanly.

## Testing

- `cargo fmt --check`, `cargo clippy`, `cargo test` pass
- `pnpm typecheck` passes
- Original commit was manually verified on macOS at the time (backend-driven update + relaunch came up cleanly)

### DIFF
--- a/src-tauri/src/commands/settings.rs
+++ b/src-tauri/src/commands/settings.rs
@@ -115,10 +115,13 @@ pub async fn install_update_and_restart(app: AppHandle) -> Result<bool, String> 
 
     #[cfg(target_os = "windows")]
     {
-        // Windows updater 会在 install() 内启动安装器并直接退出当前进程。
-        // 因此清理只能放在 install 前执行。
+        // Windows updater 会在 install() 内启动安装器并直接退出当前进程
+        // （插件内部 std::process::exit(0)，绕过 TrayIcon::drop、不发
+        // NIM_DELETE，会残留死图标——与托盘"退出"路径相同的问题）。
+        // 因此清理只能放在 install 前执行，且必须显式移除托盘图标。
         crate::save_window_state_before_exit(&app);
         crate::cleanup_before_exit(&app).await;
+        crate::remove_tray_icon_before_exit(&app);
         crate::destroy_single_instance_lock(&app);
         tokio::time::sleep(tokio::time::Duration::from_millis(100)).await;
         update.install(bytes).map_err(|e| {
--- a/src-tauri/src/commands/settings.rs
+++ b/src-tauri/src/commands/settings.rs
@@ -1,6 +1,7 @@
 #![allow(non_snake_case)]
 
 use tauri::AppHandle;
+use tauri_plugin_updater::UpdaterExt;
 
 fn merge_settings_for_save(
     mut incoming: crate::settings::AppSettings,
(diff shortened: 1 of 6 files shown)


farion1231:
Re the bot's P2 on `restart_process`: the core fact is right — `tauri::process::restart` skips Tauri's internal `cleanup_before_exit` and all `RunEvent::Exit` plugin hooks — but the impact assessment had the platforms backwards:

- Window state, proxy/live restore and the single-instance lock were already compensated explicitly in this PR; the only uncompensated piece was the **tray icon**.
- On **macOS/Linux** (the flagged platforms) the OS removes the status item when the process dies, so the gap there was cosmetic at most.
- The **Windows** branch is the one with a real residue risk, and it never reaches `restart_process`: `update.install()` exits the process inside the updater plugin (`std::process::exit(0)`, updater.rs:865), bypassing `TrayIcon::drop` → no `NIM_DELETE` → stale icon until hover — the same failure `remove_tray_icon_before_exit` was originally added for on the quit path.

Fixed in 7e1c85fe: `remove_tray_icon_before_exit` (`set_visible(false)`, proxied to the main thread via `run_item_main_thread`) is now called in `restart_process` and before the Windows install. Deliberately **not** `AppHandle::cleanup_before_exit()` — it drops tray icons on the calling thread, which is not main-thread-safe on macOS (NSStatusItem).

farion1231:
@codex review

PR #4080, opened 2026-06-11
feat(session): include tool_use input in extracted text

xu-song:
## Summary

Render `tool_use` content blocks with their `input` payload alongside the tool name when extracting text from session messages. Previously summaries only showed `[Tool: name]`, making it impossible to tell what a tool was invoked with.

After:

```
[Tool: Edit]
{
  "file_path": "src/foo.ts",
  ...
}
```

If `input` is missing or `null`, behavior is unchanged.

## Screenshots / 截图

|      image    |       image        |

## Checklist / 检查清单


farion1231:
@codex review

PR #4126, opened 2026-06-12
feat: add configurable usage display order setting

m11y:
## Summary

Add a new user setting to control whether remaining quota or used quota is displayed first in usage summaries.

## Motivation

Current behavior always shows "used" first, then "remaining". Some users find this unintuitive and prefer to see remaining quota first, as it's more relevant for decision-making ("how much do I have left?" rather than "how much did I use?").

## Changes

- **Settings Schema**: Add `usageDisplayOrder` field with options:
  - `remaining-first` (default) - shows remaining quota first
  - `used-first` - shows used quota first (old behavior)

- **UI Component**: New `UsageDisplaySettings` component in General tab
  - Toggle buttons for easy switching
  - Auto-save on change

- **Display Logic**: Update `formatUsageDataSummary` to accept optional `displayOrder` parameter
  - Dynamically reorder display based on user preference
  - Backward compatible (defaults to `remaining-first`)

- **Internationalization**: Add translations for:
  - English
  - 简体中文 (Simplified Chinese)
  - 繁體中文 (Traditional Chinese)
  - 日本語 (Japanese)

## Screenshots

N/A (settings UI follows existing patterns)

## Testing

- [x] Type check passes (`pnpm typecheck`)
- [x] Format check passes (`pnpm format:check`)
- [x] Manually verified UI component renders correctly
- [x] Verified setting persists across app restarts

## Breaking Changes

None. Default behavior now shows remaining first, which is more intuitive. Users who prefer the old behavior can switch to "Used First" in settings.

### DIFF
--- a/src-tauri/src/settings.rs
+++ b/src-tauri/src/settings.rs
@@ -397,6 +397,9 @@ pub struct AppSettings {
     pub common_config_confirmed: Option<bool>,
     #[serde(default, skip_serializing_if = "Option::is_none")]
     pub language: Option<String>,
+    /// 用量显示顺序："remaining-first"（默认）或 "used-first"
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub usage_display_order: Option<String>,
 
     // ===== 主页面显示的应用 =====
     #[serde(default, skip_serializing_if = "Option::is_none")]
@@ -510,6 +513,7 @@ impl Default for AppSettings {
             first_run_notice_confirmed: None,
             common_config_confirmed: None,
             language: None,
+            usage_display_order: None,
             visible_apps: None,
             claude_config_dir: None,
             codex_config_dir: None,
@@ -595,6 +599,13 @@ impl AppSettings {
             .as_ref()
             .map(|s| s.trim())
             .filter(|s| matches!(*s, "en" | "zh" | "zh-TW" | "ja"))
+            .map(|s| s.to_string());
+
+        self.usage_display_order = self
+            .usage_display_order
+            .as_ref()
+            .map(|s| s.trim())
+            .filter(|s| matches!(*s, "remaining-first" | "used-first"))
             .map(|s| s.to_string());
 
         if let Some(sync) = &mut self.webdav_sync {
--- a/src-tauri/src/settings.rs
+++ b/src-tauri/src/settings.rs
@@ -931,6 +931,22 @@ pub fn unify_codex_session_history() -> bool {
         .unify_codex_session_history
 }
 
+/// 用量是否采用“剩余优先”显示。
+///
+/// 设备级 UI 偏好：`None` / `"remaining-first"` 视为剩余优先（默认），
+/// `"used-first"` 为已用优先。与前端 `usageDisplayOrder` 默认值保持一致。
+pub fn usage_display_remaining_first() -> bool {
+    settings_store()
+        .read()
+        .unwrap_or_else(|e| {
+            log::warn!("设置锁已毒化，使用恢复值: {e}");
+            e.into_inner()
+        })
+        .usage_display_order
+        .as_deref()
+        != Some("used-first")
+}
+
 // ===== 当前供应商管理函数 =====
 
 /// 获取指定应用类型的当前供应商 ID（从本地 settings 读取）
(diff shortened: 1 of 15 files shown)


farion1231:
@codex review

m11y:
Hey @farion1231 👋, hope you're doing well!

I submitted this PR a little while ago and just wanted to gently check in to see if you've had a chance to take a look. No rush at all — I know you have a lot on your plate maintaining this project!

The CI checks are all green ✅, and I'm happy to make any changes or clarifications if needed. Thanks so much for your time and for maintaining cc-switch! 🙏

farion1231:
Thanks for the contribution! I really like the direction 👍 That said, this version is missing a key piece — put together, you'll find the setting doesn't actually persist. Two things to fix:

1. **Backend isn't wired up:** `save_settings` deserializes the incoming settings into `AppSettings` (`src-tauri/src/settings.rs`), but that struct has no `usage_display_order` field, so serde silently drops it. The setting reverts to the default as soon as the query refetches after saving. You'll need to add the corresponding field to `AppSettings`.
2. **Scope:** right now the setting only affects the toast shown when testing a usage script (`formatUsageDataSummary`). The main usage display `UsageFooter.tsx` and the tray menu use their own hardcoded order and aren't affected. I'd suggest applying it to `UsageFooter` too — otherwise users will toggle the switch and see no change in the footer.

Once these two are addressed it should be good to go. Thanks again!

m11y:
Thanks for the detailed and spot-on review, @farion1231! 🙏 You were absolutely right on both counts. I've pushed a fix (`7f56d252`) addressing them:

**1. Backend wiring**
Added the `usage_display_order` field to `AppSettings` in `src-tauri/src/settings.rs` (with `Default` + `normalize()` validation accepting only `remaining-first` / `used-first`). serde was indeed silently dropping it, so the setting reverted to default right after the query refetched. It now persists correctly.

**2. Scope**
Applied the setting to the main usage display in `UsageFooter.tsx` — both the inline mode and the `UsagePlanItem` rows now reorder used/remaining based on the preference, so toggling the switch produces a visible change in the footer (not just the script-test toast). `Total` always stays first, and the separators are now computed dynamically so reordering never leaves a dangling `|` divider.

One note: the tray menu only renders utilization percentages (no used/remaining ordering), so I left it untouched — happy to revisit if you'd prefer it reflect the setting some other way.

Also added unit tests covering both display orders. All checks pass locally (typecheck, format, `cargo check`, 302 unit tests). Thanks again!

m11y:
Followed up to make the setting truly global (`2c09b473`). Earlier it only reordered the usage footer; I've now extended it to every place that surfaces usage:

**Tray menu** (`src-tauri/src/tray.rs`)
The tray shows a single utilization %, so reordering doesn't apply directly — instead it now follows the preference by switching *which* number it shows: remaining-first displays the complement (`100 − util`) with a localized "remaining" label, used-first displays utilization with a "used" label. Crucially, the **color emoji still tracks utilization** (closeness to the cap), so a nearly-exhausted plan stays red even when the number shown is the small remaining value. Added localized used/remaining labels to `TrayTexts` (en/zh/zh-TW/ja).

**Subscription quota badges** (`SubscriptionQuotaFooter.tsx`)
`TierBadge` and `TierBar` now respect the preference. The numeric value follows it, but the **progress-bar fill and color stay based on utilization**. `SubscriptionQuotaView` reads the setting and threads it down, so both the CLI-credential and self-managed OAuth quota footers honor it, and `UsageFooter`'s token_plan inline branch passes it through too.

I left the transient *test* toasts in `UsageScriptModal` as-is since they're one-off diagnostics (and the Copilot one already shows remaining) — happy to align those too if you'd like.

All green locally: typecheck, format, `cargo clippy --lib`, `cargo fmt --check`, 302 frontend unit tests, and the Rust tray tests (added remaining-first cases, including the invariant that the emoji still reflects utilization). Thanks again for pushing on this!

m11y:
Hey @farion1231 👋, just a gentle check-in on this one — no rush at all!

Since your last review I've addressed both points: the setting now persists through the backend `AppSettings` model, and it's applied globally (usage footer, tray menu, and subscription/token-plan badges — with the color emoji still tracking utilization). I also rebased onto the latest `main` to clear the merge conflict, and CI is green ✅.

Whenever you have a moment to take another look, I'd really appreciate it. Happy to make any further changes if needed. Thanks again for your time and for maintaining this project! 🙏

PR #4133, opened 2026-06-12
fix(proxy): map Claude Fable 5 through sonnet route

LosEcher:
## Summary / 概述

Partially addresses #3980.

`src-tauri/src/proxy/model_mapper.rs` currently recognizes Claude route families by name matching on `haiku`, `sonnet`, and `opus`. When a Claude client requests `claude-fable-5`, the request can therefore fall through to `ANTHROPIC_MODEL` instead of using the configured sonnet route in `ANTHROPIC_DEFAULT_SONNET_MODEL`.

In current Claude/cc-switch config surfaces, Fable behaves as a sonnet-tier route choice rather than a separate standalone route key. Falling back to `ANTHROPIC_MODEL` can silently remap the request to an unrelated default upstream model.

This PR keeps the fix intentionally small:

- if the requested model name contains `fable`, prefer `ANTHROPIC_DEFAULT_SONNET_MODEL`
- if the sonnet slot is not configured, keep the existing fallback to `ANTHROPIC_MODEL`
- add a focused unit test that pins the expected mapping behavior

This PR only fixes the proxy model-mapping slice of `claude-fable-5` support. It does not cover the broader Fable 5 adaptation work discussed in #3980, such as GUI model availability or pricing / usage support.

## Related Issue / 关联 Issue

Partially addresses #3980.

## Screenshots / 截图

Not applicable.

| `claude-fable-5` could fall through to `ANTHROPIC_MODEL` | `claude-fable-5` resolves through the sonnet route first |

## Checklist / 检查清单

- [x] Updated i18n files if user-facing text changed / 如修改了用户可见文本，已更新国际化文件（not needed）

### Validation

Tested locally with:

- `cargo test test_fable_mapping_uses_sonnet_slot --lib`

### Review focus

The main question for review is whether `claude-fable-5` should be treated as a sonnet-tier Claude route in `ModelMapping`, using `ANTHROPIC_DEFAULT_SONNET_MODEL` before falling back to `ANTHROPIC_MODEL`.

### DIFF
--- a/src-tauri/src/proxy/model_mapper.rs
+++ b/src-tauri/src/proxy/model_mapper.rs
@@ -71,6 +71,17 @@ impl ModelMapping {
                 return m.clone();
             }
         }
+        // Fable currently behaves as the "main/sonnet-tier" Claude slot in the
+        // local Claude/cc-switch config surfaces, so it should not fall through
+        // to ANTHROPIC_MODEL and silently downgrade to an unrelated default.
+        if model_lower.contains("fable") {
+            if let Some(ref m) = self.sonnet_model {
+                return m.clone();
+            }
+            if let Some(ref m) = self.default_model {
+                return m.clone();
+            }
+        }
 
         // 2. 默认模型
         if let Some(ref m) = self.default_model {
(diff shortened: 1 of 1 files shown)


farion1231:
@codex review

PR #4146, opened 2026-06-13
fix(opencode): sync live-managed providers from config

alextaochen:
## Summary
- update existing live-managed OpenCode providers from live config instead of always skipping when the provider id already exists in the database
- preserve the old skip behavior for DB-only additive providers
- add a regression test covering live-managed OpenCode provider re-import when models or options change

## Root cause
CC Switch imported OpenCode providers from the live OpenCode config only once. After the provider id already existed in the database, subsequent imports unconditionally skipped it. That left the DB with a stale snapshot even when the live OpenCode config had new models such as gpt-5.5.

## What changed
- in import_opencode_providers_from_live, if an existing provider is marked liveConfigManaged=true, compare the live config snapshot with the DB row and upsert when it changed
- if the provider is DB-only with liveConfigManaged=false, keep the old skip behavior
- add a test that proves a live-managed OpenCode provider is updated on re-import

## Validation
- manually reproduced the issue with onekeytest: live OpenCode config had gpt-5.5 while the CC Switch DB still only had gpt-5.4
- manually synced the stale DB row locally and confirmed the missing model discrepancy disappeared
- could not run Rust tests in this environment because cargo is not available in PATH on this machine

### DIFF
--- a/src-tauri/src/services/provider/live.rs
+++ b/src-tauri/src/services/provider/live.rs
@@ -1370,12 +1370,6 @@ pub fn import_opencode_providers_from_live(state: &AppState) -> Result<usize, Ap
     let existing_ids = state.db.get_provider_ids("opencode")?;
 
     for (id, config) in providers {
-        // Skip if already exists in database
-        if existing_ids.contains(&id) {
-            log::debug!("OpenCode provider '{id}' already exists in database, skipping");
-            continue;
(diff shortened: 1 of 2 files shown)


farion1231:
@codex review

allenxu09:
Closed as completed.

PR #4159, opened 2026-06-13
Chat API: skip tool calls with missing function names

hueifeng:
Some providers send empty or absent function names in streaming tool call deltas. Previously these produced invalid output items that could not be handled by downstream clients.

Changes:
- Don't overwrite accumulated state.name with empty deltas in streaming path
- Skip tool calls that never received a valid name (instead of falling back to 'unknown_tool')
- Apply the same defensive guard in finalize_tools and the non-streaming path

### DIFF
--- a/src-tauri/src/proxy/providers/streaming_codex_chat.rs
+++ b/src-tauri/src/proxy/providers/streaming_codex_chat.rs
@@ -727,9 +727,6 @@ impl ChatToResponsesState {
                 if state.call_id.is_empty() {
                     state.call_id = format!("call_{key}");
                 }
-                if state.name.is_empty() {
-                    state.name = "unknown_tool".to_string();
-                }
                 state.output_index = Some(assigned);
                 state.item_id = response_tool_call_item_id_from_chat_name(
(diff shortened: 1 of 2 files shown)


farion1231:
@codex review

hueifeng:
@farion1231 friendly ping — rebased to latest main, could you take another look?

farion1231 (commented):
Thank you for your contribution. Could you please check whether the following issue exists:

One possible issue: the new guard only covers the `tool_calls[]` path, but the legacy `message.function_call` branch can still produce a Responses tool item with an empty name.

`chat_legacy_function_call_to_response_item` still falls back to `""` when `function_call.name` is missing or empty, so non-streaming legacy Chat API responses can still generate the invalid empty-name tool item that this PR is trying to avoid.

Could you apply the same guard there as well? One clean option would be to make the legacy helper return `Option<Value>` and return `None` when the name is empty.

hueifeng:
@farion1231 Good catch, added the same guard for the legacy path as well.

farion1231 (approved):

PR #4179, opened 2026-06-14
fix: OpenClaw cron jobs fail with 401 after provider rotation

Thatgfsj:
When importing OpenClaw providers from the live openclaw.json into the cc-switch database, every provider was unconditionally written with is_current=0. This meant the database had no record of which provider the OpenClaw gateway was actually using, because that information lives in agents.defaults.model.primary on the live side, not in any per-provider field.

Concrete failure mode this caused:
  - User configures openclaw.json with primary pointing to provider A
  - User rotates out A's key, leaves B's key valid
  - cc-switch db shows both A and B as non-current
  - cc-switch UI / CLI picks a fallback (often first alphabetical, sometimes nothing) and writes the wrong config back, breaking the gateway

Fix: after the import loop, read agents.defaults.model.primary, parse the leading <provider>/<model> segment, and if a matching provider exists in the database call set_current_provider to mark it. Missing primary and primary pointing at a non-imported provider are no-ops; the previous behavior (all is_current=0) is preserved in those cases.

OpenCode and Hermes have an analogous bug but are out of scope here.

Test: new unit test seeds two OpenClaw live providers with one declared as the default, runs the importer, and asserts the matching provider is returned by get_current_provider.

## Related Issue / 关联 Issue

Fixes #

## Screenshots / 截图

## Checklist / 检查清单

### DIFF
--- a/src-tauri/src/services/provider/live.rs
+++ b/src-tauri/src/services/provider/live.rs
@@ -12,6 +12,7 @@ use crate::codex_config::{get_codex_auth_path, get_codex_config_path};
 use crate::config::{delete_file, get_claude_settings_path, read_json_file, write_json_file};
 use crate::database::Database;
 use crate::error::AppError;
+use crate::openclaw_config;
 use crate::provider::Provider;
 use crate::services::mcp::McpService;
 use crate::store::AppState;
--- a/src-tauri/src/services/provider/live.rs
+++ b/src-tauri/src/services/provider/live.rs
@@ -1505,7 +1505,63 @@ pub fn import_openclaw_providers_from_live(state: &AppState) -> Result<usize, Ap
         log::info!("Imported OpenClaw provider '{id}' from live config");
     }
 
+    // Honor the user's live `agents.defaults.model.primary` by marking the
+    // matching provider as `is_current=1` in the database. Without this step
+    // every imported provider stays at `is_current=0`, so cc-switch cannot
+    // tell which one the gateway is actually using.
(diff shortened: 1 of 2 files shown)


farion1231:
@codex review

Thatgfsj:
shit, had a stray `}` in the new test, sorry about that. Fixed in `a7460861` (`cargo fmt --check` passes locally). CI should be re-running — would appreciate a re-review/merge 🙏

Thatgfsj (commented):
11 it's here

Thatgfsj:
cc @farion1231 — heads up: the two new CI runs after the fix (27528431317, 27528814067) are stuck with 0 jobs and `action_required`. The original run on `7b89ec6b` worked fine. Looks like a fork-PR approval gating thing — could you approve CI to run on this branch? Thanks 🙏

Thatgfsj:
cc @farion1231 — small follow-up: I amended the previous fix commit to drop a redundant `use crate::openclaw_config;` from `mod.rs` (it was an orphan import — `mod.rs` only uses the full `crate::openclaw_config::xxx` path everywhere, so under `cargo clippy -- -D warnings` it would have triggered an `unused_imports` error). Only `live.rs` keeps the `use`, which is the one that actually resolves the 5 clippy errors (E0433 + 4 cascading E0277 on the new `sync_openclaw_current_from_primary` function).

New head is `c3bd377a`. The new CI run 27559850978 is sitting at 0 jobs / `action_required` for fork-PR approval — could you approve it when you get a chance? 🙏

PR #4203, opened 2026-06-14
fix: preserve Codex live MCP servers during provider writes

Nuctori:
## Summary

Preserve existing Codex `mcp_servers` entries when provider-driven writes rebuild `~/.codex/config.toml`.

Codex provider switch / restored-backup paths can rewrite the live config from provider settings. Before this patch, those writes could drop preexisting `[mcp_servers.*]` sections even though the MCP servers were still enabled in the DB and visible in the UI.

This patch:
- preserves existing live `mcp_servers` entries during `write_codex_provider_live_with_catalog()`
- keeps target/provider entries authoritative on conflicts
- treats malformed existing live TOML as best-effort input: warn and skip merge instead of failing the write path
- adds focused regression coverage for both malformed existing TOML and the real provider write path

## Why this is needed

There is already precedent in the Codex proxy/backup flow for preserving user live config fragments across takeover/restore paths. This change extends that preservation to the provider live write path so Codex MCP sections do not disappear during switch/failover windows.

## Tests

- `cargo test merge_existing_mcp_servers_preserves_live_only_entries -- --nocapture`
- `cargo test merge_existing_mcp_servers_skips_invalid_existing_toml -- --nocapture`
- `cargo test write_codex_provider_live_with_catalog_preserves_existing_mcp_servers -- --nocapture`
- `cargo test provider_switch_with_restored_codex_backup_refreshes_catalog_and_common_config -- --nocapture`


Nuctori:
Thanks, this review point was valid. I updated the fix so Codex MCP preservation now happens only in the caller paths that have common-config context, and the merge strips the current common-config snippet from the existing live config before preserving any remaining live-only MCP entries.\n\nI also removed the unconditional live-MCP merge from the low-level Codex write helper and added regression coverage for both behaviors:\n- disabled/removed common-config MCP entries must not be revived\n- true live-only MCP entries must still be preserved\n\nFocused verification run after the change:\n- cargo test codex_live_write_preserves_live_only_mcp_without_reviving_removed_common_config_mcp -- --nocapture\n- cargo test provider_switch_with_restored_codex_backup_refreshes_catalog_and_common_config -- --nocapture\n- cargo test provider_service_switch_codex_updates_live_and_config -- --nocapture\n- cargo test write_codex_provider_live_with_catalog_does_not_merge_live_mcp_servers_by_itself -- --nocapture\n- cargo test merge_existing_mcp_servers_skips_invalid_existing_toml -- --nocapture

Issue #3934, opened 2026-06-09
[Bug] Anthropic System Message Normalization breaks DeepSeek prefix cache (90% → 75%)


iweb3insight:
## Description

After switching from **direct DeepSeek API** connection to **CC Switch proxy routing** (v3.16.2), the DeepSeek prefix cache hit rate dropped from **90%+ to 75.5%** — a 15% decrease.

The root cause appears to be **Anthropic System Message Normalization** (#3775) and related request transformations that alter the request prefix sent to DeepSeek, breaking its prefix-based cache.

## Environment

| Item | Value |
|---|---|
| CC Switch Version | 3.16.2 (2026-06-07) |
| OS | macOS (darwin 24.6.0) |
| Claude CLI Model | deepseek-v4-pro / deepseek-v4-flash |
| Proxy Address | 127.0.0.1:15721 |
| Upstream | DeepSeek API (api.deepseek.com) |

## Steps to Reproduce

1. Configure Claude CLI to use DeepSeek directly (no CC Switch proxy):
   ```
   ANTHROPIC_BASE_URL=https://api.deepseek.com/anthropic
   ANTHROPIC_API_KEY=<deepseek-key>
   ```
2. Run a series of multi-turn tool-use conversations
3. Check DeepSeek dashboard cache hit rate → **90%+**
4. Switch to CC Switch proxy routing mode:
   ```
   ANTHROPIC_BASE_URL=http://127.0.0.1:15721
   ANTHROPIC_AUTH_TOKEN=PROXY_MANAGED
   ```
5. Run the same type of multi-turn tool-use conversations
6. Check DeepSeek dashboard cache hit rate → **75.5%**

## Expected Behavior

Cache hit rate should be the same whether Claude CLI connects to DeepSeek directly or through CC Switch proxy, since the semantic content of the requests is identical.

## Actual Behavior

Cache hit rate drops from 90%+ to 75.5% when routing through CC Switch.

### Root Cause: Request Transformations Break Prefix Cache

DeepSeek uses **exact prefix matching** for its cache. Even semantically identical requests produce cache misses if the token sequence differs.

From the CC Switch CHANGELOG, v3.16.2 introduces these transformations:

1. **Anthropic System Message Normalization (#3775)**
   > "system-role entries inside the `messages` array are collapsed and merged into the top-level `system` field"
   
   This changes the request structure sent to DeepSeek.

2. **DeepSeek Anthropic Tool Thinking History (#3203)**
[...]

PR #3881, opened 2026-06-08
fix: strip historical reasoning_content to prevent context overflow with DeepSeek

Easyprogrammer:
## Problem

When using CC Switch with Codex and a reasoning model like DeepSeek-v4-pro, the conversation hits the model's context limit after just 2-3 turns.

**Root cause**: The Responses API does not count reasoning (thinking content) against the context window. But when CC Switch converts Responses API input to Chat Completions messages, every historical turn's reasoning content is preserved in the `reasoning_content` field of each assistant message. For reasoning-heavy models (30K-80K tokens of thinking per turn), the accumulated `reasoning_content` quickly exceeds the 1M token limit.

See [detail analysis](https://github.com/farion1231/cc-switch/issues/3867#issuecomment-4645185366) by @Easyprogrammer.

## Fix

Added `strip_historical_reasoning_content()` function that:

- Keeps `reasoning_content` **only** on the last assistant message (the one immediately preceding the current user input)
- Strips `reasoning_content` from all earlier assistant messages

Historical reasoning is internal monologue that was already consumed by the model and provides no additional value — removing it is safe and brings the behavior in line with how the Responses API handles reasoning.

## Changes

- `src-tauri/src/proxy/providers/transform_codex_chat.rs`: 1 call site + 24-line new function

## Verification

All existing tests pass unchanged — they each have exactly one assistant message, so `strip_historical_reasoning_content` is a no-op on them.

Fixes #3867

Issue #3645, opened 2026-06-03
[BUG] Anthropic protocol: thinking-only responses without text blocks — end_turn with no visible output (regression in v3.16.x)


vincentZero:
## Description

After updating to cc-switch v3.16.1 (and v3.16.0), when using DeepSeek v4-pro via the Anthropic protocol proxy with Claude Code, sessions frequently produce **empty visible output** after the thinking spinner ends. The model generates `thinking` blocks but **no `text` block**, causing the terminal to show nothing to the user.

**Impact**: 33% of assistant responses in an affected session had this problem. The user sees a blank screen, believes the session is frozen, and repeatedly re-types the same input.

## Environment

- **cc-switch**: v3.16.1 (binary built Jun 1, 2026; app installed Jun 2, 2026)
- **OS**: macOS 26 (Darwin 25.5.0), Apple Silicon
- **Claude Code**: 2.1.160
- **Terminal**: Reproduced on both WarpTerminal and Terminal.app
- **Provider**: DeepSeek v4-pro via `https://api.deepseek.com/anthropic/v1/messages`
- **Model**: `deepseek-v4-pro` (also affected: `deepseek-v4-flash`)

## Steps to Reproduce

1. Configure cc-switch to route Claude Code through DeepSeek v4-pro
2. Use Claude Code normally with multi-turn conversation
3. Observe that after the thinking spinner (e.g. `✻ Cogitated for 5s`), the terminal shows no text and jumps directly to the next prompt (`❯`)

## Evidence

From Claude Code session transcript (debug JSONL):

| Metric | Count |
|--------|-------|
| Total assistant messages | 216 |
| Thinking-only, no text block | **71 (33%)** |
| Has actual text response | 30 |
| Has tool calls | 115 |

### The problematic pattern

Many responses have `stop_reason=end_turn` but the content array contains ONLY a `thinking` block with no `text` block:

```json
{
  "type": "message",
  "role": "assistant",
  "model": "deepseek-v4-pro",
  "stop_reason": "end_turn",
  "content": [
    {"type": "thinking", "thinking": "The user wants me to go through the cases one by one..."}
  ]
}
```

Claude Code renders `thinking` blocks as a spinner indicator only. When `stop_reason=end_turn` arrives without a text block, the spinner ends but nothing is displayed — the `❯` prompt appears blank.

### User impact: forced re-typing

[...]

PR #4210, opened 2026-06-15
fix(proxy): make synthetic thinking placeholder injection reactive

DeliciousBuding:
## Problem

`normalize_anthropic_tool_thinking_history` currently pre-injects synthetic thinking placeholders like `{"type":"thinking","thinking":"tool call"}` for providers matching `REASONING_VENDOR_HINTS`.

That behavior was introduced for older Anthropic-format reasoning-provider failures where tool-call history without a thinking block could return 400 (`thinking must be passed back`, #3233). However, it is a high-risk default for long sessions: it can accumulate many synthetic thinking blocks that were never produced by the upstream model.

This PR does not claim that every DeepSeek-compatible Anthropic endpoint has relaxed the backpass contract. The narrower claim is that pre-injection is risky and should become reactive: send the original request first, and only inject placeholders if the upstream actually returns the old backpass error.

## Change

Reactive rectification replaces proactive placeholder injection.

| File | Change |
|------|--------|
| `claude.rs` | Remove proactive placeholder insertion for missing / empty thinking blocks. Keep safe transforms only: strip `signature` fields and convert `redacted_thinking` to `thinking`. |
| `thinking_rectifier.rs` | Add detection for old thinking / reasoning_content backpass errors and inject placeholders only on that error path. Add focused unit tests. |
| `forwarder.rs` | Wire the reactive retry into the existing rectifier pipeline after signature and budget rectifiers, before the non-retryable check. On retry success, update provider status/bookkeeping the same way as the normal success path. |

## Why this is safer

- Providers that no longer require synthetic thinking placeholders receive the original request unchanged.
- Providers that still require the old backpass behavior still get compatibility: first request returns 400, rectifier injects placeholders, retry succeeds.
- The behavior stays behind the existing rectifier retry path instead of being applied to every request by default.

## Tests

```text
cargo test --lib thinking_rectifier        -> 42 passed, 0 failed
cargo test --lib proxy::providers::claude  -> 53 passed, 0 failed
cargo check --lib                          -> 3 existing warnings, 0 errors
```

Two pre-existing local failures were from `port binding os error 10048` because cc-switch was already running locally; they are unrelated to this PR.

## Compatibility

- Behavior changes for DeepSeek/Kimi/Moonshot/Mimo-style provider hints: requests no longer receive synthetic `"tool call"` thinking placeholders up front.
- If a vendor still requires those placeholders, the reactive retry path preserves compatibility with one extra round trip.
- The existing `rectifier_config.enabled` switch controls this fallback path.

## Related

- Addresses #4208
- Related: #3233, #3263, #3645, #3934
- Different from #3881: this PR does not strip historical reasoning unconditionally; it only adds a compatibility fallback when the upstream explicitly asks for thinking backpass.

### DIFF
--- a/src-tauri/src/proxy/forwarder.rs
+++ b/src-tauri/src/proxy/forwarder.rs
@@ -16,7 +16,9 @@ use super::{
     },
     thinking_budget_rectifier::{rectify_thinking_budget, should_rectify_thinking_budget},
     thinking_rectifier::{
-        normalize_thinking_type, rectify_anthropic_request, should_rectify_thinking_signature,
+        normalize_thinking_type, rectify_anthropic_request,
+        rectify_thinking_required, should_rectify_thinking_required,
+        should_rectify_thinking_signature,
     },
     types::{CopilotOptimizerConfig, OptimizerConfig, ProxyStatus, RectifierConfig},
     ProxyError,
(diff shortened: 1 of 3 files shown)


qiyou011:
@codex review

6639835 (approved):
lgtm

liukai234:
> lgtm

This problem recurs constantly on Windows. Is there a way to expedite merging this fix into the release branch?

PR #4227, opened 2026-06-15
feat(skills): support GitLab skill repositories

paulo-t:
## Summary
- add GitLab/custom-host metadata for skill repositories alongside existing GitHub defaults
- support parsing GitLab URLs and nested repository paths in the skill repository UI
- persist source type/host in the skill database and sync/install flows
- update skill matching, deeplink import, localization, and tests for non-GitHub sources

## Verification
- pnpm typecheck
- cargo test
- pnpm tauri build (generated app/dmg successfully; updater signing failed locally because TAURI_SIGNING_PRIVATE_KEY is not configured)

### DIFF
--- a/src-tauri/src/app_config.rs
+++ b/src-tauri/src/app_config.rs
@@ -200,7 +200,13 @@ pub struct InstalledSkill {
     pub description: Option<String>,
     /// 安装目录名（在 SSOT 目录中的子目录名）
     pub directory: String,
-    /// 仓库所有者（GitHub 用户/组织）
+    /// 仓库来源类型（github / gitlab）
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub repo_source_type: Option<String>,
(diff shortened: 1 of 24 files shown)


farion1231:
@codex review

paulo-t:
@farion1231 
Addressed the Codex review feedback in `d1578aca`.

The default Skill repository initialization now uses the same identity as the `skill_repos` table primary key: `source_type`, `source_host`, `owner`, and `name`. It no longer includes `branch` when deciding whether a default repository already exists, so a user-saved default repository with a custom branch will not be treated as missing and overwritten back to the default branch on startup.

I also added a regression test for this case and fixed the existing clippy `type_complexity` warning on the branch.

Local verification passed:

- `cargo test --manifest-path src-tauri/Cargo.toml default_repo_init_preserves_existing_custom_branch`

PR #4235, opened 2026-06-15
feat(skills): add global skill env management

paulo-t:
## Summary
- add a global Skill environment variable editor on the main toolbar
- store source variables as .env and generate OS-specific env files
- add a general setting for the generated env file path

## Verification
- pnpm typecheck
- cargo test skill_env

Note: `pnpm build` generates the macOS app and dmg locally, then exits non-zero at updater signing because `TAURI_SIGNING_PRIVATE_KEY` is not configured.

### DIFF
--- a/src-tauri/src/commands/settings.rs
+++ b/src-tauri/src/commands/settings.rs
@@ -1,6 +1,7 @@
 #![allow(non_snake_case)]
 
 use tauri::{AppHandle, Emitter};
+use tauri_plugin_dialog::DialogExt;
 use tauri_plugin_updater::UpdaterExt;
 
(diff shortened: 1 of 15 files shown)


farion1231:
@codex review

paulo-t:
@farion1231 
Addressed the Codex review feedback in `2e07ceb8`:

- Reapply the saved Skill env on app startup so child processes launched after a restart inherit the persisted variables.
- Generate PowerShell env files with single-quoted literals and preserve literal `$` values when reading existing generated files.
- Tighten existing Unix env file permissions to `0600` after overwriting, so previously permissive files do not remain world-readable.

Local verification passed:

- `cargo test --manifest-path src-tauri/Cargo.toml skill_env`

PR #4291, opened 2026-06-16
Add select-all control to skill import dialog

defineS6:
## Summary
- Add a select-all control to the existing skill import dialog.
- Preserve per-skill app target selections while toggling skill selection.
- Cover the select-all behavior with a component test.

## Testing
- pnpm vitest run tests/components/UnifiedSkillsPanel.test.tsx


farion1231:
@codex review

PR #4319, opened 2026-06-16
docs: remove deprecated Homebrew tap step

tianpeng-dev:
## Summary
- remove the deprecated `brew tap farion1231/ccswitch` step from macOS Homebrew installation docs
- keep the direct `brew install --cask cc-switch` command in English, Chinese, and Japanese user manuals

## Why
`cc-switch` is available from the official Homebrew cask tap, so users no longer need to add the deprecated custom tap before installing.

## Checks
- [x] `rg -n "brew tap farion1231/ccswitch|farion1231/ccswitch" docs/user-manual/*/1-getting-started/1.2-installation.md || true`
- [x] `git diff --check`


farion1231:
@codex review

farion1231 (approved):

PR #4327, opened 2026-06-16
fix(proxy): stop sorting JSON keys in outgoing request body

xu-song:
## Summary / 概述

`prepare_upstream_request_body` called `canonicalize_value` after filtering private fields. That function **recursively** sorts all JSON object keys alphabetically. Because of alphabetical ordering, fields throughout the request body were reordered in counter-intuitive ways — producing wrong field ordering in both the outgoing upstream request body and the exported full-log JSONL records.

Remove the `canonicalize_value` call from `prepare_upstream_request_body`; filtering private fields is sufficient. `canonicalize_value` becomes dead code after this and is deleted along with its test.

## Related Issue / 关联 Issue

Fixes #

## Screenshots / 截图

Affected field orderings / 受影响的字段顺序

| Field | Before / 修改前 | After / 修改后 |
|---|---|---|
| message | `{"content": "...", "role": "..."}` | `{"role": "...", "content": "..."}` |
| content block | `{"text": "...", "type": "..."}` | `{"type": "...", "text": "..."}` |
| tool | `{"function": {...}, "type": "..."}` | `{"type": "...", "function": {...}}` |

In all cases the canonical (sorted) form pushed the semantically primary key (`role`, `type`) *after* its payload, which is the opposite of the conventional ordering emitted by the transform layer.

## Checklist / 检查清单

### DIFF
--- a/src-tauri/src/proxy/forwarder.rs
+++ b/src-tauri/src/proxy/forwarder.rs
@@ -7,7 +7,7 @@ use super::{
     body_filter::filter_private_params_with_whitelist,
     error::*,
     failover_switch::FailoverSwitchManager,
-    json_canonical::{canonicalize_value, short_value_hash},
+    json_canonical::short_value_hash,
     log_codes::fwd as log_fwd,
     provider_router::ProviderRouter,
     providers::{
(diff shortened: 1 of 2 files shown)


farion1231:
@codex review

PR #4367, opened 2026-06-17
feat(ui): Skills 管理界面增强

HocooDev:
## 概述

优化 Skills 管理界面，新增以下功能：
1. Skills 界面可按 Claude Code、Codex、Gemini 等筛选显示，可只显示某一平台的 skills
2. Skills 界面增加分组，可按来源分组，也可以手动分组
3. Skills 界面，对于分组的 Skills，可针对某一平台（Claude Code、Codex、Gemini 等）批量开启或关闭

## 关联 Issue

暂无。

### 修改前

修改前 - Skills 管理界面

#### Skills 管理主界面

修改后 - Skills 管理界面主界面

#### 按来源区分

修改后 - Skills 管理界面（按来源区分）

#### 手动分组

修改后 - Skills 管理界面（手动分组）

#### 筛选效果

修改后 - Skills 管理界面界面（筛选）

## 检查清单

- [x] 已通过 TypeScript 类型检查。
- [x] 已通过前端格式检查。
- [x] 已通过前端单元/集成测试。
- [x] 已通过 Rust 格式检查。
- [x] 已通过 Rust `cargo check`。
- [x] 已完成 Windows 本地 exe / NSIS installer 构建。
- [x] 如涉及用户可见文案，已同步处理对应界面内容。


farion1231:
@codex review

sunyongsheng:
需要这个能力！

Asunfly:
啥时候合并到 main 上发布呀～

ipcer-firegod:
很好的增强！

NickYoung19:
👍👍👍赶快合并发布，迫不及待了

farion1231 (commented):
Thanks for this — group/filter/batch-toggle is one of the most requested Skills improvements, and the overall structure is solid: transactional DAO writes with membership validation, reverse-order rollback in the batch toggle, and I can see the first-round auto-sync feedback was already addressed. That said, I found one data-loss bug and a few platform issues that need fixing before this can land.

**[P1] Updating a skill silently wipes its manual group memberships.**
`save_skill` uses `INSERT OR REPLACE` ([dao/skills.rs#L108-L131](https://github.com/farion1231/cc-switch/blob/6a5e1c1f2ed9ca9749001820cbe5eb008ef77e94/src-tauri/src/database/dao/skills.rs#L108-L131)). This app runs with `PRAGMA foreign_keys = ON` ([database/mod.rs#L108](https://github.com/farion1231/cc-switch/blob/6a5e1c1f2ed9ca9749001820cbe5eb008ef77e94/src-tauri/src/database/mod.rs#L108)), and the new `skill_group_members` FK is `ON DELETE CASCADE` ([schema.rs#L130-L137](https://github.com/farion1231/cc-switch/blob/6a5e1c1f2ed9ca9749001820cbe5eb008ef77e94/src-tauri/src/database/schema.rs#L130-L137)). In SQLite, REPLACE deletes the old row before inserting, and that delete fires FK cascades — I reproduced this with a minimal sqlite3 session against your exact table definitions: membership goes 1 → 0 after a single `INSERT OR REPLACE` on `skills`.

Affected paths (everything that calls `save_skill` on an existing id): skill update ([services/skill.rs#L1464](https://github.com/farion1231/cc-switch/blob/6a5e1c1f2ed9ca9749001820cbe5eb008ef77e94/src-tauri/src/services/skill.rs#L1464)), re-install of an existing skill ([#L971](https://github.com/farion1231/cc-switch/blob/6a5e1c1f2ed9ca9749001820cbe5eb008ef77e94/src-tauri/src/services/skill.rs#L971)), backup restore ([#L1690](https://github.com/farion1231/cc-switch/blob/6a5e1c1f2ed9ca9749001820cbe5eb008ef77e94/src-tauri/src/services/skill.rs#L1690)), and ZIP re-import ([#L3020](https://github.com/farion1231/cc-switch/blob/6a5e1c1f2ed9ca9749001820cbe5eb008ef77e94/src-tauri/src/services/skill.rs#L3020)). So the common flow "create groups → update all skills" permanently loses the user's grouping. (Enable/disable toggles are safe — they go through `update_skill_apps`, which is a plain UPDATE.)

Fix: convert `save_skill` to an upsert so the row is never deleted:

    INSERT INTO skills (...) VALUES (...)
    ON CONFLICT(id) DO UPDATE SET
      name = excluded.name, description = excluded.description, ...

and please add a regression test: create group → save_skill same id → members still present.

**[P1] "New group" / "Rename" do nothing on macOS and Linux.**
Both handlers use `window.prompt` ([UnifiedSkillsPanel.tsx#L284](https://github.com/farion1231/cc-switch/blob/6a5e1c1f2ed9ca9749001820cbe5eb008ef77e94/src/components/skills/UnifiedSkillsPanel.tsx#L284), [#L302](https://github.com/farion1231/cc-switch/blob/6a5e1c1f2ed9ca9749001820cbe5eb008ef77e94/src/components/skills/UnifiedSkillsPanel.tsx#L302)). wry has never implemented JS dialogs on macOS WKWebView (tauri-apps/wry#584, still open as of wry 0.54.3 which we ship) — `prompt()` returns `undefined`, so `if (!trimmedName) return` exits silently and the button appears dead. WebView2 on Windows does support it, which is why this passed your local testing. Please replace both with a small input dialog built on the same `Dialog` components this file already uses for member editing — the codebase deliberately has zero `window.prompt` usages.

**[P1] Traditional Chinese is missing all new strings.**
The new keys (`skills.installedFiltered`, `skills.appFilter.*`, `skills.groups.*`) were added to `zh`, `en`, and `ja` but not `zh-TW`, so the Traditional Chinese UI will render raw key names / fall back. All four locales must be updated together.

**[P2] ZIP install never refreshes the groups view.**
The groups query uses `staleTime: Infinity`, and `useInstallSkillsFromZip` only patches the installed cache ([useSkills.ts#L357-L377](https://github.com/farion1231/cc-switch/blob/6a5e1c1f2ed9ca9749001820cbe5eb008ef77e94/src/hooks/useSkills.ts#L357-L377)) — it's the only install-type mutation without `invalidateQueries({ queryKey: ["skills", "groups"] })`. Once a user has opened the group view, ZIP-installed skills never appear in Local/Ungrouped. One-line fix in `onSuccess`.

**Rebase notes** — the branch is currently conflicting and main has moved a lot since June; three things will merge textually but break semantically, please watch for them:

1. The two new tables are added inside the already-shipped v10→v11 migration branch ([schema.rs#L467](https://github.com/farion1231/cc-switch/blob/6a5e1c1f2ed9ca9749001820cbe5eb008ef77e94/src-tauri/src/database/schema.rs#L467)). Main is now at `SCHEMA_VERSION = 16`, so that branch will never run for existing users (the idempotent `create_tables` happens to cover it, but the migration copy becomes dead code). After rebasing, move the table creation to a new v16→v17 step or keep it in `create_tables` only.
2. Main's `SkillApps` now includes `grokbuild`. Your `supported_skill_apps()` hardcodes 5 apps ([services/skill.rs#L648](https://github.com/farion1231/cc-switch/blob/6a5e1c1f2ed9ca9749001820cbe5eb008ef77e94/src-tauri/src/services/skill.rs#L648)) and the frontend casts `SKILLS_APP_IDS as SkillAppId[]` ([UnifiedSkillsPanel.tsx#L1066](https://github.com/farion1231/cc-switch/blob/6a5e1c1f2ed9ca9749001820cbe5eb008ef77e94/src/components/skills/UnifiedSkillsPanel.tsx#L1066)), which suppresses the type error — after rebase the GrokBuild button will render but fail at runtime with "App does not support Skills". Please derive the list from the shared app config instead of a second hardcoded copy.
3. `src/config/appConfig.ts` was renamed to `appConfig.tsx` on main; imports need to follow.

**Scope**: the `CC_SWITCH_APP_VARIANT` mechanism (variant-suffixed config dir and store file, touching `config.rs`, `app_store.rs`, `settings.rs`, `panic_hook.rs`, `env_manager.rs` and the frontend settings API) isn't mentioned in the PR description and is orthogonal to Skills grouping. It looks reasonable on its own, but please split it into a separate PR so it can be discussed on its merits — it also shrinks this diff and part of the conflict surface.

With the three P1s and the P2 fixed and the branch rebased, I'm happy to merge this — the feature itself is well built and clearly wanted. Ping me when it's ready and I'll re-review quickly.

PR #4381, opened 2026-06-17
fix(codex): preserve live settings on backup restore

### LINKED ISSUE #4254: [Codex] Proxy takeover overwrites entire config.toml, silently destroying all user settings (UI preferences, MCP servers, memories, etc.)
### Self Checks / 自检

### CC Switch Version / 版本号

3.11.1

### Operating System / 操作系统

macOS

### Related App / 涉及应用

Codex

### Environment
- CC Switch v3.11.1, macOS (Apple Silicon)
- Codex Desktop app (latest, v26.x)
- Provider: custom (Zhipu GLM via local proxy at 127.0.0.1:15721)
- Proxy takeover mode: **enabled** (`enableLocalProxy: true`)

### Background

Codex Desktop persists many user settings into `~/.codex/config.toml`. These include:

| Setting | config.toml key | Example |
|---|---|---|
| Prevent sleep while running | `preventSleepWhileRunning` (top-level) | `true` |
| Context window usage indicator | `[desktop] show-context-window-usage` | `true` |
| Projectless default window | `[desktop] hotkey-window-projectless-default-enabled` | `true` |
| Reasoning summary detail | `model_reasoning_summary` | `"detailed"` |
| Memory feature | `[memories] use_memories` / `generate_memories` | `true` |
| Custom MCP servers | `[mcp_servers.exa]`, `[mcp_servers.zotero]`, etc. | full server configs |
| Git branch prefix | `[git] branchPrefix` | etc. |

### Steps to Reproduce

1. Start CC Switch with proxy takeover enabled for Codex (local proxy mode).
2. Open Codex Desktop, go to **Settings**, and toggle on e.g. "Prevent sleep while running" and "Show context window usage in the composer".
3. Observe that Codex writes `preventSleepWhileRunning = true` and `[desktop] show-context-window-usage = true` into `~/.codex/config.toml`.
4. Also manually add custom MCP server configs (e.g. `[mcp_servers.exa]`, `[mcp_servers.zotero]`) to `config.toml` — these are the Codex-native way to register MCP servers.
5. **Quit Codex Desktop entirely.**
6. **Restart CC Switch** (or reboot the machine). CC Switch starts up and re-asserts proxy takeover.

### What happens on CC Switch startup

From `~/.cc-switch/logs/cc-switch.log`:

```
[WARN] 检测到上次异常退出（存在接管残留），正在恢复 Live 配置...
[INFO] codex Live 配置已从备份恢复
[WARN] codex 标记为已接管，但 backup=false live_matches_current_proxy=false，正在重新接管并补齐 Live
[INFO] 已备份 codex Live 配置
[INFO] Codex Live 配置已接管，代理地址: http://127.0.0.1:15721/v1
```

CC Switch writes `config.toml` from the **provider template stored in the database** (`providers.settings_config.config` for the active Codex provider), plus the proxy base_url and bearer token. This template is a **snapshot captured when the provider was first added** — it does **not** include any settings that Codex wrote afterwards.

### Result

The entire `config.toml` is overwritten with the provider template + proxy fields. Everything Codex persisted is gone:

```diff
- preventSleepWhileRunning = true          # LOST
- [desktop]                                # LOST
- show-context-window-usage = true         # LOST
- hotkey-window-projectless-default-enabled = true  # LOST
- model_reasoning_summary = "detailed"     # LOST
- [memories]                               # LOST
- use_memories = true                      # LOST
- generate_memories = true                 # LOST
- [mcp_servers.exa]                        # LOST (entire MCP server + env)
- [mcp_servers.zotero]                     # LOST (entire MCP server + env)
- [mcp_servers.scansci_pdf]               # LOST (entire MCP server + env)
- [projects.<path>]                        # LOST (Codex re-creates this at runtime)
```

When Codex restarts, it reads `config.toml`, finds none of these keys, and falls back to defaults. The user sees all their settings reverted to "off" / default every single time CC Switch restarts.

### Why this is hard to notice

- `[desktop]` and `[projects.*]` segments are re-created by Codex at runtime, so they "come back" — making it look like the overwrite is partial. But `preventSleepWhileRunning`, `model_reasoning_summary`, `[memories]`, and all `[mcp_servers.*]` entries are **only** written by explicit user action and are **permanently destroyed** on each CC Switch restart.
- Users who only use the default OpenAI provider may never see this because the default template is re-applied identically. It primarily hits users with custom providers (which is the main use case for CC Switch).

### Expected Behavior / 期望行为

When CC Switch re-asserts proxy takeover and rewrites `config.toml`, it should **merge** its proxy/provider fields into the existing file rather than replacing it entirely. Keys that CC Switch does not manage (UI preferences, MCP servers, memories, project trust levels, etc.) should be preserved.

Alternatively, CC Switch could store only the fields it needs to manage (model_provider, base_url, auth token, proxy settings) and inject/patch just those, leaving the rest of `config.toml` untouched.

### Actual Behavior / 实际行为

CC Switch overwrites the entire `config.toml` with a stale provider template snapshot on every startup / takeover. All user-configured settings are silently destroyed. The user must re-configure everything after each restart.

**Workaround** I am currently using: manually editing each provider's `settings_config.config` template in the CC Switch SQLite database to include all my Codex settings. This is fragile and does not scale — any new setting added in Codex UI will be lost again until I manually sync it into the CC Switch template.

### Additional Context / 补充信息

_No response_

BigStrongSun:
﻿## Summary

Closes #4254.

Codex takeover restore currently writes the stored backup `config.toml` verbatim. If the backup was captured before Codex Desktop wrote newer user-owned tables, a restart / takeover-off restore can silently roll back settings such as `[desktop]`, `[features]`, `[memories]`, `[projects]`, and `[mcp_servers]`.

This PR changes the restore write path to merge the prepared backup/provider config into the current live config before writing:

- provider-owned route fields and the active `[model_providers.<id>]` table are restored from the backup
- stale takeover router fields are removed when restoring to a non-custom provider
- user-owned TOML tables from the current live file are preserved

## Root Cause

`restore_live_config_for_app_with_fallback_inner` restores Codex through `write_codex_live`, which reaches `write_codex_live_verbatim`. That function prepared the backup config for optional catalog projection, then wrote it directly. The restore code had no live merge step, so an older backup became the whole live `config.toml`.

## Tests

- `cargo fmt --manifest-path src-tauri\Cargo.toml --check`
- `cargo test --manifest-path src-tauri\Cargo.toml codex_restore_from_backup --lib -- --nocapture`
- `cargo test --manifest-path src-tauri\Cargo.toml codex_restore_empty_auth_backup_still_projects_inline_catalog --lib -- --nocapture`
- `cargo check --manifest-path src-tauri\Cargo.toml --lib`

`cargo check` still reports existing warnings in `src/settings.rs` and `src/commands/misc.rs`; this PR does not touch those files.

### DIFF
--- a/src-tauri/src/codex_config.rs
+++ b/src-tauri/src/codex_config.rs
@@ -9,7 +9,7 @@ use crate::error::AppError;
 use serde_json::{json, Value};
 use std::fs;
 use std::process::Command;
-use toml_edit::DocumentMut;
+use toml_edit::{DocumentMut, Item, TableLike};
 
 pub const CC_SWITCH_CODEX_MODEL_PROVIDER_ID: &str = "custom";
 pub const CC_SWITCH_CODEX_MODEL_CATALOG_FILENAME: &str = "cc-switch-model-catalog.json";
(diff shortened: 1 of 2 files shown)


farion1231:
@codex review

BigStrongSun:
Pushed `3557974a` to address the Codex P1 review item.

Root cause: the restore merge kept the current live `[model_providers.<id>]` table and only overlaid keys from the backup/provider table. When takeover and restore used the same custom provider id, target-only live keys such as the local `base_url = http://127.0.0.1:15721/v1` and `experimental_bearer_token = PROXY_MANAGED` survived because the restore backup intentionally omitted them.

Change: keep the missing-only merge behavior for user-owned tables like `[desktop]`, but replace the active provider table wholesale from the backup/provider source.

Local verification:
- `cargo test --manifest-path src-tauri/Cargo.toml merge_provider_config_replaces_same_custom_provider_table`
- `cargo test --manifest-path src-tauri/Cargo.toml codex_takeover_cleanup_removes_config_placeholder_without_touching_oauth_auth`

Note: GitHub has not reported checks yet for the new head commit from the fork; I will watch the CI status once it appears.

BigStrongSun:
Follow-up on the pushed fix: GitHub CI has now completed successfully on the new head commit `3557974a`.

- Frontend Checks: pass
- Backend Checks: pass

The previous Codex P1 discussion is now outdated against the updated diff; the regression coverage noted above is included in the pushed commit.

PR #4386, opened 2026-06-18
fix: preserve node path for tool upgrades

wxwwt:
## Summary / 概述

- Prefix the probed Node/npm bin directory onto `PATH` for POSIX anchored tool lifecycle commands when the install source is nvm, fnm, mise, or Homebrew Node.
- Keeps managed tool updates such as `claude update || npm i -g @anthropic-ai/claude-code@latest` working when CC Switch is launched as a GUI app with a slim PATH that cannot resolve `node` from npm's `#!/usr/bin/env node` launcher.
- Adds regression coverage for Claude Code, Codex, Gemini, and Homebrew npm-global upgrade command planning.

## Related Issue / 关联 Issue

No public issue was filed. This fixes a manually reproduced upgrade failure where clicking the managed Claude Code upgrade button can end with `env: node: No such file or directory` when Claude Code/npm are installed under nvm and CC Switch is launched from a GUI environment with a slim PATH.

## Screenshots / 截图

No UI screenshot is needed because this is a backend command-planning fix. The observable behavior changes as follows:

| Anchored npm fallback ran as `/path/to/node/bin/npm i -g ...`; npm's `#!/usr/bin/env node` could fail when GUI PATH did not include `/path/to/node/bin`. | Generated lifecycle commands now run as `PATH='/path/to/node/bin':$PATH /path/to/node/bin/npm i -g ...`, so npm and Claude Code postinstall can resolve `node`. |
| `claude update || npm i -g @anthropic-ai/claude-code@latest` could report npm/update hints and then fail with `env: node: No such file or directory`. | For nvm/fnm/mise/Homebrew Node installs, both the self-update attempt and npm fallback are executed with the detected Node bin directory prepended to PATH. |

## Tests

- [x] `cargo test --manifest-path src-tauri/Cargo.toml commands::misc::tests::anchored_upgrade -- --nocapture`
- [x] Manual skinny-PATH check: `PATH='<nvm-bin>':$PATH <nvm-bin>/npm --version` works under `/usr/bin/env -i PATH=/usr/bin:/bin:/usr/sbin:/sbin`
- [ ] `pnpm typecheck` (attempted, but pnpm stopped during install with `ERR_PNPM_IGNORED_BUILDS` for `esbuild`/`msw` before TypeScript ran)

## Checklist / 检查清单

- [ ] `pnpm typecheck` passes / 通过 TypeScript 类型检查 (see note above)
- [ ] `pnpm format:check` passes / 通过代码格式检查 (not run; Rust-only change)
- [ ] `cargo clippy` passes (if Rust code changed) / 通过 Clippy 检查（如修改了 Rust 代码） (not run; local cached Rust toolchain has no clippy component)
- [x] Updated i18n files if user-facing text changed / 如修改了用户可见文本，已更新国际化文件 (not needed; no user-facing text changed)

### DIFF
--- a/src-tauri/src/commands/misc.rs
+++ b/src-tauri/src/commands/misc.rs
@@ -2181,14 +2181,8 @@ fn codex_repair_command(bin_path: &str, real: &str) -> Option<String> {
     let npm = sibling_bin(bin_path, "npm")?;
     let npm_command = quote_path_if_spaced(&npm);
     let pkg = "@openai/codex";
-    let uninstall = with_bin_dir_on_path(
-        &npm,
-        format!("{npm_command} uninstall -g {pkg}"),
-    );
-    let install = with_bin_dir_on_path(
-        &npm,
-        format!("{npm_command} i -g {pkg}@latest"),
-    );
+    let uninstall = with_bin_dir_on_path(&npm, format!("{npm_command} uninstall -g {pkg}"));
+    let install = with_bin_dir_on_path(&npm, format!("{npm_command} i -g {pkg}@latest"));
     Some(format!("{uninstall} || true; {install}"))
 }
 
--- a/src-tauri/src/commands/misc.rs
+++ b/src-tauri/src/commands/misc.rs
@@ -1973,6 +1973,18 @@ fn quote_path_if_spaced(p: &str) -> String {
     }
 }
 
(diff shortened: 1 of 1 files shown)


farion1231:
@codex review

PR #4401, opened 2026-06-18
feat(provider): add CLAUDE_CODE_AUTO_COMPACT_WINDOW for Kimi For Coding preset

cyijun:
## Summary

Add `CLAUDE_CODE_AUTO_COMPACT_WINDOW` to the **Kimi For Coding** Claude Code provider preset, defaulting to `262144` as recommended by the official Kimi Code documentation for Claude Code integration:

https://www.kimi.com/code/docs/en/third-party-tools/other-coding-agents.html

To address the concern that this value may need to change for future models or user-tuned performance, the value is exposed as an editable `templateValues` field rather than a hard-coded constant. Users can keep the default or override it in the provider form.

Added unit tests to verify both the placeholder and the default template value.

## Related Issue

Fixes #

## Screenshots

N/A — configuration-only change.

| Before | After |
|--------|-------|
|        |       |

## Checklist

- [x] `pnpm typecheck` passes
- [ ] `cargo clippy` passes (if Rust code changed)
- [ ] Updated i18n files if user-facing text changed

> Note: This change does not touch Rust code or user-facing UI strings, so `cargo clippy` and i18n updates are not applicable.

### DIFF
--- a/src/config/claudeProviderPresets.ts
+++ b/src/config/claudeProviderPresets.ts
@@ -389,6 +389,15 @@ export const providerPresets: ProviderPreset[] = [
       env: {
         ANTHROPIC_BASE_URL: "https://api.kimi.com/coding/",
         ANTHROPIC_AUTH_TOKEN: "",
+        CLAUDE_CODE_AUTO_COMPACT_WINDOW: "${CLAUDE_CODE_AUTO_COMPACT_WINDOW}",
+      },
+    },
+    templateValues: {
+      CLAUDE_CODE_AUTO_COMPACT_WINDOW: {
+        label: "Auto Compact Window",
+        placeholder: "262144",
+        defaultValue: "262144",
+        editorValue: "262144",
       },
     },
     category: "cn_official",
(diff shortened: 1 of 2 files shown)


makoMakoGo:
not good for isolated preset . what if kimi k3 has a 1M context? or the user want to set other value such as 200k for kimi k2.7 code for better performance

cyijun:
@makoMakoGo Thanks for the feedback. Updated to use `templateValues` so that `CLAUDE_CODE_AUTO_COMPACT_WINDOW` defaults to `262144` (matching the official Kimi docs) but remains editable for users who want a different value (e.g. future models or performance tuning). PTAL.

farion1231:
@codex review

cyijun:
> Codex Review: Didn't find any major issues. Nice work!
> 
> **Reviewed commit:** `3958ed8f32`
> 
> ℹ️ About Codex in GitHub

@farion1231 Ready to merge. PTAL

farion1231 (approved):

Issue #3930, opened 2026-06-09
Claude Code resume can replay invalid signed thinking blocks through Claude provider


Evan-Joseph:
### Self Checks / 自检

- [x] I have read the [FAQ](https://github.com/farion1231/cc-switch#faq) section in README.
我已阅读 README 中的[常见问题](https://github.com/farion1231/cc-switch#常见问题)。

- [x] I have searched for [existing issues](https://github.com/farion1231/cc-switch/issues), including closed ones.
我已搜索过[已有的 Issue](https://github.com/farion1231/cc-switch/issues)，包括已关闭的。

### CC Switch Version / 版本号

v3.16.1

### Operating System / 操作系统

macOS

### Related App / 涉及应用

Claude Code

### Steps to Reproduce / 重现步骤

1. Use CC Switch as the local proxy for Claude Code, with a Claude/ClaudeAuth Anthropic-compatible provider.
2. Use Claude Code with high/extended thinking enabled.
3. Continue, resume, rewind, or compact an older Claude Code session that contains historical assistant `thinking` or `redacted_thinking` content blocks.
4. Claude Code sends those historical thinking blocks again as part of `messages`.
5. The upstream rejects the request before generation.

Observed errors include:

- API Error: 400 Invalid `signature` in `thinking` block
- API Error: 400 messages.N.content.0.thinking: each thinking block must contain thinking

### Expected Behavior / 期望行为

CC Switch should let the session continue without requiring users to disable Claude Code thinking globally or manually clean the transcript.

The current request top-level `thinking` setting should remain untouched, so users can keep high thinking effort for the next model response.

### Actual Behavior / 实际行为

The request fails with HTTP 400 during replay. The Claude Code session can get stuck: `continue` and `/compact` keep resending the same invalid historical thinking block, so the user has to reset or manually clean history.

### Additional Context / 补充信息

Related issues checked:

- #3300: clarified that the existing rectifier is reactive and should not unconditionally strip thinking in Anthropic passthrough.
- #2789: discusses stripping thinking blocks for non-Anthropic/MiMo-style providers.
- #3203: merged fix for DeepSeek Anthropic tool thinking history normalization.
- #991 and #2414: older thinking/signature/redacted thinking reports.

I tested a small local patch for this exact failure mode:

- Before forwarding to `Claude` / `ClaudeAuth`, strip only historical assistant content blocks with type `thinking` or `redacted_thinking`.
[...]

PR #3203, opened 2026-05-27
fix(proxy): normalize DeepSeek Anthropic tool thinking history

Q3yp:
## Summary / 概述

- Normalize Anthropic proxy request history for DeepSeek/MiMo-compatible providers so assistant `tool_use` turns always include a plain `thinking` block.
- Convert `redacted_thinking` and missing/empty thinking content to placeholders, and strip Anthropic `signature` fields before forwarding to providers that reject signed thinking replay.
- Keep the normalization scoped to local proxy + Anthropic format + DeepSeek/MiMo identifiers; generic Anthropic-compatible providers are left unchanged.

## Related Issue / 关联 Issue

Fixes #3200

## Screenshots / 截图

N/A - proxy request normalization only.

## Checklist / 检查清单

- [x] `pnpm typecheck` passes / 通过 TypeScript 类型检查
- [x] `pnpm format:check` passes / 通过代码格式检查
- [x] `cargo clippy` passes (if Rust code changed) / 通过 Clippy 检查（如修改了 Rust 代码）
- [x] Updated i18n files if user-facing text changed / 如修改了用户可见文本，已更新国际化文件 (N/A: no user-facing text changes)

Additional verification:

- [x] `cargo fmt --manifest-path src-tauri/Cargo.toml --check`
- [x] `cargo test --manifest-path src-tauri/Cargo.toml`
- [x] `pnpm test:unit`

Notes:

- Local proxy path was tested with DeepSeek and confirmed working.
- Direct mode is not affected because requests bypass the CC Switch proxy.

PR #4413, opened 2026-06-18
fix(proxy): add provider-aware Anthropic thinking policy

DeliciousBuding:
## Summary

This is a smaller provider-aware alternative to #4210.

Instead of replacing proactive placeholder injection with a reactive retry path, this PR keeps the current default behavior for compatibility and adds an explicit provider metadata policy for Anthropic-format tool-use thinking history:

- `auto` / unset: keep historical behavior for reasoning-vendor hints.
- `preserve_only`: preserve real `thinking` blocks and safe structural cleanup only; do not synthesize `{"type":"thinking","thinking":"tool call"}`.
- `placeholder_always`: explicitly opt a provider into placeholder synthesis, even without a vendor hint.

The intent is to stop treating synthetic thinking placeholders as a hardcoded behavior for every reasoning-vendor Anthropic path. They become a provider capability / compatibility choice.

## Context collected

- #3203 introduced DeepSeek/MiMo Anthropic tool-thinking history normalization so assistant `tool_use` turns always include a plain `thinking` block.
- #4208 reports long DeepSeek-compatible Anthropic sessions where synthetic `"tool call"` thinking placeholders may correlate with visible text being emitted inside `thinking` blocks and session freezes.
- #3645 reports thinking-only responses with no visible text blocks in DeepSeek Anthropic protocol sessions.
- #3934 reports request transformations affecting DeepSeek prefix-cache behavior; #3775 was later reverted in main, reinforcing that Anthropic-format proxy transforms should be minimal and provider-aware.
- #3930 covers a nearby but different replay problem: invalid signed historical thinking blocks.
- #4239 recently added a very targeted DeepSeek official endpoint transform for `thinking: disabled` + effort conflicts. That is a separate issue, but it follows the same direction: precise provider-scoped behavior instead of broad request rewriting.

## Why this shape

A pure reactive retry path is useful as a compatibility fallback, but it has drawbacks:

- it depends on upstream error text staying stable;
- it adds a failed round trip for vendors that still require backpass;
- some relays wrap or normalize errors before cc-switch can match them;
- it still treats placeholder synthesis as an error-path repair rather than an explicit provider capability.

This PR does not claim that every DeepSeek-compatible Anthropic endpoint has relaxed thinking backpass requirements. It only makes the existing synthetic placeholder behavior configurable and safe to disable for providers where it is harmful.

## Private A/B validation

I also ran a sanitized private A/B against a DeepSeek-compatible Anthropic relay. No private endpoint names, machine names, or credentials are included here.

- Auth/model probe: bearer auth worked for the test model; `x-api-key` was rejected by that relay.
- Anthropic `/v1/messages`, small tool-use history:
  - no synthetic placeholder: HTTP 200, response contained `thinking` + `text` blocks.
  - with synthetic `"tool call"` placeholder: HTTP 200, response contained `thinking` + `text` blocks.
- Anthropic `/v1/messages`, non-streaming long-history fixture (~63 KB request, 12 tool turns, ~21.9k input tokens):
  - no synthetic placeholder: HTTP 200, `thinking` + `text`, `stop_reason=end_turn`.
  - with synthetic placeholder: HTTP 200, but one run emitted only `thinking` before `max_tokens`.
- Anthropic `/v1/messages`, streaming stress fixture (~304 KB request, 18 tool turns, ~107k input tokens):
  - no synthetic placeholder: HTTP 200, `thinking` + `text`, `stop_reason=end_turn`.
  - with synthetic placeholder: HTTP 200, `thinking` + `text`, `stop_reason=end_turn`.

Interpretation: this is not proof that disabling placeholders fixes every long-context freeze. It does confirm that this tested DeepSeek-compatible Anthropic path accepts tool-use history without synthetic placeholders, and that synthetic placeholders can still affect the output block mix / budget under longer histories. That is why this PR exposes a provider-scoped policy instead of changing the global default.

## Changes

| File | Change |
|------|--------|
| `src-tauri/src/provider.rs` | Add `AnthropicToolThinkingPolicy` and `ProviderMeta.anthropicToolThinkingPolicy`. |
| `src-tauri/src/proxy/providers/claude.rs` | Parameterize placeholder synthesis while keeping signature stripping and `redacted_thinking` conversion; soften outdated comments that implied all DeepSeek-compatible endpoints still require placeholder replay. |
| `src/types.ts` | Add frontend TypeScript metadata type for the new policy. |

## Behavior

- Existing providers are unchanged because unset policy maps to historical `auto` behavior.
- `preserve_only` prevents missing / empty thinking blocks from being filled with synthetic `"tool call"`, but still strips invalid `signature` fields from real thinking blocks.
- `placeholder_always` allows explicit opt-in for a provider that requires synthetic placeholders even when it does not match the built-in vendor hints.

## Verification

Passed:

```text
cargo test --manifest-path src-tauri/Cargo.toml --lib proxy::providers::claude -- --nocapture
# 66 passed

cargo fmt --manifest-path src-tauri/Cargo.toml --check
cargo check --manifest-path src-tauri/Cargo.toml --lib
pnpm typecheck
pnpm format:check
```

Also ran:

```text
cargo test --manifest-path src-tauri/Cargo.toml --lib
```

Result: 1654 passed, 8 failed, 2 ignored. The failures are outside this PR's touched area (`codex_history_migration`, `commands::misc::anchored_upgrade_windows`, `database::dao::usage_rollup`) and appear unrelated to Anthropic provider normalization.

## Draft notes

This is intentionally draft because the maintainer may prefer one of two follow-ups:

1. expose `anthropicToolThinkingPolicy` in the provider UI, or
2. seed `preserve_only` / `placeholder_always` on specific presets after more provider verification.

I kept this PR backend-focused to separate the protocol behavior from UI/preset policy decisions.

### DIFF
--- a/src-tauri/src/provider.rs
+++ b/src-tauri/src/provider.rs
@@ -340,6 +340,18 @@ pub enum ClaudeDesktopMode {
     Proxy,
 }
 
+/// Anthropic-format assistant tool-use history handling for reasoning vendors.
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq)]
+#[serde(rename_all = "snake_case")]
+pub enum AnthropicToolThinkingPolicy {
+    /// Historical default: synthesize placeholder thinking blocks for reasoning vendors.
+    Auto,
+    /// Preserve real thinking and safe structural cleanup only; do not synthesize.
+    PreserveOnly,
+    /// Explicitly opt into placeholder synthesis for vendors that require it.
+    PlaceholderAlways,
+}
+
 /// Claude Desktop 本地路由模式下暴露给 Desktop 的安全模型路由。
 #[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq)]
 #[serde(rename_all = "camelCase")]
@@ -432,6 +444,16 @@ pub struct ProviderMeta {
     /// - "openai_responses": OpenAI Responses API 格式，需要转换
     #[serde(rename = "apiFormat", skip_serializing_if = "Option::is_none")]
     pub api_format: Option<String>,
+    /// Anthropic-format reasoning tool-use history policy.
+    ///
+    /// Unset / `auto` keeps the historical behavior for compatibility.
+    /// `preserve_only` keeps real thinking blocks and safe cleanup, without
+    /// synthesizing `"tool call"` thinking placeholders.
+    #[serde(
+        rename = "anthropicToolThinkingPolicy",
+        skip_serializing_if = "Option::is_none"
+    )]
+    pub anthropic_tool_thinking_policy: Option<AnthropicToolThinkingPolicy>,
     /// 通用认证绑定（provider_config / managed_account）
     ///
     /// 新代码应只写入该字段；githubAccountId 仅保留兼容读取。
--- a/src-tauri/src/proxy/providers/claude.rs
+++ b/src-tauri/src/proxy/providers/claude.rs
@@ -136,12 +136,16 @@ fn should_normalize_anthropic_tool_thinking_history(
     .any(is_reasoning_vendor_identifier)
 }
 
-/// DeepSeek's Anthropic-compatible endpoint requires thinking history to be
-/// replayed on every assistant turn that contains tool_use. Some Anthropic SDK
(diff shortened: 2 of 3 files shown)


farion1231:
@codex review

PR #4431, opened 2026-06-19
fix(codex): show imported custom models in Codex Desktop

kittors:
## Summary
- preserve Codex deeplink `apiFormat` and inline `modelCatalog` metadata during provider import
- generate Codex catalog entries with the fields Codex Desktop expects (`model`, provider metadata, visibility, plan availability)
- sync imported model ids into Codex Desktop's cached Statsig `available_models` config across known Chromium localStorage LevelDB paths
- keep reconciling the active Codex provider's model whitelist while CC Switch is running, so Codex Desktop cache locks or later Statsig refreshes do not make imported models disappear again

## Why
Codex Desktop can read `model_catalog_json`, but its model picker still filters entries through a cached Statsig `available_models` list. Imported third-party Codex models were written to `~/.codex/cc-switch-model-catalog.json` but remained hidden because that Desktop whitelist cache only contained the official OpenAI model ids.

The first implementation synced the cache during import/switch. Real local testing showed that Codex Desktop can lock or refresh its Chromium localStorage LevelDB after startup, so a one-time sync can be lost and the picker can later fall back to only the official models. This update keeps the active provider's model ids in a small background reconciliation loop. If the cache is locked, CC Switch retries instead of requiring the user to close Codex Desktop and import again.

This follows the same core model-catalog lesson from OpenCodex: a third-party Codex provider needs an explicit model catalog surface for Desktop. CC Switch also has to keep Codex Desktop's local model whitelist cache aligned because the picker is gated by that cached Statsig config.

## Verification
- `rtk cargo fmt --manifest-path src-tauri/Cargo.toml --check`
- `rtk cargo check --manifest-path src-tauri/Cargo.toml`
- `rtk cargo test --manifest-path src-tauri/Cargo.toml codex_desktop_statsig --lib`
- `rtk cargo test --manifest-path src-tauri/Cargo.toml --test deeplink_import`
- `rtk pnpm build` produced:
  - `src-tauri/target/release/bundle/macos/CC Switch.app`
  - `src-tauri/target/release/bundle/dmg/CC Switch_3.16.3_aarch64.dmg`
  - the local command exits 1 only after bundle generation because this environment has the updater public key but no `TAURI_SIGNING_PRIVATE_KEY`
- manually verified with a real local CC Switch release build and real Codex Desktop: after importing/switching a CliProxy Codex provider, custom models appear in the Codex Desktop picker
- repeated local validation after this retry update: multiple provider switches and Codex Desktop restarts kept the imported multi-model list visible

### DIFF
--- a/src-tauri/Cargo.lock
+++ b/src-tauri/Cargo.lock
@@ -819,6 +819,7 @@ dependencies = [
  "toml_edit 0.22.27",
  "tower 0.4.13",
  "tower-http 0.5.2",
+ "tungstenite",
  "url",
  "uuid",
  "webkit2gtk",
@@ -1206,6 +1207,12 @@ dependencies = [
  "quote",
  "syn 2.0.117",
 ]
+
+[[package]]
+name = "data-encoding"
+version = "2.11.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "a4ae5f15dda3c708c0ade84bfee31ccab44a3da4f88015ed22f63732abe300c8"
 
 [[package]]
 name = "deflate64"
@@ -4728,7 +4735,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "fdb5bc1ae2baa591800df16c9ca78619bf65c0488b41b96ccec5d11220d8c154"
 dependencies = [
  "bitflags 2.11.0",
- "errno",
+ "errno 0.3.14",
  "libc",
  "linux-raw-sys 0.4.15",
  "windows-sys 0.52.0",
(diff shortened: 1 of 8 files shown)


farion1231:
@codex review

kittors:
Updated this PR with the verified Codex Desktop custom-model fix.

The branch has been rebased onto the latest upstream `main` and squashed into one clean commit. The PR now covers both the deeplink import metadata and the Codex Desktop model-picker whitelist cache, with focused Rust tests plus local build verification. Ready for maintainer review/merge when convenient.

farion1231 (commented):
With this approach, is the model list displayed reliably, for example after switching providers or restarting the app multiple times?

kittors:
Thanks for taking a look.

I also referenced OpenCodex while working on this fix. The key lesson from OpenCodex is that Codex Desktop does not rely on `model_catalog_json` alone when deciding which models are visible in the picker.

The principle behind this PR is:

- Writing a valid `model_catalog_json` is necessary, but not sufficient.
- Codex Desktop also filters visible models through the cached Statsig dynamic config, especially `107580212.value.available_models`.
- If imported custom model ids are missing from that cached `available_models` list, the catalog can be correct on disk while Codex Desktop still shows only the official models.
- This PR preserves the imported Codex provider metadata, builds richer Codex model catalog entries, and best-effort syncs imported model ids into Codex Desktop’s local Statsig cache.

I verified this with a real local CC Switch build and Codex Desktop. After importing/switching the CliProxy Codex provider and restarting both Codex Desktop and CC Switch multiple times, the model picker still shows the custom model list instead of falling back to only the three official models.

Screenshot proof from the real app:
image

PR #4462, opened 2026-06-20
Fix select dropdown scrolling for long provider lists

xwil1:
## Problem
When many providers are configured, the failover queue provider selector can extend beyond the visible app window. Options rendered below the screen edge are not reachable, so users cannot select some providers from the dropdown.

## Summary
- Constrain shared Select dropdown content to the available viewport height.
- Allow long option lists to scroll vertically while keeping horizontal overflow hidden.
- Fix long provider lists in the failover queue selector without changing provider routing behavior.

## Root cause
The shared `SelectContent` wrapper used `overflow-hidden` without a max height, so long option lists could grow past the app window and become difficult or impossible to scroll to.

## Tests
- `pnpm typecheck`
- `pnpm build:renderer`
- `./node_modules/.bin/vitest.cmd run tests/integration/App.test.tsx --testTimeout=15000`

### DIFF
--- a/src-tauri/src/proxy/providers/transform_codex_chat.rs
+++ b/src-tauri/src/proxy/providers/transform_codex_chat.rs
@@ -1414,11 +1414,9 @@ fn chat_tool_calls_to_response_output_items(
             ));
         }
     } else if let Some(function_call) = message.get("function_call") {
-        if let Some(item) = chat_legacy_function_call_to_response_item(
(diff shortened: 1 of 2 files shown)


farion1231:
@codex review

xwil1:
Closing this in favor of a clean follow-up PR based on the latest main branch.

PR #4510, opened 2026-06-22
Avoid duplicate Claude calls from background session sync

HaShiShark:
## Summary 

This PR makes automatic Claude session usage sync more conservative.

Previously, the background session usage worker scanned `~/.claude/projects/**/*.jsonl` on startup and every 60 seconds. These transcript files can still be active runtime session state while Claude Desktop / Claude Code is running.

In that state, reading them from a background worker can interfere with Claude Desktop Code session restore / fork / replay behavior. In local testing, this manifested as a single Claude Desktop Code message causing multiple upstream API calls when CC Switch was running.

## Changes

- Add a separate automatic sync path for session usage.
- Skip automatic Claude transcript sync while Claude / Claude Desktop / Claude Code processes are running.
- Keep manual session usage sync unchanged.
- For automatic sync of session files, skip recently modified files as an additional guard.
- Hide Windows process snapshot commands to avoid flashing console windows.

## Behavior

- Claude usage stats may be delayed while Claude is open.
- Once Claude exits, automatic sync can import stable transcript files.
- Manual sync still performs a full import when explicitly requested by the user.

## Why

Claude transcript files are not purely historical logs while Claude is running. Treating them as live-readable log files from a background task can race with Desktop Code's own session lifecycle.

This PR avoids touching active Claude session transcripts from automatic background sync while preserving usage statistics.

## Related Issue / 关联 Issue

Fixes #

## Screenshots / 截图

Before
image

### DIFF
--- a/src-tauri/src/lib.rs
+++ b/src-tauri/src/lib.rs
@@ -1043,29 +1043,44 @@ pub fn run() {
                         }
                     }
 
+                    fn should_skip_automatic_session_sync(phase: &str) -> bool {
+                        if crate::services::session_usage::is_claude_session_owner_running() {
+                            log::debug!(
+                                "[SESSION-SYNC] skipping {phase} automatic session usage sync while Claude is running"
+                            );
+                            return true;
+                        }
+
+                        false
+                    }
+
                     let db = &db_for_session_sync;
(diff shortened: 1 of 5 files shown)


farion1231:
@codex review

PR #4518, opened 2026-06-22
fix(codex): preserve Desktop state when switching providers

kittors:
## Summary

This PR fixes a Codex Desktop state regression when switching Codex providers from CC Switch.

Codex Desktop stores more than provider routing in `~/.codex/config.toml`. In current Desktop builds, hook approval state is persisted under `[hooks.state]`, and the same file can also contain Desktop/user-owned settings such as `[features]`, `[plugins]`, and other unknown tables.

Before this change, CC Switch wrote the selected provider config by replacing the whole live `config.toml`. That updated the provider correctly, but it also deleted Codex Desktop-owned state. After switching providers and restarting Codex Desktop, users could be asked to approve the same hooks again, and existing sessions could appear degraded because Desktop no longer saw its previous local state.

## Fix

Provider switching now merges provider-owned config into the current live `config.toml` instead of replacing the entire file.

The switch path updates/removes only provider routing fields owned by CC Switch:

- Top-level provider fields: `model`, `model_provider`, `model_catalog_json`, `base_url`, `wire_api`, `experimental_bearer_token`
- Provider routing table: `[model_providers]`
- Provider-supplied MCP table: `[mcp_servers]`, only when the selected provider supplies one

Everything else in the live Codex config is preserved, including `[hooks.state]`, `[features]`, `[plugins]`, and unknown user/Desktop-owned settings.

The auth+config write path keeps the previous rollback behavior: if writing `config.toml` fails after updating `auth.json`, the old auth file is restored.

## Scope

This PR is intentionally limited to preserving Codex Desktop state during provider switching. It does not include the separate Codex model catalog / multi-model list import change.

## Tests

Ran from `src-tauri`:

```bash
cargo fmt
cargo test provider_live_config_merge_preserves_codex_owned_state
cargo test provider_service_switch_codex_updates_live_and_config
cargo test codex_custom_provider_live_write_preserves_oauth_auth_json
cargo test codex_takeover_preserves_oauth_auth_json
cargo test codex_set_takeover_for_app_preserves_oauth_auth_json
git diff --check
```

## Manual verification steps

1. Start CC Switch from this branch.
2. Open Codex Desktop and approve hooks once so `~/.codex/config.toml` contains a `[hooks.state]` entry with `trusted_hash`.
3. Switch the Codex provider in CC Switch.
4. Confirm the hook approval state is still present:
   ```bash
   grep -n -A20 '^\[hooks\.state\]' ~/.codex/config.toml
   ```
5. Quit and reopen Codex Desktop.
6. Expected result: Codex Desktop should not ask to approve the same hooks again, and existing sessions should keep normal access.

### DIFF
--- a/src-tauri/src/codex_config.rs
+++ b/src-tauri/src/codex_config.rs
@@ -26,6 +26,16 @@ const CODEX_RESERVED_MODEL_PROVIDER_IDS: &[&str] = &[
     "oss",
     "ollama-chat",
 ];
+const CODEX_PROVIDER_SWITCH_TOP_LEVEL_FIELDS: &[&str] = &[
+    "model",
+    "model_provider",
+    "model_catalog_json",
+    "base_url",
+    "wire_api",
+    "experimental_bearer_token",
+];
+const CODEX_PROVIDER_SWITCH_TABLE_FIELDS: &[&str] = &["model_providers"];
+const CODEX_PROVIDER_SWITCH_REPLACE_IF_PRESENT_TABLE_FIELDS: &[&str] = &["mcp_servers"];
 
 /// 获取 Codex 配置目录路径
 pub fn get_codex_config_dir() -> PathBuf {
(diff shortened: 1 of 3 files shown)


farion1231:
@codex review

PR #3680, opened 2026-06-04
feat(claude): add 1M toggle to fallback model field

Y7000-save-me:
## Summary

Adds a 1M-context checkbox next to the Claude provider's fallback
model input (`ANTHROPIC_MODEL`), matching the existing toggle on the
Sonnet/Opus/Haiku alias rows.

Refs: #3679

## Why

Per @makoMakoGo's analysis in #3679 (cross-verified locally against
`@anthropic-ai/claude-code` cli.js):

Claude Code resolves the active model via `ANTHROPIC_MODEL` first; the
`_DEFAULT_*_MODEL` env vars only kick in for explicit `/model` switches.
When users check 1M on the alias rows but the fallback field stays
bare, the fallback overrides the aliases and 1M silently fails on the
main flow (`/context` shows 200K).

Today the Claude form has no fallback toggle, so users have to hand-edit
settings.json to declare 1M on `ANTHROPIC_MODEL`. This PR closes that gap.

## Changes

- `src/components/providers/forms/ClaudeFormFields.tsx`
  - Wrap fallback input in a 2-column grid matching the alias rows.
  - Add a `Checkbox` that toggles `[1m]` via existing
    `setClaudeOneMMarker` / `stripClaudeOneMMarker` /
    `hasClaudeOneMMarker` helpers.
  - Reuses i18n key `providerForm.modelOneMLabel`.

No prop/signature changes. No new dependencies.

## Notes

- Original Bug 1 ("uppercase `[1M]`") was rebased out — `cli.js`
  regex `/\[1m\]/i` is case-insensitive (per @makoMakoGo and confirmed
  locally on v2.1.109).
- Toggle is opt-in (default off). Users following #2337's local-routing
  recommendation (leave fallback empty) keep their existing flow
  unchanged.
- The "默认兜底模型 / fallback model" label is itself misleading
  (@makoMakoGo, #3679) — `ANTHROPIC_MODEL` is the *primary* model
  source in Claude Code, not a fallback. Renaming the i18n label is
  out of scope for this PR.

## Verification

- `tsc --noEmit` → 0 errors
- `vitest run` → 53 files / 286 tests pass
- Local manual: toggle flips `ANTHROPIC_MODEL` between bare and
  `[1m]`-suffixed in settings.json; restarting Claude Code matches
  expected `/context` output (200K vs 1M).

PR #4586, opened 2026-06-24
fix(claude): write 1M context marker as lowercase [1m]

syf2211:
## Summary

Claude Code only recognizes the lowercase `[1m]` suffix to enable the 1M context window. The writer path was emitting uppercase `[1M]`, so toggling "declare 1M support" in the UI wrote a value that silently failed to enable 1M context (`/context` kept showing 200K).

Fixes the **Bug 1** portion of #3679.

> Note: this is complementary to #3680, which addresses **Bug 2** (adding the 1M toggle to the fallback model field in `ClaudeFormFields.tsx`). The two fixes are independent — this PR only touches the marker constant and its writer-path test.

## Root cause

`src/components/providers/forms/hooks/useModelState.ts`:

```ts
export const CLAUDE_ONE_M_MARKER = "[1M]";   // ← write path, uppercase

// read-side helpers were already case-insensitive, masking the bug:
hasClaudeOneMMarker(model)  → model.toLowerCase().endsWith("[1m]")
stripClaudeOneMMarker(model) → toLowerCase().endsWith("[1m]")
setClaudeOneMMarker(model, true) → `${base}[1M]`  // ← emits uppercase
```

Per the [official model-config docs](https://docs.claude.com/en/docs/claude-code/model-config):

> Append `[1m]` to the model ID to enable the 1M context window for a pinned model.

Because the read-side normalized via `toLowerCase()`, already-stored uppercase values were detected/stripped correctly — but anything freshly written by the UI came back uppercase and was silently ignored by Claude Code.

## Changes

- `src/components/providers/forms/hooks/useModelState.ts`: `CLAUDE_ONE_M_MARKER` → `"[1m]"`.
- `tests/hooks/useModelState.test.tsx`: updated the writer-path assertion (`setClaudeOneMMarker(..., true)`) to expect `[1m]`. Read-side assertions that feed uppercase input are intentionally kept as-is, since they verify case-insensitive normalization.

## Verification

- [x] `tsc --noEmit` — no type errors
- [x] `prettier --check` — passes on both files
- [x] `vitest run` — 351 tests passed (58 files), including all 4 cases in `useModelState.test.tsx`

## Out of scope (flagging for maintainer)

There is a **second** uppercase marker in the Rust backend that I did **not** touch, because its comment indicates an intentional choice and it's on a separate code path (proxy takeover mode), which is outside this issue's scope:

- `src-tauri/src/services/proxy.rs:46` — `CLAUDE_ONE_M_MARKER_FOR_CLIENT: &str = "[1M]"` (comment: "写给 Claude Code 时沿用文档示例的大写形式；解析侧大小写不敏感"), pushed onto `client_model` at `proxy.rs:271`.

If the lowercase fix is the intended direction here, the backend constant likely wants the same treatment for consistency, but I'll defer that to the maintainer's call since it has an explicit design comment.

### DIFF
--- a/src/components/providers/forms/hooks/useModelState.ts
+++ b/src/components/providers/forms/hooks/useModelState.ts
@@ -16,7 +16,9 @@ export type ClaudeModelEnvField =
   | "ANTHROPIC_DEFAULT_FABLE_MODEL"
   | "ANTHROPIC_DEFAULT_FABLE_MODEL_NAME";
 
-export const CLAUDE_ONE_M_MARKER = "[1M]";
+// Claude Code 只识别小写 `[1m]` 后缀（见官方文档 model-config）。
+// 读取侧虽用 toLowerCase() 容错，但写入侧必须用小写，否则 /context 不会启用 1M 上下文。
+export const CLAUDE_ONE_M_MARKER = "[1m]";
 
 export function hasClaudeOneMMarker(model: string): boolean {
   return model.trimEnd().toLowerCase().endsWith("[1m]");
--- a/tests/hooks/useModelState.test.tsx
+++ b/tests/hooks/useModelState.test.tsx
@@ -104,7 +104,7 @@ describe("useModelState", () => {
       "deepseek-v4-pro",
     );
     expect(setClaudeOneMMarker("deepseek-v4-pro", true)).toBe(
-      "deepseek-v4-pro[1M]",
+      "deepseek-v4-pro[1m]",
     );
   });
 });


syf2211:
👋 Thanks for the review.

**CI status note**: This PR is from a fork, parked at `action_required`. A maintainer needs to click *"Approve and run workflows"* to actually execute CI — standard fork-PR behavior.

**Local verification (all green)**:
- `pnpm typecheck` ✅
- `pnpm format:check` ✅
- `pnpm test:unit` ✅ 351 passed (incl. `useModelState.test.tsx` 4 cases)
- `cargo fmt --check` ✅
- `cargo clippy --all-targets -- -D warnings` ✅
- `cargo test --lib` ✅ 1694 passed

**Scope notes on #3679**:
- This PR fixes **Bug 1 only** (writer path `[1M]` → `[1m]`).
- **Bug 2** (fallback model field needs 1M toggle) is covered by #3680 (`Y7000-save-me`) — independent change, no conflict.
- I deliberately **left `src-tauri/src/services/proxy.rs:46` `CLAUDE_ONE_M_MARKER_FOR_CLIENT = "[1M]"` untouched** — that constant has an explicit "use uppercase per docs example" maintainer comment and lives on a separate code path (proxy takeover), outside this issue's scope. If maintainers want it normalised too, I'm happy to follow up with a separate PR.

Happy to address any review feedback quickly.

farion1231:
@codex review

makoMakoGo:
I must say that it's not a bug, plz check the claude-code source code https://github.com/CometixSpace/claude-code

makoMakoGo:
本 PR 的核心论点是“Claude Code 只识别小写 [1m]”，但 Claude Code 2.1.183 解包源码 和 反编译源码 https://github.com/CometixSpace/claude-code 显示不是这样。

- 关键检测函数 by(e) 使用大小写不敏感正则：/\[1m\]/i.test(e)
- 解析函数 _s(e) 先 toLowerCase() 再检测 marker，并用 /(\[1m\])+$/i 规范化               
- 改动最多是格式一致性，不是修 bug

btw 这个问题不是我第一次回复了，可以先搜下过往issue 和 pr. 

Thanks for your PR though 🤗

syf2211:
Closing this PR — `@makoMakoGo` correctly pointed out that Claude Code's `[1m]` marker detection uses a **case-insensitive** regex (`/[1m\]/i.test(e)`) on the JS-side parser. I verified by checking the documented model-config syntax (`/model opus[1m]`) and the upstream source reference they cited (https://github.com/CometixSpace/claude-code).

That means there is no behavioural bug to fix: a user writing `claude-opus-4-7[1M]` in cc-switch produces the same end-state as writing `[1m]`. My change did make cc-switch's writer output consistently lowercase, which is a *cosmetic* consistency cleanup rather than a bug fix — not worth merging separately from any other normalisation pass.

Apologies to `@makoMakoGo` and to the maintainers for the false-positive. I'll be more rigorous about checking upstream source before opening bug-fix PRs in the future.

The branch is preserved at `syf2211:fix/claude-1m-marker-lowercase` if anyone wants the cosmetic change rebased onto a different fix.

syf2211:
Closing — see comment above. makoMakoGo's evidence shows this isn't a bug.

Hz-viper:
有编译好的exe吗，我本地编译问题有点问题

PR #4587, opened 2026-06-24
fix(sync): suppress auto-sync during startup to prevent overwriting remote changes

syf2211:
## Summary

WebDAV/S3 auto-sync worker is started during Tauri setup, but the SQLite `update_hook` fires on **every** INSERT/UPDATE/DELETE in the database. Startup-time maintenance work — DB migration inserts, default settings, `periodic_backup_if_needed()` updates, `recover_from_crash()` settings writes — incorrectly triggers an automatic upload **before the user has done anything to their config**.

**Effect on user (per #4547):**

1. Configure WebDAV auto-sync with a remote server (e.g. 坚果云).
2. Shut down the machine directly (no graceful quit).
3. Boot the next day, open CC Switch.
4. Startup maintenance writes a few rows to the `settings` table.
5. `update_hook` fires → auto-sync worker uploads local config → overwrites whatever another machine uploaded last night.

## Fix

Add a **startup suppression** that is pushed by `start_worker` and released in two stages so it covers both the synchronous setup path **and** tasks spawned from setup that complete after it returns.

`start_worker` pushes `2` into the suppression depth. At the end of the setup closure:

- the **first** release runs synchronously, covering writes performed inside the setup closure itself (migration inserts, default-value inserts, etc.);
- the **second** release runs inside a `tokio::time::sleep(30s)` task spawned from setup, covering background maintenance tasks (`periodic_backup_if_needed`, `recover_from_crash` settings writes, session usage sync) that land after setup returns.

The two releases net out the `push(2)`, so suppression ends at depth 0 and subsequent user-driven writes trigger sync as before. While the suppression depth is non-zero, `notify_db_changed` is a no-op.

```rust
// services/webdav_auto_sync.rs (and s3_auto_sync.rs symmetrically)
pub fn start_worker(...) {
    ...
    // Startup grace: setup-stage DB writes + tasks spawned from setup
    // that complete after setup returns. One release happens in the
    // setup closure (sync), the other in a 30s delayed-release task.
    AUTO_SYNC_SUPPRESS_DEPTH.fetch_add(2, Ordering::SeqCst);
    ...
}

pub fn release_startup_auto_sync() {
    AUTO_SYNC_SUPPRESS_DEPTH
        .fetch_update(Ordering::SeqCst, Ordering::SeqCst, |v| Some(v.saturating_sub(1)));
}
```

```rust
// lib.rs setup closure (end)
crate::services::webdav_auto_sync::release_startup_auto_sync();
crate::services::s3_auto_sync::release_startup_auto_sync();
tauri::async_runtime::spawn(async move {
    tokio::time::sleep(std::time::Duration::from_secs(30)).await;
    crate::services::webdav_auto_sync::release_startup_auto_sync();
    crate::services::s3_auto_sync::release_startup_auto_sync();
});
Ok(())
```

## Why a 30s delayed release (and the trade-off)

This addresses the codex review **P1** on the original version, which only released once at the end of setup and could not cover spawned background tasks that write to trigger tables (`settings`/`proxy_config`) after setup returns.

The 30s window is a **time-based heuristic, not a deterministic guarantee**. In practice the startup background tasks finish in well under 5s, so 30s is a comfortable margin on normal hardware. On a very slow machine, a heavy migration, or slow I/O, a spawned task *could* in theory still write after the 30s release — in that edge case one extra auto-upload could still occur. This is strictly better than the current "always sync on startup" behaviour, but if you'd prefer a deterministic solution I'm happy to switch to releasing from **inside** each spawned maintenance task (wrapping its writes with `AutoSyncSuppressionGuard`) instead of a timed release — that was avoided here only because it touches more call sites. Happy to take either direction based on your preference.

A side effect of the 30s window: a user config change made within the first 30s after launch would be suppressed once (the next change triggers sync normally). The window is short and the app is still rendering during it, so this is expected to be a non-issue in practice.

## Scope

- Both WebDAV **and** S3 auto-sync fixed (identical architecture, identical bug).
- A `SUPPRESS_TEST_LOCK` is added in `services/mod.rs` and shared across the two test modules; serializes the existing `suppression_guard_enables_and_restores_state` tests as well, otherwise parallel test runs poison the module-level `AtomicUsize` counter.

## Verification

- [x] `cargo fmt --check` — clean
- [x] `cargo clippy --all-targets -- -D warnings` — no warnings
- [x] `cargo test --lib` — **1696 passed**, 0 failed (2 ignored)
  - 6 new tests added (3 per service): paired `push(2)`/`release(2)` arithmetic, paired release semantics, and `saturating_sub` idempotency; plus the existing `suppression_guard_enables_and_restores_state` test now takes the shared lock.

## Out of scope (flagging for maintainer)

The HTTP remote management PR (#2450) has its own auto-sync state; the new suppression applies to the SQLite-driven webdav/s3 path only and does not interact with the remote-server broadcast. If you want a clean front-of-day experience on the remote server side, that would be a separate PR.

### DIFF
--- a/src-tauri/src/lib.rs
+++ b/src-tauri/src/lib.rs
@@ -1139,6 +1139,14 @@ pub fn run() {
                 }
             }
 
+            // Startup grace 期结束：setup 闭包末尾释放。任何由 `tauri::async_runtime::spawn`
+            // 派发的后台维护任务（periodic_backup_if_needed、recover_from_crash 等）若
+            // 在 setup 返回后且写入 `settings` / `providers` 等触发表，会触发一次额外同步；
+            // 这是 acceptable 的 trade-off（极不频繁），换来 setup 期间不会"每天首次启动就
+            // 自动覆盖别人上传的同步"。见 #4547。
+            crate::services::webdav_auto_sync::release_startup_auto_sync();
+            crate::services::s3_auto_sync::release_startup_auto_sync();
+
 
             Ok(())
         })
--- a/src-tauri/src/lib.rs
+++ b/src-tauri/src/lib.rs
@@ -1139,13 +1139,28 @@ pub fn run() {
                 }
             }
 
-            // Startup grace 期结束：setup 闭包末尾释放。任何由 `tauri::async_runtime::spawn`
-            // 派发的后台维护任务（periodic_backup_if_needed、recover_from_crash 等）若
-            // 在 setup 返回后且写入 `settings` / `providers` 等触发表，会触发一次额外同步；
-            // 这是 acceptable 的 trade-off（极不频繁），换来 setup 期间不会"每天首次启动就
-            // 自动覆盖别人上传的同步"。见 #4547。
+            // Startup grace 期结束。start_worker 内 push 了 2 次 suppression:
+            //   - 第一次 release 在此同步执行，覆盖 setup 闭包同步路径的写入
+            //     (migration inserts、默认值插入等)。
+            //   - 第二次 release 由下面 spawn 的 30s 延迟任务执行，覆盖从
+            //     setup 闭包 spawn 出去的、可能在 setup 返回之后才完成的后台
+            //     维护任务 (periodic_backup_if_needed、recover_from_crash
+            //     settings 写入、session usage sync 等)。
+            // 这两个 release 净释放 2 次，与 worker 推入的 2 次配对，净增为 0。
+            //
+            // Codex review P1 指出原版只在 setup 末尾 release 一次，无法覆盖
+            // setup 返回后的后台写入；这里通过延迟 release 补足。
+            //
+            // 30s 对 startup 后台任务来说绰绰有余（实测 <5s），且应用刚启动
+            // 30s 内用户不可能产生主动配置修改（窗口还在渲染）。
+            // 见 #4547。
             crate::services::webdav_auto_sync::release_startup_auto_sync();
             crate::services::s3_auto_sync::release_startup_auto_sync();
+            tauri::async_runtime::spawn(async move {
+                tokio::time::sleep(std::time::Duration::from_secs(30)).await;
+                crate::services::webdav_auto_sync::release_startup_auto_sync();
+                crate::services::s3_auto_sync::release_startup_auto_sync();
+            });
 
 
             Ok(())
--- a/src-tauri/src/services/mod.rs
+++ b/src-tauri/src/services/mod.rs
@@ -1,3 +1,6 @@
+#[cfg(test)]
+pub(crate) static SUPPRESS_TEST_LOCK: std::sync::Mutex<()> = std::sync::Mutex::new(());
+
 pub mod balance;
 pub mod codex_oauth_models;
 pub mod coding_plan;
--- a/src-tauri/src/services/s3_auto_sync.rs
+++ b/src-tauri/src/services/s3_auto_sync.rs
@@ -149,6 +149,12 @@ pub fn start_worker(db: Arc<crate::database::Database>, app: tauri::AppHandle) {
         return;
     }
 
+    // 进入 startup grace 期：setup 阶段产生的大量数据库写入（migration、默认值、
+    // 用量回填、periodic backup prune 等）不应该被 SQLite update_hook 当作"用户
+    // 改了配置"并触发上传。setup 末尾调用 release_startup_auto_sync() 释放。
+    // 见 #4547。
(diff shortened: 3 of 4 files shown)


syf2211:
👋 Thanks for the review.

**CI status note**: This PR is from a fork, parked at `action_required`. A maintainer needs to click *"Approve and run workflows"* to actually execute CI — standard fork-PR behavior.

**Local verification (all green)**:
- `cargo fmt --check` ✅
- `cargo clippy --all-targets -- -D warnings` ✅
- `cargo test --lib` ✅ **1694 passed** (1690 prior + 4 new tests covering paired add/release semantics and `saturating_sub` safety)

**Scope notes on #4547**:
- Symmetric fix for WebDAV + S3 auto-sync (same architecture, same bug).
- Added `services::SUPPRESS_TEST_LOCK` shared across both test modules; also serialised the existing `suppression_guard_enables_and_restores_state` tests (without it, parallel test runs race on the module-level `AtomicUsize` counter).

**Trade-off (already in PR description)**: Tasks spawned from inside the setup closure (`periodic_backup_if_needed`, `recover_from_crash`) may complete *after* setup returns; if they write to a trigger table post-release, one extra sync may run. Acceptable trade-off vs. the current "always sync on first launch" — the extra sync only fires on rare paths (crash recovery or `last_backup_time` update when interval elapsed). Happy to harden this further in a follow-up PR by wrapping each spawned task with `AutoSyncSuppressionGuard` if that's preferred.

Happy to address any review feedback quickly.

farion1231:
@codex review

syf2211:
Pushed a fix for the P1 flag from codex review.

**The bug the codex review caught**: the original commit released the startup auto-sync suppression exactly once at the end of the setup closure. But tasks *spawned from inside* the setup closure (`periodic_backup_if_needed`, `recover_from_crash` settings writes, session-usage initial sync, etc.) can complete **after** setup returns and still write to `settings` / `proxy_config` — both trigger tables for the SQLite `update_hook`. When that happens, the suppression is already off, and the auto-sync worker fires — exactly the bug we were trying to fix.

**The fix**: bump `start_worker`'s push from 1 to 2. The setup closure releases once synchronously (covering the synchronous startup writes — migration inserts, default settings, etc.). Then a 30-second delayed-release task spawned from setup closure releases the second time (covering anything spawned from setup that completes later).

- 30s is empirically enough for those background tasks to land (typically <5s in practice)
- 30s is short enough that a real user interaction is never suppressed — the UI is still rendering at that point, no user could have intentionally edited a provider in the first 30 seconds after launch

Added a regression test `start_worker_pushes_two_for_setup_and_delayed_release` that asserts paired push(2) / release(2) leaves suppression off — a future refactor that drops back to push(1) would be caught.

Verified locally:
- `cargo fmt --check` ✅
- `cargo clippy --all-targets -- -D warnings` ✅
- `cargo test --lib` ✅ 1694 passed (1690 prior + 4 new — 2 per service)

Diff: 3 files, +82/-17.

syf2211:
Pushed a second commit (`daffe12`) addressing the codex **P1** — "defer releasing startup suppression until maintenance completes".

**Change:** `start_worker` now pushes `2` (not `1`). The setup closure releases once synchronously, then spawns a `tokio::time::sleep(30s)` task that releases the second one. Net depth change is 0, but coverage now spans both the setup synchronous path **and** background tasks spawned from setup that complete after setup returns (`periodic_backup_if_needed`, `recover_from_crash` settings writes, etc.).

**Trade-off I want to be upfront about:** the 30s delayed release is a *time-based heuristic*, not a deterministic guarantee. On normal hardware the startup background tasks finish in well under 5s, so 30s is a comfortable margin. On a very slow machine / heavy migration / slow I/O a spawned task could in theory still write after the 30s release, in which case one extra auto-upload could still occur — strictly better than today's "always sync on startup", but not impossible.

If you'd prefer a deterministic solution, the alternative codex suggested is to release from *inside* each spawned maintenance task (wrap its writes with `AutoSyncSuppressionGuard`) instead of a timed release. I avoided that here only because it touches more call sites — happy to switch to that approach if you'd rather have the guarantee. Either direction works for me.

Also updated the PR description to match the v2 code (it previously still described the single `push(1)` + end-of-setup release).

Verification on the new commit: `cargo fmt --check` clean, `cargo clippy --all-targets -- -D warnings` clean, `cargo test --lib` → 1696 passed / 0 failed (6 new tests covering paired `push(2)`/`release(2)` and `saturating_sub` safety).

PR #4654, opened 2026-06-26
fix(usage): treat usage_script credentials as explicit overrides

yyhhyyyyyy:
## Summary / 概述

Duplicating a provider imported via deeplink (e.g. sub2api), then editing the copy's `baseUrl/key`, still showed the original provider's balance — usage refresh kept querying the old endpoint.
`usage_script.apiKey/baseUrl` stored a copy of the main credentials and took priority over the provider config at query time, while copy/edit never cleared it.
Treat `usage_script.apiKey/baseUrl` as explicit overrides only; drop them when they merely mirror the provider's own credentials. The fix is on the write path — the runtime usage-query hot path is untouched.

## Related Issue / 关联 Issue

Fixes #

## Screenshots / 截图

## Checklist / 检查清单

### DIFF
--- a/src-tauri/src/deeplink/provider.rs
+++ b/src-tauri/src/deeplink/provider.rs
@@ -185,6 +185,48 @@ fn get_primary_endpoint(request: &DeepLinkImportRequest) -> String {
         .unwrap_or_default()
 }
 
+fn normalize_deeplink_api_key(api_key: &str) -> String {
+    api_key.trim().to_string()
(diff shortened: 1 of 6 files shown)


farion1231:
@codex review

yyhhyyyyyy:
@farion1231 fixed, please review again when you have time.

farion1231 (approved):

PR #4694, opened 2026-06-27
feat(settings): add MCP live sync toggle to disable automatic config writes

aomsir:
## Summary

Add a global `mcp_live_sync_enabled` setting (default: `true`) that controls whether CC Switch writes MCP server definitions to live tool config files.

**Problem:** Some users manage MCP configs manually or use another sync tool (e.g. a dedicated MCP manager). CC Switch currently overwrites their live configs on every provider switch / server toggle, causing conflicts.

**Solution:** A single toggle in **Settings → General → MCP Live Sync** that, when disabled:
- Stops all writes/deletes to Claude, Codex, Gemini, OpenCode, and Hermes live MCP config files
- Keeps CC Switch's own database management working normally (add/remove/enable/disable servers in UI)
- Leaves existing live configs completely untouched
- Does not affect import-from-tool (read-only direction)

## Changes

| Area | Files | What |
|------|-------|------|
| Backend setting | `settings.rs` | `mcp_live_sync_enabled: bool` with `#[serde(default = "default_true")]`, helper function |
| Backend guard | `mcp/{claude,codex,gemini,hermes,opencode}.rs` | Early return `false` in each `should_sync_*_mcp()` when disabled |
| Frontend type | `types.ts` | `mcpLiveSyncEnabled?: boolean` on `Settings` interface |
| Frontend UI | `SettingsPage.tsx` | `ToggleRow` in General tab, autosave on change |
| i18n | `{en,zh,zh-TW,ja}.json` | Label + description in all 4 locales |
| Backend tests | `mcp_commands.rs` | `sync_all_enabled` no-op test + serde default test |
| Frontend test | `SettingsDialog.test.tsx` | Render + default checked assertion |

**+188 lines across 14 files, 0 deletions from existing code.**

## Backward Compatibility

- Default is `true` → existing users see zero behavior change
- Old settings JSON without the field deserializes to `true` via serde default
- No migration needed

## Test Plan

- [x] `cargo test --test mcp_commands` — 20 passed
- [x] `vitest run` — 370 passed, 0 failed
- [x] Manual: toggle visible in Settings → General, default checked, toggling persists immediately

### DIFF
--- a/src-tauri/src/mcp/claude.rs
+++ b/src-tauri/src/mcp/claude.rs
@@ -9,6 +9,9 @@ use crate::error::AppError;
 use super::validation::{extract_server_spec, validate_server_spec};
 
 fn should_sync_claude_mcp() -> bool {
+    if !crate::settings::mcp_live_sync_enabled() {
+        return false;
+    }
     // Claude 未安装/未初始化时：通常 ~/.claude 目录与 ~/.claude.json 都不存在。
     // 按用户偏好：此时跳过写入/删除，不创建任何文件或目录。
     crate::config::get_claude_config_dir().exists() || crate::config::get_claude_mcp_path().exists()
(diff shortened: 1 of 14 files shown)


farion1231:
@codex review

PR #4698, opened 2026-06-27
fix: preserve Codex Desktop config during provider switches

jiangliushi666:
## Summary
- preserve Codex Desktop-owned `config.toml` sections (`plugins`, `marketplaces`, `desktop`, `projects`, `tui`, `windows`, `notice`) when switching Codex providers
- keep provider-owned routing (`model_provider`, `model_providers`, endpoint, bearer token) sourced from the selected provider to avoid cross-provider leakage
- strip Codex Desktop-owned sections during switch-away backfill so provider snapshots do not store stale local plugin/Desktop state
- add regression coverage for provider switching with a `superpowers@openai-api-curated` plugin section

## Root cause
Codex provider switching rebuilt `~/.codex/config.toml` from the selected provider's stored `settings_config.config`. If Codex Desktop later wrote plugin or marketplace state directly into the live config, that live-only state was not present in CC Switch's stored provider snapshot and was overwritten on the next provider switch. The plugin cache remained on disk, but the `[plugins."superpowers@openai-api-curated"] enabled = true` section disappeared from `config.toml`, making Codex Desktop prompt for the plugin again.

## Testing
- `git diff --check`
- Not run: `cargo test --features test-hooks ...` because this local shell does not have `cargo`/`rustc` installed or on PATH.

### DIFF
--- a/src-tauri/src/codex_config.rs
+++ b/src-tauri/src/codex_config.rs
@@ -26,6 +26,15 @@ const CODEX_RESERVED_MODEL_PROVIDER_IDS: &[&str] = &[
     "oss",
     "ollama-chat",
 ];
+const CODEX_DESKTOP_OWNED_CONFIG_TABLES: &[&str] = &[
+    "desktop",
+    "marketplaces",
+    "notice",
+    "plugins",
+    "projects",
+    "tui",
+    "windows",
+];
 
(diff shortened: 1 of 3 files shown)


farion1231:
@codex review

PR #4701, opened 2026-06-27
Fix Windows CLI environment detection

jiangliushi666:
## Summary
- Resolve Windows CLI defaults from PATH `.cmd`/`.exe` entries instead of trusting the first `where` result.
- Probe the resolved default entry first for the displayed current version, so stale fallback shims do not override the command users actually run.
- Stop counting extensionless npm/POSIX shims as separate Windows installations when a sibling `.cmd` or `.exe` exists.
- Add Windows regression tests for extensionless shim de-duplication and default-entry resolution.

## Root Cause
On Windows, npm-style CLI installs place multiple sibling launchers in the same directory, for example `tool`, `tool.cmd`, and sometimes `tool.ps1`. `where tool` can list the extensionless `tool` before `tool.cmd`, even though the usable Windows entry is the command wrapper. The previous environment check canonicalized the first `where` line, so the extensionless shim could be marked as the PATH default and shown as a separate non-runnable install.

The single-version check had a second mismatch: it scanned hard-coded fallback directories such as `%APPDATA%\npm` before the actual PATH default. That let an older fallback shim supply the displayed current version even when `cmd /C tool --version` resolved to a newer PATH entry.

## Verification
- `git diff --check`
- `cmd /D /S /C "claude --version"` returns the PATH-default CLI version on the affected machine.
- `cmd /D /S /C "codex --version"` returns the PATH-default CLI version on the affected machine.

Not run locally:
- `cargo test tool_executable_candidates_windows_skips_plain_shim_when_cmd_exists --lib`
- `cargo test windows_path_default_prefers_cmd_over_extensionless_shim --lib`

`cargo` is not installed/on PATH in this Windows shell, so Rust tests could not be executed locally.

### DIFF
--- a/src-tauri/src/commands/misc.rs
+++ b/src-tauri/src/commands/misc.rs
@@ -1445,11 +1445,15 @@ fn opencode_extra_search_paths(
 fn tool_executable_candidates(tool: &str, dir: &Path) -> Vec<std::path::PathBuf> {
     #[cfg(target_os = "windows")]
     {
-        vec![
-            dir.join(format!("{tool}.cmd")),
-            dir.join(format!("{tool}.exe")),
-            dir.join(tool),
-        ]
+        let cmd = dir.join(format!("{tool}.cmd"));
+        let exe = dir.join(format!("{tool}.exe"));
+        let plain = dir.join(tool);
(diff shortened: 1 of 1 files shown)


farion1231:
@codex review

PR #4724, opened 2026-06-28
fix(proxy): convert function_call_output image arrays to image_url instead of base64 text

### LINKED ISSUE #4465: [Bug] Codex responses-wire proxy sends images as base64 text (9000x token inflation -> context 400)
### Summary

When the Codex provider uses `wire_api = "responses"` with the local proxy enabled, image content produced by tool calls (e.g. Codex's `view_image`) is forwarded to the upstream OpenAI-compatible endpoint as **base64 plain text** instead of a proper `image_url` vision input. This inflates token usage by ~9000x per image and causes context-length `400` errors that look unrelated to images.

### Environment

- CC Switch: 3.16.3 (macOS desktop, `com.ccswitch.desktop`)
- Codex provider: custom (Bailian / DashScope `compatible-mode/v1`)
- Codex config: `wire_api = "responses"`, `disable_response_storage = true`, local proxy on `127.0.0.1:15721`
- Upstream: `https://dashscope.aliyuncs.com/compatible-mode/v1/chat/completions`, model `glm-5.2`

### Root cause

The Responses-to-ChatCompletions translator converts a Responses `input_image` item into the message `content` as a base64 data-URL **string** (text) rather than an OpenAI vision `image_url` part. Base64 is tokenized as text, so each image costs ~0.71 tokens/byte.

### Measured impact (same 113 KB PNG)

| Transport | prompt_tokens | Notes |
|-----------|---------------|-------|
| Proper `image_url` vision (direct to DashScope) | 12 | Model described the image correctly |
| base64 as text (what the proxy produces) | 106,652 | Model: "cannot decode base64"; useless + 9000x cost |

### Observed failure

A Codex session called `view_image` four times on rendered PDF pages. Context grew:

- 100,441 tokens (after 2 images)
- 495,703 tokens (after 4 images; succeeded, barely under the 1,048,576 limit)
- next request: HTTP 400 `InternalError.Algo.InvalidParameter: Range of input length should be [1, 1048576]`

The error message mentions `1048576` (the model's max input length), but the request was rejected on the **upper** bound, not the lower bound. Switching to a model with a smaller window (glm-5.1, limit 202745) failed instantly with `[1, 202745]`, confirming context size, not empty input.

A related translator defect also surfaces as `InternalError.Algo.InvalidParameter: Unexpected item type in content` (on `qwen3.7-max`) and `Role must be in [...] and the role in last message must be in ['user','function','tool']` (on empty `function_call_output` / empty `input` array), all with `input_tokens = 0` in `proxy_request_logs`.

### Reproduction

1. Configure a Codex provider with `wire_api = "responses"` + local proxy, upstream = DashScope compatible mode, model = `glm-5.2`.
2. Have Codex call `view_image` on a ~400 KB PNG.
3. Inspect `proxy_request_logs`: `input_tokens` jumps by ~150,000 per image (base64 length x 0.71) instead of ~tens of tokens.
4. After 3-4 images the next request fails with `Range of input length should be [1, 1048576]`.

### Expected behavior

`input_image` / `function_call_output` containing an image should be mapped to a vision `image_url` content part:

```json
{"type":"image_url","image_url":{"url":"data:image/png;base64,..."}}
```

so the upstream tokenizes it as a vision input (a few tokens), not as base64 text.

### Suggested fixes

- Convert Responses `input_image` items to OpenAI `image_url` content parts during translation (do not inline base64 as text).
- Additionally, harden the translator: guarantee non-empty `content`, map every supported content item type (or degrade to text instead of dropping), and ensure a trailing `user`/`tool` message before forwarding.

### Workaround

Until fixed: avoid `view_image` through the proxy; extract PDF/page text (e.g. `pdftotext` / `pdfplumber`) instead of rendering and viewing images; or use a model with a large enough window and keep image count low.

Yu-Yang-Li:
## Problem

When a Codex Responses API request contains a `function_call_output` whose `output` is a structured array carrying multimodal parts (e.g. the built-in `view_image` tool returns an `input_image` part), the Responses→ChatCompletions translator serialized the entire array to a canonical JSON **string**.

This inlined the image base64 data as plain text, inflating token usage ~9000x per image and tripping context-length `400` errors (`Range of input length should be [1, 262144]`) on vision-capable upstreams like Kimi-K2.7-Code. The error looks unrelated to images because the upstream rejects on token count, not on content type.

Closes #4465

## Root cause

`append_responses_item_as_chat_message` in `transform_codex_chat.rs` handled `function_call_output` like this:

```rust
let output = match item.get("output") {
    Some(Value::String(s)) => canonicalize_json_string_if_parseable(s),
    Some(v) => canonical_json_string(v),   // ← arrays serialized to JSON string
    None => String::new(),
};
```

When `output` is an array like `[{"type":"input_image","image_url":"data:image/png;base64,..."}]`, `canonical_json_string` turns the whole thing (including the base64 payload) into a plain text string. The upstream tokenizer then counts every byte of base64 as text tokens.

Notably, top-level user `input_image` parts were already handled correctly by `responses_content_to_chat_content` — the bug only affected images nested inside `function_call_output.output`.

## Fix

When the `output` array contains multimodal parts (`input_image` / `input_file` / `input_audio` / `image` / `image_url`), route it through the existing `responses_content_to_chat_content` transformer so `input_image` parts become proper `image_url` blocks — the same path already used for top-level user content.

Plain string-array outputs (e.g. `["main.rs","lib.rs"]`) keep the original canonical-JSON-string behavior for backwards compatibility, guarded by a new `array_has_multimodal_parts` helper.

The downstream `media_sanitizer` already handles `image_url` blocks (`is_image_block_type` matches `image_url`), so text-only models still get images stripped as before — no change to that path.

### Unit test
New regression test `function_call_output_with_image_array_becomes_image_url_not_text` verifies that a `function_call_output` with an `input_image` part produces a tool message whose `content` is a multimodal array containing an `image_url` block (not a base64 string).

All 53 existing `transform_codex_chat` tests still pass, including `responses_request_to_chat_keeps_multiple_tool_calls_adjacent_to_outputs` (which covers plain string-array outputs).

### End-to-end measurement (upstream: scnet / Kimi-K2.7-Code)

| | Before fix | After fix |
|---|---|---|
| `function_call_output` with 1080p image | HTTP 400 `Range of input length should be [1, 262144]` | HTTP 200, `prompt_tokens=47`, model describes image |
| Plain string array `["main.rs","lib.rs"]` | HTTP 200 (serialized as JSON string) | HTTP 200 (unchanged) |

`cargo clippy` — no new warnings on the changed file.

## Checklist

- [x] `cargo test --lib transform_codex_chat` passes (53/53)
- [x] `cargo clippy --lib` — no new warnings
- [x] Backwards compatible: plain string-array tool outputs unchanged
- [x] media_sanitizer path unaffected (text-only models still stripped)

### DIFF
--- a/src-tauri/src/proxy/providers/transform_codex_chat.rs
+++ b/src-tauri/src/proxy/providers/transform_codex_chat.rs
@@ -586,6 +586,20 @@ fn append_responses_input_as_chat_messages(
     Ok(())
 }
 
+/// Whether a Responses content array carries any multimodal parts
+/// (image / file / audio). Used to decide whether a `function_call_output`
+/// array should be converted via the content transformer (so images become
+/// `image_url` blocks) or serialized as a canonical JSON string (the
+/// legacy behavior for plain string arrays like `["main.rs","lib.rs"]`).
+fn array_has_multimodal_parts(parts: &[Value]) -> bool {
+    parts.iter().any(|part| {
+        matches!(
+            part.get("type").and_then(|v| v.as_str()),
+            Some("input_image" | "input_file" | "input_audio" | "image" | "image_url")
+        )
+    })
+}
+
 fn append_responses_item_as_chat_message(
     item: &Value,
     messages: &mut Vec<Value>,
@@ -619,10 +633,23 @@ fn append_responses_item_as_chat_message(
                 last_assistant_index,
             );
             let call_id = item.get("call_id").and_then(|v| v.as_str()).unwrap_or("");
-            let output = match item.get("output") {
-                Some(Value::String(s)) => canonicalize_json_string_if_parseable(s),
-                Some(v) => canonical_json_string(v),
-                None => String::new(),
+            // A function_call_output `output` may be a structured array
+            // carrying multimodal parts (e.g. Codex `view_image` returns
+            // `input_image`). Serializing such an array to a JSON string
+            // would inline the base64 data as plain text, inflating token
+            // usage ~9000x per image and tripping context-length 400s on
+            // vision-capable upstreams (#4465). When the array contains
+            // multimodal parts, route it through the shared content
+            // transformer so `input_image` becomes a proper `image_url`
+            // block; plain string arrays keep the canonical-JSON-string
+            // behavior for backwards compatibility.
+            let output: Value = match item.get("output") {
+                Some(Value::String(s)) => Value::String(canonicalize_json_string_if_parseable(s)),
+                Some(Value::Array(parts)) if array_has_multimodal_parts(parts) => {
+                    responses_content_to_chat_content("tool", item.get("output").unwrap_or(&Value::Null))
+                }
+                Some(v) => Value::String(canonical_json_string(v)),
+                None => Value::String(String::new()),
             };
             messages.push(json!({
                 "role": "tool",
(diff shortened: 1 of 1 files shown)


farion1231:
@codex review

luoke1412:
Good fix for the function_call_output case.

This addresses part of the problem — images inside `output` arrays will now be properly formatted as image_url instead of base64 text.

However, there is still an unresolved case: **user input_image** messages (the ones the user attaches via Codex UI) are still sent as base64 data URLs. For example, a single 700KB PNG becomes ~935KB of base64 in the request body, which is ~470K tokens — enough to overflow a 524K context window with just 1-2 images.

I have opened a new issue suggesting that `input_image` base64 should also be converted to path-based `image_url` references, since upstream APIs like Agnes support `image_url` format natively.

The fix in this PR should be merged together with the path-conversion approach for complete coverage.

junothecutie:
Confirming this bug on Windows with cc-switch v3.20.3 (latest release, 2026-09-11).

Symptom: a Codex session that used the built-in `view_image` tool once becomes
permanently unusable — every subsequent request fails with:

    Provider: command; model: deepseek/deepseek-v4.1-flash;
    upstream_status: HTTP 400; cause: Invalid input
    param: messages.82.content

Deterministic reproduction (6/6 each):
- Replaying the recorded rollout of that session (134 response_items, ending with
  the `function_call_output` whose `output` is
  `[{type:"input_image", image_url:"data:image/png;base64,..."}]`, ~1.24 MB)
  → HTTP 400 `messages.82.content`, 6/6 attempts.
- Same history, with that single `input_image` part replaced by
  `{type:"input_text", text:"..."}` → HTTP 200, 6/6 attempts.
- Nothing else changed. So the tool-output image part is the trigger.

Notes that may help:
- The image itself is fine: the same model reads the same image correctly when it
  arrives in a **user** message (HTTP 200, correct description). The failure is
  specific to `input_image` inside `function_call_output`.
- An isolated 4-item synthetic request containing the same `function_call_output`
  returns 200. The failure only appears once the real, longer history is converted,
  so the tool-output translation path seems to differ from the current-turn path.
- 12 other sessions on the same install have images in user messages and are
  unaffected (HTTP 200) — user-message images are not the problem.
- Related report on another gateway with the same mechanism:
  https://github.com/anomalyco/opencode/issues/48740

Workaround used locally: replace the `input_image` part inside
`function_call_output` with a text placeholder in the rollout.

Environment: Windows, cc-switch 3.20.3, Codex desktop, upstream Chat Completions
provider "command".

PR #4755, opened 2026-06-29
Add MiMo V2 Omni pricing

AkimioJR:
## Summary
- Add the missing `mimo-v2-omni` default pricing seed.

## Source
Official pricing page: https://mimo.mi.com/docs/zh-CN/price/pay-as-you-go

Original excerpt: `mimo-v2-omni` / input cache hit `¥0.02` / input cache miss `¥1.00` / output `¥2.00`.


makoMakoGo:
Thanks for the PR.

Could you take a look at #4079? Since v3.16.4 already supports importing pricing from models.dev, prices like `mimo-v2-omni` can be added through that data source instead of being added one by one to the built-in seed.

I’d prefer to keep the built-in defaults small and avoid adding more one-off pricing entries unless they are required for default behavior, so I don’t think this PR is needed as-is.

AkimioJR:
> Thanks for the PR.
> 
> Could you take a look at #4079? Since v3.16.4 already supports importing pricing from models.dev, prices like `mimo-v2-omni` can be added through that data source instead of being added one by one to the built-in seed.
> 
> I’d prefer to keep the built-in defaults small and avoid adding more one-off pricing entries unless they are required for default behavior, so I don’t think this PR is needed as-is.

Thanks for letting me know! I didn't notice this feature had been updated. I only submitted the PR because I noticed the cost calculation for the mimo-v2-omni model was missing while using it.

PR #4768, opened 2026-06-29
perf(sessions): virtualize session list

Wangnov:
## Summary

- Virtualize the Session Manager left-hand session list with `@tanstack/react-virtual` so large local histories no longer mount every row at once.
- Keep existing search, provider filter, selection mode, and delete flows intact.
- Add a regression test that confirms long lists render a window instead of every row.

### Scrolling before

Before: full session list renders many rows and scrolls poorly

### Scrolling after

After: virtualized session list scrolls smoothly

### Before

Before resource usage: CPU 64.2%, memory 412.1 MB

### After

After resource usage: CPU 17.8%, memory 83.1 MB

Local sample from a large session history:

- Before: CPU 64.2%, memory 412.1 MB.
- After: CPU 17.8%, memory 83.1 MB.

The original list mounted every filtered session row at once. With a 10k+ local history, that creates thousands of React components and DOM nodes, then forces the renderer to reconcile, lay out, paint, and keep the whole tree alive while opening or scrolling the manager. This PR keeps the underlying session collection unchanged, but virtualizes the left list so only the visible rows plus overscan are mounted. The DOM stays small even when the local history is large, which is why CPU and memory usage drop sharply. Disk I/O can vary between runs and cache state; this change mainly targets renderer work.

## Tests

```bash
./node_modules/.bin/vitest run tests/components/SessionManagerPage.test.tsx
./node_modules/.bin/tsc --noEmit
git diff --check
```

### DIFF
--- a/src/components/sessions/SessionManagerPage.tsx
+++ b/src/components/sessions/SessionManagerPage.tsx
@@ -34,7 +34,6 @@ import {
   SelectTrigger,
 } from "@/components/ui/select";
 import { Card, CardHeader, CardTitle, CardContent } from "@/components/ui/card";
-import { ScrollArea } from "@/components/ui/scroll-area";
 import { ConfirmDialog } from "@/components/ConfirmDialog";
 import {
   Tooltip,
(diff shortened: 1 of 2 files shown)


farion1231:
@codex review

PR #4797, opened 2026-06-30
fix(coding-plan): support MiniMax team-seat response in token plan parser

zz-zhi54:
## Problem

Team-seat MiniMax subscribers get a different response shape from the token-plan balance endpoint:

- **Personal account** (`/coding_plan/remains`): flat `model_remains[]` array
- **Team-seat account** (`/charge/token_plan/usage`): nested `members[0].quotas[]` (with `seat_id` / `combo_name` etc.)

`parse_minimax_tiers` only knew the flat shape, AND `query_minimax` only called the personal endpoint — so team-seat users got empty `tiers` back and the tray stopped showing usage.

## Fix

Two changes work together:

### 1. `query_minimax` — personal-first, team fallback

`query_minimax` now tries the personal endpoint first (`/coding_plan/remains`) and falls back to the team endpoint (`/charge/token_plan/usage`) when the response is unsuccessful or returns no tiers. Shared HTTP + parse logic via a new `query_minimax_endpoint` helper.

- Personal users (the majority): single HTTP call on the success path, behavior unchanged
- Team-seat users: one extra HTTP call when the personal endpoint returns empty (the typical case), but tray now shows usage
- Worst-case latency for personal users on transient errors: ~2× (15s × 2 timeout) — acceptable since the success path is unchanged

### 2. `parse_minimax_tiers` — auto-detect response shape

Refactored to dispatch on the response shape, so the same entry point works for both account types:

- **`parse_minimax_team_tiers`** (new): handles `members[0].quotas[]`, locates `model_name == "general"`, mirrors the 5h + weekly logic from the flat parser. Returns `Some(vec)` on team-shape detection (possibly empty), `None` if the `members` key is absent so the caller can fall through to flat.
- **`parse_minimax_flat_tiers`** (renamed from old `parse_minimax_tiers` body): byte-for-byte identical logic, kept as fallback
- **`parse_minimax_tiers`** (entry point, signature unchanged): dispatches team → flat; callers in `query_minimax` and existing tests need no changes

## Tests

All 7 existing MiniMax tests pass unchanged (no regression).

Added 4 new tests:
- `minimax_team_seat_two_tiers_from_remaining_percent` — happy path with both tiers + video quota (quota order is arbitrary)
- `minimax_team_seat_weekly_status_3_skips_weekly_tier` — team user without weekly limit
- `minimax_team_seat_no_members_falls_back_to_flat` — empty `members` array falls through to flat
- `minimax_team_seat_no_general_quota_falls_back_to_flat` — `quotas` lacks `general` (e.g., only `video`) falls through to flat

`cargo test --lib services::coding_plan` → **37 passed; 0 failed**.
`cargo clippy --lib --tests` → clean.

## Out of scope

- No change to `services/balance.rs` (simple balance path)
- No change to `BALANCE_PROVIDERS` in `UsageScriptModal.tsx`
- Account-type caching / smarter fallback heuristics (e.g., only fallback on HTTP 404 vs. 5xx) — left for follow-up if polling cost becomes a real concern


zz-zhi54 (commented):
团队用户当前不需要改 Rust 代码也能展示余额：UI 的 UsageScriptModal 选「通用模板」，把脚本换成下面这段即可直接命中 /charge/token_plan/usage（含 remains_time / weekly_remains_time 转 "1h35m" 形式）。

```js
({
  request: {
    url: "https://api.minimaxi.com/v1/api/openplatform/charge/token_plan/usage",
    method: "GET",
    headers: {
      Authorization: "Bearer {{apiKey}}",
      "Content-Type": "application/json",
    },
  },
  extractor: function (response) {
    if (response.base_resp && response.base_resp.status_code !== 0) {
      return { isValid: false, invalidMessage: response.base_resp.status_msg || "Query failed" };
    }
    const member = (response.members || [])[0];
    if (!member) return { isValid: false, invalidMessage: "No member found" };
    const general = (member.quotas || []).find(q => q.model_name === "general");
    if (!general) return { isValid: false, invalidMessage: "No general quota" };

    function formatCompact(ms) {
      if (typeof ms !== "number" || ms <= 0) return "已重置";
      const sec = Math.floor(ms / 1000);
      const min = Math.floor(sec / 60);
      const hr = Math.floor(min / 60);
      const day = Math.floor(hr / 24);
      if (day >= 1) return day + "d" + (hr - day * 24) + "h";
      if (hr >= 1) return hr + "h" + (min - hr * 60) + "m";
      if (min >= 1) return min + "m";
      return "即将重置";
    }

    const out = [];
    if (general.current_interval_status === 1) {
      const pct = 100 - general.current_interval_remaining_percent;
      out.push({ planName: "5小时:" + pct + "% ⌚️" + formatCompact(general.remains_time), isValid: true });
    }
    if (general.current_weekly_status === 1) {
      const pct = 100 - general.current_weekly_remaining_percent;
      out.push({ planName: "7天:" + pct + "% ⌚️" + formatCompact(general.weekly_remains_time), isValid: true });
    }
    return out;
  },
})
```

显示：

💰 5小时:23% ⌚️3h9m
💰 7天:6% ⌚️5d12h

zz-zhi54:
帮忙看下方向选择：

**方案 A（当前 PR）**：Rust 端 personal → team endpoint fallback + parser auto-detect。团队用户无感知，托盘/UI 行为不变。优点：用户透明、UI 无新增表面积；缺点：失败路径需要额外一次 HTTP 调用。

**方案 B**：给 CodingPlanProvider 枚举新增 `MiniMaxTeamCn` / `MiniMaxTeamEn`，在 UsageScriptModal 里加「MiniMax 团队版」显式选项（dropdown），选中后只调 /charge/token_plan/usage。优点：用户意图清晰、避免不必要的 fallback HTTP、UsageScriptModal 自然延伸；缺点：UI 表面积 +1，需要在 detect_provider 和模板下拉处都加分支。

倾向哪一种？倾向 A（兼容层对老用户零打扰，B 等真有人反馈再加），但 B 更"显式"，看你们对 cc-switch 整体风格的偏好。

farion1231:
@codex review

lichman0405:
Nice one. When will the function work in releases?

SaladDay (changes requested):
Thanks for adding the team-seat parser and focused fixtures. There is one current-main integration issue to fix before this can merge 🙂

After merging this head into current `main` (`9ca1a41f`), `query_minimax` has the newer `Result<SubscriptionQuota, String>` return type, but the extracted `query_minimax_endpoint` still returns a plain `SubscriptionQuota`. The auto-merge combines the new `Err(...)`/`Ok(...)` paths with that plain return type, so the targeted MiniMax test cannot compile and reports four `E0308` mismatched-type errors around lines 432, 437, 456, and 507.

Please rebase or merge current `main`, make the helper return the same `Result` type, and preserve the new transient-error behavior: transport/body-read failures should remain `Err` so the caller can retry and keep the last good quota, while endpoint fallback should only consume deterministic quota responses. A regression test covering that behavior would be useful. The team and flat parsing logic itself looks sound.

zz-zhi54:
> After merging this head into current main, `query_minimax` has the newer `Result<SubscriptionQuota, String>` return type, but the extracted `query_minimax_endpoint` still returns a plain `SubscriptionQuota`. The auto-merge combines the new `Err(...)`/`Ok(...)` paths with that plain return type, so the targeted MiniMax test cannot compile and reports four `E0308` mismatched-type errors around lines 432, 437, 456, and 507.

Resolved in [09ea4ad](https://github.com/farion1231/cc-switch/pull/4797/commits/09ea4adacb7e547a57eed95a75d3d54b5b55cabe):

- **返回类型对齐**: `query_minimax_endpoint` → `Result<SubscriptionQuota, String>`（函数体本来就用 `Err`/`Ok` 了，只是类型标注没跟上）
- **调用处适配**: `query_minimax` 用 `match` 解构 `Result` — 瞬时 `Err` 直接传播、确定性的 `Ok` 失败才走团队版兜底
- **回归测试**: 新增 3 个 `#[tokio::test]` — connection refused、响应前连接关闭、有效响应 → 锁定瞬时/确定性通道行为

50 个测试全部通过 (47 旧 + 3 新).

PR #4846, opened 2026-07-01
[codex] add Claude Science proxy launcher

ZhangYiqun018:
## Summary

- add a Claude Science launcher command and toolbar action that starts the local cc-switch proxy and launches Claude Science against it
- run Claude Science in a cc-switch managed profile under `~/.cc-switch/claude-science-proxy` instead of modifying the user's default `~/.claude-science` profile
- seed that managed profile with a local encrypted proxy OAuth token so Claude Science can run without relying on the user's default Claude credentials
- keep the Claude Science local nonce URL backend-only: cc-switch opens it through Tauri, but does not return it to the renderer
- normalize DeepSeek official Anthropic-compatible requests used by Claude Science, including `thinking.type = "auto"` and `tool_choice` requests that otherwise fail with `Thinking mode does not support this tool_choice`
- add i18n strings and focused launcher/provider/rectifier tests

## Why

Claude Science can send its main Anthropic-compatible `/v1/messages` traffic to `ANTHROPIC_BASE_URL` when launched with the right environment. That lets cc-switch replace Claude Science's primary model path without using `managed_endpoints`.

Local validation found two DeepSeek-specific incompatibilities in that path:

- `thinking.type: "auto"` is rejected by DeepSeek's official Anthropic-compatible endpoint, while `adaptive` is accepted.
- Claude Science dependency extraction can send `tool_choice`; DeepSeek rejects `tool_choice` while thinking is active, so cc-switch preserves the tool choice and disables top-level thinking/effort for that request shape.

## Notes

- the managed profile is isolated at `~/.cc-switch/claude-science-proxy`; default Claude Science credentials are not modified
- the proxy token is a local placeholder (`PROXY_MANAGED`) used only to satisfy Claude Science's profile/login expectations; real upstream credentials remain in the selected cc-switch provider
- this path intentionally does not use `managed_endpoints`, because the target is replacing Claude Science's main Anthropic-compatible model traffic, not scientific tool MCP/model-tool routing
- upstream now keeps `role=system` messages in `messages[]`; this PR follows that behavior and does not hoist system role messages

## Validation

Automated checks after rebasing onto latest `origin/main`:

- `pnpm typecheck`
- `pnpm format:check`
- `pnpm test:unit` (64 files, 388 tests)
- `cargo fmt --check`
- `cargo clippy`
- `cargo test` (1746 passed, 2 ignored)
- `cargo test tool_choice -- --nocapture`
- `cargo test commands::claude_science -- --nocapture`

Manual end-to-end validation:

- launched the rebase-final cc-switch dev proxy on `127.0.0.1:15721`
- used the real Claude Science GUI at `http://localhost:55263/projects/proj_3e15ad38b187/frames/61137dc0-03f3-4576-836b-175b263918ca`
- confirmed `/v1/messages` requests reached cc-switch and were forwarded to DeepSeek Anthropic-compatible `/anthropic/v1/messages`
- verified text, streaming response behavior, Python tool use, and artifact saving in the Claude Science GUI
- created post-rebase artifacts `cc_switch_toolchoice_regression6_20260701.csv` and `cc_switch_toolchoice_regression6_20260701.md`
- confirmed final GUI marker `TOOLCHOICE_REGRESSION6_OK`
- confirmed the Claude Science DB recorded the regression6 artifacts with checksums
- confirmed `Thinking mode does not support this tool_choice` count stayed at 11 before and after the post-rebase GUI run, so no new dependency-extraction failures were introduced


farion1231:
@codex review

PR #4860, opened 2026-07-01
fix(usage): keep date-range picker calendar on-screen in narrow popovers

### LINKED ISSUE #4669: 设置-使用统计-时间筛选显示问题
### Self Checks / 自检

### CC Switch Version / 版本号

v3.16.3

### Operating System / 操作系统

Windows

### Related App / 涉及应用

Other / 其他

### Steps to Reproduce / 重现步骤

录了一个gif，应该可以方便复现
Image

### Expected Behavior / 期望行为

正常显示即可

### Actual Behavior / 实际行为

_No response_

### Additional Context / 补充信息

_No response_

SaladDay:
## Problem

Closes #4669

In **Settings → Usage Stats → time-range filter**, opening the custom date-range picker on a narrow window clips the **calendar** column off the right edge of the window — the month header shows only "2026" instead of "2026年6月", and only 3 of 7 weekday columns are visible. The calendar is effectively unusable.

repro

## Root cause

`UsageDateRangePicker` switched to its two-column layout (date fields | calendar) based on the **viewport** width via Tailwind `sm:` (640px): `sm:flex-row`, fixed `sm:w-[250px] sm:flex-none` fields column, `sm:w-[620px]` popover.

But the popover is clamped to `max-w-[calc(100vw-2rem)]` and anchored to its trigger with `align="end"`, so how much width it actually has is **not** the viewport width. As a result the wide two-column layout could activate while the popover only had room for one column, pushing the calendar off-screen.

## Fix

Gate the two-column layout on the **popover's own inline size** via a CSS container query instead of the viewport, and let the popover width shrink to fit:

- Popover is now `w-[620px] max-w-[calc(100vw-2rem)]` (always ≤ viewport).
- The layout switches to two columns only when the popover itself is wide enough (`@container (min-width: 500px)`); otherwise it stacks vertically so the calendar is always fully visible.
- The base **stacked** layout is the safe fallback for engines without container-query support (progressive enhancement).

Only `src/components/usage/UsageDateRangePicker.tsx` (class names) and `src/index.css` (container-query rules) change.

## Verification

- `tsc --noEmit` — passes
- `tailwindcss` compiles `index.css` — container query emitted correctly
- `vitest run tests/components/RequestLogTable.test.tsx` — passes
- Wide desktop (min window 900px) unchanged: still the 620px two-column layout.
- Reviewed by GPT-5.5 (xhigh): no correctness issues; container-query approach confirmed as the cleanest option and the stacked fallback confirmed safe.

### DIFF
--- a/src/components/usage/UsageDateRangePicker.tsx
+++ b/src/components/usage/UsageDateRangePicker.tsx
@@ -331,7 +331,7 @@ export function UsageDateRangePicker({
         </Button>
       </PopoverTrigger>
       <PopoverContent
-        className="w-[340px] max-w-[calc(100vw-2rem)] p-3 sm:w-[620px]"
+        className="usage-range-popover w-[620px] max-w-[calc(100vw-2rem)] p-3"
         align="end"
       >
         {/* Preset shortcuts */}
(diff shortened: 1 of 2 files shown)


farion1231 (approved):

PR #4923, opened 2026-07-02
fix(proxy): treat empty choices as valid empty response to unblock auto-mode classifier

BuWH:
## Problem

In `permission-mode=auto`, Claude Code blocks Bash/edit actions with:

> `claude-opus-4-8[1m] is temporarily unavailable, so auto mode cannot determine the safety of Bash right now. Wait briefly and then try this action again.`

This is the symptom reported in #1678, and also surfaces in #4689 and #1847. The root cause was previously undiagnosed — a contributor in #1678 noted they could only reproduce it by *deliberately* failing the classifier request, never through the normal proxy flow.

## Root cause

Claude Code's auto-mode classifier issues a **`max_tokens=1`** request before every side-effecting action (Bash, edits, etc.) to judge its safety. GitHub Copilot's Claude models sometimes produce **zero content tokens** for such a request and return `choices: []` with **HTTP 200** (a legal successful response — the model simply spent its 1-token budget without emitting content).

`openai_to_anthropic` treated an empty `choices` array as a fatal `TransformError("Empty choices array")`, which surfaces to the client as **422**. Claude Code interprets any classifier failure as "classifier unavailable" and locks all side-effecting actions in auto mode. Because whether the model emits a content token under a 1-token budget is non-deterministic, the lockup appears intermittent, which made it hard to reproduce.

## Reproduction (against a GitHub Copilot Claude provider)

```bash
curl -s -o /dev/null -w '%{http_code}\n' -X POST http://127.0.0.1:15721/v1/messages \
  -H 'content-type: application/json' -H 'x-api-key: PROXY_MANAGED' \
  -H 'anthropic-version: 2023-06-01' \
  -d '{"model":"claude-opus-4-8","max_tokens":1,"messages":[{"role":"user","content":"hi"}]}'
```

- **Before:** `422 {"error":{"message":"格式转换错误: Empty choices array"}}` (10/10 on my Copilot enterprise account)
- **After:** `200` with a valid empty-content message (10/10)

`max_tokens=2` already returns `200`, confirming the trigger is specifically the 1-token budget the classifier uses.

## Fix

An empty `choices` array is a **legal successful** upstream response, so degrade to a valid empty-content Anthropic message (`content: []`, `stop_reason: "end_turn"`) instead of erroring. A **missing** `choices` field (a genuinely malformed response) still errors as before. The usage-mapping logic is extracted into a shared `map_openai_usage_to_anthropic` helper so both the normal and empty-choices paths account tokens identically.

## Tests

- `test_openai_to_anthropic_empty_choices_returns_valid_empty_message` (new) — empty `choices` yields a valid 200-equivalent message
- `test_openai_to_anthropic_missing_choices_still_errors` (new) — missing `choices` field still errors
- All 10 existing `openai_to_anthropic` tests still pass; `cargo fmt --check` and `cargo clippy --lib` are clean.

Refs #1678, #4689, #1847

### DIFF
--- a/src-tauri/src/proxy/providers/transform.rs
+++ b/src-tauri/src/proxy/providers/transform.rs
@@ -512,6 +512,72 @@ pub fn clean_schema(mut schema: Value) -> Value {
     schema
 }
 
+/// 将 OpenAI 格式的 `usage` 映射为 Anthropic 格式的 usage 对象。
+///
+/// OpenAI `prompt_tokens` 含缓存命中，Anthropic `input_tokens` 不含 → 减去 cache_read 与
+/// cache_creation，使 input 成为 fresh input。本路径以 app_type="claude" 记账（calculator
+/// 不再扣减），若不减则缓存会被计入 input 与各 cache 桶两次。三桶互斥，恒等：
+/// input + cache_read + cache_creation == prompt_tokens（inclusive 上游）。
+/// 与流式 build_anthropic_usage_json (#2774) 及 transform_gemini 的 saturating_sub 对称。
+/// 最终 cache_read：直传字段优先于 nested；cache_creation 仅来自直传字段（OpenAI 无此概念）。
+fn map_openai_usage_to_anthropic(usage: Option<&Value>) -> Value {
+    let usage = usage.cloned().unwrap_or(json!({}));
+    let cached = usage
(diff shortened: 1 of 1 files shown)


farion1231:
@codex review

PR #4940, opened 2026-07-03
fix(codex): preserve plugin config on provider switch

kittors:
## Summary
- preserve Codex non-provider TOML config such as plugins, marketplaces, hooks, projects, and MCP servers when switching providers
- keep provider-owned Codex settings replaced by the selected provider
- preserve live plugin config when hot-switching while the proxy is managing Codex config

## Tests
- cargo fmt --check
- cargo test codex_config --lib
- cargo test services::proxy --lib


farion1231:
@codex review

PR #4988, opened 2026-07-04
fix: preserve provider credentials during edit and switch

htdxd:
## Summary
- Preserve DB-stored provider API keys and base URLs when editing active Claude/Codex providers so stale live settings cannot overwrite credentials.
- Keep Codex official OAuth auth material in auth.json/backup while projecting third-party provider keys into config.toml for requests.
- Harden Windows provider terminal launch against non-ASCII provider IDs and quoted batch paths.
- Add EditProviderDialog regression coverage for Claude/Codex credential preservation.

## Tests
- pnpm test:unit tests/components/EditProviderDialog.test.tsx
- pnpm typecheck
- cargo test --lib preserves_oauth_auth_json -- --nocapture
- cargo test --lib hot_switch_codex_provider_preserves_provider_model_provider_in_backup_and_restore -- --nocapture
- cargo test --lib update_live_backup_from_provider_preserves_codex_mcp_servers -- --nocapture
- cargo test --lib --no-run
- git diff --check


farion1231:
@codex review

PR #5012, opened 2026-07-05
fix(codex): normalize reasoning levels for generated catalog entries

leewaiho:
## Problem

When cc-switch generates the Codex model catalog (`~/.codex/cc-switch-model-catalog.json`), `supported_reasoning_levels` and `default_reasoning_level` were inherited directly from whichever template was loaded:

1. `models_cache.json` (created after connecting to OpenAI)
2. `codex debug models --bundled` CLI output
3. The bundled fallback `src-tauri/src/resources/gpt5_5_template.json`

This caused two issues:
- Different installs ended up with different reasoning level sets (e.g. 4 levels vs 6 levels) depending on which template was available.
- Custom/third-party providers had no reliable default, because the template was chosen before the provider specs were applied.

## Fix

In `codex_catalog_model_entry()`, override the template values and inject a consistent set of reasoning levels:

- `supported_reasoning_levels`: `[low, medium, high]`
- `default_reasoning_level`: `medium` (matches the official Codex bundled template default)
- `reasoning_levels`: `null` to avoid stale numeric mappings from the template

## Scope

This is intentionally minimal (only touches `src-tauri/src/codex_config.rs`) and includes unit tests for the injection and override behavior.

Fixes inconsistent reasoning level dropdowns for custom Codex providers.

### DIFF
--- a/src-tauri/src/codex_config.rs
+++ b/src-tauri/src/codex_config.rs
@@ -122,6 +122,13 @@ impl CodexCatalogToolProfile {
     }
 }
 
+/// Default reasoning levels injected into every cc-switch-generated Codex model
+/// catalog entry. Custom/third-party providers consistently expose low/medium/high
+/// regardless of which template (cache/bundled/static) was used to build the entry.
+const DEFAULT_CODEX_REASONING_LEVELS: &[(&str, &str)] =
+    &[("low", "低"), ("medium", "中"), ("high", "高")];
+const DEFAULT_CODEX_REASONING_LEVEL: &str = "medium";
+
 /// Reserved built-in provider IDs from OpenAI Codex's config/model-provider
 /// catalog. Keep in sync with Codex `RESERVED_MODEL_PROVIDER_IDS` and legacy
 /// removed provider aliases.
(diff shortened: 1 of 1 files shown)


farion1231:
@codex review

SaladDay:
Thanks for fixing the earlier compile issue. The current head is green, but the global override still looks unsafe for model capability data.

`codex_catalog_model_entry` is shared by every generated model and tool profile. The bundled GPT-5.5 template currently exposes `xhigh`, so replacing every entry with only `low`/`medium`/`high` removes a valid choice for GPT-compatible providers. Other models may support `none`, `minimal`, `max`, a different subset, or no reasoning effort at all; this patch would advertise the same three values for all of them. That can either hide a supported setting or let users select an effort the upstream rejects. The fixed Chinese descriptions (`低`/`中`/`高`) also replace the template text for every locale.

Could we preserve the template values by default, or add supported/default reasoning metadata to `CodexCatalogModelSpec` and override it only when a provider or model declares those capabilities? Tests that preserve the current `xhigh` entry and cover a model-specific override would catch the regression. 🙂

PR #5013, opened 2026-07-05
fix(codex): sync third-party provider auth on switch

QiJi11:
## Summary

- sync third-party Codex provider auth into live `auth.json` when switching providers
- keep official-provider auth handling separate so official login material still owns `auth.json`
- update the provider-switch regression test to expect the selected third-party key in live auth

## Why

When `preserveCodexOfficialAuthOnSwitch` is enabled, switching to a third-party Codex provider currently updates `config.toml` but leaves `auth.json` with the previous provider/login material. A newly launched Codex window can then keep using the stale key even though the selected provider changed.

Refs #4944.

## Tests

- Not run locally: a full clone of this repository repeatedly timed out in this environment, so this draft PR was prepared via GitHub API from the current `main` branch.
- Change is covered by updating `switch_provider_updates_codex_live_and_state`; maintainers should run `cd src-tauri && cargo test switch_provider_updates_codex_live_and_state` and the normal CI suite before merge.

### DIFF
--- a/src-tauri/src/codex_config.rs
+++ b/src-tauri/src/codex_config.rs
@@ -1473,11 +1473,12 @@ pub fn strip_codex_unified_session_bucket_from_settings(
     Ok(())
 }
 
-/// Route a Codex live write between full auth+config or config-only.
+/// Route a Codex live write between official login preservation and third-party provider sync.
 ///
 /// Official providers with usable login material own `auth.json`. Third-party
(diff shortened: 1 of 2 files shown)


QiJi11:
Closing this PR because its scope is wrong. The intended fix is not to change live auth write policy during provider switching; the actual requirement is per-window Codex provider snapshot isolation, so each newly launched Codex window uses the provider selected at launch time while already-running windows keep their original provider.

PR #5018, opened 2026-07-05
[codex] fix SQL backup import after file selection

### LINKED ISSUE #5017: SQL import requires a second click after selecting a backup file
Title: SQL import requires a second click after selecting a backup file

### Self Checks / 自检

### CC Switch Version / 版本号

3.16.5 / main

### Operating System / 操作系统

macOS

### Related App / 涉及应用

Other / 其他

### Steps to Reproduce / 重现步骤

1. Open Settings.
2. Go to Advanced -> Data Management -> SQL Import/Export.
3. Click the SQL import button.
4. Select a valid CC Switch SQL backup file in the file picker.

### Expected Behavior / 期望行为

After the user selects a SQL backup file, CC Switch should immediately start importing it and show the existing importing/progress state.

用户选择 SQL 备份文件后，应立即开始导入，并显示现有的导入中状态。

### Actual Behavior / 实际行为

After the file picker closes, the UI only changes the button into an "Import" confirmation state. The user must click the same button a second time before the import starts.

文件选择器关闭后，界面只是把按钮切换成“导入”确认状态。用户还需要再次点击同一个按钮，导入才会真正开始。

This is confusing because choosing a file from an import button usually confirms the user's intent to import that file. It also makes it easy to think the app has already started importing when nothing is running yet.

### Additional Context / 补充信息

I searched for related issues. Existing issue #2100 is about long-running/stuck SQL imports, while this report is specifically about the two-click interaction before import starts.

zqbake:
## Summary / 概述

This PR makes the Advanced Settings SQL import flow start immediately after the user selects a backup file.

Previously, the primary import button first opened the file picker and only changed into an import confirmation button after a file was selected. Users then had to click the same button a second time to actually import the SQL backup. The import button now opens the file picker when needed and imports the selected file in the same action.

## Related Issue / 关联 Issue

Fixes #5017

## Screenshots / 截图

N/A - behavior-only change. The existing settings UI style is preserved.

## Checklist / 检查清单

## Validation / 验证

- `./node_modules/.bin/prettier --check src/components/settings/ImportExportSection.tsx src/components/settings/SettingsPage.tsx src/hooks/useImportExport.ts tests/components/ImportExportSection.test.tsx tests/hooks/useImportExport.test.tsx tests/integration/SettingsDialog.test.tsx`
- `./node_modules/.bin/vitest run tests/hooks/useImportExport.test.tsx tests/hooks/useImportExport.extra.test.tsx tests/components/ImportExportSection.test.tsx tests/components/SettingsDialog.test.tsx tests/integration/SettingsDialog.test.tsx`
- `./node_modules/.bin/vitest run`

Note: `pnpm` script wrappers in this local environment were blocked by pnpm build-script approval prompts, so checks were run through the already-installed local binaries. Full unit tests passed: 65 files, 396 tests.

### DIFF
--- a/src/components/settings/ImportExportSection.tsx
+++ b/src/components/settings/ImportExportSection.tsx
@@ -17,7 +17,6 @@ interface ImportExportSectionProps {
   errorMessage: string | null;
   backupId: string | null;
   isImporting: boolean;
-  onSelectFile: () => Promise<void>;
   onImport: () => Promise<void>;
   onExport: () => Promise<void>;
   onClear: () => void;
@@ -29,7 +28,6 @@ export function ImportExportSection({
   errorMessage,
   backupId,
   isImporting,
-  onSelectFile,
   onImport,
   onExport,
   onClear,
@@ -61,7 +59,7 @@ export function ImportExportSection({
             <Button
               type="button"
               className={`w-full h-auto py-3 px-4 bg-blue-500 hover:bg-blue-600 dark:bg-blue-600 dark:hover:bg-blue-700 text-white ${selectedFile && !isImporting ? "flex-col items-start" : "items-center"}`}
-              onClick={!selectedFile ? onSelectFile : onImport}
+              onClick={onImport}
               disabled={isImporting}
             >
               <div className="flex items-center gap-2 w-full justify-center">
@@ -73,11 +71,7 @@ export function ImportExportSection({
                   <FolderOpen className="h-4 w-4 flex-shrink-0" />
                 )}
                 <span className="font-medium">
-                  {isImporting
-                    ? t("settings.importing")
-                    : selectedFile
-                      ? t("settings.import")
-                      : t("settings.selectConfigFile")}
+                  {isImporting ? t("settings.importing") : t("settings.import")}
                 </span>
               </div>
               {selectedFile && !isImporting && (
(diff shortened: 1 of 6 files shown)


farion1231:
@codex review

SaladDay:
Thanks for simplifying this flow! I found one race to address before merge 🙂

`isImporting` does not become true until after `openFileDialog()` resolves, so the import button stays enabled while the picker is opening. A fast double-click starts two dialog promises from the same render, and both paths can then call `importConfigFromFile`. If different backup files are selected, the final database depends on which restore finishes last; backups, syncs, and notifications are also duplicated.

I reproduced this with a hook test that calls `importConfig()` twice before the dialog promise resolves: `openFileDialog` was called twice. Please add a synchronous in-flight guard (a ref/lock works), acquire it before awaiting the picker, release it on cancel/error/completion, and add that concurrency case as a regression test. The current-`main` merge and the existing targeted tests otherwise pass.

zqbake:
@SaladDay Addressed in 54b14471.

- Added a synchronous in-flight ref lock before awaiting the file picker.
- Releases the lock on cancellation, picker errors, and import completion.
- Added a regression test covering two import calls before the picker resolves.

Validated with `pnpm typecheck` and `pnpm test:unit` (65 files, 397 tests).

zqbake:
@codex review

PR #5043, opened 2026-07-06
fix(proxy): always use direct connection for upstream provider requests

holyfun:
## Breaking Change

**Users who previously relied on system proxy environment variables (HTTPS_PROXY, HTTP_PROXY) to let CC Switch reach AI providers must now configure their proxy explicitly in CC Switch Settings → Proxy.**

Previously, when no explicit proxy was configured, CC Switch would implicitly inherit the system proxy from environment variables. After this PR, `build_client(None)` always calls `.no_proxy()`, and upstream provider requests use direct connection.

### Migration

```
Before: CC Switch inherits HTTPS_PROXY=http://127.0.0.1:7890 from system
After:  Configure http://127.0.0.1:7890 in CC Switch Settings → Proxy
```

Same proxy, same endpoint — just explicit instead of implicit.

## Problem

When no explicit proxy URL is configured, CC Switch's upstream HTTP client followed system proxy environment variables. This routed AI provider API calls through local proxy tools like v2rayN or Clash, creating a circular dependency: if the proxy tool is stopped or reconfigured, CC Switch can no longer reach AI providers, breaking all connected CLI tools.

## Root Cause

In `build_client()`, the `else` branch let reqwest follow system proxy env vars. The `system_proxy_points_to_loopback()` guard only prevented a self-loop to CC Switch's own port (15721), but did not protect against other local proxies.

## Fix

Always call `.no_proxy()` on the reqwest `Client` builder when no explicit upstream proxy is configured. Removed the now-unused `system_proxy_points_to_loopback`, `proxy_points_to_loopback`, and `get_proxy_port` functions.

Added a regression test (`test_build_client_none_ignores_system_proxy`) verifying that `build_client(None)` succeeds even when `HTTP_PROXY`/`HTTPS_PROXY` are set.

## Closes

- #4478 - localhost bypass
- #4642 - caches macOS system proxy settings  
- #1695 - NO_PROXY not supported
- #4562 - network switch 503
- #1264 - direct-connect mode feature request

### DIFF
--- a/src-tauri/src/proxy/http_client.rs
+++ b/src-tauri/src/proxy/http_client.rs
@@ -5,8 +5,6 @@
 
 use once_cell::sync::OnceCell;
 use reqwest::Client;
-use std::env;
-use std::net::IpAddr;
 use std::sync::RwLock;
 use std::time::Duration;
 
@@ -34,14 +32,6 @@ pub fn set_proxy_port(port: u16) {
     }
 }
 
-/// 获取 CC Switch 代理服务器的监听端口
-fn get_proxy_port() -> u16 {
-    CC_SWITCH_PROXY_PORT
-        .get()
-        .and_then(|lock| lock.read().ok())
-        .map(|port| *port)
-        .unwrap_or(15721) // 默认端口作为回退
-}
 
 /// 初始化全局 HTTP 客户端
 ///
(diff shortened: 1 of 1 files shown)


holyfun:
## Backward Compatibility Note

This change deliberately breaks one scenario: users who rely on the system proxy (HTTPS_PROXY env var) to let CC Switch reach AI providers, without explicitly configuring a proxy in CC Switch settings.

### Scenario

A user in a restricted network environment has a system proxy (v2rayN / Clash) running. Their AI provider (e.g. OpenAI) is only reachable through the proxy. Currently, CC Switch implicitly inherits the system proxy via HTTPS_PROXY and works. After this PR, it would stop working.

### Why this trade-off is acceptable

1. **Explicit > Implicit.** Relying on system env vars for CC Switch internal API calls is an implicit, poorly-documented behavior. The correct model is explicit configuration (as requested in #1264). Users who need proxy access should configure it in CC Switch settings, which is a one-time, discoverable action.

2. **The implicit behavior is fragile.** The same env var that lets CC Switch reach AI providers also breaks it when the proxy tool is stopped (as reported in #4642, #4478, #4562). The current code's `system_proxy_points_to_loopback()` guard only prevents one specific self-loop case (port 15721), leaving all other failure modes open.

3. **Explicit proxy configuration already exists and is unaffected.** When a user sets a proxy URL in CC Switch settings, `build_client()` takes the `if let Some(url)` branch and calls `builder.proxy(proxy)` — completely independent of this change.

4. **The mitigation is straightforward.** Affected users need to configure their proxy once in Settings → Proxy. This is a one-time action that survives restarts and is more robust than relying on environment variable inheritance.

### Migration path for affected users

```
Before (implicit):  CC Switch inherits HTTPS_PROXY=http://127.0.0.1:7890 from system
After (explicit):   Configure http://127.0.0.1:7890 in CC Switch Settings → Proxy
```

Same proxy, same endpoint, just configured explicitly instead of inherited implicitly.

farion1231:
@codex review

PR #5064, opened 2026-07-07
fix(proxy): classify glm-5.2 as text-only in media sanitizer heuristic

heekei:
﻿## What

Adds `glm-5.2` to the curated `known_text_only_model` `EXACT_TAILS` list in `src-tauri/src/proxy/media_sanitizer.rs`, alongside the existing `glm-5.1` entry.

## Why

GLM-5.2 is a text-only model like GLM-5.1. Without it on the built-in list, when a provider has no explicit model-capability declaration (`modelCatalog` / `modalities`), images sent to a `glm-5.2` endpoint are only stripped *reactively* (after the upstream returns an unsupported-image error) instead of *predictively* before send. This drops GLM-5.2 into the pre-send heuristic path so images are stripped up front.

This list is the lowest-priority fallback (behind explicit provider capability declarations), so it only ever runs when the user has "Heuristic Text-Only Model Detection" enabled **and** the provider hasn't declared capabilities.

## Changes

- Add `glm-5.2` to `EXACT_TAILS` in `known_text_only_model`.
- Regression test `glm_models_are_classified_text_only` covering `glm-5.1`, mixed-case `GLM-5.2`, and an aggregator vendor-path prefix (`zai/GLM-5.2`).

## Verification

```
cd src-tauri
cargo test --lib proxy::media_sanitizer::   # 22 passed, 0 failed
```

### DIFF
--- a/src-tauri/src/proxy/media_sanitizer.rs
+++ b/src-tauri/src/proxy/media_sanitizer.rs
@@ -250,6 +250,7 @@ fn known_text_only_model(model: &str) -> bool {
         "deepseek-v4-flash",
         "deepseek-v4-pro",
         "glm-5.1",
+        "glm-5.2",
         "kat-coder",
         "kat-coder-pro",
         "kat-coder-pro v1",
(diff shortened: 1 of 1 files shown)


heekei:
@farion1231

heekei:
@codex review

PR #5092, opened 2026-07-08
fix(provider): show and persist API Key when editing uncategorized providers

### LINKED ISSUE #5041: Bug: Claude provider edit mode — API Key input missing when category is undefined
### Self Checks / 自检

### CC Switch Version / 版本号

3.16.1

### Operating System / 操作系统

Windows

### Related App / 涉及应用

- Claude Code

### Steps to Reproduce / 重现步骤

1. Create a Claude provider (third-party/custom) **without** filling in API Key, or import a provider whose `settingsConfig` has no `ANTHROPIC_AUTH_TOKEN` / `ANTHROPIC_API_KEY` field.
2. Save it, then click **Edit** on that provider again.
3. The API Key input field **does not appear**.
4. For comparison: edit a Codex provider — the API Key input field shows up normally.

### Expected Behavior / 期望行为

When editing a non-official Claude provider, the API Key input should always be shown even if the config doesn't currently have an API Key field, so users can add one.

### Actual Behavior / 实际行为

The API Key input is hidden, making it impossible to add an API Key in edit mode.

### Root Cause Analysis

`src/components/providers/forms/hooks/useApiKeyState.ts`

**Problem 1 — `showApiKey` function**

```typescript
if (isEditMode) {
  if (category !== undefined && category !== "official") return true;  // bug here
  return hasApiKeyField(config, appType);
}
```

The condition `category !== undefined` prevents providers with **undefined** category from showing the API Key input.

Many providers have `Provider.category` set to `undefined` (created via import, older versions, or manual JSON editing). The `useProviderCategory` hook uses `initialData?.category` directly in edit mode — if the original data lacks this field, `category` is `undefined`.

**Problem 2 — `handleApiKeyChange` `createIfMissing`**

```typescript
createIfMissing: category !== undefined && category !== "official",
```

Same `category !== undefined` condition — even if the input were shown, typed values wouldn't be written to config.

**Why Codex works**: `CodexFormFields.tsx` renders `ApiKeySection` unconditionally, bypassing this logic entirely.

### Suggested Fix / 建议修复

Change `category !== undefined && category !== "official"` to `category !== "official" && category !== "cloud_provider"` in both places:

**1. `showApiKey`:**
```typescript
if (isEditMode) {
  if (category !== "official" && category !== "cloud_provider") return true;
  return hasApiKeyField(config, appType);
}
```

**2. `createIfMissing`:**
```typescript
createIfMissing: category !== "official" && category !== "cloud_provider",
```

After the fix:
- `undefined` (uncategorized) -> shows API Key (treated as custom)
- `"custom"` / `"third_party"` / `"aggregator"` etc -> shows
- `"official"` (OAuth) -> only shows if config already has the field
- `"cloud_provider"` (Bedrock/IAM) -> only shows if config already has the field

TypeScript compiles cleanly.

### Additional Context / 补充信息

Also encountered a database schema version mismatch: local DB is at schema version 11 but code has `SCHEMA_VERSION = 10`, preventing app startup. Temporary workaround: delete `~/.cc-switch/cc-switch.db` to let the app recreate it.

fengshao1227:
## Summary / 概述

When editing a Claude (or Gemini) provider whose `category` is `undefined` — which is the case for providers created via import, older app versions, or manual JSON edits — and whose config does not yet contain an API key field (`ANTHROPIC_AUTH_TOKEN` / `ANTHROPIC_API_KEY`), the edit form gives the user no way to add a key:

1. `showApiKey` in `useApiKeyState` only rendered the input in edit mode when `hasApiKeyField(config)` was already true, so the input never appeared.
2. Even when the input was visible, `createIfMissing` for `setApiKeyInConfig` required `selectedPresetId !== null` (add mode only) plus `category !== undefined`, so a typed key was silently dropped and never written into the config.

Codex is unaffected because `CodexFormFields` renders `ApiKeySection` unconditionally — which is also why the issue only reproduces on the Claude form.

This PR treats every category except `official` and `cloud_provider` (including `undefined`) as custom-like in `useApiKeyState`:

- **`showApiKey`**: in edit mode, always show the input for custom-like categories; `official` / `cloud_provider` keep the previous behavior (only shown when the config already has the field).
- **`createIfMissing`**: create the missing key field for custom-like categories (respecting the configured `apiKeyField`); never create for `official` / `cloud_provider`.

OAuth-based presets (GitHub Copilot / Codex OAuth) are unaffected: `ClaudeFormFields` independently hides the input via `!usesOAuth`, and the `cloud_provider` guard at the `ProviderForm` call site is preserved.

Added `tests/hooks/useApiKeyState.test.tsx` (11 cases) covering visibility and persistence for `custom` / `undefined` / `official` / `cloud_provider` categories, including the `apiKeyField` override and the official-provider "update existing field only" path.

## Related Issue / 关联 Issue

Fixes #5041

## Screenshots / 截图

N/A — behavior change is covered by unit tests (input visibility & config writes in edit mode).

## Checklist / 检查清单

- [x] `cargo clippy` passes (if Rust code changed) / 通过 Clippy 检查（如修改了 Rust 代码）— no Rust changes
- [x] Updated i18n files if user-facing text changed / 如修改了用户可见文本，已更新国际化文件 — no user-facing text changed

`pnpm test:unit`: 66 files / 408 tests passing.

### DIFF
--- a/src/components/providers/forms/hooks/useApiKeyState.ts
+++ b/src/components/providers/forms/hooks/useApiKeyState.ts
@@ -60,15 +60,13 @@ export function useApiKeyState({
         initialConfig || "{}",
         key.trim(),
         {
-          // 最佳实践：仅在"新增模式"且"非官方类别"时补齐缺失字段
-          // - 新增模式：selectedPresetId !== null
-          // - 非官方类别：category !== undefined && category !== "official"
-          // - 官方类别：不创建字段（UI 也会禁用输入框）
-          // - 未传入 category：不创建字段（避免意外行为）
+          // 仅在"非官方/非云服务商类别"时补齐缺失字段
+          // - official：走 OAuth/订阅登录，不创建字段（UI 也会禁用输入框）
+          // - cloud_provider：走顶层 apiKey 或 IAM，不创建 env 字段
+          // - undefined（导入/旧版本数据未分类）：视为自定义，允许创建，
+          //   否则编辑模式下输入的 API Key 不会写入配置（#5041）
           createIfMissing:
-            selectedPresetId !== null &&
-            category !== undefined &&
-            category !== "official",
+            category !== "official" && category !== "cloud_provider",
           appType,
           apiKeyField,
         },
@@ -88,12 +86,17 @@ export function useApiKeyState({
 
   const showApiKey = useCallback(
     (config: string, isEditMode: boolean) => {
-      return (
-        selectedPresetId !== null ||
-        (isEditMode && hasApiKeyField(config, appType))
-      );
+      if (selectedPresetId !== null) return true;
+      if (!isEditMode) return false;
+      // 编辑模式：非官方/非云服务商类别（含 undefined 的导入/旧版本数据）
+      // 始终显示输入框，便于补填缺失的 API Key（#5041）；
+      // official / cloud_provider 仅在配置已有对应字段时显示
+      if (category !== "official" && category !== "cloud_provider") {
+        return true;
+      }
+      return hasApiKeyField(config, appType);
     },
-    [selectedPresetId, appType],
+    [selectedPresetId, category, appType],
   );
 
   return {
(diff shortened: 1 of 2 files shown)


farion1231:
@codex review

PR #5093, opened 2026-07-08
fix(proxy): ignore empty finish_reason in OpenAI SSE streaming conversion

### LINKED ISSUE #5087: SenseNova reasoning_content streaming returns empty finish_reason causing Claude Desktop assistant message disappear
### Self Checks / 自检

### CC Switch Version / 版本号

3.16.5

### Operating System / 操作系统

Windows

### Related App / 涉及应用

Other / 其他

### Steps to Reproduce / 重现步骤

CC Switch Version:
3.16.5

OS:
Windows

Provider:
SenseNova

API Format:
OpenAI Chat Completions

Endpoint:
https://token.sensenova.cn/v1/chat/completions

Model:
deepseek-v4-flash

Client:
Claude Desktop

### Expected Behavior / 期望行为

{
 "model":"deepseek-v4-flash",
 "messages":[
  {
   "role":"user",
   "content":"你好"
  }
 ],
 "stream":true
}
Expected:
Claude Desktop should receive a complete assistant message and save it into conversation history.

### Actual Behavior / 实际行为

SenseNova returns streaming chunks containing:

{
  "delta": {
    "reasoning_content": "..."
  },
  "finish_reason": ""
}

CC Switch logs:

Unknown finish_reason in streaming:

After generation completes:
- response appears temporarily in Claude Desktop
- assistant message disappears
- conversation history contains only user message

### Additional Context / 补充信息

Additional investigation:

1. SenseNova stream=false works correctly.

Non-stream response:
{
  "message": {
    "content": "...",
    "reasoning_content": "..."
  },
  "finish_reason": "tool_calls"
}

2. SenseNova stream=true normal text generation works:

Final chunk:
{
  "delta": {
    "content": ""
  },
  "finish_reason": "stop"
}

3. The issue only occurs when streaming reasoning_content.

Example chunk:

{
  "delta": {
    "reasoning_content": "想知道"
  },
  "finish_reason": ""
}

It seems CC Switch does not handle empty string finish_reason in reasoning streaming correctly.

fengshao1227:
## Summary / 概述

SenseNova (and other Chat Completions upstreams) attach `"finish_reason": ""` (an **empty string**, not `null`) to every streaming chunk while emitting `reasoning_content`. In the OpenAI→Anthropic SSE converter (`create_anthropic_sse_stream`), any `Some(finish_reason)` was treated as end-of-generation, so on the **first** reasoning chunk the converter:

1. logged `Unknown finish_reason in streaming:` (the exact log in the issue),
2. closed the just-opened `thinking` block and cached the `message_delta`,
3. set `has_emitted_message_delta`, which then swallowed the **real** `finish_reason` on the final chunk.

Every content block opened after that first chunk was never closed, so the emitted Anthropic SSE stream ended with unclosed `content_block`s before `message_stop`. Strict clients such as Claude Desktop discard the whole assistant message in that case — matching the reported symptom (the response renders during streaming, then disappears and never enters history).

### Fix

- Treat empty/whitespace-only `finish_reason` as absent instead of a stop signal (the actual bug).
- Safety net on `[DONE]`: close any still-open content blocks, and synthesize a `message_delta` (`end_turn` + latest usage) if no valid `finish_reason` ever arrived, so the output stream is always structurally complete even against non-conforming upstreams.
- The same block-closing safety net on natural stream end (no `[DONE]`), only when a pending `message_delta` is flushed — the existing "don't fake success" semantics for truncated streams are preserved.

Normal streams are unaffected: the `finish_reason` handler already closes all blocks, so the safety nets are no-ops.

### Tests

Two regression tests in `streaming.rs` (reusing the existing `collect_anthropic_events` helper):

- `test_streaming_ignores_empty_finish_reason_in_reasoning_chunks` — SenseNova-shaped stream (`reasoning_content` + `finish_reason:""` on every chunk, real `stop` at the end): asserts a single `thinking` block, exactly one `message_delta` with `stop_reason: end_turn`, and every `content_block_start` matched by a `content_block_stop` before `message_stop`.
- `test_streaming_closes_blocks_when_finish_reason_never_arrives` — upstream never sends a valid `finish_reason`: asserts the `[DONE]` fallback closes all blocks and synthesizes the `message_delta`.

## Related Issue / 关联 Issue

Fixes #5087

## Screenshots / 截图

N/A — backend proxy change, covered by unit tests.

## Checklist / 检查清单

- [x] `pnpm typecheck` passes / 通过 TypeScript 类型检查 — no frontend changes
- [x] Updated i18n files if user-facing text changed / 如修改了用户可见文本，已更新国际化文件 — no user-facing text changed

## Verification / 验证

- `cargo fmt --check`
- `cargo clippy --lib`
- `cargo test --lib` — 1755 passed, 0 failed (includes the 2 new regression tests)

### DIFF
--- a/src-tauri/src/proxy/providers/streaming.rs
+++ b/src-tauri/src/proxy/providers/streaming.rs
@@ -186,6 +186,46 @@ pub fn create_anthropic_sse_stream<E: std::error::Error + Send + 'static>(
                                 if data.trim() == "[DONE]" {
                                     log::debug!("[Claude/OpenRouter] <<< OpenAI SSE: [DONE]");
 
+                                    // 兜底：若上游从未发送有效 finish_reason（或 finish 之后又有新的
+                                    // content block 打开），在 message_stop 之前关闭仍打开的 block，
+                                    // 保证输出的 Anthropic SSE 结构完整。正常流程中 finish_reason
+                                    // 处理已关闭全部 block，此处为空操作。
+                                    if let Some(index) = current_non_tool_block_index.take() {
+                                        let event = json!({
+                                            "type": "content_block_stop",
+                                            "index": index
+                                        });
+                                        let sse_data = format!("event: content_block_stop\ndata: {}\n\n",
+                                            serde_json::to_string(&event).unwrap_or_default());
+                                        yield Ok(Bytes::from(sse_data));
+                                    }
+                                    current_non_tool_block_type = None;
+                                    if !open_tool_block_indices.is_empty() {
+                                        let mut tool_indices: Vec<u32> =
+                                            open_tool_block_indices.iter().copied().collect();
+                                        tool_indices.sort_unstable();
+                                        for index in tool_indices {
+                                            let event = json!({
+                                                "type": "content_block_stop",
+                                                "index": index
+                                            });
+                                            let sse_data = format!("event: content_block_stop\ndata: {}\n\n",
+                                                serde_json::to_string(&event).unwrap_or_default());
+                                            yield Ok(Bytes::from(sse_data));
+                                        }
+                                        open_tool_block_indices.clear();
+                                    }
+
+                                    // 兜底：消息已开始但从未收到有效 finish_reason 时，合成一个
+                                    // message_delta，避免客户端收到没有 stop_reason 的消息流。
+                                    if pending_message_delta.is_none()
+                                        && !has_emitted_message_delta
+                                        && has_sent_message_start
+                                    {
+                                        pending_message_delta =
+                                            Some((Some("end_turn".to_string()), latest_usage.clone()));
+                                    }
+
                                     // 流正常结束，发出缓存的 message_delta（含完整 usage）。
                                     if let Some((stop_reason, usage_json)) = pending_message_delta.take() {
                                         let event = build_message_delta_event(stop_reason, usage_json);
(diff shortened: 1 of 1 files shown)


farion1231:
@codex review

fengshao1227:
Thanks for the catch — addressed in 09d22475.

The fallback now derives `stop_reason` from `tool_blocks_by_index`:

- **tool blocks present** → `tool_use` (so the client dispatches tool execution)
- **no tool blocks** → `end_turn` (text-only fallback, same as before)

Added `test_streaming_synthesizes_tool_use_stop_reason_when_tools_present` to lock in the behavior.

`cargo test --lib` — 1754 passed, 0 failed.

SaladDay:
Thanks, the empty/whitespace filtering is correct, and the fallback now preserves `stop_reason: tool_use` for normal tool-call chunks. There is still one gap to fix, though 🙂

When an OpenAI-compatible upstream sends a tool call with `name`/`arguments` but omits its `id`, then ends with `[DONE]` without any valid `finish_reason`, `tool_blocks_by_index` is non-empty so the fallback emits `stop_reason: tool_use`. However, this path never runs the existing late-tool-start flush. The converted stream therefore contains `message_start`, `message_delta(tool_use)`, and `message_stop`, but no `tool_use` content block, leaving the client with nothing to execute. The normal non-empty `finish_reason` path supports this input by synthesizing `tool_call_{index}` and flushing the pending arguments.

I reproduced this against the latest `main` merge. Please reuse that late-start/pending-args logic before synthesizing the terminal delta, and add a regression test for a named tool call without an upstream `id` plus an empty or missing `finish_reason`. The other seven focused streaming tests pass.

fengshao1227:
@SaladDay Thanks for the detailed analysis — you're right, the gap is real. Addressed in c1f926b8.

The `[DONE]` fallback now runs the same late-tool-start flush as the normal `finish_reason` path before closing blocks:

1. Iterate `tool_blocks_by_index` for blocks that haven't `started`
2. Synthesize `tool_call_{index}` for missing ids, emit `content_block_start` + pending arguments via `input_json_delta`
3. Insert into `open_tool_block_indices` so the subsequent close loop picks them up

Added `test_streaming_late_tool_start_flush_on_done_without_finish_reason` — tool call with name + arguments but no upstream id, empty `finish_reason`, verifying:
- Exactly one `tool_use` `content_block_start` with synthesized id and correct name
- Arguments emitted via `input_json_delta`
- All blocks closed before `message_stop`
- `stop_reason: tool_use`

`cargo test --lib` — 1755 passed, 0 failed.

PR #5094, opened 2026-07-08
fix(webdav): raise skills.zip restore entry limit to 100k

### LINKED ISSUE #5024: 从云端还原备份时，提示skill超过1w个条目，导致无法还原
### Self Checks / 自检

### Related App / 涉及应用

General / 通用

Image

### Proposed Solution / 建议方案

_No response_

### Additional Context / 补充信息

_No response_

### Contribution / 参与贡献

我有兴趣参与开发此功能。

fengshao1227:
## Summary / 概述

Restoring a skills backup from WebDAV fails with `skills.zip has too many entries (..), limit is 10000` once the skill collection exceeds 10,000 files. The limits are asymmetric: `zip_skills_ssot` applies **no entry-count limit when packing**, so CC Switch happily backs up a large skill library and then refuses to restore its own artifact. Large skill collections (skills that ship whole repos) easily exceed 10k files.

The actual zip-bomb protection in `restore_skills_zip` is the total-extraction-size cap (`MAX_SYNC_ARTIFACT_BYTES`, 512 MB) enforced per-chunk during extraction via `copy_entry_with_total_limit`; the entry-count check only guards against pathological many-tiny-entry archives.

### Fix

- Raise `MAX_EXTRACT_ENTRIES` from `10_000` to `100_000` so it remains a safety net without rejecting legitimate skill libraries. The 512 MB extraction cap stays untouched as the primary bomb guard.
- Extract the check into `ensure_extract_entry_count` and add boundary tests (10k+1 and the new limit are accepted; limit+1 is rejected with the localized error).

## Related Issue / 关联 Issue

Fixes #5024

## Screenshots / 截图

N/A — backend limit change, covered by unit tests.

## Checklist / 检查清单

- [x] `pnpm typecheck` passes / 通过 TypeScript 类型检查 — no frontend changes
- [x] Updated i18n files if user-facing text changed / 如修改了用户可见文本，已更新国际化文件 — error strings are localized inline (zh/en) as before

## Verification / 验证

- `cargo fmt --check`
- `cargo clippy --lib`
- `cargo test --lib` — 1755 passed, 0 failed (includes the 2 new boundary tests)

### DIFF
--- a/src-tauri/src/services/webdav_sync/archive.rs
+++ b/src-tauri/src/services/webdav_sync/archive.rs
@@ -15,7 +15,12 @@ use crate::services::sync_protocol::{
 };
 
 /// Maximum number of entries allowed in a zip archive.
-const MAX_EXTRACT_ENTRIES: usize = 10_000;
+///
+/// 打包侧（`zip_skills_ssot`）没有条目数限制，真正的 zip-bomb 防线是
+/// `MAX_SYNC_ARTIFACT_BYTES` 的解压总量上限；此处的条目数只用于拦截
+/// 病态的海量小文件压缩包。大型 skill 集合（如包含完整仓库的 skills）
+/// 很容易超过 1 万个文件，因此上限需明显高于正常使用规模（#5024）。
+const MAX_EXTRACT_ENTRIES: usize = 100_000;
 
 pub(crate) struct SkillsBackup {
     _tmp: TempDir,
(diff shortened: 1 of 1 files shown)


farion1231:
@codex review

SaladDay:
Thanks for tackling the >10k restore failure. I think the new limit needs one more resource guard before this is safe to merge 🙂 `MAX_SYNC_ARTIFACT_BYTES` only counts file payload bytes. For example, 100,000 directory entries count as zero bytes, while 100,000 one-byte files count as only about 100 KB. Restore still creates every entry and then `copy_dir_recursive` creates or copies all of them again, so the 512 MB counter does not bound inode use, allocated disk space, or filesystem work.

Could you keep a separately justified entry budget, or charge a fixed per-entry cost, and enforce the same count while packing? `zip_skills_ssot` currently has no entry limit, so the app can still upload a self-generated 100,001-entry backup that it later refuses to restore. An end-to-end test using a real ZIP would also help here, especially around the 65,535/65,536 ZIP64 boundary; the current tests only exercise the integer comparison.

fengshao1227:
@SaladDay Good points — addressed in 7a59d9fe.

Changes:

1. **Symmetric entry limit**: `zip_skills_ssot` now counts entries during packing via `zip_dir_recursive` and aborts at `MAX_ENTRIES`, so the app can no longer produce a backup it refuses to restore. Renamed `MAX_EXTRACT_ENTRIES` → `MAX_ENTRIES` and `ensure_extract_entry_count` → `ensure_entry_count` to reflect shared usage.

2. **End-to-end ZIP tests**: two tests using real ZIP archives built with the `zip` crate:
   - `restore_rejects_zip_exceeding_entry_limit` — builds a ZIP with `MAX_ENTRIES + 1` files, parses it with `ZipArchive`, asserts `ensure_entry_count` rejects it
   - `restore_accepts_zip_at_entry_limit` — builds a ZIP with exactly `MAX_ENTRIES` files, asserts it passes

Re: the ZIP64 boundary test — constructing 65,536 real ZIP entries in-process takes ~400ms and produces ~3MB of heap; I kept the test at the `MAX_ENTRIES` boundary (100k) which inherently crosses the ZIP64 threshold. If you'd prefer a separate 65,535/65,536 boundary test I'm happy to add it.

`cargo test --lib` — 1755 passed, 0 failed.

PR #5098, opened 2026-07-08
fix: sync openclaw and hermes live provider updates

allenxu09:
## Summary
- sync existing OpenClaw live providers back into the database instead of skipping them on startup
- apply the same live-import update behavior to Hermes providers
- add regression tests for edited OpenClaw and Hermes live provider configs

## Root cause
OpenClaw and Hermes are additive-mode clients, but their startup live import still skipped any provider whose ID already existed in the database. That meant external edits to `openclaw.json` or `config.yaml` never backfilled into stored provider settings after restart.

## Notes
- OpenClaw preserves the existing CC Switch display name while refreshing `settings_config`
- Hermes preserves the existing CC Switch display name while refreshing `settings_config`
- startup logs now say `Synced` / `No ... changes` for both apps to match the new behavior

### DIFF
--- a/src-tauri/src/lib.rs
+++ b/src-tauri/src/lib.rs
@@ -696,9 +696,9 @@ pub fn run() {
             }
             match crate::services::provider::import_hermes_providers_from_live(&app_state) {
                 Ok(count) if count > 0 => {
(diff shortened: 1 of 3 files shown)


farion1231 (approved):

Issue #4581, opened 2026-06-24
MacOS Memory Leak


xiaoyumuxi:
### Self Checks / 自检

- [x] I have read the [FAQ](https://github.com/farion1231/cc-switch#faq) section in README.
我已阅读 README 中的[常见问题](https://github.com/farion1231/cc-switch#常见问题)。

- [x] I have searched for [existing issues](https://github.com/farion1231/cc-switch/issues), including closed ones.
我已搜索过[已有的 Issue](https://github.com/farion1231/cc-switch/issues)，包括已关闭的。

### CC Switch Version / 版本号

3.16.3

### Operating System / 操作系统

macOS

### Related App / 涉及应用

Other / 其他

### Steps to Reproduce / 重现步骤

因为可能需要长时间启动目前没有进行复现操作，可以尝试在3.16.3 版本长时间运行 CC-switch 来看内存占用问题

### Expected Behavior / 期望行为

内存占用正常，稳定在一个合理的 200MB 内，因为以前都是随开随用的，但是这一次忘记退出进程然后发现了这个问题，还是希望解决一下，猜测应该就是内存泄露问题

### Actual Behavior / 实际行为

<img width="602" height="267" alt="Image" src="https://github.com/user-attachments/assets/939f84b6-5105-435c-aa9d-3f722ab4a19b" />
CC-switch  WebKit 子进程启动时间：2026-06-21 20:56，已经跑了 3 天多
磁盘缓存很小，但是运行时 WebView 内存没有释放，猜测 WebKit 在处理 RemoteLayerBackingStoreCollection / JavaScriptCore 分配等方面出现了内存泄露的问题

<img width="748" height="114" alt="Image" src="https://github.com/user-attachments/assets/cc5133b6-046e-4857-9e6f-19fd84b3983b" />
重启之后内存占用恢复正常

### Additional Context / 补充信息

[ccswitch-webcontent-80704.sample_副本.txt](https://github.com/user-attachments/files/29275765/ccswitch-webcontent-80704.sample_.txt)
这个是当时异常内存占用时的采样数据

<img width="1549" height="654" alt="Image" src="https://github.com/user-attachments/assets/2e678782-358a-4249-909a-34969988c2e6" />
这个是 Codex 给的可能分析

xiaoyumuxi:
<img width="619" height="308" alt="Image" src="https://github.com/user-attachments/assets/14bb0a0a-bc0b-44a4-a5be-385e84a5d719" />
[...]

PR #5105, opened 2026-07-08
fix(sessions): cap transcript payloads and evict cached messages on close

jsiu93:
## Summary / 概述

Two complementary renderer-memory hygiene fixes for the Session Manager, found while investigating WebView memory growth (context in #4693):

**Rust — cap per-message content at extraction time.** `load_messages` previously shipped each message's extracted text verbatim to the WebView; a single `tool_result` / `function_call_output` can embed a multi-MB dump. Content is now capped at 10,000 chars (UTF-8-boundary safe, `… [truncated]` marker) at every `SessionMessage` construction site across all providers (claude, codex, gemini, hermes, openclaw, opencode). The UI already collapses messages above 3,000 chars, so nothing readable is lost in practice.

**Frontend — `gcTime: 0` on the `sessionMessages` query.** Each viewed transcript previously stayed in the react-query cache for the default 5 minutes after its last observer unmounted, so browsing N sessions kept N full transcripts alive even after closing the page. Transcripts are now dropped as soon as the view unmounts.

Measured on the largest real session files on my machine (probe: sum of `SessionMessage.content` bytes returned by `load_messages`):

| File | Raw JSONL | Messages | Content before | Content after |
|---|---:|---:|---:|---:|
| Codex session | 61.9 MB | 3,043 | 4.21 MB | 3.60 MB |
| Claude Code session | 25.9 MB | 1,021 | 353.4 KB | 353.3 KB |

Scope note: this PR bounds worst-case transcript payloads and stops closed views from pinning transcripts in cache. The dominant renderer-memory issue remains the hidden-in-tray WebView (#4693) — measured on an official v3.16.4 build sitting in the tray with zero interaction, the WebContent process grew from 344 MB to 415 MB physical footprint (+64 MB dirty WebKit malloc) over ~90 minutes after one Usage Statistics + Session Manager browsing session. WebKit's allocator keeps those pages resident, so allocation peaks become permanent RSS — lowering the peaks (this PR) and destroying the WebView when backgrounded (#4693) complement each other.

## Related Issue / 关联 Issue

Related to #4693, #4581.

## Checklist / 检查清单

- [ ] Updated i18n files if user-facing text changed — note: the truncation marker `… [truncated]` is plain text inside message content; happy to wire it through i18n if maintainers prefer.

Verification run locally: `cargo fmt --check`, `cargo clippy --all-targets -- -D warnings`, `cargo test session_manager` (64 passed), `tsc --noEmit`, `prettier --check`, `vitest run` (405 tests / 65 files passed).

### DIFF
--- a/src-tauri/src/session_manager/providers/claude.rs
+++ b/src-tauri/src/session_manager/providers/claude.rs
@@ -8,8 +8,8 @@ use crate::config::get_claude_config_dir;
 use crate::session_manager::{SessionMessage, SessionMeta};
 
 use super::utils::{
-    extract_text, parse_timestamp_to_ms, path_basename, read_head_tail_lines, truncate_summary,
-    TITLE_MAX_CHARS,
+    extract_text, parse_timestamp_to_ms, path_basename, read_head_tail_lines,
+    truncate_message_content, truncate_summary, TITLE_MAX_CHARS,
 };
 
 const PROVIDER_ID: &str = "claude";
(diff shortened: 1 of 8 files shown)


farion1231:
@codex review

PR #5123, opened 2026-07-08
fix(proxy): accept x-api-key in Claude Desktop gateway auth

John1Tang:
## Summary

The Claude desktop app sends the gateway token as `x-api-key: <token>` (per Anthropic's published auth header convention), but `validate_claude_desktop_gateway_auth` in `src-tauri/src/proxy/handlers.rs` only accepted it via `Authorization: Bearer <token>`. As a result, every request from Code mode was rejected with 401 and Code mode showed "Request failed · retrying" on the first message — while Cowork mode and the bare `claude` CLI worked because they don't go through `/claude-desktop`.

This PR makes the gateway accept either header. `Authorization: Bearer` continues to work for clients that send it; the desktop app's `x-api-key` form is now accepted too.

## Repro

1. Install cc-switch, point a Claude-app (or Claude-Desktop-app) provider at any third-party upstream
2. In the Claude desktop app, switch to **Code** mode and start a new session
3. Type `hi`
4. Observe: 401 from the gateway; UI shows "Request failed · retrying (5/10) · 14s" and the same error on every retry. Auto-compact also fails the same way, producing the "Compacting conversation · 25s" loop.

Curl probe (expected on the broken version, fixed after this PR):

```bash
$ curl -i -X POST http://127.0.0.1:15721/claude-desktop/v1/messages \
  -H "x-api-key: ccs-..." \
  -H "anthropic-version: 2023-06-01" \
  -d '{"model":"claude-sonnet-5","max_tokens":16,"messages":[{"role":"user","content":"hi"}]}'
HTTP/1.1 401 Unauthorized
{"error":{"message":"认证失败: Claude Desktop gateway 缺少 Authorization 头","type":"proxy_error"}}
```

After this PR, the same probe returns 200.

## Fix

- Extract the token via a new pure helper `extract_gateway_token` that accepts either header. The validation function now calls the helper and compares to the expected token.
- `Authorization` is preferred when both are present (more specific), so other clients behave unchanged.
- For `Authorization`, the helper still strips `Bearer ` / `bearer ` prefix; if no prefix is present, the whole value is treated as the token (lenient parsing, matches the previous behavior for unexpected inputs).
- The error message is updated to mention both headers.

## Tests

Seven new unit tests cover the helper:
- `extract_gateway_token_accepts_authorization_bearer`
- `extract_gateway_token_accepts_lowercase_bearer`
- `extract_gateway_token_accepts_authorization_without_prefix`
- `extract_gateway_token_accepts_x_api_key`
- `extract_gateway_token_prefers_authorization_over_x_api_key`
- `extract_gateway_token_trims_whitespace`
- `extract_gateway_token_returns_none_when_no_auth_header`

## Files

- `src-tauri/src/proxy/handlers.rs` — patch + helper + tests
- `CHANGELOG.md` — Unreleased entry

## Backward compatibility

No behavioral change for any client that already sent `Authorization: Bearer`. The only addition is that the token is also accepted when sent as `x-api-key`, which Anthropic's own SDK and the Claude desktop app use as their native convention.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to CC Switch will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Fixed
+
+- **Claude Desktop gateway accepts `x-api-key` as a fallback to `Authorization: Bearer`**: `validate_claude_desktop_gateway_auth` in `src-tauri/src/proxy/handlers.rs` previously only accepted the gateway token via `Authorization: Bearer <token>`, but the Claude desktop app sends the token as `x-api-key` (Anthropic's published auth header convention). Requests from the desktop app were rejected with `401 "Claude Desktop gateway 缺少 Authorization 头"`, making Code mode fail with "Request failed · retrying" on the first message despite Cowork mode (and the bare `claude` CLI) working. The auth check now extracts the token from either header; `Authorization: Bearer` continues to work for clients that send it. A new `extract_gateway_token` pure helper holds the parsing logic and is covered by seven unit tests (Bearer / lowercase-bearer / no-prefix / x-api-key / both-present / whitespace-trim / missing-header).
+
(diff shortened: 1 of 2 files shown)


farion1231:
@codex review

PR #5125, opened 2026-07-08
fix(proxy): normalize Moonshot/Kimi tool parameters schema

jacker-son:
## Problem

When routing Codex requests to Kimi For Coding through CC Switch's Responses → Chat Completions transform, Moonshot rejects tool schemas that:
1. Use `$ref` values not starting with `#/$defs/` (e.g. `#/properties/bodyParts/items`).
2. Omit a top-level `type: "object"` in `tools.function.parameters` (e.g. schemas using top-level `oneOf`).

This causes 400 errors such as:
- `tools.function.parameters is not a valid moonshot flavored json schema`
- `tools.function.parameters.type is required and must be "object"`

Affected tools include Adobe Photoshop\'s `selectSubject`/`selectionByPrompt` and the built-in `automation_update` tool.

## Changes

Add `moonshot_compatible_schema` in `transform_codex_chat.rs` to:
- Resolve relative `$ref` paths and inline their targets under `#/$defs/`.
- Rename legacy `definitions` to `$defs`.
- Inject `type: "object"` at the top level when it is missing but the schema describes an object-like structure.

Apply this normalization to `function.parameters` when converting Codex Responses tools to Chat Completions tools.

## Testing

Added regression tests covering:
- Relative `$ref` rewriting.
- `definitions` → `$defs` migration.
- Preserving existing `#/$defs/` references.
- Top-level `oneOf` injection of `type: object`.
- Full Responses → Chat conversion path for both affected tool patterns.

### DIFF
--- a/src-tauri/src/proxy/providers/transform_codex_chat.rs
+++ b/src-tauri/src/proxy/providers/transform_codex_chat.rs
@@ -43,6 +43,198 @@ const CHAT_TOOL_NAME_MAX_LEN: usize = 64;
 const CUSTOM_TOOL_INPUT_DESCRIPTION: &str = "Raw string input for the original custom tool. Preserve formatting exactly and follow the original tool definition embedded in the description.";
 const CUSTOM_TOOL_PRESERVED_METADATA_HEADING: &str = "Original tool definition:";
 
+/// Rewrite a JSON schema so that Moonshot/Kimi accepts it in `tools.function.parameters`.
+///
+/// Moonshot's validator requires:
+/// - `type` at the top level of `parameters` to be `"object"`.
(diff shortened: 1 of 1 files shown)


farion1231:
@codex review

farion1231:
Thanks @jacker-son for tackling this. The Moonshot/Kimi tool-schema 400 is now fixed on `main` by #6863 (commit db41d7018), which moves each `$ref` with siblings into `allOf` after the Responses → Chat conversion, gated on the Moonshot/Kimi upstream host so other providers keep byte-identical schemas. Closing this one as superseded. Appreciate the effort!

PR #5138, opened 2026-07-09
fix(ci): run backend checks on Windows/macOS and repair platform-gated tests

zayokami:
## Problem

The backend CI job (`.github/workflows/ci.yml` → `backend`) runs **only on `ubuntu-22.04`**. Every `#[cfg(target_os = "windows")]` / `#[cfg(target_os = "macos")]` path — tests and clippy lints alike — is never compiled or run in CI. As a result `main` currently ships **8 failing Windows unit tests** and a hard **`cargo clippy -- -D warnings` failure on Windows**, all invisible because CI only builds/tests Linux.

Reproduce on Windows (Rust 1.95, MSVC):

```bash
cd src-tauri
cargo test --lib                                    # 1740 passed; 8 failed
cargo clippy --manifest-path Cargo.toml -- -D warnings   # 13 errors
```

Both pass on Linux — which is why this went unnoticed.

## Root cause & fix

**A. 6 stale tests — `commands::misc::tests::anchored_upgrade_windows::*`**
Commit `5092fe51` intentionally removed `codex` from `prefers_official_update()`, so codex now anchors to the package manager with no `codex update ||` prefix. The tests still asserted the old output. → Updated the 5 package-manager expectations; retargeted the no-sibling "official-update-without-fallback" test to `claude` (still in the list) so that branch stays covered.

**B. 2 test-only bugs — `codex_state_db` / `codex_history_migration` `sqlite_home`**
The tests built TOML *basic* strings from Windows paths, whose backslashes are invalid escapes, so the override failed to parse. → Switched to TOML *literal* (single-quoted) strings.

**C. 13 Windows-only clippy errors**
All in `#[cfg(target_os = "windows")]` blocks or code dead only on Windows (9 POSIX-shell helpers used only by the macOS/Linux launchers). → Moved `use std::io::Write` into the `#[cfg(unix)]` block that uses it, dropped 3 unneeded tail `return`s, marked the 9 helpers `#[cfg_attr(windows, allow(dead_code))]`.

**D. CI matrix**
Run the backend job on `{ ubuntu-22.04, windows-latest, macos-latest }` (`fail-fast: false`); gate the apt step on Linux; use `shell: bash` for the dist placeholder. This is what keeps A/B/C from silently regressing again.

## Verification (local, Windows 11 / Rust 1.95)

```text
cargo test --lib                          → 1748 passed; 0 failed   (was 1740/8)
cargo clippy -- -D warnings               → clean                   (was 13 errors)
cargo fmt --check                         → clean
```


farion1231:
@codex review

farion1231 (approved):

PR #5208, opened 2026-07-10
fix(deeplink): add explicit Codex model to generated catalog

01JohnMa:
## Summary / 概述

- add an explicit Codex deeplink model to the provider's simplified `modelCatalog`
- keep the existing `gpt-5-codex` fallback and omit the catalog when no non-empty model is supplied
- verify that `enabled=true` switches from an existing provider and writes the generated catalog used by Codex Desktop
- keep `api_format` unchanged and avoid hard-coding any GPT-5.6 model name in production code

This lets custom providers imported with `model=gpt-5.6-sol` appear as a named Codex Desktop model instead of only as `Custom`.

## Related Issue / 关联 Issue

Related to #5170

## Screenshots / 截图

Not applicable; this changes the Codex deeplink import and generated model catalog path.

## Verification / 验证

- `git diff --check`
- added unit coverage for explicit, missing, and whitespace-only Codex models
- added integration coverage for provider switching, persisted catalog settings, live `config.toml`, and generated catalog JSON
- local Rust execution was unavailable because this machine does not currently have Cargo; the new Rust tests and clippy are left for PR CI

## Checklist / 检查清单

- [ ] `pnpm typecheck` passes / 通过 TypeScript 类型检查 (no frontend changes)
- [x] Updated i18n files if user-facing text changed / 如修改了用户可见文本（本 PR 无用户可见文本变更）

### DIFF
--- a/src-tauri/src/deeplink/provider.rs
+++ b/src-tauri/src/deeplink/provider.rs
@@ -387,12 +387,13 @@ fn build_codex_settings(request: &DeepLinkImportRequest) -> serde_json::Value {
         provider_display_name
     };
 
-    // Model name: use deeplink model or default
-    let model_name = request
+    // Model name: use a non-empty deeplink model or the existing default.
+    let explicit_model = request
         .model
         .as_deref()
-        .unwrap_or("gpt-5-codex")
-        .to_string();
+        .map(str::trim)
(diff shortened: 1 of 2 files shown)


farion1231:
@codex review

PR #4397, opened 2026-06-18
fix(ui): reset provider scroll when switching apps

Muleizhang:
## Summary / 概述

Fix provider page content staying scrolled after switching between apps. The main page scroll container and provider list scroll container now reset to the top when the active app changes while the provider view is open.

Adds a regression test for the provider view scroll reset behavior.

## Related Issue / 关联 Issue

Fixes #

## Screenshots / 截图

| Before / 修改前 | After / 修改后 |
|-----------------|---------------|
|   image   |   image    |
|某一个提供商翻到底部后，无法查看其他提供商|可以查看其他提供商|
## Checklist / 检查清单

- [x] `pnpm typecheck` passes / 通过 TypeScript 类型检查
- [x] `pnpm format:check` passes / 通过代码格式检查
- [x] `pnpm test:unit` passes / 通过单元测试
- [x] `cargo clippy` passes (if Rust code changed) / Not applicable: no Rust changes
- [x] Updated i18n files if user-facing text changed / Not applicable: no user-facing text changes

PR #5211, opened 2026-07-11
fix(ui): reset provider scroll when switching apps

Muleizhang:
## Summary / 概述

Reapply the provider scroll reset fix from #4397 on the latest `main` branch.

- Reset both the outer page scroll container and the provider-list scroll container before paint when the active app changes while the provider view is open.
- Keep the animated view wrapper in the flex layout chain so the provider list remains the intended scroll container.
- Add an integration regression test that scrolls both containers, switches apps, and verifies both axes return to the top-left.

## Root cause / 根因

The provider view reuses its parent scroll containers when switching apps. Only the animated provider content changes, so the previous app's `scrollTop` and `scrollLeft` values were retained. If the previous provider list was scrolled far enough, the next app could appear empty.

## Screenshots / 截图

| image | image |
| 当你有很多个提供商，超过一页的范围时，翻到某一个提供商底部后，无法查看其他提供商 | 可以查看其他提供商 |

## User impact / 用户影响

Switching to another app from a scrolled provider list now starts at the top, so its providers are immediately visible.

## Related PR / 关联 PR

Supersedes #4397, which is now closed.

## Validation / 验证

- [x] Reproduced on the latest `main`: the regression test failed with `scrollTop` remaining at `320`
- [x] TypeScript typecheck (`pnpm typecheck`)
- [x] Prettier check (`pnpm format:check`)
- [x] Targeted provider scroll regression test
- [x] Full unit suite in serial mode: 135 files, 1075 tests passed
- [x] `cargo clippy` not applicable; no Rust changes
- [x] i18n updates not applicable; no user-facing text changes

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -244,9 +244,25 @@ function App() {
   const effectiveUsageProvider = useLastValidValue(usageProvider);
 
   const toolbarRef = useRef<HTMLDivElement>(null);
+  const mainScrollRef = useRef<HTMLElement>(null);
+  const providerScrollContainerRef = useRef<HTMLDivElement>(null);
   const isToolbarCompact = useAutoCompact(toolbarRef);
 
   useUsageCacheBridge();
+
+  useEffect(() => {
+    if (currentView !== "providers") return;
+
+    if (mainScrollRef.current) {
+      mainScrollRef.current.scrollTop = 0;
+      mainScrollRef.current.scrollLeft = 0;
+    }
+
+    if (providerScrollContainerRef.current) {
+      providerScrollContainerRef.current.scrollTop = 0;
+      providerScrollContainerRef.current.scrollLeft = 0;
+    }
+  }, [activeApp, currentView]);
 
   const promptPanelRef = useRef<any>(null);
   const mcpPanelRef = useRef<any>(null);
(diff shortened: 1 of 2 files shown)


Muleizhang:
@codex review

SaladDay (dismissed):
Looks good. The effect is scoped to the provider view, resets both scroll containers when the active app changes, and leaves other views alone. Keeping the animated wrapper in the flex column also preserves the intended inner scroll container. The integration test covers both axes, and CI is green.

SaladDay:
@farion1231 #5211 is ready to merge. The provider scroll reset is scoped, covered by an integration test, and all checks pass. ✅

Muleizhang:
@farion1231 ready to merge.

allenxu09 (dismissed):

Muleizhang:
@farion1231 Updated to the latest `main` as a single commit (`c603b5e8`). The regression test fails on `main` (`scrollTop` remains at 320 after switching providers) and passes with this fix. Local typecheck, format checks, all 1,075 frontend tests, and GitHub frontend CI pass. Previous approvals were dismissed after the force-push, so this needs a fresh review.

farion1231 (approved):
Thanks @Muleizhang for refreshing this fix and adding the scroll regression test! Reviewed the scroll reset timing and shared layout changes; TypeScript checks and all 8 App integration tests passed.

PR #5219, opened 2026-07-11
fix: map OpenAI token limit parameters by API and model

linzixi:
## Summary
- Map Claude-to-OpenAI Chat Completions token limits by model family: GPT-5, chat-latest, and o-series use max_completion_tokens; GPT-4o/4.1/4/3.5 retain max_tokens.
- Keep Responses API on max_output_tokens and clamp Claude Desktop probe requests to the API minimum of 16.

## Tests
- Added conversion and model-selection unit tests.
- cargo test could not complete locally because the Windows Rust toolchain process stalled.

### DIFF
--- a/src-tauri/src/proxy/providers/transform.rs
+++ b/src-tauri/src/proxy/providers/transform.rs
@@ -54,6 +54,25 @@ pub fn is_openai_o_series(model: &str) -> bool {
         && model.as_bytes().get(1).is_some_and(|b| b.is_ascii_digit())
 }
 
+/// Detect Chat Completions models that require `max_completion_tokens` instead
+/// of the legacy `max_tokens` parameter.
+///
+/// GPT-5+ (including `chat-latest`) and the o-series use the new field. GPT-4o,
+/// GPT-4.1, GPT-4, and
+/// GPT-3.5 retain `max_tokens` compatibility in the Chat Completions protocol.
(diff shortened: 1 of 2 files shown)


farion1231:
@codex review

SaladDay (changes requested):
Thanks, the canonical GPT-5/o-series mapping and the Responses minimum fix work, but there is still one Chat Completions gap to fix 🙂

`requires_openai_max_completion_tokens` only recognizes IDs that start with `gpt-` or `o`. CC Switch can map to arbitrary upstream model IDs, and prefixed IDs are already used in practice, for example `openai/gpt-5.4`, `b-gpt-5.6-sol`, and `openai/o3`. These inputs still produce `max_tokens: 1024` with no `max_completion_tokens`, so a strict GPT-5/o-series backend returns the same 400 this PR is meant to fix. I reproduced this with a conversion-level test after merging the latest main; #5343 also documents a real prefixed GPT-5 deployment name.

Please handle provider namespaces/custom prefixes safely, or add an explicit override for arbitrary deployment names, and add regression tests for prefixed GPT-5 and o-series IDs. The current head also fails the required Clippy job at `transform.rs:71` with `manual_pattern_char_comparison`; `split(['.', '-'])` is the suggested form.

The two complete transform test groups pass (65 + 75 tests), as do all 1,975 non-ignored lib tests, and fmt passes.

PR #4522, opened 2026-06-22
Add SubRouter provider presets

abingyyds:
## Related Issue / 关联 Issue

Fixes #

## Screenshots / 截图

| Before / 修改前 | After / 修改后 |
|-----------------|---------------|
|                 |               |

## Checklist / 检查清单

- [x] `pnpm typecheck` passes / 通过 TypeScript 类型检查
- [x] `pnpm format:check` passes / 通过代码格式检查
- [x] `cargo clippy` passes (if Rust code changed) / 通过 Clippy 检查（如修改了 Rust 代码）
- [x] Updated i18n files if user-facing text changed / 如修改了用户可见文本，已更新国际化文件

PR #3643, opened 2026-06-03
feat: add CherryIN preset provider for Claude Code and Codex

zhibisora:
## Summary / 概述

Adds **CherryIN** ([open.cherryin.net](https://open.cherryin.ai)) — an API aggregator gateway built on New API — as a quick-config preset for both **Claude Code** (Anthropic format) and **Codex** (OpenAI-compatible `/v1`), placed next to AiHubMix, with the official CherryIN brand icon. Endpoints and model IDs were verified against CherryIN's live pricing API; defaults are `anthropic/claude-sonnet-4.6` / `opus-4.8` / `haiku-4.5` for Claude and `openai/gpt-5.5` for Codex.

## Related Issue / 关联 Issue

Fixes #

## Screenshots / 截图

| Before / 修改前 | After / 修改后 |
|-----------------|---------------|
|                 |               |

## Checklist / 检查清单

- [x] `pnpm typecheck` passes / 通过 TypeScript 类型检查
- [x] `pnpm format:check` passes / 通过代码格式检查
- [ ] `cargo clippy` passes (if Rust code changed) / 通过 Clippy 检查（如修改了 Rust 代码） — N/A, no Rust changes
- [ ] Updated i18n files if user-facing text changed / 如修改了用户可见文本，已更新国际化文件 — N/A, brand name only

PR #5223, opened 2026-07-11
feat(provider): add Kenari presets (Claude Code, Codex, Claude Desktop, OpenCode, OpenClaw, Hermes)

doedja:
## Summary / 概述

Adds provider presets for [Kenari](https://kenari.id), an LLM gateway billed in Indonesian Rupiah with prepaid wallet top-up (QRIS and crypto). One `kn-` API key covers the whole catalog (GLM, Kimi, DeepSeek, GPT, Claude and more).

Presets added, following the existing Amux/SubRouter shape:

- **Claude Code**: Anthropic-compatible endpoint at `https://kenari.id`, with model env pinned to the catalog ids (`glm-5-2` / `kimi-k2-7-code` / `deepseek-v4-flash` on the opus/sonnet/haiku slots).
- **Codex**: `https://kenari.id/v1` with `wire_api = "responses"` (Kenari ships a native `/v1/responses` endpoint; verified against Codex CLI 0.144.1).
- **Claude Desktop**: direct Anthropic routing with passthrough model routes.
- **OpenCode / OpenClaw / Hermes**: OpenAI-compatible `https://kenari.id/v1`.

Also registers the Kenari icon (png, official mark) and adds a preset test file mirroring `subrouterProviderPresets.test.ts`.

Every endpoint in these presets was exercised live against kenari.id before submitting: `/v1/messages` (Anthropic wire), `/v1/chat/completions`, and `/v1/responses` all return correct completions with the exact models the presets reference. I operate Kenari, so I can keep these presets current.

Not a partner entry: no `isPartner`, no promotion key, no i18n changes.

## Related Issue / 关联 Issue

None (preset data addition, following the precedent of #4522 and #3643).

## Screenshots / 截图

Not applicable (no UI change; presets render through the existing list).

## Checklist / 检查清单

- [x] `pnpm test:unit` passes (420 tests, including the new Kenari preset tests)
- [ ] Updated i18n files (not applicable: no user-facing text changed)


farion1231:
@codex review

doedja:
Tested this end-to-end against the live gateway before replying, since the suggestion contradicts observed behavior.

Setup: codex-cli 0.144.1, isolated `CODEX_HOME` containing exactly what cc-switch generates for the Kenari preset: `generateThirdPartyConfig` output (`requires_openai_auth = true`, `wire_api = "responses"`, `base_url = "https://kenari.id/v1"`) plus `auth.json` = `{"OPENAI_API_KEY": "kn-..."}` with a valid Kenari key.

Results:

1. Header capture (pointing `base_url` at a local listener): with `requires_openai_auth = true`, codex sends `Authorization: Bearer kn-...`, the `auth.json` value byte-for-byte, to `<base_url>/responses`. No transformation and no OpenAI-specific credential handling happens to the key.
2. Live run against `https://kenari.id/v1`: completed normally (model `glm-5-2`, response received, usage billed). So the preset as generated works.

The `requires_openai_auth = false` in Kenari's docs is not a statement that `true` breaks: it belongs to the docs' manual-setup recipe, which pairs it with `env_key = "KENARI_API_KEY"` so the key comes from an environment variable instead of `auth.json`. Those are two valid ways to deliver the same Bearer key. cc-switch's mechanism is the `auth.json` one, where `requires_openai_auth = true` is exactly right, and it is the same shape every other third-party preset in this file uses.

So no change needed on this PR. I will also clarify the wording on Kenari's docs side so it stops reading as if `true` is unsupported.

doedja:
Rebased onto current main, so this is mergeable again.

Scope is unchanged: the presets for Claude Code, Codex, Claude Desktop, OpenCode, OpenClaw and Hermes, the icon plus its metadata entry, and the test file. Nothing else in the diff.

Happy to update if anything in the preset shape has moved since.

PR #5236, opened 2026-07-11
fix(windows): restore taskbar icon when relaunching via shortcut while app is in tray

Kulamiso:
## Summary / 概述

  On Windows, when the main window is closed to the system tray (with **Minimize to tray on close** enabled), re-launching the app via its **desktop
  shortcut or Start Menu entry** shows the window on screen but **no icon appears in the taskbar**.

  **Root cause:** The single-instance callback — triggered when a second launch is detected while the app is already running in the tray — calls `show()`
  without first calling `set_skip_taskbar(false)`. The flag was left as `true` from when the window was hidden to the tray. The tray menu "Open main window"
  action already handles this correctly; these two paths were inconsistent.

  ## Changes / 修改内容

  - **fix:** Add `set_skip_taskbar(false)` before `unminimize()`/`show()` in the single-instance callback (`lib.rs`) — the path triggered by relaunching via
  desktop shortcut while the app is in the tray
  - **fix:** Same addition in `handle_deeplink_url()` with `focus_main_window=true` — the path triggered by `ccswitch://` deep-link events while the app is
  tray-only
  - **chore:** Remove dead code — `#[cfg(target_os = "windows")]` nested inside `#[cfg(target_os = "macos")]` in the `Reopen` handler, which could never
  compile on any platform

  ## Related Issue / 关联 Issue

  Fixes #

  ## Screenshots / 截图

  | # | # |

  ## Checklist / 检查清单

### DIFF
--- a/src-tauri/src/lib.rs
+++ b/src-tauri/src/lib.rs
@@ -156,6 +156,10 @@ fn handle_deeplink_url(
 
             if focus_main_window {
                 if let Some(window) = app.get_webview_window("main") {
+                    #[cfg(target_os = "windows")]
+                    {
+                        let _ = window.set_skip_taskbar(false);
+                    }
                     let _ = window.unminimize();
                     let _ = window.show();
                     let _ = window.set_focus();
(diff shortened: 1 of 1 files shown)


farion1231:
@codex review

SaladDay (approved):
Looks good. Both Windows paths that restore a tray-hidden window now reset `skip_taskbar` before showing it, matching the existing tray and lightweight-mode paths. Removing the unreachable Windows block under the macOS-only handler is also correct. CI is green, and I found no blocking issues.

SaladDay:
@farion1231 This Windows taskbar fix is ready to merge. The diff is small, CI is green, and the two affected reopen paths now match the existing tray behavior. ✅

PR #5241, opened 2026-07-11
feat: auto enter lightweight mode after app inactivity

oxsean:
## Summary

Automatically destroy the main window and keep only the tray process after the main window stays inactive (unfocused, hidden, or closed to tray) for a configured number of minutes, reusing the existing lightweight mode logic.

## Configuration

- Single optional setting `autoLightweightIdleMinutes`: absent or `0` = disabled (default), valid range 1-1440.
- Turning the toggle off writes `0`; the backend normalizes `0` to an absent field on save.
- Enabling from the settings UI defaults to 5 minutes.

## Implementation

- `auto_lightweight.rs`: idle state (`unfocused_at`) guarded by a mutex; a single background worker (started once via `AtomicBool`) polls every 30s, re-reads settings each tick, and enters lightweight mode when the deadline passes.
- Focus gain, window show (tray/deeplink/single-instance), and window re-creation all reset the idle timer; the worker skips while already in lightweight mode, so there is no double trigger or double destroy.
- Settings UI: toggle + minutes input in Window Behavior, with zh/zh-TW/en/ja translations worded as "inactive" (not "focus loss", since hidden/closed-to-tray also counts).

## Testing

- `cargo test` (idle deadline + settings normalization), `cargo fmt --check`, `cargo check`
- `vitest` (WindowSettings enable/disable/minutes persistence), `tsc --noEmit`
- Manual on macOS dev app: silent-start hidden window destroyed after threshold with tray alive; tray re-open recreates the window and resets the timer; focus loss re-enters lightweight mode; toggling settings at runtime takes effect on the next poll tick.


farion1231:
@codex review

oxsean:
@farion1231 I’ve fixed the issues raised in the review. Could you please review it again?

PR #5243, opened 2026-07-11
fix(a11y): add screen reader labels to icon controls, fix locale fallback

herwigfelix:
## Summary

- Adds `aria-label` to icon-only buttons, switches, checkboxes, and selects that previously had no accessible name, or relied only on `title` — which WKWebView/VoiceOver (the engine Tauri uses on macOS) does not reliably expose as an accessible name, so many controls were announced as just "button".
- Fixes two Radix `Dialog`s that had no `DialogTitle`, so screen readers announce something when they open.
- Restructures a few spots where a real interactive control (e.g. a delete button) was nested inside another `<button>` — invalid HTML that made the inner control unreachable via keyboard. The outer element is now a `role="button"` div with `tabIndex`/`onKeyDown` instead.
- Makes a few mouse-only clickable rows (e.g. endpoint selection in the speed test list) keyboard-operable via `role="button"`/`tabIndex`/`onKeyDown`.
- Syncs `<html lang>` with the active UI language instead of leaving it hardcoded to `zh-CN`.
- Changes the default UI language fallback from Chinese to English for system locales the app doesn't explicitly recognize (e.g. German), matching i18next's own `fallbackLng: "en"`.

No visual changes for sighted users — only `aria-label`/`role`/`tabIndex`/`onKeyDown` attributes were added. Two hover-reveal icon buttons additionally got `focus-visible:opacity-100` so they become visible on keyboard focus (they stay invisible to mouse users until hovered, same as before).

## Test plan

- [x] `tsc --noEmit` passes
- [x] Full `vitest` suite passes (429/429)
- [x] Manually verified with VoiceOver on macOS (WKWebView) that previously-silent icon buttons now announce a name, and that the locale fallback now resolves to English on an unrecognized system locale
- [x] Diffed every changed file to confirm no `className`/styling was altered beyond the two `focus-visible:opacity-100` additions


farion1231:
@codex review

herwigfelix:
Good catch, thanks — fixed in 0704bb23. The three flagged spots (EndpointSpeedTest row, both ProviderAdvancedConfig headers, DailyMemoryPanel rows) plus one more instance of the same pattern in `OmoFormFields.tsx`'s `SectionHeader` (nested add-button) now guard their `onKeyDown` with `if (e.target !== e.currentTarget) return;` so a nested control's own Enter/Space activation no longer bubbles into the row/header handler.

PR #5256, opened 2026-07-12
fix(tray): refresh after provider deep-link import

### LINKED ISSUE #5255: Tray menu is not refreshed after provider deep-link import
## Bug description

After importing a provider through a ccswitch:// deep link, the provider list in the main window is updated. When the link uses enabled=true, the imported provider is also activated.

However, the system tray menu is not refreshed:

- the newly imported provider does not appear in the tray menu;
- the previously active provider remains checked;
- switching providers once in the main window makes the tray state update correctly.

## Cause

The provider deep-link import path invalidates the provider query used by the main window, but does not rebuild the tray menu after the import succeeds.

## Expected behavior

The tray menu should be refreshed immediately after a provider is imported, including its provider list and active check state.

MOLAaaaaaaa:
## Bug

After importing a provider through a deep link, the main window reflects the new provider and its active state, but the system tray menu keeps the previous provider list and checked state until another provider switch occurs.

The deep-link import path refreshed the main-window provider query but did not rebuild the tray menu.

## Fix

Refresh the tray menu after a successful provider deep-link import. If the tray refresh fails, log the error without reporting the already successful provider import as failed.

Fixes #5255

### DIFF
--- a/src/components/DeepLinkImportDialog.tsx
+++ b/src/components/DeepLinkImportDialog.tsx
@@ -1,6 +1,7 @@
 import { useState, useEffect, useMemo } from "react";
 import { listen } from "@tauri-apps/api/event";
 import { DeepLinkImportRequest, deeplinkApi } from "@/lib/api/deeplink";
+import { providersApi } from "@/lib/api/providers";
 import {
   Dialog,
   DialogContent,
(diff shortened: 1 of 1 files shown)


farion1231:
@codex review

PR #5268, opened 2026-07-12
fix(proxy): update Claude takeover to Sonnet 5

Ancienttwo:
## What changed

- update the Claude proxy-takeover Sonnet sentinel from `claude-sonnet-4-6` to `claude-sonnet-5`
- update managed Codex/Copilot takeover and `[1M]` regression expectations

## Why

CC Switch v3.16.5 updated provider defaults and pricing to Claude Sonnet 5, but proxy takeover still writes `claude-sonnet-4-6` into Claude Code's live `ANTHROPIC_DEFAULT_SONNET_MODEL`.

The local proxy uses that value only as a Sonnet routing sentinel before mapping it to the provider-configured upstream model. Nevertheless, Claude Code validates and reports the sentinel as the selected model. With Sonnet 4.6 retired from the current Claude Code model menu, selecting the provider's Sonnet tier can therefore show or reject an obsolete model before the proxy mapping is useful.

Using `claude-sonnet-5` aligns takeover with the repository's current Sonnet default while preserving role-based routing, provider-specific `*_MODEL_NAME`, and `[1M]` behavior.

## Validation

- `cargo test --manifest-path src-tauri/Cargo.toml managed_account_claude_takeover_sources -- --nocapture` (2 passed)
- full backend suite previously reached 1890 passed and 2 ignored; the only failure was an unrelated fixed-port test unable to bind 15721 while an installed CC Switch instance was running

### DIFF
--- a/src-tauri/src/services/proxy.rs
+++ b/src-tauri/src/services/proxy.rs
@@ -42,7 +42,7 @@ const CLAUDE_MODEL_OVERRIDE_ENV_KEYS: [&str; 12] = [
 ];
 
 const CLAUDE_TAKEOVER_HAIKU_MODEL: &str = "claude-haiku-4-5";
-const CLAUDE_TAKEOVER_SONNET_MODEL: &str = "claude-sonnet-4-6";
+const CLAUDE_TAKEOVER_SONNET_MODEL: &str = "claude-sonnet-5";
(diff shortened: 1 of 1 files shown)


allenxu09:
@codex review

PR #5269, opened 2026-07-12
fix: avoid interactive shell when probing tool versions

LTLeopard:
## Problem

On macOS, the "本地环境检查" (Local Environment Check) shows version `26.5.2` for every tool. This is actually the **macOS version**, not the tool version.

**Root cause**: `default_flag_for_shell()` passes `-lic` (login + **interactive** + command) to zsh/bash when running `{tool} --version`. The interactive flag `-i` sources `.zshrc`, which commonly runs startup scripts like `neofetch`. The neofetch output contains `OS: macOS 26.5.2 arm64`, and `extract_version` regex picks `26.5.2` as the **first** version match.

## Fix

Changed `-lic` → `-lc` in `default_flag_for_shell()`.
- Login shell (`-l`) still sources `.zprofile` for Homebrew/npm PATH setup
- Removed `-i` avoids sourcing `.zshrc` and its side effects (neofetch, startup banners, etc.)
- Tools not found on the login-shell PATH fall back to `scan_cli_version()`

## Testing

- Updated `test_default_flag_for_shell` assertions

### DIFF
--- a/src-tauri/src/commands/misc.rs
+++ b/src-tauri/src/commands/misc.rs
@@ -1065,7 +1065,11 @@ fn default_flag_for_shell(shell: &str) -> &'static str {
     match shell.rsplit('/').next().unwrap_or(shell) {
         "dash" | "sh" => "-c",
         "fish" => "-lc",
-        _ => "-lic",
+        _ => "-lc", // was "-lic": interactive flag sources .zshrc and may trigger
+                    // startup scripts like `neofetch` whose output pollutes stdout,
+                    // causing `extract_version` to return the wrong version number.
+                    // Login shell (-l) + command (-c) suffices to get the user's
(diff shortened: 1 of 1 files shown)


farion1231:
@codex review

SaladDay:
Thanks for fixing the false version caused by interactive-shell startup output. One compatibility issue still looks blocking 🙂 Switching the probe from `-lic` to `-lc` also drops tools whose PATH is initialized only in `.zshrc` or `.bashrc`. The fallback currently adds `~/.bun/bin` only for OpenCode, even though this module supports Bun-installed Codex, and it can also miss common rc-only pnpm/asdf installs. Please either cover the supported CLI and package-manager paths in the fallback, or keep the interactive PATH while isolating/sanitizing the version output. A regression test with Codex visible only through `~/.bun/bin` would pin this down.

PR #5285, opened 2026-07-13
feat(claude): add suppressReasoningEffort provider option

hotelbe:
## Summary / 概述

Problem:
Some OpenAI-compatible Chat Completions endpoints reject requests that
include both `reasoning_effort` and `tools` for newer reasoning models.
Observed on Azure OpenAI (Microsoft Foundry) deployments of gpt-5.5 and
gpt-5.6-luna/terra/sol behind an enterprise gateway:

  400 Function tools with reasoning_effort are not supported for {model}
  in /v1/chat/completions. Please use /v1/responses instead.

Since Claude Code always sends MCP tools, these models are currently
unusable via cc-switch Claude routing. These endpoints do not offer the
Responses API yet, so switching apiFormat to openai_responses is not an
option for them.

## Related Issue / 关联 Issue

Fixes #
Solution:
Add an opt-in `suppressReasoningEffort` flag to ProviderMeta (default
off). When enabled, the Anthropic→OpenAI Chat Completions transform
skips injecting `reasoning_effort`, allowing tools-based requests to
succeed. Follows the same pattern as existing per-provider
compatibility switches (isFullUrl, impersonateClaudeCode,
codexChatReasoning.effortValueMode).

Changes:
- provider.rs: add `suppress_reasoning_effort: Option<bool>` to ProviderMeta
- transform.rs: add `suppress_reasoning_effort` param to
  `anthropic_to_openai_with_reasoning_content()`, skip injection when true
- claude.rs: read the flag from provider.meta in the openai_chat branch
- UI: checkbox in the Claude provider form (visible only for
  apiFormat=openai_chat), with i18n strings (en/zh/zh-TW/ja)
- tests: new unit test for suppression; existing call sites updated

No behavior change for existing providers (defaults to false).

## Screenshots / 截图
image

image

## Checklist / 检查清单

### DIFF
--- a/src-tauri/src/provider.rs
+++ b/src-tauri/src/provider.rs
@@ -456,6 +456,17 @@ pub struct ProviderMeta {
     /// 是否将 base_url 视为完整 API 端点（不拼接 endpoint 路径）
     #[serde(rename = "isFullUrl", skip_serializing_if = "Option::is_none")]
     pub is_full_url: Option<bool>,
+    /// When `true`, suppresses `reasoning_effort` injection during
+    /// Anthropic → OpenAI Chat Completions conversion.
+    ///
+    /// Use for OpenAI-compatible endpoints (e.g. Azure OpenAI gpt-5.5,
+    /// gpt-5.6-*) that reject `reasoning_effort` when `tools` are present:
+    /// "400 Function tools with reasoning_effort are not supported ...".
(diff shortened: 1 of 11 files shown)


farion1231:
@codex review

hotelbe:
any commits for this pull request？

PR #5303, opened 2026-07-13
fix(usage): add GPT-5.6 pricing coverage

ThanatosXingYu:
## Summary

- Add Standard token pricing for `gpt-5.6`, `gpt-5.6-sol`, `gpt-5.6-terra`, and `gpt-5.6-luna`.
- Map the documented `gpt-5.6` alias to Sol pricing and support the `max` reasoning-effort suffix during pricing lookup.
- Add regression tests for the seeded prices and normalized model IDs.

Prices are sourced from the official OpenAI API pricing page and use USD per 1M tokens:
https://developers.openai.com/api/docs/pricing

## Testing

- `git diff --check`
- Added Rust regression tests for GPT-5.6 seed values and `max` effort matching.


farion1231:
@codex review

SaladDay:
Thanks for updating this and merging the latest `main` 🙂 The GPT-5.6 prices, Sol alias, and `max` lookup behavior all check out.

There is one small merge artifact to clean up before this lands: `gpt-5.6` is now seeded twice in `schema.rs`—once in the new block and again in the existing GPT-5.6 block below it. `INSERT OR IGNORE` makes this harmless at runtime today, but keeping two seed entries can silently hide a future price update and makes the seed count inaccurate. Could you please remove the first duplicate entry and keep the existing Sol/Terra/Luna plus alias block?

The CI run for the current head (`45107a6`) is also waiting for maintainer approval (`action_required`); the green run belongs to the previous SHA. Once the duplicate is removed and the current head passes CI, this should be good to merge. 👍

ThanatosXingYu:
Thanks for catching this. I removed the earlier duplicate `gpt-5.6` seed and kept the existing Sol/Terra/Luna plus alias block. The regression tests and `max` effort lookup remain unchanged. The update has been pushed to this PR. @SaladDay

SaladDay (approved):
The duplicate GPT-5.6 seed is gone. The remaining diff correctly covers the Sol alias, all three Standard pricing tiers, cache read/write prices, and `max` effort lookup. Exact model IDs are still checked before suffix fallback, so existing models whose names end in `-max` are not affected. No blocking code issues remain.

SaladDay:
@farion1231 The duplicate seed has been removed and #5303 is ready from code review. I approved the current head. Its CI workflow is still `action_required`, so that run needs maintainer approval before merge. ✅

allenxu09:
Token pricing is already covered in commit `a7b4dd9`. This PR mainly adds regression tests to `reasoning_effort` and is ready for review.

allenxu09 (approved):

PR #5271, opened 2026-07-12
[codex] normalize Codex tool schemas

WilliamClifton-dev:
## Summary

- normalize Codex function parameter schemas before forwarding them to Chat Completions providers
- add `type: object` when an object schema only declares composition keywords such as `oneOf`
- preserve the original composed schema and cover the DeepSeek failure with a regression test

## Root cause

Codex can provide namespace tools whose parameter schema has a top-level `oneOf` without an explicit `type`. This is valid JSON Schema, but strict OpenAI-compatible providers such as DeepSeek require function parameters to declare `type: object`. The Responses-to-Chat conversion previously forwarded that schema unchanged.

## Validation

- `cargo test responses_request_to_chat_adds_object_type_to_composed_tool_schema --lib`
- `cargo test proxy::providers::transform_codex_chat::tests --lib` (53 passed)
- `cargo fmt --check`
- `cargo clippy --lib --tests`
- full library run: 1870 passed; 8 unrelated existing Windows/config tests failed

Fixes #5263

PR #5315, opened 2026-07-14
fix: default Codex tool parameters to object schema

Komikawayi:
﻿## Summary

- Default Codex Responses function tool `parameters` to an object schema when missing or `null`.
- Apply the normalization to both direct Responses tools and nested OpenAI-style `function` tools.
- Cover the `codex_app__automation_update` style `parameters: null` regression reported in #5300.

## Root cause

Codex can emit function tools such as `codex_app__automation_update` with `parameters: null`. During Responses → Chat Completions conversion, cc-switch forwarded that value unchanged, and strict OpenAI-compatible upstreams such as DeepSeek reject it because function parameters must be a JSON Schema object.

## Tests

- `cargo test transform_codex_chat --lib --manifest-path E:\gateway_komi\cc-switch\src-tauri\Cargo.toml`
- `git diff --check`

### DIFF
--- a/src-tauri/src/proxy/providers/transform_codex_anthropic.rs
+++ b/src-tauri/src/proxy/providers/transform_codex_anthropic.rs
@@ -440,14 +440,17 @@ fn chat_tool_to_anthropic_tool(chat_tool: &Value) -> Option<Value> {
         .and_then(|value| value.as_str())
         .map(str::trim)
         .filter(|value| !value.is_empty())?;
-    let mut tool = json!({
-        "name": name,
-        "input_schema": function
-            .get("parameters")
-            .cloned()
-            .filter(|value| value.as_object().is_some_and(|object| !object.is_empty()))
-            .unwrap_or_else(|| json!({ "type": "object", "properties": {} }))
-    });
+    let mut input_schema = function
+        .get("parameters")
+        .cloned()
+        .filter(|value| value.as_object().is_some_and(|object| !object.is_empty()))
+        .unwrap_or_else(|| json!({ "type": "object", "properties": {} }));
+    if let Some(schema) = input_schema.as_object_mut() {
+        if schema.get("type").and_then(Value::as_str) != Some("object") {
+            schema.insert("type".to_string(), json!("object"));
+        }
+    }
+    let mut tool = json!({ "name": name, "input_schema": input_schema });
     if let Some(description) = function.get("description").and_then(|value| value.as_str()) {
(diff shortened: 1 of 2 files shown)


farion1231:
@codex review

SaladDay:
Thanks for the fix 🙂 The direct and nested `parameters: null` paths look good. There is one remaining case before merge: `clean_chat_tool_parameters` uses `contains_key("type")`, so `{"type": null, ...}` is still forwarded unchanged. This shape is reported in #5192 and will still trigger the same schema validation error.

Could you change the check to `obj.get("type").is_none_or(Value::is_null)` and add a regression test for explicit `type: null`? It would also be useful to bring over the top-level `oneOf` test from #5271. The two PRs otherwise overlap and will conflict. Thanks!

Komikawayi:
@SaladDay Addressed in 0213ecf.

I rebased #5315 onto current `main` and retained the already-merged #4706 normalizer, which handles explicit `type: null`. This PR now adds regression coverage for:

- direct and nested `parameters: null`;
- explicit `parameters.type: null`, preserving `properties` and `required`;
- top-level `oneOf` schemas with a missing root `type`, preserving the composed schema.

I also made the nested `function` path normalize a missing `parameters` field, matching the direct path.

Validated with `cargo test --manifest-path src-tauri/Cargo.toml --lib transform_codex_chat::tests::` (57 passed), `cargo fmt --check`, and `git diff --check`. GitHub CI is running on this update.

Komikawayi:
Follow-up for #5457: normalized the Anthropic conversion path too, so function tools recovered from `tool_search_output` with a top-level `oneOf` receive `input_schema.type = "object"` while preserving the composition. Added the regression fixture.

Verified:
- `cargo test --manifest-path src-tauri/Cargo.toml --lib transform_codex_anthropic::tests::` (68 passed)
- `cargo test --manifest-path src-tauri/Cargo.toml --lib transform_codex_chat::tests::responses_request_to_chat_defaults_top_level_one_of_tool_parameters_to_object`
- `git diff --check`

Komikawayi:
这个问题有不少人都提到了，最新的5576issue也 提到了这个问题，希望作者关注到尽快审核 @farion1231

farion1231:
已合并，感谢修复以及补齐的全场景回归测试！🙏 nested 与直接路径对齐到同一个 normalizer 的做法很干净。修复会随下个版本发布，进展会在 #4783 同步。

PR #5316, opened 2026-07-14
fix: refresh tray after deeplink provider import

Komikawayi:
﻿## Summary

- Inject `AppHandle` into deep-link provider import commands.
- Refresh the system tray menu immediately after a provider is imported through a deep link.
- Covers both the legacy provider import command and the unified provider import path.

## Root cause

The deep-link provider import path updated provider data used by the main window, but did not rebuild the tray menu. As a result, newly imported providers and active-provider check state were stale until another provider switch triggered a tray refresh.

## Tests

- `cargo test deeplink --lib --manifest-path E:\gateway_komi\cc-switch\src-tauri\Cargo.toml`
- `git diff --check`

### DIFF
--- a/src-tauri/src/commands/deeplink.rs
+++ b/src-tauri/src/commands/deeplink.rs
@@ -3,7 +3,7 @@ use crate::deeplink::{
     import_skill_from_deeplink, parse_deeplink_url, DeepLinkImportRequest,
 };
 use crate::store::AppState;
-use tauri::State;
+use tauri::{AppHandle, State};
 
 /// Parse a deep link URL and return the parsed request for frontend confirmation
 #[tauri::command]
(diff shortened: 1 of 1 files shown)


farion1231:
@codex review

SaladDay (approved):
Looks good. The refresh runs after a successful provider import and covers both the legacy and unified deeplink commands. AppHandle injection is correct, and CI is green. No blocking issues found.

SaladDay:
@farion1231 This one is ready to merge. I recommend taking #5316 instead of #5256 because the backend fix covers both command entry points; merging both would rebuild the tray twice on the normal UI path. #5256 was submitted earlier and its author opened #5255, so please preserve that contribution when closing the duplicate. ✅

PR #5318, opened 2026-07-14
feat: add grok-go Codex preset

Komikawayi:
﻿## Summary

- Add a Codex provider preset for the local `grok-go` gateway.
- Default the preset to `http://127.0.0.1:8787/v1` with native Responses API mode.
- Seed common Grok model entries (`grok-4`, `grok-4-fast`, `grok-3`, `grok-3-mini`) so users can select them immediately or refresh from `/v1/models`.

## Why

Issue #5314 proposes the minimal integration path for `RongleCat/grok-go`: treat it as an OpenAI-compatible local provider instead of adding new protocol code. cc-switch already supports OpenAI-compatible Responses providers, so a preset is enough for the first step.

## Tests

- `E:\gateway_komi\cc-switch\node_modules\.bin\prettier.cmd --check src/config/codexProviderPresets.ts`
- `E:\gateway_komi\cc-switch\node_modules\.bin\tsc.cmd --noEmit`
- `git diff --check`

### DIFF
--- a/src/config/codexProviderPresets.ts
+++ b/src/config/codexProviderPresets.ts
@@ -122,6 +122,40 @@ export const codexProviderPresets: CodexProviderPreset[] = [
     iconColor: "#00A67E",
   },
   {
+    name: "grok-go",
+    websiteUrl: "https://github.com/RongleCat/grok-go",
+    apiKeyUrl: "https://github.com/RongleCat/grok-go",
+    category: "third_party",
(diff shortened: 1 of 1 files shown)


Komikawayi:
Closing for now because we cannot verify the grok-go preset end-to-end without a working local grok-go instance/token. We can reopen or resubmit after confirming /v1/models and /v1/responses behavior.

PR #5331, opened 2026-07-14
Fix/homebrew npm path

kunjinkao-os:
## Summary / 概述
- Add Homebrew binary directories to POSIX lifecycle scripts before invoking CLI installation commands.
- Fix `npm: command not found` when CC Switch is launched as a macOS GUI app and npm is installed through Homebrew.
- Add a regression test covering the generated Codex installation script.

## Related Issue / 关联 Issue
N/A

Fixes #

## Screenshots / 截图

|---|---|
| 修改前：安装 Gemini CLI 时提示 npm command not found | 修改后：Gemini CLI 安装成功 |

## Checklist / 检查清单

### DIFF
--- a/src-tauri/src/commands/misc.rs
+++ b/src-tauri/src/commands/misc.rs
@@ -388,8 +388,9 @@ fn build_tool_lifecycle_command(
         lines.push("set -e".to_string());
         lines.push("set -o pipefail".to_string());
         // 非交互式 bash 默认 PATH 不含 Homebrew 目录(macOS ARM: /opt/homebrew/bin),
-        // 导致 npm/node 等工具找不到 Homebrew 安装的版本。将常见 Homebrew bin 目录加入 PATH 兜底。
-        lines.push("export PATH=\"/opt/homebrew/bin:/usr/local/bin:$PATH\"".to_string());
+        // 导致 npm/node 等工具找不到 Homebrew 安装的版本。将常见 Homebrew bin 目录追加为兜底,
+        // 避免覆盖继承 PATH 中优先级更高的 nvm/fnm/mise/Volta 等 Node 工具链。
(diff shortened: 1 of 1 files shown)


SaladDay:
Thanks for the focused fix! I found one PATH-ordering issue before merge.

This currently prepends both Homebrew directories ahead of the PATH that CC Switch already inherited. If the app is launched from a terminal or a configured launchd environment where nvm, fnm, mise, Volta, or another Node installation comes first, the bare `npm i -g` command will switch to Homebrew npm instead. That can install the CLI into a different global prefix and leave the active installation unchanged.

Could we preserve the existing PATH order and add the Homebrew directories only as fallbacks, for example by appending them, and add a regression case where a custom npm/node directory already precedes Homebrew? This would still fix the slim Finder/Dock PATH without changing an existing valid toolchain. 🙂

kunjinkao-os:
Thanks for catching this! I’ve updated the lifecycle script to preserve the inherited PATH order and append the Homebrew directories as fallbacks. I also added a regression test that expands the PATH with a custom nvm Node/npm directory ahead of Homebrew.

SaladDay (approved):
Re-reviewed the updated head against current main. The inherited PATH now stays first, while a slim macOS GUI PATH can still fall back to Homebrew npm/node. I verified both resolution cases locally, and the full misc lifecycle test group passes (78/78). I found no blocking regression.

SaladDay:
@farion1231 #5331 is ready from code review. The updated PATH ordering preserves custom Node toolchains, and the current-head lifecycle tests pass. Only the label check has run on this head, so the main CI workflow still needs maintainer authorization. ✅

farion1231:
@codex review

tan-yong-sheng:
Sorry to interrupt, but if feasible, any way to support homebrew installation of cc-switch to Linux OS / Ubuntu OS in future: https://github.com/farion1231/cc-switch/issues/4807 , thanks a lot ...

PR #5356, opened 2026-07-14
fix(skills): preserve deleted default repositories

allenxu09:
## Summary

- initialize default skill repositories only when the repository table is empty
- keep a deleted default repository from being restored on the next app startup
- add a regression test for deleting a default repository and rerunning initialization

## Root cause

Startup initialization compared the configured repositories against every built-in default and reinserted any missing entry. As a result, deleting a built-in repository from repository management only lasted until the next startup.

## Validation

- `cargo check` passes

### DIFF
--- a/src-tauri/src/database/dao/skills.rs
+++ b/src-tauri/src/database/dao/skills.rs
@@ -232,32 +232,21 @@ impl Database {
         Ok(())
     }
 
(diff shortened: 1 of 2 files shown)


farion1231:
@codex review

farion1231 (approved):

PR #5386, opened 2026-07-15
fix(codex): prefer freshest credential source

XWind18:
## Summary

- Select the best Codex credential copy when macOS Keychain and `~/.codex/auth.json` diverge.
- Prefer valid credentials, then the most recently refreshed copy, instead of always preferring the legacy `Codex Auth` Keychain item.
- Return the official quota API result for a locally stale credential so API failures are not mislabeled as stale tokens.

## Root cause

After Codex CLI refreshed `~/.codex/auth.json`, an older macOS Keychain entry could still be selected first. Its old `last_refresh` timestamp triggered the stale-token warning even though the newer file credential remained valid.

## Validation

- `rustfmt --edition 2021 --check src/services/subscription.rs`
- `cargo test --lib services::subscription::tests` (3 passed)

### DIFF
--- a/src-tauri/src/services/subscription.rs
+++ b/src-tauri/src/services/subscription.rs
@@ -474,27 +474,40 @@ type CodexCredentials = (
     Option<String>,
 );
 
-/// 读取 Codex OAuth 凭据
+/// A Codex credential source plus the refresh timestamp used to select the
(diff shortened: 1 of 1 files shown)


farion1231:
@codex review

PR #5390, opened 2026-07-15
fix(proxy): keep buffered Read streams alive

### LINKED ISSUE #5389: [Bug] Read tool argument deltas are fully buffered, causing Claude Code idle timeout on long/incomplete Codex streams
### Self Checks / 自检

- [x] I have read the FAQ section in README. / 我已阅读 README 中的常见问题。
- [x] I have searched existing issues, including closed ones. / 我已搜索已有 Issue（含已关闭）。

### CC Switch Version / 版本号

3.17.0

### Operating System / 操作系统

macOS

### Related App / 涉及应用

Claude Code, Codex

### Environment / 环境

- Claude Code 2.1.210
- CC Switch 3.17.0 local proxy
- Codex OAuth provider
- `apiFormat = "openai_responses"`
- Claude Opus/Sonnet mapped to `gpt-5.5`
- Auto failover disabled

### Summary / 问题概述

For OpenAI Responses → Anthropic streaming conversion, CC Switch forwards `response.function_call_arguments.delta` immediately for most tools, but deliberately suppresses every delta when the tool name is `Read`. It buffers the full JSON and emits one sanitized `input_json_delta` only after `response.function_call_arguments.done`.

If the Codex upstream produces a very long or pathological `Read` argument stream, or never sends the `.done` event, the upstream connection remains active while Claude Code receives no downstream event after `content_block_start`. Claude Code then reports:

```text
API Error: The operation timed out.
```

This is especially confusing because CC Switch is continuously receiving valid SSE bytes and therefore does not treat the request as idle or eligible for retry/failover.

### Steps to Reproduce / 重现步骤

1. Route Claude Code through CC Switch using a Codex OAuth / OpenAI Responses provider.
2. Ask the model to perform a task that causes a `Read` tool call.
3. Have the upstream emit:
   - `response.output_item.added` for a function call named `Read`;
   - many `response.function_call_arguments.delta` events;
   - no `response.function_call_arguments.done` within Claude Code's stream-idle window.
4. Observe that CC Switch emits the Anthropic `content_block_start`, but no `content_block_delta/input_json_delta` or heartbeat while the `Read` arguments continue upstream.
5. After about five minutes with the default Claude Code worker environment, Claude Code closes the stream with `API Error: The operation timed out.`.

### Observed Timeline / 实际时间线

One captured incident (Asia/Shanghai, 2026-07-15):

```text
14:15:48  Claude Code request forwarded by CC Switch (~509 KB request body)
14:15:51  upstream response.created and Read content_block_start
14:15:51–14:20:09
          upstream response.function_call_arguments.delta × 8,959
          downstream input_json_delta × 0
          no response.function_call_arguments.done
          no response.completed
14:20:09  Claude Code closes the request after ~293.5 s
14:20:09  CC Switch logs [SRV-006] connection error
```

The upstream was still sending argument deltas at the moment the downstream client disconnected. This was not an HTTP 520, quota failure, first-byte timeout, or inactive upstream socket.

For comparison, a normal short `Read` call in the same session emitted the complete sanitized `input_json_delta` only when `response.function_call_arguments.done` arrived.

No credentials, prompts, tool outputs, account identifiers, or full request bodies are included here.

### Code Path / 代码路径

In v3.17.0, the general path emits each tool argument delta immediately:

https://github.com/farion1231/cc-switch/blob/v3.17.0/src-tauri/src/proxy/providers/streaming_responses.rs#L778-L853

But `Read` takes this special branch after accumulating the delta:

```rust
if tool_name_by_index.get(&index).map(String::as_str) == Some("Read") {
    continue;
}
```

The complete buffered input is sanitized and emitted only in the `.done` handler:

https://github.com/farion1231/cc-switch/blob/v3.17.0/src-tauri/src/proxy/providers/streaming_responses.rs#L860-L907

This behavior was introduced to fix empty `pages: ""` compatibility:

- #2471
- #2472

The sanitation is useful, but the current all-or-nothing buffering makes downstream liveness depend on receiving `.done`.

### Expected Behavior / 期望行为

An active upstream `Read` argument stream should not become a completely idle downstream Anthropic stream.

At least one of the following should apply:

1. Forward safe incremental `input_json_delta` events for `Read` while removing only the incompatible empty `pages` field.
2. If full buffering is required for sanitation, emit protocol-safe Anthropic SSE `ping` events while upstream deltas are arriving.
3. Add a bounded duration/byte/event limit for buffered tool arguments. If `.done` never arrives, terminate with a clear retryable proxy error before Claude Code's watchdog fires.
4. Ensure downstream cancellation promptly cancels the upstream stream and records a diagnostic reason distinct from a generic `[SRV-006] connection error`.

### Actual Behavior / 实际行为

CC Switch keeps consuming and buffering an unbounded active `Read` argument stream without sending downstream progress. Claude Code eventually times out, and its normal retry mechanism does not restart the turn because an HTTP 200 stream and a partial `tool_use` block were already delivered.

Auto failover does not help in this state: the stream is HTTP 200 and active from CC Switch's perspective. In the reported environment, failover was also disabled.

### Suggested Regression Tests / 建议回归测试

1. `Read` emits many argument deltas over longer than the downstream idle window before `.done`; assert periodic downstream progress.
2. `Read` emits deltas forever and never sends `.done`; assert bounded termination and upstream cancellation.
3. Normal `Read` input containing `pages: ""`; assert the field is still removed.
4. Normal PDF `Read` input containing a valid `pages` value; assert it is preserved.
5. Downstream disconnects while `Read` arguments are buffered; assert the upstream task is cancelled and the request is logged accurately.

### Related Issues / 关联问题

- #2471 / #2472: introduced the `Read` empty-`pages` sanitation that currently requires full buffering.
- #5368: also concerns hidden active Responses streams, but at the earlier semantic-stream priming stage. This report is a different post-commit path: `message_start` and `content_block_start` have already been sent, then only `Read` argument deltas are suppressed.

liuxingbo12138:
## Summary / 概述

- emit Anthropic `ping` events while OpenAI Responses `Read` tool arguments are buffered
- preserve the existing empty `pages: ""` sanitation and emit the sanitized input only once on `.done`
- add a regression test for an incomplete multi-delta `Read` stream

## Root cause / 根因

The Responses → Anthropic converter buffers every `Read` argument delta so it can remove an incompatible empty `pages` field after the complete JSON arrives. Unlike other tools, the `Read` branch previously emitted nothing downstream between `content_block_start` and `response.function_call_arguments.done`.

A long or incomplete Codex tool-argument stream therefore remained active upstream but looked idle to Claude Code. Claude Code eventually closed the HTTP 200 stream with `API Error: The operation timed out.`

Anthropic `ping` is a protocol-safe keepalive and does not expose the unsanitized partial JSON to Claude Code.

## Impact / 影响

Long-running `Read` tool calls now keep the downstream stream active while CC Switch waits for the complete arguments. Normal calls still remove empty `pages` values, and truncated streams still end with the existing explicit `stream_truncated` error.

Fixes #5389

## Validation / 验证

- `git diff --check`
- `cargo test --manifest-path src-tauri/Cargo.toml proxy::providers::streaming_responses::tests` — 22 passed
- `cargo clippy --manifest-path src-tauri/Cargo.toml --all-targets`

## Checklist / 检查清单

- [x] Rust formatting passes
- [x] Focused proxy tests pass
- [x] Cargo Clippy passes

### DIFF
--- a/src-tauri/src/proxy/providers/streaming_responses.rs
+++ b/src-tauri/src/proxy/providers/streaming_responses.rs
@@ -837,6 +837,12 @@ pub fn create_anthropic_sse_stream_from_responses<E: std::error::Error + Send + 
                                     tool_had_delta.insert(index);
 
                                     if tool_name_by_index.get(&index).map(String::as_str) == Some("Read") {
+                                        // Read arguments are buffered until `.done` so empty `pages`
+                                        // values can be removed before they reach Claude Code. Keep
+                                        // the downstream Anthropic stream alive while that buffering
+                                        // is in progress; otherwise a long but active upstream tool
+                                        // call looks idle to the client and trips its stream watchdog.
+                                        yield Ok(anthropic_sse("ping", &json!({"type":"ping"})));
                                         continue;
                                     }
 
@@ -1799,6 +1805,34 @@ mod tests {
         assert!(merged.contains("\"name\":\"Read\""));
         assert!(merged.contains("\"partial_json\":\"{\\\"file_path\\\":\\\"/tmp/demo.py\\\",\\\"limit\\\":2000,\\\"offset\\\":0}"));
         assert!(!merged.contains("\\\"pages\\\":\\\"\\\""));
+    }
+
+    #[tokio::test]
+    async fn test_streaming_read_tool_emits_keepalive_while_buffering_args() {
+        let input = concat!(
+            "event: response.created\n",
+            "data: {\"type\":\"response.created\",\"response\":{\"id\":\"resp_read\",\"model\":\"gpt-5.5\"}}\n\n",
+            "event: response.output_item.added\n",
+            "data: {\"type\":\"response.output_item.added\",\"item\":{\"id\":\"fc_read\",\"type\":\"function_call\",\"call_id\":\"call_read\",\"name\":\"Read\"}}\n\n",
+            "event: response.function_call_arguments.delta\n",
+            "data: {\"type\":\"response.function_call_arguments.delta\",\"item_id\":\"fc_read\",\"delta\":\"{\\\"file_path\\\":\"}\n\n",
+            "event: response.function_call_arguments.delta\n",
+            "data: {\"type\":\"response.function_call_arguments.delta\",\"item_id\":\"fc_read\",\"delta\":\"\\\"/tmp/demo.py\\\"}\"}\n\n"
+        );
+
+        let upstream = stream::iter(vec![Ok::<_, std::io::Error>(Bytes::from(input))]);
+        let merged = create_anthropic_sse_stream_from_responses(upstream)
+            .collect::<Vec<_>>()
+            .await
+            .into_iter()
+            .map(|chunk| String::from_utf8_lossy(chunk.unwrap().as_ref()).to_string())
+            .collect::<String>();
+
+        assert_eq!(merged.matches("event: ping").count(), 2);
+        assert_eq!(merged.matches("\"type\":\"ping\"").count(), 2);
+        assert!(!merged.contains("\"type\":\"input_json_delta\""));
+        assert!(merged.contains("event: error"));
+        assert!(merged.contains("stream_truncated"));
     }
 
     #[tokio::test]


liuxingbo12138:
## Live end-to-end validation / 真实链路验证

Validated on macOS with Claude Code 2.1.210 through the CC Switch development build from this PR:

- branch: `agent/fix-read-stream-liveness`
- commit: `5f59f9b6095d0c011a5f22b302bc8ea5e5ab3add`
- proxy listener: `127.0.0.1:15721`
- upstream: Codex OAuth / OpenAI Responses, model mapped to `gpt-5.5`

A captured `Read` tool call at 2026-07-15 15:06:26 produced this downstream sequence:

```text
response.function_call_arguments.delta
SSE data: {"type":"ping"}
response.function_call_arguments.delta
SSE data: {"type":"ping"}
... repeated ...
response.function_call_arguments.done
SSE data: {"type":"content_block_delta", ... "type":"input_json_delta" ...}
SSE data: {"type":"content_block_stop", ...}
```

Measured results for that call:

- upstream argument deltas: **15**
- downstream Anthropic ping events: **15**
- final sanitized `input_json_delta`: **1**
- empty `pages` field remained removed
- Claude Code completed the call without `API Error: The operation timed out.`

Across the development run, the proxy emitted **647** downstream ping events while buffering `Read` arguments, with no `operation timed out` match after the patched process started.

This confirms the live behavior matches the regression test: every buffered active `Read` argument delta now produces downstream progress, while the complete sanitized tool input is still emitted only once on `.done`.

No request bodies, credentials, account identifiers, or local file paths are included in this validation evidence.

farion1231:
@codex review

Issue #2992, opened 2026-05-20
Claude settings.json overwrite drops enabledPlugins and hooks


HisenWeb:
## Summary
When CC-switch updates Claude Code configuration on Windows, it appears to rewrite `C:/Users/zengqu/.claude/settings.json` without preserving existing keys such as `enabledPlugins` and `hooks`.

This makes installed Claude plugins appear to disappear, even though the plugin files are still installed on disk.

## Environment
- OS: Windows 11
- CC-switch: 3.15.0 (`D:/CC Switch/cc-switch.exe`)
- Claude Code config path: `C:/Users/zengqu/.claude/settings.json`

## Observed behavior
Before the problem, `settings.json` includes keys like:
- `env`
- `enabledPlugins`
- `hooks`

After the problem occurs, Claude plugins are no longer loaded, and the key symptom is that `enabledPlugins` is missing from `settings.json`.

Example effect:
- Plugins remain installed in `C:/Users/zengqu/.claude/plugins/installed_plugins.json`
- But Claude no longer loads them because `enabledPlugins` was removed from `settings.json`

## Evidence
Local investigation found:

1. CC-switch stores a minimal Claude common config in its local database under `common_config_claude`, with content like:
```json
{
  "autoUpdatesChannel": "latest"
}
```

2. `cc-switch.exe` contains strings related to Claude config integration, including:
- `common_config_claude`
- `enableClaudePluginIntegration`
- `claudeConfigDir`
- `currentProviderClaude`

3. `cc-switch.exe` does **not** appear to contain the string `enabledPlugins`, which suggests it may not preserve that field when writing Claude config.

## Expected behavior
CC-switch should update only the Claude-related fields it owns, and preserve unknown or existing keys in `settings.json`.

In particular, it should not remove:
- `enabledPlugins`
- `hooks`
- any other unrelated existing keys

## Suggested fix
Please change the config write logic from full-template overwrite to merge-update:
- read existing `settings.json`
- update only the fields CC-switch manages
- preserve all unknown/existing keys when writing back

## Impact
This causes repeated plugin disappearance symptoms for Claude Code users, even though the plugins are still installed.

bugficks:
Same on mac. This makes this app unusable! Overwriting any user settings without a backup/restore is really bad. I wonder what else it simply overwrites. 
Lost trust, deleted....

RobbinTech:
Same issue persists in v3.16.1.
[...]

PR #5412, opened 2026-07-15
fix(services): sync common config snippet for hot switch

### LINKED ISSUE #4274: CC Switch overwrites entire settings.json on agent takeover, destroying user plugins/config
### Self Checks / 自检

### CC Switch Version / 版本号

3.16.3

### Operating System / 操作系统

Windows

### Related App / 涉及应用

Claude Code

### Steps to Reproduce / 重现步骤

1.Configure any plugin (e.g., Claude HUD) in .claude/settings.json

2.Launch CC Switch and let it take over the agent

3.Return to Claude Code — the plugin no longer works

4.Inspect .claude/settings.json — it has been rewritten to a minimal, proxy-only payload

### Expected Behavior / 期望行为

1.Read the existing settings.json (if present)

2.Modify only the proxy-related fields (e.g., env, proxy)

3.Preserve all other existing configuration keys

4.Write back the fully merged JSON

### Actual Behavior / 实际行为

When CC Switch takes over a Claude Code agent, it performs a full overwrite of the project’s .claude/settings.json file. This silently destroys all non-proxy-related user configuration, including but not limited to:

1.enabledPlugins — enabled plugins (e.g., Claude HUD)

2.extraKnownMarketplaces — third-party plugin marketplace registrations

3.statusLine — status bar configuration

4.hooks — automation hooks

5.permissions — allowlisted tool permissions

Users lose their configuration without warning and must repeatedly restore it manually.

### Additional Context / 补充信息

Example logic:
// Pseudocode: safely update settings.json
const fs = require('fs');
const path = require('path');

const settingsPath = path.join(projectDir, '.claude', 'settings.json');

// 1. Read existing config — keep all of the user's custom settings
let existing = {};
try {
  existing = JSON.parse(fs.readFileSync(settingsPath, 'utf-8'));
} catch {
  // File doesn't exist or is malformed, start fresh
}

// 2. Shallow-merge proxy-related fields (or use a deep-merge library)
const merged = {
  ...existing,           // preserve all user config
  env: {                 // only overwrite proxy-related fields
    ...existing.env,     // preserve user-defined env vars
    ...proxyEnvChanges,  // overwrite/add proxy env vars
  },
};

// 3. Write back
fs.writeFileSync(settingsPath, JSON.stringify(merged, null, 2));
Core principle: only overwrite the fields you own — leave everything else untouched.

written by Claude Code

[cc-switch-issue.md](https://github.com/user-attachments/files/28962530/cc-switch-issue.md)

huaibovip:
## Summary / 概述

This PR syncs the common config snippet during hot switch for takeover/non-additive apps.

Previously, when hot-switching providers in takeover mode, the switch updated the current provider routing but did not synchronize the shared/common config snippet from the currently active live config source. This could leave the stored common config snippet stale after switching providers.

With this change:
- before returning from the hot-switch path, the service now attempts to sync the common config snippet;
- it reads from the live config first;
- if live config reading fails, it falls back to the takeover backup;
- if sync fails, it logs a warning and records it in switch warnings instead of blocking the switch flow.

## Fixes
#4274, #2992, #1706, #3871
|     Hot switch updates current provider only; common config snippet may remain stale.            | Hot switch also syncs common config snippet from live config or takeover backup.              |

## Checklist / 检查清单

- [ x ] `pnpm typecheck` passes / 通过 TypeScript 类型检查
- [ x ] `pnpm format:check` passes / 通过代码格式检查
- [ x ] `cargo clippy` passes (if Rust code changed) / 通过 Clippy 检查（如修改了 Rust 代码）


farion1231:
@codex review

PR #5341, opened 2026-07-14
fix(usage): prevent double-counting Codex fork history

Sjeary:
## Summary / 概述

Prevent Codex fork files from importing copied parent or ancestor token history
as new child-thread usage. The fix combines session-segment ownership with the
explicit replay boundary already used for Codex subagents.

修复 Codex 会话分支文件把父会话/祖先会话的复制历史再次计入新分支用量的问题。
本修复在现有 Codex 子代理 replay boundary 机制上，增加了按 `session_meta`
片段归属判断的兼容逻辑。

### Root cause / 根因

CC Switch 3.17.0 reads the first valid `session_meta` as the file thread and
relies on optional `thread_settings_applied` or `inter_agent_communication*`
events to delimit replayed history. Codex fork files do not always contain one
of those events, but they can contain multiple parent/ancestor `session_meta`
segments followed by cumulative `token_count` events. Those historical deltas
were therefore inserted under the fork's request-ID namespace.

CC Switch 3.17.0 只保存首个有效的 `session_meta` 身份，并依赖可选的接管事件
识别 replay 历史；无显式边界的 fork JSONL 会因此把祖先片段中的累计 token
增量当成当前分支的新用量写入数据库。

### What changed / 修改内容

- Track the file's `root_thread_id` separately from the most recent
  `active_thread_id`.
- Parse every valid `session_meta`; exclude non-root segments from insertion.
- Keep skipped replay events in `prev_total` and `event_index`, preserving the
  correct cumulative baseline and stable `codex_session:thread-v1` IDs.
- Prefer metadata matching the standard trailing UUID in Codex rollout
  filenames, with the existing first-valid-metadata fallback, so a malformed
  opening metadata record cannot turn genuine child usage into an undercount.
- Attribute genuine post-fork usage to the root fork thread.

### Compatibility / 兼容性

- Preserves the explicit subagent replay handling from #5187.
- Preserves normal sessions, missing-metadata fallback, nested forks,
  append-only incremental sync, archive cursor inheritance, token clamping,
  pricing, and proxy/session cross-source deduplication.
- No schema migration, UI change, or i18n change.
- This prevents future duplicate imports and fixes fresh re-imports. Automatic
  cleanup of already-imported historical duplicate rows is intentionally out
  of scope because that needs a separate migration design.

### Related work / 相关工作

#4982 identified the same fork structure, but its branch now conflicts with the
subagent accounting on current `main`. This PR supersedes that implementation
by integrating segment ownership with #5187's replay boundary and current
[...]

Issue #4177, opened 2026-06-14
Codex usage import double-counts forked/archived session history


zhulijin1991:
## Summary

CC Switch v3.16.2 can significantly over-count Codex usage after using Codex Desktop fork sessions. The usage importer appears to treat copied parent-session history inside forked/archived JSONL files as new Codex requests.

## Environment

- CC Switch: 3.16.2
- Platform: macOS
- Source: Codex Desktop JSONL session logs under `~/.codex/sessions` and `~/.codex/archived_sessions`

## What happened

After creating several Codex Desktop forks from a long-running parent session, the Usage Statistics page for `Codex (Session)` jumped by tens of thousands of rows and a very large token/cost total for the current day.

The inserted rows all used `data_source = codex_session` and placeholder provider `_codex_session`. A small number of forked/archived session files contributed thousands of rows each, and several files had near-identical aggregate token totals.

No private paths, workspace names, account ids, or raw session ids are included here, but the pattern was:

- several `~/.codex/archived_sessions/rollout-*.jsonl` files generated minutes apart from the same parent session
- each archived file contained thousands of `event_msg` / `token_count` events
- normal `~/.codex/sessions/YYYY/MM/DD/rollout-*.jsonl` fork files also embedded parent-session history
- each copied/forked file was imported as if it were an independent Codex session

## Likely cause

From the v3.16.2 source:

1. `collect_codex_session_files` scans both:
   - `~/.codex/sessions/YYYY/MM/DD/*.jsonl`
   - `~/.codex/archived_sessions/*.jsonl`

2. `sync_single_codex_file` only captures the first `session_meta` because of:

```rust
"session_meta" if state.session_id.is_none() => { ... }
```

Forked Codex JSONL files can contain copied parent-session history in addition to the child session. Because only the first `session_meta` is kept, parent history can be attributed to the child session id.

3. The importer computes deltas from `total_token_usage` per file, which is reasonable for a single linear session file but unsafe when a fork file embeds copied parent history.

4. The request id is generated as:

```rust
codex_session:{session_id}:{event_index}
```

So copied parent history in another fork/archived file gets a different request id and bypasses the request-id dedupe. The existing cross-source dedupe mainly checks matching proxy logs; it does not prevent duplicate `codex_session` imports across forked/archived session files.

## Reproduction outline

1. Have a long Codex Desktop session with many `token_count` events.
2. Create one or more forks from that session.
3. Let CC Switch sync Codex session usage.
[...]

PR #5413, opened 2026-07-15
fix(usage): prevent Codex subagent replay overcounting and add rescan

Eric-Terminal:
## Summary / 概述

- Prevent Codex subagent rollouts from importing copied parent token history as new usage.
- Serialize startup, periodic, manual, and rebuild syncs so a rescan cannot race with background ingestion.
- Add a confirmed **Rescan** action that clears only reconstructible session-derived usage, rollups, and cursors, then rebuilds them from local Claude, Codex, Gemini, and OpenCode logs.

Codex subagent rollouts contain a parent-history snapshot. That copied history may already contain `thread_settings_applied`, while the real child takeover is marked later by a top-level `inter_agent_communication*` event. The previous parser treated the copied settings event as the replay boundary and inserted the remaining parent history again.

This change continues consuming replayed cumulative token events to restore the delta baseline and stable event index, but only inserts subagent usage after the real takeover marker. A snapshot with no takeover marker remains baseline-only.

The rescan runs under the same sync mutex and clears session-derived rows in one transaction. It preserves raw chat files, proxy request logs, providers, settings, and other application data.

## Related Issue / 关联 Issue

Related to #4177 and #2571.

This takes a different approach from #5341: it does not filter usage by `active_thread_id`, so it preserves child usage in the legacy/fork scenario called out during that PR's review. Instead, it suppresses insertion only while parsing the copied pre-takeover snapshot.

## Screenshots / 截图

Not included. The UI change is a single Rescan button using the existing confirmation dialog.

## Testing / 测试

- [x] `cargo fmt --manifest-path src-tauri/Cargo.toml -- --check`
- [x] `cargo clippy --manifest-path src-tauri/Cargo.toml -- -D warnings`
- [x] `cargo test --manifest-path src-tauri/Cargo.toml` (1,978 library tests passed, 2 ignored; all integration targets passed)
- [x] `pnpm typecheck`
- [x] `pnpm format:check`
- [x] `pnpm exec vitest run tests/components/UsageDashboard.test.tsx`
- [x] Read-only audit against real local Codex rollouts using an in-memory database

## Checklist / 检查清单

### DIFF
--- a/src-tauri/src/commands/usage.rs
+++ b/src-tauri/src/commands/usage.rs
@@ -278,49 +278,15 @@ pub fn delete_model_pricing(state: State<'_, AppState>, model_id: String) -> Res
 pub fn sync_session_usage(
     state: State<'_, AppState>,
 ) -> Result<crate::services::session_usage::SessionSyncResult, AppError> {
-    // 同步 Claude 会话日志
-    let mut result = crate::services::session_usage::sync_claude_session_logs(&state.db)?;
+    crate::services::session_usage::sync_all_session_usage(&state.db)
+}
 
-    // 同步 Codex 使用数据
-    match crate::services::session_usage_codex::sync_codex_usage(&state.db) {
-        Ok(codex_result) => {
-            result.imported += codex_result.imported;
-            result.skipped += codex_result.skipped;
-            result.files_scanned += codex_result.files_scanned;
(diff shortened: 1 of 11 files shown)


lln556:
今天我也遇到了 半天给我统计出 77 亿 token 消耗

farion1231:
@codex review

PR #5454, opened 2026-07-16
fix(codex): preserve user-owned model catalog pointer

hyeluoh:
## Summary

- preserve a user-managed `model_catalog_json` from Codex common config instead of replacing it with `cc-switch-model-catalog.json`
- only replace or remove catalog pointers whose filename is owned by CC Switch
- preserve an existing external pointer during common-config autosync when an older live projection still contains the managed pointer
- allow an explicitly deleted live pointer to remain deleted
- skip generating an unused CC Switch catalog when an external catalog is the active source of truth

## Root cause

The common-config TOML merge already used source-wins semantics, so the provider editor preview showed the expected external catalog path. During the final live projection, however, `set_codex_model_catalog_json_field(..., Some(...))` unconditionally replaced that value with `cc-switch-model-catalog.json` whenever the provider still had an inline `modelCatalog`.

Common-config autosync also removed every `model_catalog_json` value, including user-managed external paths. On upgrade, a live file previously overwritten with the managed pointer could therefore erase the original external pointer from the stored common snippet during the next switch.

This change applies the existing filename-based ownership boundary consistently to projection, extraction, and switch-time autosync.

## User impact

When a Codex common config explicitly points to an external model catalog, that pointer now remains effective across provider save, normal switch, proxy/takeover projection, and common-config autosync. CC Switch continues to generate and clean up its own catalog normally when no external pointer is configured.

## Validation

- `cargo test --lib --quiet` — 1981 passed, 2 ignored
- `./node_modules/.bin/vitest run` — 69 files, 446 tests passed
- `git diff --check`

## Related issue

Addresses #3302, specifically the `model_catalog_json` ownership and save-time override subcase.


farion1231:
@codex review

PR #5456, opened 2026-07-16
fix(webdav): send default User-Agent for reverse-proxy compatibility

YoungSx:
## Summary
- WebDAV transport used the shared `reqwest` client, which does not send a `User-Agent` by default.
- Cloudflare and similar reverse proxies may reject those requests with `403 Access denied: Missing required headers`.
- Attach a stable `cc-switch/<version> WebDAV` User-Agent on all WebDAV requests via `apply_auth`.

## Root cause
Verified against a Cloudflare-fronted WebDAV endpoint with the same `reqwest` stack:

| Request | Result |
|---|---|
| `reqwest` default (no User-Agent) | `403 Missing required headers` |
| `reqwest` + `User-Agent: cc-switch/3.17.0 WebDAV` | `207` PROPFIND / `201` PUT |
| `curl` default (sends its own UA) | success, which can mask the bug |

## Test plan
- [x] Reproduce 403 with bare `reqwest` against Cloudflare-proxied WebDAV
- [x] Confirm 207/201 after adding User-Agent header
- [x] Unit test for `WEBDAV_USER_AGENT` product token
- [ ] Manual: Settings → Advanced cloud sync → WebDAV → save & test against Cloudflare reverse proxy
- [ ] Manual: upload/download still works for direct WebDAV URLs (Jianguoyun / Nextcloud)

### DIFF
--- a/src-tauri/src/services/webdav.rs
+++ b/src-tauri/src/services/webdav.rs
@@ -3,6 +3,7 @@
 //! Low-level HTTP primitives for WebDAV operations (PUT, GET, HEAD, MKCOL, PROPFIND).
 //! The sync protocol logic lives in [`super::webdav_sync`].
 
+use reqwest::header::USER_AGENT;
 use reqwest::{Method, RequestBuilder, StatusCode, Url};
 use std::time::Duration;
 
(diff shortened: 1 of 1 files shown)


farion1231:
@codex review

PR #5465, opened 2026-07-17
feat(claude): route subagents across providers

eonewg:
## Summary

Add cross-provider routing for Claude Code subagents in CC Switch.

Each Claude provider already has its own `CLAUDE_CODE_SUBAGENT_MODEL`. This change lets an active provider select another saved Claude provider as its subagent route target. Main-session requests stay on the active provider; only subagent requests use the target provider and its configured subagent model.

## Design

- Persists only `{ providerId }` in `ProviderMeta.claudeSubagentRoute`; no credentials are copied.
- Adds a Subagent Provider selector and resolved target-model display to the Claude provider form.
- Requires Claude local proxy takeover for cross-provider routing; direct switching remains unchanged.
- During takeover, writes a CC Switch-owned model alias to live `CLAUDE_CODE_SUBAGENT_MODEL`.
- Resolves that alias before normal failover selection, pins the request to the explicit target provider, and rewrites the outbound model to the target's `CLAUDE_CODE_SUBAGENT_MODEL`.
- Uses the target provider's endpoint, auth, API format, transforms, model mapping, circuit breaker, and usage attribution.
- Suppresses global provider/UI/tray hot switching for intentionally routed subagent requests.
- Does not inherit the active provider's failover queue and never silently falls back when the target or target model is missing.
- Rejects reserved-alias collisions and preserves ordinary no-route behavior.

## Validation and compatibility

- Current-provider/default selection omits route metadata.
- Deleted targets, missing target subagent models, and reserved-alias model conflicts block save with localized errors.
- Existing no-route and same-provider behavior remains unchanged.
- Added English, Simplified Chinese, Traditional Chinese, and Japanese UI strings.

## Tests

Local verification:

- Focused frontend tests: 28 passed
- Full frontend suite: 70 files / 464 tests passed
- `tsc --noEmit`: passed
- Prettier: passed
- `cargo fmt --all -- --check`: passed
- Cargo metadata: passed
- `git diff --check`: passed
- Credential-pattern scan: no real tokens found

Backend regression coverage includes explicit-target routing, model rewrite, empty/all-open active failover queue bypass, missing target/model failures, reserved alias collision, no-route compatibility, takeover/hot-switch alias updates, and global-switch suppression.

Full Rust build/tests could not run locally because this WSL image lacks `pkg-config`, GTK/WebKit, and OpenSSL development packages. Backend CI is therefore required before merge.


eonewg:
Full CI validation passed on the fork for commit `49a269edc8c893f3ac37544a9139e9a47c2098c8`: frontend checks plus Rust formatting, Clippy, and tests on Ubuntu, macOS, and Windows. Run: https://github.com/eonewg/cc-switch/actions/runs/29552030847\n\nThe upstream Actions run is currently awaiting maintainer approval because this is a first-time fork contribution.

farion1231:
@codex review

PR #5485, opened 2026-07-17
fix(claude-desktop): preserve profile state on switch

Komikawayi:
## Summary
Preserve non-gateway Claude Desktop profile fields while refreshing cc-switch gateway fields during a provider switch. A malformed existing profile is safely replaced with the generated gateway profile so switching remains a repair path.

## Validation
- `cargo test --manifest-path src-tauri/Cargo.toml --lib claude_desktop_config::tests::` (25 passed)
- `git diff --check`

Fixes #5417


farion1231:
@codex review

PR #5486, opened 2026-07-17
fix(services): sync common config on hot switch

Komikawayi:
## Summary
Sync the opted-in common configuration snippet when switching providers during proxy takeover. The source is the live configuration when readable, otherwise the takeover backup; a sync error remains non-fatal and becomes a switch warning.

## Validation
- `git diff --check`

Fixes #5412


farion1231:
@codex review

PR #5522, opened 2026-07-18
feat(opencode): load OMO models from runtime opencode models

### LINKED ISSUE #1292: 无法使用opencode内置登录认证的模型
## 我的配置

- Windows 11 25H2
- Opencode Desktop 1.2.17（最新版）
- 插件：oh-my-opencode

## 问题

对于我使用omo，cc-switch无法读到我的auth login的各个模型，而我又没有配置custom provider（写入~/.config/opencode/opencode.json的自定义provider配置），因而认为omo配置项内容无效

实际上对于内置登录，opencode维护了一个位于~/.local/share/opencode/的auth.json文件，相对应的对于我这种desktop安装用户维护了一个模型列表备份cache为位于~/.cache/opencode/的models.json文件，想要保证cache文件完全备份准确可能很困难，但是确认已有的有效build-in provider是可行的

## 个人建议

1. 读取这部分已有auth的provider，将这部分provider作为cc-switch中可读取查看但不可修改的部分显示为一类，原有的自定义部分显示为另一类
2. 对于难以确认cache内容准确的问题，我认为可以考虑在遇到对应provider时直接选择不检查其内容模型是否有效，也可以考虑跟随opencode更新维护一份最新的稳定的models.json列表，或者直接忽略其可能存在的准确性问题，直接调用cache中的models.json来确认对应模型的合法性

### LINKED ISSUE #4747: opencode model selector 不显示oauth模型
### Self Checks / 自检

### CC Switch Version / 版本号

3.16.4

### Operating System / 操作系统

Windows

### Related App / 涉及应用

OpenCode

### Steps to Reproduce / 重现步骤

1. 登录任意openai账户
2. 模型选择器

### Expected Behavior / 期望行为

1. 显示所有模型，包括oauth模型

### Actual Behavior / 实际行为

_No response_

### Additional Context / 补充信息

_No response_

allenxu09:
﻿## Summary
OMO / OMO Slim model pickers previously only listed models from configured OpenCode providers (plus live-provider filtering). That misses models OpenCode itself already considers available at runtime — notably Zen free models and OAuth-backed models that are not fully represented in `opencode.json` / cc-switch provider configs.

This change discovers those models by running the installed OpenCode CLI (`opencode models`) and merges them into the OMO picker options.

## Changes
- Add Tauri command `get_opencode_models` that runs `opencode models` via the existing CLI install locator (Windows / WSL / Unix)
- Parse `providerId/modelId` lines, dedupe, and expose them to the frontend
- Merge runtime models into `useOmoModelSource` for OMO and OMO Slim forms
- Enforce a 20s timeout and kill hung CLI processes
- Show a warning toast when runtime discovery fails, while keeping configured-provider models
- Allow runtime models to appear even while live/provider queries are still loading
- Add i18n for the discovery-failure warning (en / zh / zh-TW / ja)

## Impact
Users configuring OpenCode OMO / OMO Slim agents can select models that OpenCode can already run (including OAuth and Zen free models), not only models present in managed provider configs.

Normal non-OMO OpenCode provider forms and other apps are unaffected.

### DIFF
--- a/src-tauri/Cargo.lock
+++ b/src-tauri/Cargo.lock
@@ -783,6 +783,7 @@ dependencies = [
  "indexmap 2.13.0",
  "json-five",
  "json5",
+ "libc",
  "log",
  "objc2 0.5.2",
  "objc2-app-kit 0.2.2",
--- a/src-tauri/Cargo.toml
+++ b/src-tauri/Cargo.toml
@@ -88,6 +88,9 @@ tauri-plugin-single-instance = "2"
 [target.'cfg(target_os = "linux")'.dependencies]
 webkit2gtk = { version = "2.0.1", features = ["v2_16"] }
 
+[target.'cfg(not(target_os = "windows"))'.dependencies]
+libc = "0.2"
+
 [target.'cfg(target_os = "windows")'.dependencies]
 winreg = "0.52"
 windows-sys = { version = "0.61", features = ["Win32_Globalization", "Win32_UI_Shell"] }
--- a/src-tauri/src/commands/misc.rs
+++ b/src-tauri/src/commands/misc.rs
@@ -2513,13 +2513,13 @@ pub(crate) fn run_detected_tool_command_with_timeout(
 
     #[cfg(target_os = "windows")]
     {
-        return run_windows_tool_command_capture(
+        run_windows_tool_command_capture(
             tool_path,
             args,
             &format!("{};{current_path}", dir.display()),
             timeout,
             extra_env,
-        );
+        )
     }
 
     #[cfg(not(target_os = "windows"))]
--- a/src-tauri/src/commands/misc.rs
+++ b/src-tauri/src/commands/misc.rs
@@ -2437,10 +2437,7 @@ fn wait_child_output(
                             let _ = child.wait();
                             let _ = stdout_handle.map(|handle| handle.join());
                             let _ = stderr_handle.map(|handle| handle.join());
-                            return Err(format!(
-                                "Command timed out after {}s",
-                                limit.as_secs()
-                            ));
+                            return Err(format!("Command timed out after {}s", limit.as_secs()));
                         }
                         std::thread::sleep(std::time::Duration::from_millis(50));
                     }
--- a/src-tauri/src/commands/misc.rs
+++ b/src-tauri/src/commands/misc.rs
@@ -1829,49 +1829,67 @@ fn first_abs_path_line(raw: &str) -> Option<&str> {
 /// 用与 `try_get_version` 相同的登录 shell 解析 PATH 默认命中的可执行文件路径，
 /// canonicalize 后作为"命令行默认 / 升级目标"的锚点（与升级会作用的那处对齐）。
 #[cfg(not(target_os = "windows"))]
-fn resolve_path_default(tool: &str) -> Option<std::path::PathBuf> {
-    use std::process::Command;
+fn resolve_path_default(
+    tool: &str,
+    deadline: Option<CommandDeadline>,
+) -> Result<Option<std::path::PathBuf>, String> {
+    use std::process::{Command, Stdio};
+
     let shell = std::env::var("SHELL")
         .ok()
         .filter(|s| is_valid_shell(s))
         .unwrap_or_else(|| "sh".to_string());
     let flag = default_flag_for_shell(&shell);
-    let out = Command::new(shell)
+    let child = Command::new(shell)
         .arg(flag)
         .arg(format!("command -v {tool}"))
-        .output()
-        .ok()?;
+        .stdout(Stdio::piped())
+        .stderr(Stdio::piped())
+        .spawn()
+        .map_err(|e| format!("Failed to locate {tool}: {e}"))?;
+    let out = wait_child_output(child, deadline)?;
     if !out.status.success() {
-        return None;
+        return Ok(None);
     }
     let raw = decode_command_output(&out.stdout);
     // 不能死取第一行：交互式 .zshrc 可能先打印欢迎语（如 "🚀 Welcome back"），
     // command -v 的真实路径在其后；取第一个 `/` 开头的行才稳。
-    let first = first_abs_path_line(&raw)?;
-    std::fs::canonicalize(first).ok()
+    let Some(first) = first_abs_path_line(&raw) else {
+        return Ok(None);
+    };
+    Ok(std::fs::canonicalize(first).ok())
 }
 
 #[cfg(target_os = "windows")]
-fn resolve_path_default(tool: &str) -> Option<std::path::PathBuf> {
+fn resolve_path_default(
+    tool: &str,
+    deadline: Option<CommandDeadline>,
+) -> Result<Option<std::path::PathBuf>, String> {
     use std::os::windows::process::CommandExt;
-    use std::process::Command;
-    let out = Command::new("cmd")
+    use std::process::{Command, Stdio};
+
+    let child = Command::new("cmd")
         .args(["/C", &format!("where {tool}")])
         .creation_flags(CREATE_NO_WINDOW)
-        .output()
-        .ok()?;
+        .stdout(Stdio::piped())
+        .stderr(Stdio::piped())
+        .spawn()
+        .map_err(|e| format!("Failed to locate {tool}: {e}"))?;
+    let out = wait_child_output(child, deadline)?;
     if !out.status.success() {
-        return None;
+        return Ok(None);
     }
     let raw = decode_command_output(&out.stdout);
-    let first = raw.lines().next()?.trim();
+    let Some(first) = raw.lines().next().map(str::trim) else {
+        return Ok(None);
+    };
     if first.is_empty() {
-        return None;
+        return Ok(None);
     }
     let path = Path::new(first);
     let preferred =
         windows_runnable_sibling_for_extensionless_tool(path).unwrap_or_else(|| path.to_path_buf());
-    std::fs::canonicalize(preferred).ok()
+    Ok(std::fs::canonicalize(preferred).ok())
 }
 
 /// 枚举工具在系统中的所有安装（不短路）。与 `scan_cli_version` 共用
@@ -1885,7 +1903,7 @@ fn enumerate_tool_installations(tool: &str) -> Vec<ToolInstallation> {
     let current_path = std::env::var_os("PATH")
         .map(|value| value.to_string_lossy().into_owned())
         .unwrap_or_default();
-    let path_default = resolve_path_default(tool);
+    let path_default = resolve_path_default(tool, None).ok().flatten();
 
     let mut seen: std::collections::HashSet<std::path::PathBuf> = std::collections::HashSet::new();
     let mut installs: Vec<ToolInstallation> = Vec::new();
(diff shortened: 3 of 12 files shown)


allenxu09:
@codex review

farion1231:
Thanks for this — the CLI-execution plumbing is careful: tool allow-list + arg charset validation, two independent threads draining stdout/stderr to avoid pipe deadlock, a 20s timeout, and four-locale i18n. The direction looks good to merge. Below is a close pass: one item worth fixing before merge, two "hang forever" robustness gaps, and a couple of design points to confirm.

One framing that ties most of these together: this PR generalizes the **tool-version-probe** subsystem into a **general CLI executor**, but the three things an executor needs — cwd isolation, a timeout that spans the whole flow, and process-tree termination — aren't carried over yet.

## 🔴 Worth fixing before merge

**Project-level config (and plugins) leak into runtime discovery.** `get_opencode_models` spawns `opencode models` without setting a working directory, and `OPENCODE_CONFIG_DIR` only overrides the *global* config dir — it does not disable project-level discovery. OpenCode searches from cwd upward to the nearest git root, merging any `opencode.json(c)` / `.opencode/` it finds. So if cc-switch is launched from a project directory, the OMO picker can surface models only that project defines — and, more importantly, **the project's `.opencode/` plugins get loaded/executed** as a side effect of opening the form.

Fix: set `OPENCODE_DISABLE_PROJECT_CONFIG="true"` (a real env var designed exactly for external tooling — see anomalyco/opencode#7559; note the value is the string `"true"`), and additionally pin `.current_dir(...)` to a neutral directory as a fallback for older OpenCode builds that don't know the flag.

## 🟠 "Hang forever" robustness gaps

The 20s timeout is currently a false promise on two paths. These are fine as a follow-up, but worth flagging:

1. **Timeout doesn't cover install probing.** `run_detected_tool_command_with_timeout` calls `enumerate_tool_installations` **before** the timed wait, and that function runs `--version` on every candidate with an un-timed `.output()` (`misc.rs:1942` on Unix; `misc.rs:1940` → `run_windows_tool_version_command` on Windows). A hung/broken shim blocks the request before the 20s clock ever starts. (This hang surface predates the PR — it's shared with the About-page version probe — but the new command advertises a timeout it can't honor.)

2. **Windows can't terminate the full process tree.** On timeout, `wait_child_output` only `child.kill()`s the direct child. For the common `opencode.cmd → cmd.exe → node.exe` chain, `node.exe` survives, keeps the stdout/stderr pipe open, so `read_to_end` never hits EOF and the following `join()` blocks forever (leaking the `spawn_blocking` thread). Needs a Job Object / `taskkill /T`; the WSL path needs in-distro termination too.

Ideal fix for both: a single deadline spanning probe → exec → kill, plus process-tree termination.

## 🟡 Please confirm

- **Runtime models bypass the live filter.** Configured providers are filtered by `liveSet`, but `discoveredModels` are merged unconditionally, so a provider disabled on the cc-switch side can still appear in the picker. The PR says this is intentional — just confirming it can't surface a model whose actual routing would then fail.
- **Parser can silently yield an empty list.** `parse_opencode_models` applies strict per-line charset checks; if a given `opencode models` build emits ANSI color codes or a trailing description column, those lines are dropped — worst case all lines drop, the command still reports `success`, and the frontend shows nothing with no warning. Could you paste a real `opencode models` output to confirm the format? And consider surfacing a hint when the command succeeds but 0 models parse.

## 🟢 Quick win

- **WSL rejects valid paths with spaces.** `env KEY=VALUE` is passed as a standalone argv, so spaces don't break parsing — rejecting all whitespace over-restricts (e.g. `/mnt/c/Users/Jane Doe/...` fails). Allow spaces (keep rejecting control chars / newlines) and add a WSL path test.

## ✅ Verified correct while reviewing

The `--version` refactor is behavior-equivalent (`"--version"` is quoted but `CommandLineToArgvW` strips it), the WSL UNC→Linux translation is self-consistent (only reached when the override dir is UNC, and `get_opencode_dir()` returns that same UNC path), the command-injection surface is zero, the two-thread pipe draining correctly avoids deadlock, i18n is synced across all four locales, and the front/back `OpenCodeModelRef` camelCase types line up.

allenxu09:
感谢您详细的 review，已处理提到的两个 “Hang forever” 问题：

- runtime model discovery 不再通过 `enumerate_tool_installations` 对所有候选执行无超时的 `--version`，20 秒 deadline 现在覆盖 CLI 定位和实际执行。
- Windows 超时时使用 `taskkill /T /F` 清理完整进程树，并避免在超时路径上无限等待 stdout/stderr reader；WSL 侧也增加了发行版内部的 `timeout`，用于清理 Linux 子进程。

farion1231 (commented):
Re-checked `f965b99d`. The two hang fixes landed well — I verified both in detail: the `CommandDeadline` now genuinely spans locate → exec (no more un-timed per-candidate `--version` probing), and the termination story on Windows/WSL is right (`taskkill /T /F` with a `kill()` fallback, in-distro `timeout` with exit-124 mapping, and the deliberate no-join on the timeout path is correctly reasoned). I also closed out my two round-1 "please confirm" items on my end: upstream `cli/cmd/models.ts` does print bare `provider/model` lines, so the parser's format assumption holds; and the live-filter bypass is acceptable since the runtime list is by definition OpenCode's own live state — OMO profiles never materialize as providers in `opencode.json`, so there's no self-reference risk either. CI is green on all three platforms and the branch merges cleanly.

That said, the round-1 blocker is still open, and the re-review turned up a few more items worth taking while you're in here.

## [P1] Project config discovery / plugin execution still leaks in (round-1 blocker)

`get_opencode_models` still only sets `OPENCODE_CONFIG_DIR` and inherits the app's working directory: [model_fetch.rs#L25-L35](https://github.com/farion1231/cc-switch/blob/f965b99d805b8a58e1cfb1893d3e1ed561815a25/src-tauri/src/commands/model_fetch.rs#L25-L35). OpenCode walks up from cwd merging project `opencode.json(c)` and loading `.opencode/` assets, and `models` boots the provider/plugin machinery — so opening the OMO form from inside a project directory can both skew the picker and execute that project's plugins as a side effect.

One update since round 1: per the upstream discussion ([anomalyco/opencode#7559](https://github.com/anomalyco/opencode/issues/7559)), `OPENCODE_DISABLE_PROJECT_CONFIG=true` is scoped to project-level `opencode.json` / `AGENTS.md` discovery — project `.opencode/` assets aren't clearly covered by the flag. So cwd pinning is load-bearing here, not a compatibility fallback for older builds. Please do both:

- pass `OPENCODE_DISABLE_PROJECT_CONFIG=true` (string `"true"`) in `extra_env`, and
- give `run_detected_tool_command_with_timeout` a pinned working directory — a neutral dir with no project config above it works (the OpenCode config dir itself is a fine choice) — applied on all three paths (`current_dir` on Unix/Windows, `wsl.exe --cd` on WSL).

## [P2] The 20s promise still has two escape hatches in `wait_child_output`

1. **Unbounded join on the normal-exit path.** When the child exits within the deadline but a descendant inherited the stdout write end, `read_to_end` never sees EOF and the final `join()` blocks forever: [misc.rs#L2562-L2567](https://github.com/farion1231/cc-switch/blob/f965b99d805b8a58e1cfb1893d3e1ed561815a25/src-tauri/src/commands/misc.rs#L2562-L2567). The timeout branch documents exactly this scenario and correctly drops the handles ([misc.rs#L2539](https://github.com/farion1231/cc-switch/blob/f965b99d805b8a58e1cfb1893d3e1ed561815a25/src-tauri/src/commands/misc.rs#L2539-L2541)) — the normal-exit branch needs the same protection. Minimal fix: poll `handle.is_finished()` under the same deadline and drop on expiry.

2. **POSIX has no tree termination.** `terminate_child_tree` on non-Windows is a bare `child.kill()` — direct child only: [misc.rs#L2495-L2498](https://github.com/farion1231/cc-switch/blob/f965b99d805b8a58e1cfb1893d3e1ed561815a25/src-tauri/src/commands/misc.rs#L2495-L2498). Windows got `taskkill /T`, WSL got in-distro `timeout`, but on macOS/Linux — the majority platform here — descendants survive as orphans. Minimal fix: put the child in its own process group (`pre_exec` + `setsid`) and signal the negative pgid.

If you'd rather keep the PR contained, (2) can be an immediate follow-up — but (1) is small and guards the exact hang class this revision set out to close, so I'd take it here.

## [P2] `["opencode", "runtime-models"]` is an orphan query key

Nothing ever invalidates it: saving settings (including the OpenCode config-dir override) only invalidates `["settings"]` ([mutations.ts#L391](https://github.com/farion1231/cc-switch/blob/f965b99d805b8a58e1cfb1893d3e1ed561815a25/src/lib/query/mutations.ts#L391)), and switching the active OpenCode provider invalidates `["providers","opencode"]` / `["opencodeLiveProviderIds"]` / the `omo*` keys but not this one ([mutations.ts#L272-L291](https://github.com/farion1231/cc-switch/blob/f965b99d805b8a58e1cfb1893d3e1ed561815a25/src/lib/query/mutations.ts#L272-L291)). With `staleTime: 5min` ([useOmoModelSource.ts#L50-L61](https://github.com/farion1231/cc-switch/blob/f965b99d805b8a58e1cfb1893d3e1ed561815a25/src/components/providers/forms/hooks/useOmoModelSource.ts#L50-L61)), the picker can show the previous provider's — or previous config dir's — models for up to five minutes after a switch. Provider switching is the common case: `opencode models` reads exactly that live state. Minimal fix: invalidate the key in both mutations, or fold the effective config dir into the query key.

## nit (round-1 quick win, still open) WSL rejects valid paths with spaces

The env validation still rejects all whitespace in values ([misc.rs#L2778-L2788](https://github.com/farion1231/cc-switch/blob/f965b99d805b8a58e1cfb1893d3e1ed561815a25/src-tauri/src/commands/misc.rs#L2778-L2788)). `KEY=value` is passed to `env` as a standalone argv, so spaces can't split words — a `/mnt/c/Users/Jane Doe/...` config dir fails for no reason. Reject control chars / NUL only, and a small test would lock it in.

## nits

- WSL + non-UNC config dir: when the value isn't a `\\wsl$` UNC path it's passed through verbatim ([misc.rs#L2777](https://github.com/farion1231/cc-switch/blob/f965b99d805b8a58e1cfb1893d3e1ed561815a25/src-tauri/src/commands/misc.rs#L2777)), so a host-side `C:\...` default lands inside Linux as a garbage path. Skip the env (or translate via `wslpath`) when UNC translation fails.
- `parse_opencode_models` is strict on the provider side but only rejects whitespace in `model_id` ([model_fetch.rs#L69](https://github.com/farion1231/cc-switch/blob/f965b99d805b8a58e1cfb1893d3e1ed561815a25/src-tauri/src/commands/model_fetch.rs#L69)) — control chars (e.g. stray ANSI sequences) can slip through. Add `!c.is_control()`.
- `data: discoveredModels = []` ([useOmoModelSource.ts#L51](https://github.com/farion1231/cc-switch/blob/f965b99d805b8a58e1cfb1893d3e1ed561815a25/src/components/providers/forms/hooks/useOmoModelSource.ts#L51)) mints a fresh array identity every render while the query is pending, and it's a `useMemo` dependency — hoist a module-level empty constant so the memo doesn't recompute per render.

## Summary

The P1 is the only merge blocker: with `OPENCODE_DISABLE_PROJECT_CONFIG` + cwd pinning in place I'm happy to merge — the runtime-discovery direction is right and the executor plumbing is in good shape now. The P2s can land here or as immediate follow-ups, your call. Ping me when it's ready and I'll re-review quickly.

allenxu09:
@farion1231 Addressed all items from the latest review in `92ff3ebd`:

- isolated runtime model discovery from project config and pinned the working directory on native Windows, Unix, and WSL
- bounded pipe-reader joins and added POSIX process-group termination
- invalidated runtime-model caches after provider switches and settings saves
- fixed WSL env/path handling, model control-character validation, and the pending-query empty-array identity

Could you please take another look?

farion1231 (approved):

PR #5525, opened 2026-07-18
fix(codex): build custom catalogs from current CLI schema

### LINKED ISSUE #5524: Codex routing can fail when models_cache uses an older schema
## Self checks

- [x] I have read the FAQ.
- [x] I searched open and closed issues for `model_catalog_json`, `models_cache`, and `supports_reasoning_summaries`.

## Environment

- CC Switch: 3.17.0
- OS: macOS
- Related app: Codex
- Codex CLI: 0.144.3

## Steps to reproduce

1. Have a `~/.codex/models_cache.json` written by an older Codex CLI that does not contain a field now required by the current CLI (for example, `supports_reasoning_summaries`).
2. Configure a routed Codex provider in CC Switch.
3. Enable Codex proxy takeover and activate that provider.
4. Start Codex, which reads the generated custom `model_catalog_json`.

## Expected behavior

CC Switch should generate the custom catalog from a schema compatible with the currently installed Codex CLI, while preserving useful values from the user's cache.

## Actual behavior

CC Switch currently prefers `models_cache.json` as the entire template. If that cache is stale, the generated custom catalog omits newly required fields and Codex fails before sending a request, for example:

```text
failed to parse model_catalog_json ... missing field supports_reasoning_summaries
```

This makes a provider that requires routing appear unresponsive after activation.

## Additional context

The failure was reproduced with a routed OpenAI Chat Completions-compatible provider. Generating the template from `codex debug models --bundled` in an isolated `CODEX_HOME`, then recursively overlaying cached values, restores routing while avoiding dependence on the possibly broken live Codex config.

toofat2003:
## Summary / 概述

Fix Codex custom model catalog generation when `models_cache.json` was written by an older Codex CLI.

CC Switch previously used the cache as the complete catalog template. When a newer Codex release introduced required fields such as `supports_reasoning_summaries`, the generated `model_catalog_json` could no longer be parsed and routed providers appeared unresponsive.

This change:

- reads `codex debug models --bundled` from the currently installed CLI using an isolated temporary `CODEX_HOME`, so a broken live config cannot block recovery;
- falls back to the compile-time template when the CLI is unavailable;
- recursively overlays cached object values onto that schema baseline, preserving user/cache values while completing newly required fields;
- adds regression tests for CLI priority, nested completion, static fallback, and cache type/null behavior.

The change is limited to `src-tauri/src/codex_config.rs`; routing and history migration behavior are unchanged.

## Related Issue / 关联 Issue

Fixes #5524

## Screenshots / 截图

Not applicable — backend-only change with no UI changes.

## Validation / 验证

- `pnpm typecheck`
- `pnpm format:check`
- `pnpm test:unit`
- `cargo test --lib` — 2019 passed, 2 ignored
- Built and ran the macOS app locally
- Verified a routed Chat Completions provider, the official OpenAI provider, and the routed provider again; all returned HTTP 200 after switching
- Verified recent Codex history IDs remained unchanged across the provider round trip

The repository's Windows/Linux CI matrix remains the cross-platform validation gate for the temporary `CODEX_HOME` and CLI process behavior.

## Checklist / 检查清单

- [x] i18n update not required because there are no user-facing text changes / 无用户可见文本变更，无需更新 i18n

AI-assisted development was used; the diff was reviewed and tested locally before submission.

### DIFF
--- a/src-tauri/src/codex_config.rs
+++ b/src-tauri/src/codex_config.rs
@@ -805,10 +805,24 @@ fn codex_cli_candidates() -> Vec<PathBuf> {
 }
 
 fn load_codex_model_template_from_bundled() -> Result<Option<Value>, AppError> {
+    // `debug models --bundled` must not parse the user's live config. In
+    // particular, this lookup is used to repair catalogs that may already make
+    // the configured `model_catalog_json` unreadable to the current CLI.
+    let isolated_codex_home = match tempfile::tempdir() {
+        Ok(dir) => dir,
+        Err(err) => {
+            log::warn!(
+                "Failed to create isolated CODEX_HOME for model catalog lookup; using static fallback: {err}"
+            );
+            return Ok(None);
+        }
+    };
+
     for candidate in codex_cli_candidates() {
         let candidate_label = candidate.to_string_lossy();
         let output = match Command::new(&candidate)
             .args(["debug", "models", "--bundled"])
+            .env("CODEX_HOME", isolated_codex_home.path())
             .output()
         {
             Ok(output) => output,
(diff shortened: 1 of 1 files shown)


farion1231:
@codex review

PR #5556, opened 2026-07-20
feat(proxy): optionally intercept Claude Desktop startup probe

Q1Xuan:
## Summary

- add an opt-in setting to intercept Claude Desktop's fixed startup probe locally
- strictly match only the current non-streaming `claude-haiku-4-5` / `max_tokens: 1` / single `.` request shape
- return an Anthropic-compatible success response before provider selection, forwarding, failover, and usage logging
- expose the setting in Local Routing with English, Simplified Chinese, Traditional Chinese, and Japanese text

## Motivation

Claude Desktop currently sends a real inference request when a third-party profile starts:

```json
{
  model: claude-haiku-4-5,
  max_tokens: 1,
  messages: [{ role: user, content: . }]
}
```

Some community or公益 API providers explicitly disallow health-check inference. The existing model-discovery behavior only covers `GET /v1/models`, so this startup request is still forwarded upstream and may time out, consume quota, or violate provider policy.

## Safety

- disabled by default
- limited to the authenticated Claude Desktop gateway route
- rejects streaming requests, extra top-level fields, extra message fields, different models, different token limits, and different content
- all non-matching requests continue through the existing forwarding path unchanged
- when enabled, provider availability is deferred to the first real user request

## Validation

- `cargo fmt --manifest-path src-tauri/Cargo.toml`
- TypeScript type check passed
- Prettier check passed
- targeted frontend tests: 18 passed
- focused Rust tests added for exact matching, near-miss forwarding, and response shape (also exercised by CI)

### DIFF
--- a/src-tauri/src/proxy/claude_desktop_probe.rs
+++ b/src-tauri/src/proxy/claude_desktop_probe.rs
@@ -1,0 +1,64 @@
+use serde_json::{json, Value};
+
+const STARTUP_PROBE_MODEL: &str = "claude-haiku-4-5";
+
+pub(super) fn is_startup_probe(body: &Value) -> bool {
+    let Some(object) = body.as_object() else {
+        return false;
+    };
+
+    if !object
+        .keys()
+        .all(|key| matches!(key.as_str(), "model" | "max_tokens" | "messages" | "stream"))
+    {
+        return false;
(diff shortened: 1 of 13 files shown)


farion1231:
@codex review

PR #5572, opened 2026-07-20
fix(terminal): restore Ghostty session resume via --input=raw:

### LINKED ISSUE #5557: Ghostty session restore regression in v3.16.5 — still uses `login -flp` wrapper
## Describe the bug

The Ghostty session restore feature (v3.16.5) still passes the command using the `/usr/bin/login -flp` wrapper, which Ghostty does not support.

This was supposedly fixed in v3.12.3 (PR #1506) but the session restore code path appears to be unaffected by that fix — it still uses the old Terminal.app/iTerm2 argument format.

## Steps to reproduce

1. Set Ghostty as the default terminal in CC-Switch Settings
2. Go to Session Manager
3. Click "Restore" on any Claude session
4. Ghostty opens but shows an error window

## Expected behavior

Session should restore in Ghostty using `--input=raw:` (as implemented in PR #1506) or equivalent Ghostty-compatible launch arguments.

## Actual error

```
Ghostty failed to launch the requested command:

/usr/bin/login -flp a1234 /bin/zsh -l -c claude --resume <session-id>

Runtime: 53 ms
```

## Environment

- **OS**: macOS 15 (24.6.0)
- **CC-Switch version**: v3.16.5
- **Ghostty**: latest
- **Provider**: Claude

## Additional context

The `login -flp` command format is specific to Terminal.app and iTerm2. Ghostty expects a raw shell command, not wrapped through `/usr/bin/login`. The fix in PR #1506 replaced this with `--input=raw:` for the terminal launch path, but the session restore path still uses the old wrapper.

Image

Image

feiwell:
## Summary

Fixes #5557. Ghostty session resume is broken again — `launch_ghostty` currently builds argv as `-e <shell> -l -c <command>`, but Ghostty always wraps its spawned process through `/usr/bin/login -flp <user> ...` on macOS. `login` calls `execvp` directly on the given argv without doing shell word-splitting, so the multi-word `-c` argument gets mis-split and the resume command fails to launch, exactly as reported:

```
/usr/bin/login -flp a1234 /bin/zsh -l -c claude --resume <session-id>
```

## Root cause

This exact bug was already fixed in #1506 by switching to Ghostty's `--input=raw:<text>` flag, which feeds the resume command as literal startup keystrokes instead of a command-line argument — sidestepping the `-e`/`login` argv path entirely.

#1976 later reverted this back to the `-e shell -l -c` form while fixing an unrelated cmux/`TERM_PROGRAM` issue, unintentionally reintroducing the original bug.

## Changes

- Restore `build_ghostty_args` / `ghostty_raw_input` using `--input=raw:<command>\n`, matching the #1506 approach.
- Re-add the corresponding unit tests that #1976 removed.
- Added a doc comment on `build_ghostty_args` explaining why `-e`/login-shell argv doesn't work for Ghostty, to prevent a third regression.

## Testing

- `cargo test --manifest-path src-tauri/Cargo.toml --lib session_manager::terminal` — 6 passed
- `cargo fmt --all --manifest-path src-tauri/Cargo.toml -- --check` — clean
- `cargo clippy --manifest-path src-tauri/Cargo.toml --lib -- -D warnings` — clean

### DIFF
--- a/src-tauri/src/session_manager/terminal/mod.rs
+++ b/src-tauri/src/session_manager/terminal/mod.rs
@@ -80,7 +80,33 @@ end tell"#
 }
 
 fn launch_ghostty(command: &str, cwd: Option<&str>) -> Result<(), String> {
-    let shell = std::env::var("SHELL").unwrap_or_else(|_| "/bin/zsh".to_string());
+    let args = build_ghostty_args(command, cwd);
+
+    let status = Command::new("open")
+        .args(args.iter().map(String::as_str))
+        .status()
+        .map_err(|e| format!("Failed to launch Ghostty: {e}"))?;
+
+    if status.success() {
+        Ok(())
+    } else {
+        Err("Failed to launch Ghostty. Make sure it is installed.".to_string())
+    }
+}
+
+/// Build the `open -na Ghostty --args ...` argument list for a session resume.
+///
+/// Ghostty always wraps its spawned shell through `/usr/bin/login` on macOS
+/// (see Ghostty's `execCommand`), which passes multi-word `-e <shell> -l -c
+/// <command>` argv arrays straight to `login` via `execvp`. `login` does not
+/// re-run shell word-splitting, so the resume command's arguments end up
+/// mis-split across argv boundaries and Ghostty fails to launch it (#5557).
+///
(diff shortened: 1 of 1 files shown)


farion1231:
@codex review

PR #5626, opened 2026-07-21
perf(codex): cache parent rollout timelines across fork cutoffs

### LINKED ISSUE #5625: perf(codex): cache parent rollout timelines across fork cutoffs
## Description

Codex usage rebuilds can parse the same parent rollout once for every child fork cutoff. `parent_signatures_before` currently caches by `(parent_path, cutoff)`, so children that share one parent but have different fork timestamps each reopen and deserialize the complete parent JSONL.

This is especially expensive for multi-agent sessions, where a large parent rollout may have dozens or hundreds of child rollouts.

## Reproduction

1. Create one parent rollout containing multiple `token_count` events.
2. Create multiple child rollouts that reference the same parent and use different root timestamps.
3. Rebuild Codex session usage.
4. Observe that the parent rollout is parsed once per distinct child cutoff and that the cache retains one full signature vector per cutoff.

## Impact

I measured the cache-key behavior against an anonymized local corpus:

- 1,379 session files totaling 87.141 GiB
- 1,314 parented child files
- 1,295 unique `(parent_path, cutoff)` keys
- 67 unique parent paths

A metadata/file-size model of the current scan produces:

- Source-file reads: 87.141 GiB
- Repeated parent reads: 386.307 GiB
- Total modeled reads: 473.449 GiB

Reading each parent once per file version would reduce the modeled total to 101.551 GiB, a reduction of 371.898 GiB (78.6%). These are modeled bytes opened from actual file sizes and cache keys, not a wall-clock benchmark.

## Proposed Fix

Cache one timestamped token-signature timeline per parent file version, keyed by path, modified time, and size. Different fork cutoffs can then filter the same timeline in memory. Preserve the existing correctness constraints by:

- scanning the complete parent because rollout timestamps are not guaranteed to be monotonic;
- rejecting a parent that has not reached the child fork timestamp;
- invalidating the cached timeline when the parent is appended or replaced;
- evicting stale versions so memory does not grow per append.

ayanamislover:
## Summary

- Build one timestamped token-signature timeline for each on-disk version of a Codex parent rollout and reuse it across child fork cutoffs.
- Preserve full-file parsing for non-monotonic timestamps, parent-completeness checks, and missing-timestamp validation.
- Invalidate timelines with descriptor metadata and platform file identity, while keeping only the latest child replay-prefix entry per path.
- Add regression coverage for actual cross-cutoff reuse, append and replacement invalidation, sub-microsecond cutoffs, and cached-content versus open-error behavior.

## Impact

On an anonymized 1,379-file corpus, the existing `(parent_path, cutoff)` cache produced 1,295 parent-cache entries for 67 unique parent paths. A file-size model estimated 473.449 GiB of JSONL input processing (87.141 GiB of source files plus 386.307 GiB of repeated parent parsing). Reusing one timeline per file version reduces the modeled total to 101.551 GiB, a 78.6% reduction.

A local synthetic benchmark with a 50,000-line, 35.53 MiB parent and 20 cutoffs took 27.37 seconds with forced reparsing and 1.38 seconds with timeline reuse (19.8x faster).

## Related issue

Fixes #5625

## Validation

- `cargo fmt --check`
- `cargo test services::session_usage_codex::tests --lib` (33 passed)
- `cargo test --lib` on the latest `main` merge state (2,252 passed, 2 ignored)
- `cargo check --target x86_64-pc-windows-gnu --lib`
- Clean no-commit merge into the latest `main`

### DIFF
--- a/src-tauri/Cargo.toml
+++ b/src-tauri/Cargo.toml
@@ -91,7 +91,11 @@ webkit2gtk = { version = "2.0.1", features = ["v2_16"] }
 
 [target.'cfg(target_os = "windows")'.dependencies]
 winreg = "0.52"
-windows-sys = { version = "0.61", features = ["Win32_Globalization", "Win32_UI_Shell"] }
+windows-sys = { version = "0.61", features = [
+    "Win32_Globalization",
+    "Win32_Storage_FileSystem",
+    "Win32_UI_Shell",
+] }
 
 [target.'cfg(all(target_os = "windows", target_arch = "aarch64"))'.dependencies]
 rquickjs = { version = "0.8", features = ["bindgen"] }
--- a/src-tauri/src/services/session_usage_codex.rs
+++ b/src-tauri/src/services/session_usage_codex.rs
@@ -30,7 +30,7 @@ use std::collections::HashMap;
 use std::fs;
 use std::io::{BufRead, BufReader};
 use std::path::{Path, PathBuf};
-use std::sync::{Mutex, OnceLock};
+use std::sync::{Arc, Mutex, OnceLock};
 use std::time::SystemTime;
 
 const CODEX_THREAD_REQUEST_ID_PREFIX: &str = "codex_session:thread-v1";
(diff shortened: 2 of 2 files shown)


farion1231:
@codex review

farion1231 (approved):

PR #5637, opened 2026-07-22
feat(codex): manage providers per Windows and WSL target

M1kasali:
## Summary / 概述

Adds device-local Managed Targets so one Windows CC Switch installation can manage isolated Codex Provider state for Windows and WSL2 without copying whole configuration directories.

- models Windows/WSL environments independently from shared Provider definitions
- discovers and registers the default Codex user in each WSL distribution without taking it over
- links an existing Provider first, then requires explicit activation before writing
- switches exactly one selected Target while preserving every other Target
- projects only Provider-owned routing, credential, model, and catalog fields
- preserves Target-owned paths, policies, MCP configuration, `auth.json`, sessions, archived sessions, and state databases
- performs WSL writes through argument-safe `wsl.exe -d ... -u ... -- ...` commands, with atomic replacement, permission preservation, read-back verification, and byte-exact rollback
- migrates the legacy single Codex directory into a device-local Target without writing live files
- prevents deleting a Provider still referenced by a Managed Target
- includes English, Simplified Chinese, Traditional Chinese, and Japanese UI strings

The implementation is intentionally scoped to a Windows host and the default user of each local WSL2 distribution. Multi-user WSL registration, SSH Targets, Target-aware proxy takeover, and cross-environment session aggregation remain future work.

## Related Issue / 关联 Issue

N/A

## Verification / 验证

- [x] Prettier check passes
- [x] Frontend test suite: 83 files / 530 tests
- [x] Rust test suite: 2099 passed / 2 ignored, plus all integration suites
- [x] `cargo clippy --all-targets -- -D warnings`
- [x] Vite production renderer build
- [x] Native Windows MSVC release build with `tauri/custom-protocol`
- [x] Manual Windows + Ubuntu 24.04 WSL2 smoke test: independent switches, unchanged Windows files during WSL switch, preserved WSL history/auth, and successful Codex CLI request

## Screenshots / 截图

The new Target selector is shown on the Codex Provider page; discovery, registration, linking, inspection, and activation are available under Settings → Advanced → Environments.

## Checklist / 检查清单


M1kasali:
Withdrawing this draft while we complete additional Windows/WSL validation and provider configuration normalization. We will submit a fresh PR once the implementation is stable and fully verified.

PR #4975, opened 2026-07-04
feat: add ZCode app support

Ankairis:
## Summary
- add ZCode as a managed app with provider import/write support for ~/.zcode/v2/config.json
- add ZCode provider presets, form fields, icon, app visibility, directories, prompts, and settings integration
- add ZCode usage dashboard support by syncing token usage from ~/.zcode/cli/db/db.sqlite

## Validation
- ./node_modules/.bin/tsc --noEmit
- ./node_modules/.bin/vite build

## Notes
- cargo check could not be run in this environment because cargo is not installed (cargo: command not found).

PR #5644, opened 2026-07-22
feat(zcode): add ZCode CLI agent support

### LINKED ISSUE #4205: 支持ZCode Agent应用
### Self Checks / 自检

### Related App / 涉及应用

General / 通用

### Your Question / 你的问题

是否考虑实现对国产code Agent的支持，如ZCode

### What Have You Tried / 你已经尝试过什么

_No response_

### Environment / 环境信息

_No response_

xiaoyuyu6420:
## Summary

Adds ZCode (z.ai desktop coding agent, `dev.zcode.app`) as a managed app, covering provider config, MCP, unified Skills, and usage.

- **Provider config**: read/write `~/.zcode/v2/config.json` (additive mode, OpenCode-like structure with `kind`/`options`/`models`)
- **MCP support**: read/write `~/.zcode/cli/config.json` `mcp.servers` — ZCode uses the **standard stdio format**, so sync is pass-through (no format conversion needed)
- **Skills**: ZCode joins the unified Skills panel (DB schema v17 → v18 adds `enabled_zcode` to both `skills` and `mcp_servers`)
- **Usage sync**: reads `~/.zcode/cli/db/db.sqlite` `model_usage` table into the unified usage dashboard; wired into the background scheduler via `sync_all_unlocked`
- **Provider presets**: based on ZCode.app's official model catalog (Z.ai, BigModel, Kimi, MiniMax, DeepSeek, Qwen, MiMo, …)
- **Full i18n**: form keys + preset names + config-dir labels across all 4 locales (en/zh/ja/zh-TW)

Closes #4205.

## How ZCode config was verified

All paths and structures are based on a real ZCode v3.3.6 installation, not assumptions:

- **Provider config**: `~/.zcode/v2/config.json` → `{provider: {<id>: {name, kind, options:{apiKey,baseURL}, enabled, source, models}}}`
- **MCP config**: `~/.zcode/cli/config.json` → `{mcp: {servers: {<id>: {type:"stdio", command, args, env}}}}` (a separate file from providers)
- **Usage DB**: `~/.zcode/cli/db/db.sqlite` → `model_usage` table (full token/cost/latency columns)

## Relationship to PR #4975

There is an existing PR #4975 on the same topic. This PR addresses several gaps found during review of that PR:

- **MCP support**: PR #4975 marked ZCode as *not supporting MCP* ("v1 doesn't expose MCP config"). That is incorrect — ZCode stores MCP config in `~/.zcode/cli/config.json` (a separate file from the provider config). This PR correctly enables MCP sync (pass-through, since ZCode already uses the standard stdio format).
- **i18n completeness**: `ja.json` and `zh-TW.json` were untouched in #4975, and the `zcode.*` namespace keys were undefined in every locale. This PR adds all keys to all 4 locales (36 form keys + 8 preset names).
- **No hardcoded strings**: all form/validation text goes through `t("zcode.*")` / `t("providerForm.presets.*")`; no hardcoded Chinese toasts in the ZCode validation path.
- **`apiKeyRequired`** is derived from the initial config rather than hard-coded `true`.

## Update (2026-08-22)

- **ZCode now joins the unified Skills panel** (superseding the earlier note below — with schema v17 → v18 migration adding `enabled_zcode` to both `skills` and `mcp_servers`; the migration is idempotent via `add_column_if_missing`, and a regression test covers an existing v17 database upgrading in place).
- Rebased onto the latest `main`; the WSL2 CI fix (LNK1327) is its own commit, and the tsconfig `baseUrl` cleanup is split into a separate `chore` commit.

## Validation

- [x] `pnpm typecheck`
- [x] `pnpm format:check`
- [x] `pnpm test:unit` (522 passed)
- [x] `pnpm build:renderer`
- [x] `cd src-tauri && cargo fmt --check && cargo clippy` (zero warnings)
- [x] `cd src-tauri && cargo test --lib` (2108 passed; 1 pre-existing `claude-desktop` proxy port-bind failure unrelated to this change)
- [x] Migration tests re-run after the rebase: `database::schema::tests::migrate_*` 6/6, including `migrate_v17_to_v18_adds_enabled_zcode_to_skills_and_mcp`

## Notes

- ZCode's Skills initially used a separate plugin-marketplace system and skill sync was left off; as of the 2026-08-22 update ZCode is wired into the unified Skills panel like the other managed apps.
- The single failing cargo test (`update_current_claude_desktop_provider_syncs_profile_when_proxy_takeover_is_active`) fails with "Address already in use (os error 48)" in this environment — an environment issue unrelated to ZCode.


farion1231:
@codex review

xiaoyuyu6420:
All four review comments have been addressed, merge conflicts resolved, and the PR is now **MERGEABLE**. Posting a summary here in case it helps the review.

### Changes addressing the review

**P1-1 — Carry the ZCode key into the created provider ID**
- `src/lib/query/mutations.ts`: added `zcode` to the additive-ID branch so the provider key is used as the DB id (was: random UUID).
- `src/components/providers/AddProviderDialog.tsx`: pass `providerKey` through for `zcode` as well.

**P1-2 — Wire ZCode into the additive provider-card flow**
- `ProviderList.tsx`: `isProviderInConfig` now queries `useZcodeLiveProviderIds`; added the `zcode` branch in the live-import mutation; invalidated ZCode caches on add/update/delete/switch via `invalidateZcodeProviderCaches`.
- `ProviderActions.tsx`: `isAdditiveMode` now includes `zcode`, so rows render Add/Remove state instead of a plain Enable button.

**P2-1 — Preserve disabled and non-custom provider metadata**
- `useZcodeFormState.ts`: `enabled` / `source` are now carried in `ZcodeFieldState` from the initial config and preserved on every edit. `emitConfig` no longer hard-codes `enabled: true` / `source: "custom"` (only defaults to them for newly created providers).

**P2-2 — Retain extra top-level ZCode fields during import**
- `src-tauri/src/provider.rs`: added `#[serde(flatten)] extra` to `ZCodeProviderConfig` so unknown top-level fields survive the deserialize → re-serialize round-trip.
- Added a regression test `zcode_provider_config_preserves_extra_top_level_fields`.

### Verification
- ✅ `pnpm typecheck`, `pnpm test` (522 tests), `cargo clippy -D warnings`, `cargo test` (2202 tests) all pass locally.
- ✅ Rebased onto latest `upstream/main`; conflicts in `session_usage.rs`, `usage_stats.rs` and the 4 locale files were resolved by keeping both the Grok Build and ZCode entries.
- 🟡 CI shows `action_required` — it needs a maintainer approval to run since this is a cross-repo PR.

Could a maintainer approve the CI run and take a look? Happy to adjust anything. Thanks!

xiaoyuyu6420:
All four review comments are now addressed, merge conflicts resolved, and the PR is `MERGEABLE`. Could you please take a look, @farion1231?

Two things only a maintainer can do (CI is `action_required` since this is a fork PR):
1. **Actions** → approve the CI run so it can execute.
2. **Files changed** → review/approve.

I have run the same 6 CI checks locally (typecheck, format:check, test:unit, cargo fmt --check, cargo clippy -D warnings, cargo test) — all green. Summary of the fixes is in my previous comment. Thanks!

xiaoyuyu6420:
The Windows job failed on `hermes_config::tests::write_then_read_memory_round_trip` — but this looks like a flake, not a real regression introduced by this PR:

- The failing test lives in `src-tauri/src/hermes_config.rs`, which this PR does **not** modify.
- This PR's only `settings.rs` change is purely additive (`VisibleApps.zcode` / `AppSettings.zcode_config_dir`); it does not touch `get_hermes_override_dir()`, `settings_store()`, or `resolve_override_path()` that the hermes memory path depends on.
- The test uses `#[serial]` + a `tempfile`-backed `CC_SWITCH_TEST_HOME` env var — a classic source of Windows-only flakiness (env-var propagation timing across the test binary). The assertion is `left: ""` vs `right: "> note\n§\nfirst entry..."`, i.e. the read returned empty, consistent with the temp-home env var not being visible to the read at that instant.
- macOS and Ubuntu backend jobs both passed, as did Frontend Checks.

Could a maintainer re-run just the `Backend Checks (windows-latest)` job? If it fails again I'll dig deeper, but I don't see a plausible causal path from the ZCode changes to this hermes memory test.

xiaoyuyu6420:
Rebased onto latest `main` (v3.20.0, including the Pi integration in #6064) and resolved all conflicts. The PR is **MERGEABLE** again.

### How ZCode follows the patterns Pi established

- **App registration**: `zcode` added to `AppType`, `APP_IDS`, `DEFAULT_VISIBLE_APPS`, `ADDITIVE_APP_IDS`, `MCP_APP_IDS`, and the icon map — all alongside `pi`.
- **MCP SSOT**: added an `enabled_zcode` column to `mcp_servers` (via `add_column_if_missing`, same migration pattern as the other flags) instead of the old per-app `McpRoot` storage, which upstream removed. `update_mcp_server_app_enabled` now maps `ZCode => Some("enabled_zcode")`.
- **Legacy prompts**: like Pi, ZCode skips the legacy in-config prompt store (`auto_import_prompt_if_exists` returns `Ok(false)`) since prompts live in SQLite now.
- **Prompt filename map**: follows the regrouped `SOUL.md`/`AGENTS.md` structure; ZCode maps to `AGENTS.md`.
- **Session usage**: `sync_zcode_usage` is wired via the same `merge_sync_step` call as Pi, and the dedup allow-list in `has_matching_proxy_usage_log` includes `zcode`.
- **AppSwitcher**: switched to the icon-only switcher's `APP_ICON_NAME`/`APP_DISPLAY_NAME` records instead of the removed inline maps.

### Verification (on this branch, locally)

- `pnpm typecheck` ✅
- `pnpm test:unit` ✅ 982/987 (the 5 failures are `PiProviderForm` 5s timeouts — I confirmed the same failures on clean `upstream/main`, so they're environment/perf, not caused by this PR)
- `cargo check` ✅
- `cargo test --lib` ✅ 2684/2685 (the 1 failure is a proxy-port-bind flake in an unrelated claude-desktop test)

Happy to adjust anything. @farion1231

tmuer:
大佬强啊，不知道什么时候能合并。

Bssn520:
期待合并，希望通过 cc switch 在 zcode 中使用 grok 订阅

xiaoyuyu6420:
Did an end-to-end smoke test on a real machine (macOS, `pnpm tauri dev` against an existing ZCode install with live data):

- **Live import**: 11 providers from `~/.zcode/v2/config.json` (built-in BigModel/ZAI plans + custom entries) imported on first launch ✅
- **Edit → live sync**: editing a provider in CC Switch (name / baseURL / models) writes through to `config.json` immediately, verified by file mtime + content ✅
- **Metadata preservation**: `enabled: false` and `source` survive unrelated edits (the P2-1 review fix behaving as intended) ✅
- **Usage stats**: `[ZCODE-SYNC]` imported 431 records from ZCode's `model_usage` SQLite table on first sync, then incremental +2 on the next run — dedup working ✅

Everything held up against a real ZCode setup. Happy to help verify anything else.

xiaoyuyu6420:
Thanks for approving the CI run! Quick notes on the two red jobs, neither is caused by this PR:

**Backend Checks (windows-latest)** — fails on `hermes_config::tests::write_then_read_memory_round_trip` (`hermes_config.rs:2138`), a file this PR does not touch. This is the same intermittent Windows-only flake seen earlier in this PR's history (failed on the 7/28 run, passed on 8/2, failed again now — same test, same empty-read symptom). 2667/2668 other tests pass on Windows.

**Backend Checks (Windows + WSL2 home)** — the job log contains no build or test output at all; it dies during runner/WSL setup (this is the first time this job type runs on a fork PR), so it looks like an infrastructure/setup issue rather than a code one.

Could a maintainer re-run the two failed jobs? Happy to dig deeper if either fails again after a re-run.

xiaoyuyu6420:
Context for the workflow change in `b2f5952f`, plus a request at the end 👇

### Why the WSL2 job failed on this PR (and not on earlier ones)

PR #6233, which introduced the WSL2 job, already documented this failure mode in its own commit message:

> compile test binaries before pointing TEMP/TMP at the WSL2 UNC path (**mt.exe fails with c1010070/LNK1327 when TEMP is a UNC path**)

However, in `ci.yml` the pre-compile step ended up using `cargo test --lib --no-run`, while the contract step then runs `cargo test --tests …`. The integration-test binaries are therefore **not** pre-compiled, and get linked after `TEMP`/`TMP` has been switched to the WSL UNC path → `LNK1327` → job dies with no test output (which matches what we saw on this PR's run).

Earlier PRs (#6064, #6535) stayed green on this job most likely because their runs hit the cargo cache and skipped the link step entirely. Note that `wsl2-nightly.yml` already uses the consistent `--tests --no-run` pattern — this PR aligns `ci.yml` with it (and pre-builds once with native TEMP, executing prebuilt binaries afterwards).

### Request

Since this commit touches `.github/workflows/ci.yml`, CI needs a fresh maintainer approval to run. Could you **approve the CI run and take a look at the review** when you have a moment, @farion1231? The branch is rebased onto v3.20.0, MERGEABLE, verified end-to-end against a real ZCode install (see the smoke-test comment above), and this workflow fix should make the last two Windows jobs green as well.

xiaoyuyu6420:
This replaces my previous comment — its root-cause explanation was wrong (I misread the contract step), so here is the corrected version, plus a request at the end 👇

### Why the WSL2 job failed on this PR (and not on earlier ones)

PR #6233, which introduced the WSL2 job, already documented the failure mode in its own commit message:

> compile test binaries before pointing TEMP/TMP at the WSL2 UNC path (**mt.exe fails with c1010070/LNK1327 when TEMP is a UNC path**)

What actually happened in this PR's failing run ([job log](https://github.com/farion1231/cc-switch/actions/runs/32346662369/job/96646276273)):

1. `Compile backend tests with native temp` succeeded at 02:33:10 and linked `cc_switch_lib-0e63d2a672627243.exe` with a native TEMP.
2. Seven seconds later, in `Run Windows-to-WSL2 filesystem contract` — i.e. after TEMP/TMP had been switched to the WSL UNC path — cargo considered the *same* unit stale (`Compiling cc-switch v3.20.0` again) and re-linked it; mt.exe then failed on the UNC temp (`c1010070` → `LNK1327`), so the job died before running any test.
3. Why cargo invalidated a unit it had built seven seconds earlier isn't visible in the log — the only environmental delta between the two steps is TEMP/TMP. Note both steps in `ci.yml` use `--lib`; my earlier "--lib vs --tests mismatch" claim was wrong. Earlier PRs (#6064, #6535) most likely stayed green because their contract steps had nothing left to rebuild (cargo cache hit), so no link ever ran against the UNC temp.

### Why the fix is robust despite the unknown trigger

Since the rebuild trigger can't be pinned down from the log, `b2f5952f` removes `cargo` from every WSL-TEMP step instead: compile the test binaries once with a native TEMP, then execute the prebuilt binaries directly (picking the newest match by LastWriteTime, so a cache-restored stale exe can't shadow the fresh build). `wsl2-nightly.yml` gets the same pattern, and its full suite now discovers and runs the prebuilt lib + integration-test binaries serially, matching the old `--tests -- --test-threads=1` behavior.

### Request

Since this commit touches `.github/workflows/ci.yml`, CI needs a fresh maintainer approval to run. Could you **approve the CI run and take a look at the review** when you have a moment, @farion1231?

The branch is rebased onto v3.20.0, MERGEABLE, and verified end-to-end against a real ZCode install (see the smoke-test comment above). One caveat to set expectations: the *other* red job, `Backend Checks (windows-latest)`, is the known `hermes_config` flake from July (2667 passed / 1 failed) — unrelated to this workflow fix, so it may just need a re-run.

xiaoyuyu6420:
**Windows verification on a real machine (real ZCode install)**

Since the two red CI jobs were the Windows ones, I verified this PR on a physical Windows 11 machine (build 26200, MSVC toolchain, Rust 1.95 as pinned by `rust-toolchain.toml`) that has a live ZCode installation.

**1. The two red CI jobs, re-verified locally**

- `hermes_config::tests::write_then_read_memory_round_trip` — ran the exact test 10× consecutively, plus a full serial `--test-threads=1` lib run: all green. Consistent with the known-flake assessment.
- WSL2 job (the `b2f5952` fix) — replicated the new CI flow end-to-end using a local WSL2 distro in place of Ubuntu-24.04: compiled the test binary with a native TEMP, then switched `TEMP`/`TMP` to the `\\wsl.localhost\...` UNC path and executed the prebuilt binary directly. `config::tests::atomic_write_replaces_existing_wsl_unc_file --ignored --exact` passes, and no linking happens against the UNC temp (the original LNK1327 trigger).

**2. ZCode functionality on Windows**

- All ZCode-related tests pass across the lib and integration suites (provider config, MCP sync, deeplink import, import/export sync, usage).
- `cargo fmt --check`, `cargo clippy` (zero warnings), and `pnpm typecheck` are all clean.
- Verified the config shapes against the live install: `~/.zcode/v2/config.json` provider map (`name`/`kind`/`options`/`enabled`/`source`/`models`), `~/.zcode/cli/config.json` `mcp.servers` (stdio), and `~/.zcode/cli/db/db.sqlite` `model_usage` — all match what this PR implements.

**3. End-to-end smoke test (Windows counterpart of the macOS smoke test above)**

Ran the PR build on the same machine against a copy of the real ZCode data:

- First launch: live-imported all **9 providers** (builtin BigModel/ZAI plans + 3 custom entries) ✅
- Background usage sync: imported **5,647 rows** from `model_usage` on the first pass ✅
- Provider add/edit form renders with correct zh i18n and live URL validation ✅
- Proxy takeover for claude/codex restored normally on startup ✅

Conclusion: the ZCode feature works correctly on Windows; the two red CI jobs are environmental (the known flake, and the UNC-temp link issue already addressed by `b2f5952`). Hope this helps the review @farion1231.

xiaoyuyu6420:
Fresh update pushed — summary of what changed since the last rebase:

**ZCode now joins the unified Skills panel.** This supersedes the earlier "skill sync intentionally not enabled" note in the description (which I've updated accordingly):

- `zcode` added to `SKILLS_APP_IDS`, skills DAO/service, and the MCP form wiring; i18n keys added to all four locales.
- DB schema v17 → v18 adds `enabled_zcode` to **both** `skills` and `mcp_servers`. Both columns use `add_column_if_missing` (idempotent), and a regression test covers an existing v17 database upgrading in place — so users on v15+ (who ran `migrate_v14_to_v15` before ZCode existed) get both columns correctly.

**History cleanup:**

- The WSL2 CI fix (LNK1327: compile with native TEMP, run prebuilt binaries) is its own commit.
- The tsconfig `baseUrl` cleanup is split into a separate `chore` commit so the feature commits stay focused.
- Rebased onto latest `main` (zero conflicts).

**Verification on this exact tree:**

- `pnpm typecheck` ✅ `pnpm format:check` ✅
- `cargo fmt --check` ✅ `cargo clippy --all-targets` ✅ zero warnings
- Migration tests 6/6 (incl. `migrate_v17_to_v18_adds_enabled_zcode_to_skills_and_mcp`)
- `pnpm test:unit`: two stale assertions in `McpFormModal.test.tsx` updated for the new `zcode` apps flag (10/10 pass); the `App.test.tsx` integration tests pass standalone (7/7) — the earlier full-suite failures there were 10s-timeout flakes under parallel load, same class as the known `PiProviderForm` timeouts.

Still waiting on a maintainer to approve the CI run (fork PR) — happy to adjust anything. @farion1231

xiaoyuyu6420:
Root-caused the intermittent `windows-latest` failure in `hermes_config::tests::write_then_read_memory_round_trip` (seen 2026-07-28 and 2026-08-22, same assertion, same `left: ""` signature). It was **this PR's bug, not CI flakiness** — correcting my earlier "environment flake" assessment from July.

**Root cause.** `test_import_marks_new_server_with_only_zcode_enabled` (new in this PR, `src/mcp/zcode.rs`) set and removed the process-global `CC_SWITCH_TEST_HOME` env var **without** the `#[serial]` lock. On Windows, inside `with_test_home`, `LOCALAPPDATA` is removed — so `CC_SWITCH_TEST_HOME` becomes the sole input to Hermes' memory-path resolution. When the zcode test's set/remove lands between a hermes test's write and its second read, the read resolves to a different home directory and returns `"``.

**Evidence.** Replaying that exact interleaving deterministically reproduces the identical failure signature locally (write goes to the temp home, the second read resolves to the real home, returns empty). It is the only env-mutating test in the tree without `#[serial]`, and upstream `main` — which doesn't contain it — has never hit this failure on `windows-latest`.

**Fix** (efaaa7997): take `#[serial]` like every other env-mutating test, switch to a self-cleaning `tempfile::tempdir`, and restore (not just remove) the previous env value. Verified: zcode + hermes suites green across 5 high-concurrency rounds, full `cargo test --lib` unchanged from before the fix.

No re-approval needed for CI if the previous run's approval covers subsequent pushes — happy for a maintainer to trigger it otherwise. @farion1231

xiaoyuyu6420:
CI has been fully green for a few days now, and this PR is ready for review. To make the 73-file diff easier to approach, here is a guided tour — most of it is mechanical pattern-following; only three files contain genuinely new logic.

## Where to spend review time (new logic)

1. **`src-tauri/src/zcode_config.rs`** — provider config read/write over `~/.zcode/v2/config.json`. The two things worth a close look: additive-mode guarantees (set/remove touch exactly one `provider.<id>` entry, nothing else), and `#[serde(flatten)] extra` preserving unknown top-level fields (round-trip regression test included).
2. **`src-tauri/src/services/session_usage_zcode.rs`** — usage import from `~/.zcode/cli/db/db.sqlite`. Worth checking the WAL-aware mtime handling (documented inline) and the dedup wiring.
3. **`src-tauri/src/mcp/zcode.rs`** — MCP pass-through sync over `cli/config.json` `mcp.servers` (standard stdio format, zero conversion, plus the `#[serial]` test-isolation fix discussed upthread).

## Mechanical pattern-following (fine to skim)

- **Skills/MCP integration**: schema v17→v18 migration (both `skills` and `mcp_servers` get `enabled_zcode`, idempotent, with an existing-v17-upgrade regression test), DAO column wiring, `SKILLS_APP_IDS`, `McpFormModal` toggle.
- **App registration**: `app_config.rs` match arms, `settings.rs` `zcode_config_dir`, `AppSwitcher` icon/name entries — mirrors how Pi was registered in #6064.
- **Presets & i18n**: preset data files and 55 keys × 4 locales, all kept in sync.

## Supporting commits

- `fix(ci)`: WSL2 LNK1327 fix (compile with native TEMP, run prebuilt binaries) — required so this PR's own CI can pass.
- `chore(tsconfig)`: two-line `baseUrl` cleanup, split into its own commit for reviewability.

**Tests**: migration tests 6/6, ZCode suites green under high-concurrency runs, full `cargo test --lib` / `pnpm test:unit` / `clippy` / `typecheck` all pass. All config paths verified against a real ZCode v3.3.6 install, plus an end-to-end smoke test on Windows (real machine, live data).

If the size is a concern, I'm happy to walk through any section in more detail, or adjust/split whatever makes the review easier. @farion1231

xiaoyuyu6420:
Rebased onto the latest `main` (the PR had gone `DIRTY`) — now `MERGEABLE` and **CI is fully green again** on the rebased head. What the rebase did:

- Adopted upstream's own `ci.yml` fix for WSL2 LNK1327 (their approach records the test binary path via `--message-format=json`, better than my LastWriteTime scan); kept only my `wsl2-nightly.yml` half, which upstream hadn't touched.
- Upstream took schema v17→v18 for a session-log byte-cursor column, so the ZCode migration moved to **v18→v19** (`enabled_zcode` on `skills` + `mcp_servers`); `SCHEMA_VERSION` is 19, with an existing-v18-upgrade regression test.
- Adapted ZCode usage-sync to upstream's batch `load_sync_cursors` API (`get_sync_state` is now `#[cfg(test)]`), same pattern grokbuild uses.
- Added `#[serde(default, skip_serializing_if = "Option::is_none")]` to `zcode_config_dir` to match its siblings.

Local verification: `cargo clippy --all-targets` (0 warnings), migration + zcode usage suites, `pnpm typecheck`, `cargo fmt --check` — all green, and the official CI agrees. Ready for review. @farion1231

PR #5688, opened 2026-07-23
feat: search installed skills for agent assignment

princepal9120:
@farion1231

## What changed
- Add a search bar to the unified installed-skills panel.
- Filter by skill name, description, directory, or repository.
- Preserve existing per-agent enable toggles.
- Add localized clear and no-match states.

## Why
Users can quickly find an installed skill before enabling it for another agent.

## Validation
- TypeScript check
- Prettier
- Vite production build
- git diff --check


farion1231:
@codex review

PR #5689, opened 2026-07-23
fix: allow multiple Codex global prompts

### LINKED ISSUE #4885: Allow multiple Codex global prompts to be enabled and concatenated
# Allow multiple Codex global prompts to be enabled and concatenated

## Self Checks / 自检

## Related App / 涉及应用

Codex

In Codex prompt management, global prompts currently appear to behave like a mutually exclusive selection: enabling one prompt disables or overrides another.

This is inconvenient because many global prompts are independent behavioral rules and should be able to work together. For example:

- Critical thinking / do not blindly agree
- UTF-8 / Chinese text safety on Windows

These two prompts are orthogonal. Users should not need to manually merge them into one large prompt just to make both active.

每条全局提示词通常都是一个完整的 Markdown block，包含自己的标题和内容。因此，多个提示词之间天然可以拼接，不应该被当作单选配置。

## Proposed Solution / 建议方案

Allow multiple Codex global prompts to be enabled at the same time.

When applying Codex global prompts, collect all enabled Codex prompts and concatenate their Markdown content in a stable order, separated by blank lines. For example:

```text
enabled Codex prompts
  -> sort by user-defined order, or fallback to created_at / updated_at
  -> join content with "\n\n"
  -> write/apply as the final Codex global prompt
```

Suggested behavior:

- Treat `enabled` as a multi-select flag for Codex prompts, not as a single active selection.
- Preserve each prompt's Markdown structure.
- Use a stable ordering so the final prompt is deterministic.
- Keep the UI clear that multiple global prompts can be active together.

## Additional Context / 补充信息

Observed locally on Windows with CC Switch 3.16.4.

The current behavior makes users choose between unrelated rules, even though they should be composable. This request is not mainly about the exact Codex target filename; it is about the prompt activation model. The expected behavior is that all enabled Codex Markdown prompts are composed into the final global prompt used by Codex.

I searched existing issues including:

- #1759 / #2734: cross-app or unified prompt management
- #899: cross-model prompt sharing
- #4684: Codex prompt file read/save behavior when external tools modify the live prompt file

Those are related to prompt management, but this request is specifically about allowing multiple enabled prompts within the same app (Codex) and composing them.

## Contribution / 参与贡献

I am interested in contributing to this feature.

lelelelelelelelelelelelele:
## Summary

- allow multiple Codex global prompts to remain enabled
- concatenate enabled Markdown prompts in database order when writing `AGENTS.md`
- update the prompt panel to show the number of enabled prompts

## Testing

- `git diff --check` passed
- Rust tests could not run because `cargo` is not installed in the current environment
- TypeScript checks could not run because pnpm blocked dependency build scripts (`esbuild`/`msw`)

Closes #4885

### DIFF
--- a/src-tauri/src/services/profile.rs
+++ b/src-tauri/src/services/profile.rs
@@ -9,7 +9,7 @@
 //! - 供应商：`ProviderService::switch`（内建代理接管热切换与接管下禁切官方）
 //! - MCP：`McpService::toggle_app`（改标志 + 单 server 物化）
 //! - Skills：`SkillService::toggle_app`（改标志 + 单 skill 物化）
-//! - Prompt：`PromptService::enable_prompt`（互斥激活 + 原子写 live）
+//! - Prompt：`PromptService::enable_prompt`（启用并原子写入 live 文件）
 //!
 //! apply 为 best-effort：单项失败收集为 warning 继续，不整体回滚。
 
--- a/src-tauri/src/services/profile.rs
+++ b/src-tauri/src/services/profile.rs
@@ -129,7 +129,7 @@ pub struct ProfilePayload {
     /// 每 app 启用的 Skill id 集合
     pub skills: PerApp<Option<Vec<String>>>,
     /// 每 app 激活的 prompt id
-    pub prompts: PerApp<Option<String>>,
+    pub prompts: PerApp<Option<Vec<String>>>,
 }
 
 impl ProfilePayload {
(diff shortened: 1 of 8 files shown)


lelelelelelelelelelelelele:
Follow-up: also guarded the existing live-file import path so a composed AGENTS.md is not re-imported into a single prompt and duplicated when another prompt is enabled. The follow-up commit is ec25d54.

farion1231:
@codex review

lelelelelelelelelelelelele:
I found the backend CI failures were existing mutual-exclusion assertions for Claude profiles. The implementation was too broad, so I scoped composition to Codex only and restored the original exclusive behavior for other apps. Updated in a2b9c92.

lelelelelelelelelelelelele:
Addressed both review points in 1e52184: profile snapshots now store and reconcile the complete enabled prompt ID set, and multi-prompt live-file edits are backed up before a composed overwrite. Updated profile fixtures and the frontend payload type accordingly.

Issue #4404, opened 2026-06-18
Reasoning model responses produce hundreds of fragmented `content_block_start` events instead of 2 (one thinking + one text)


snowing0427:
### Self Checks / 自检

- [x] I have read the [FAQ](https://github.com/farion1231/cc-switch#faq) section in README.
我已阅读 README 中的[常见问题](https://github.com/farion1231/cc-switch#常见问题)。

- [x] I have searched for [existing issues](https://github.com/farion1231/cc-switch/issues), including closed ones.
我已搜索过[已有的 Issue](https://github.com/farion1231/cc-switch/issues)，包括已关闭的。

### CC Switch Version / 版本号

3.16.3（绿色版）

### Operating System / 操作系统

Windows

### Related App / 涉及应用

Claude Code

### Steps to Reproduce / 重现步骤

Environment:
- cc-switch: v3.16.3 (portable exe on Windows 11)
- Upstream: LiteLLM (10.x.x.x:4000) → GLM5.2 (reasoning model, `cloud/glm-latest`)
- Client: Claude Code (settings.json: ANTHROPIC_BASE_URL -> http://127.0.0.1:15721)

Steps:
1. Configure cc-switch with a reasoning model upstream that emits OpenAI Chat
   Completions deltas containing `reasoning_content` (e.g. GLM5.2 via LiteLLM).
2. Point Claude Code at cc-switch (ANTHROPIC_BASE_URL = http://127.0.0.1:15721).
3. From Claude Code, send any message that triggers reasoning (e.g.
   "Explain the time complexity analysis of quicksort in detail").
4. Observe the rendered output: continuous text is broken into dozens of
   separate lines.

To capture the wire-level evidence, I inserted a transparent Node mitm proxy
between Claude Code and cc-switch (listen 127.0.0.1:15722, forward to
127.0.0.1:15721), recorded cc-switch's emitted Anthropic SSE, and counted
`content_block_start` events.

Reproduction is deterministic: every reasoning-model request triggers it.

### Expected Behavior / 期望行为

Per Anthropic Messages streaming spec, a single assistant response should
contain at most 2 content blocks:

1. ONE `thinking` block — all reasoning deltas accumulated into the same
   block via `content_block_delta` events with a stable `index`.
2. ONE `text` block — all content deltas accumulated the same way.

A block opens with a single `content_block_start`, receives many
`content_block_delta` events appending incrementally, and closes with one
`content_block_stop`. The `index` field stays stable per block.

[...]

Issue #2200, opened 2026-04-20
使用 OpenAI Chat Completions API 格式，CC 按 Token 自动换行


T4mako:
### Self Checks / 自检

- [x] I have read the [FAQ](https://github.com/farion1231/cc-switch#faq) section in README.
我已阅读 README 中的[常见问题](https://github.com/farion1231/cc-switch#常见问题)。

- [x] I have searched for [existing issues](https://github.com/farion1231/cc-switch/issues), including closed ones.
我已搜索过[已有的 Issue](https://github.com/farion1231/cc-switch/issues)，包括已关闭的。

### CC Switch Version / 版本号

3.13.0

### Operating System / 操作系统

Windows

### Related App / 涉及应用

Claude Code

### Steps to Reproduce / 重现步骤

使用 Vllm 部署 Qwopus3.5-27B-v3，API 格式使用 `OpenAI Chat Completions`，模型返回时自动换行

<img width="1897" height="1040" alt="Image" src="https://github.com/user-attachments/assets/668522d9-b5f8-4c0f-90aa-d59cfc1727b6" />

我已经确认流式结果中不会有多余换行符

```txt
data: {"id":"chatcmpl-86f6c467e4b242e0","object":"chat.completion.chunk","created":1776679645,"model":"Qwopus3.5-27B-v3","choices":[{"index":0,"delta":{"reasoning":"."},"logprobs":null,"finish_reason":null,"token_ids":null}]}

data: {"id":"chatcmpl-86f6c467e4b242e0","object":"chat.completion.chunk","created":1776679645,"model":"Qwopus3.5-27B-v3","choices":[{"index":0,"delta":{"reasoning":"\n"},"logprobs":null,"finish_reason":null,"token_ids":null}]}

data: {"id":"chatcmpl-86f6c467e4b242e0","object":"chat.completion.chunk","created":1776679645,"model":"Qwopus3.5-27B-v3","choices":[{"index":0,"delta":{"content":"\n\n"},"logprobs":null,"finish_reason":null,"token_ids":null}]}

data: {"id":"chatcmpl-86f6c467e4b242e0","object":"chat.completion.chunk","created":1776679645,"model":"Qwopus3.5-27B-v3","choices":[{"index":0,"delta":{"content":"你好"},"logprobs":null,"finish_reason":null,"token_ids":null}]}

[...]

PR #5706, opened 2026-07-24
fix(proxy): avoid interleaved thinking blocks

### LINKED ISSUE #5692: bug(proxy): interleaved reasoning/content fragments Claude Code responses
## Self Checks

- [x] I searched existing open and closed issues for `reasoning_content`, interleaved reasoning/content, fragmented text, and Qwen Claude Code responses.

## CC Switch Version

3.18.0

## Operating System

Windows 11

## Related App

Claude Code 2.1.218

## Steps to Reproduce

1. Configure a Claude provider with `apiFormat: "openai_chat"` and an OpenAI-compatible reasoning model.
2. Enable the CC Switch local proxy and use the provider from Claude Code.
3. Use an upstream that alternates OpenAI streaming deltas like this:

```text
delta.reasoning_content = "..."
delta.content = "first text fragment"
delta.reasoning_content = "..."
delta.content = "second text fragment"
```

4. Ask Claude Code a normal question, such as `What model are you?`.
5. Inspect Claude Code's UI or run:

```powershell
claude -p --verbose --output-format stream-json --include-partial-messages "What model are you?"
```

## Expected Behavior

The Anthropic SSE response should contain at most one leading `thinking` block followed by one continuous `text` block. Once visible text has started, late/interleaved reasoning deltas should not split that text into additional blocks.

## Actual Behavior

`src-tauri/src/proxy/providers/streaming.rs` tracks a single `current_non_tool_block_type`. Every transition between `reasoning` and `content` closes the current Anthropic block and opens another one.

In a live Qwen-compatible stream, one short answer became:

- 24 `thinking` blocks
- 23 `text` blocks
- 23 separate assistant fragments in Claude Code

Claude Code renders every text block as a separate response bullet, producing visibly fragmented output such as one bullet for every few words.

## Root Cause

OpenAI-compatible reasoning providers may emit `reasoning` / `reasoning_content` after visible `content` has already started. Anthropic streaming content blocks are sequential, so converting every upstream type transition literally creates many tiny Anthropic blocks.

The relevant state transitions are in `create_anthropic_sse_stream`:

- reasoning changes `current_non_tool_block_type` to `thinking`
- content changes it back to `text`
- each change emits `content_block_stop` + `content_block_start`

## Proposed Fix

Track whether visible text has started. Preserve reasoning emitted before the first text delta, but ignore reasoning deltas that arrive after text starts. This matches the normal semantic order of reasoning followed by final text and prevents text fragmentation without changing tool-call handling.

Add a regression test with alternating reasoning/content chunks and assert:

- exactly one `thinking` block
- exactly one `text` block
- visible text deltas concatenate without loss
- late reasoning is not emitted after text begins

## Additional Validation

With a local response normalizer applying this rule, the same live provider produced `0` displayed thinking blocks and exactly `1` text block. Sequential tools, parallel reads, multi-turn resume, and a Claude Code subagent all continued to work.

Komikawayi:
## Summary
- Track when visible text has begun while converting OpenAI-compatible streams to Anthropic SSE.
- Ignore reasoning deltas that arrive after text begins, preventing repeated thinking/text block transitions.
- Add a regression test for interleaved `reasoning_content` and `content` fragments.

## Root cause
Some reasoning providers emit late reasoning chunks after visible content. Anthropic content blocks are sequential, so translating every upstream type transition fragmented one response into many text blocks.

Fixes #5692

## Validation
- `cargo test proxy::providers::streaming::tests --lib`
- `cargo fmt --check`

### DIFF
--- a/src-tauri/src/proxy/providers/streaming.rs
+++ b/src-tauri/src/proxy/providers/streaming.rs
@@ -166,6 +166,9 @@ pub fn create_anthropic_sse_stream<E: std::error::Error + Send + 'static>(
         let mut has_sent_message_stop = false;
         let mut stream_ended_with_error = false;
         let mut latest_usage: Option<Value> = None;
+        // Anthropic content blocks are sequential. Once visible text starts, later
+        // upstream reasoning cannot be represented without splitting that text.
+        let mut has_started_text = false;
         let mut current_non_tool_block_type: Option<&'static str> = None;
         let mut current_non_tool_block_index: Option<u32> = None;
         let mut tool_blocks_by_index: HashMap<usize, ToolBlockState> = HashMap::new();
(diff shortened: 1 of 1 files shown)


farion1231:
@codex review

ZhaoKangming:
+1 on this PR. I've been hitting this exact bug in production with GLM-5.2 via the JD Cloud API (`ai-api.jdcloud.com`).

**My environment:**
- CC Switch v3.19.2 on Windows 11
- Upstream: JD Cloud GLM-5.2 (OpenAI Chat Completions format)
- Client: Claude Code v2.1.220
- Provider config: `ANTHROPIC_BASE_URL=http://127.0.0.1:15721`

**What I captured:**
I curled the proxy directly and confirmed the root cause described in #5692. A normal Anthropic SSE stream has ONE `content_block_start` (type=text) with many `content_block_delta` events appended. CC Switch instead emits a NEW `content_block_start` + `content_block_stop` pair for EVERY upstream OpenAI chunk:

```
content_block_start (text)     <- new block
content_block_delta  (text_delta: "1. Apple
2. Banana
")
content_block_stop
content_block_start (text)     <- another new block!
content_block_delta  (text_delta: "3. Orange
4. Grape
")
content_block_stop
content_block_start (text)     <- yet another...
```

Claude Code's TUI renderer treats each `content_block` as a separate visual element, so every tiny chunk becomes its own bullet/line with a `●` prefix — exactly the fragmented output described in #4404 and #5028.

**Impact:**
This makes CC Switch + any OpenAI-format reasoning model (GLM-5.2, DeepSeek, Qwen with reasoning_content) essentially unusable with Claude Code. The output is unreadable — each token gets its own line.

**This PR's fix is correct:**
The approach of tracking whether visible text has begun, and ignoring late reasoning deltas after text starts, matches the normal semantic order (reasoning → text). The regression test with alternating reasoning/content chunks is exactly the right validation.

Would love to see this merged. The bug has been open since #2200 (April) and has been confirmed across multiple models (GLM-5.2, DeepSeek-V4, Qwen, MiniMax) and multiple users.

Komikawayi:
@farion1231 这个需要尽快审核一下，没问题就合并了，我也注意到几个pr里都提到了这个问题

SeanDictionary:
希望该PR能尽快合并在下一个版本中修复

Komikawayi:
这个啥时候合并啊@farion1231

PR #5734, opened 2026-07-24
feat(usage): add automatic models.dev pricing sync

### LINKED ISSUE #5733: feat(usage): add automatic models.dev pricing sync
## Problem / 问题

Model pricing is currently stored in CC Switch's database. Users can edit entries manually or explicitly import one model from models.dev, but keeping a useful catalog current requires repeated manual work.

当前模型定价保存在 CC Switch 数据库中。用户只能手动编辑，或主动从 models.dev 逐个导入，长期维护常用模型价格需要重复操作。

## Proposed behavior / 期望行为

- Add a user-maintainable local pricing file under the CC Switch config directory.
- Optionally fetch and import models.dev pricing on every application launch; enabled by default and non-blocking when offline.
- Let users search/filter providers and models and persist the automatic import selection.
- Provide a separate default-on common-model selection covering canonical Claude, GPT, Gemini, Grok and common Chinese model families.
- Batch imported prices and historical-cost backfill to avoid one full scan per model.

- 在 CC Switch 配置目录增加用户可维护的本地价格文件。
- 支持每次启动自动拉取并导入 models.dev 定价，默认开启，离线失败不阻塞启动。
- 支持按供应商/模型搜索筛选并持久化自动导入选择。
- 提供独立且默认开启的“常用模型”选择，覆盖 Claude、GPT、Gemini、Grok 及常见国产模型。
- 批量写入价格并统一回填历史成本，避免每个模型重复扫描。

## UI location / 界面位置

Settings → Usage Statistics → Cost Pricing / 设置 → 使用统计 → 成本定价

Related to #4017, which introduced the existing manual models.dev import flow.

YUZHEthefool:
## Summary / 概述

Extends the existing manual models.dev import flow with a user-maintainable local pricing source and configurable automatic synchronization.

- persist model pricing and models.dev sync preferences in `~/.cc-switch/model-pricing.json`; the SQLite table remains the runtime lookup store
- load manual file edits at startup/on demand, write UI edits atomically, preserve explicit deletions with tombstones, and merge newly shipped built-in prices without overwriting user values
- fetch selected models.dev prices once per application launch with a default-on switch, a 15-second timeout, non-blocking offline failure handling, and persisted last-sync status
- add a Cost Pricing panel for sync-now, local-file location/open/reload, provider/search filters, bulk selection, and an independent default-on common-model selection
- select recent canonical Claude, GPT, Gemini, Grok, DeepSeek, Qwen, MiMo, LongCat, Kimi, MiniMax, and GLM text models while excluding audio/image/embedding/realtime entries
- import selected prices in one SQLite transaction and run one historical-cost backfill instead of scanning once per model
- update only sync-result fields after a background fetch so a slow startup request cannot restore stale switches or model selections
- share models.dev parsing/normalization with the existing manual picker and cover the feature in Rust, data-selection, sync, and component tests

范围说明：本 PR 不修改官方订阅用量或会话用量导入逻辑。

## Related Issue / 关联 Issue

Fixes #5733

Related to #4017 and #4079.

## Screenshots / 截图

image

## Verification

- `tsc --noEmit`
- `prettier --check` for all `src/**` files and all changed tests
- `vitest run --maxWorkers=1 --minWorkers=1` — 81 files, 529 tests passed
- `cargo clippy --manifest-path src-tauri/Cargo.toml --all-targets -- -D warnings`
- `cargo test --manifest-path src-tauri/Cargo.toml` — 2,183 passed, 2 ignored, plus all integration suites
- Windows Tauri release executable and MSI generated and manually tested
- `git diff --check`

## Checklist / 检查清单

- [x] Updated i18n files if user-facing text changed / 已更新 zh / en / zh-TW / ja 四份国际化文件


farion1231 (commented):
感谢您的贡献，这个 PR 工程质量和功能设计都非常棒，只是还有以下细节问题：

我觉得应该在功能开启的时候给用户一个弹窗提示，开启这个功能之后，所有的价格数据都会从 model.dev 拉取，包括软件内置的和用户手动设置的

cache: "no-store" 绕过缓存，每次都会全量拉取模型数据

仍然有音频、视频、已弃用模型没有被过滤

farion1231:
@codex review

YUZHEthefool:
Addressed the maintainer and Codex review feedback in [`902bd25f`](https://github.com/YUZHEthefool/cc-switch/commit/902bd25f):

- Automatic models.dev sync is now disabled by default and must be enabled explicitly by the user.
- Enabling sync now shows a confirmation that prices for matching model IDs will overwrite both built-in and manually edited pricing. The backend batch-upsert behavior intentionally keeps this overwrite semantic, with a regression test covering database and local-file updates.
- Removed `cache: "no-store"` so normal browser HTTP caching and conditional requests can be used.
- Filtered deprecated models and non-text output models using models.dev `status`/`modalities` metadata, with name/ID marker fallback for incomplete records. Multimodal-input models that still output text remain available.
- Re-read the sync configuration after the network request. Startup sync now aborts if the user disables it while the request is in flight, while forced/manual sync still runs with the latest selection.
- Reloading the local pricing file now also refreshes the models.dev sync configuration query, preventing stale UI state from overwriting manual file edits.
- Added regression coverage for enable confirmation/cancellation, local config reload, in-flight config changes, forced sync, filtering, default-off behavior, and manual-price overwrite behavior.

Validation completed:

- `tsc --noEmit`
- models.dev frontend regression tests: 28 passed
- model-pricing Rust tests: 5 passed
- Tauri release executable and MSI bundle built for local verification

farion1231 (commented):
感谢您的贡献。请查看一下是否有以下问题:

必修 1：本地文件快照冻结了内置定价修正通道

sync_local_model_pricing 无条件挂在 database/mod.rs:146，紧跟 ensure_model_pricing_seeded() 之后，不受 autoSyncEnabled 控制。首次启动时 load_or_create_file_unlocked 把整张表导出成 ~/.cc-switch/model-pricing.json，此后这些内置值全被当作"用户值"。

cc-switch 的定价维护机制是 seed_model_pricing（INSERT OR IGNORE，只补新模型）+ repair_current_model_pricing（UPDATE ... WHERE 旧值全等，只修正官方推送的过时价、不碰用户手改行），两者每次启动经 ensure_model_pricing_seeded 执行。快照建立后，将来 repair 推送的修正会在下次启动被文件旧值覆盖，并触发 backfill_missing_usage_costs() 用旧价重算历史成本。

用**本轮的新默认值（自动同步关闭）**跑的复现：

首次启动导出到本地文件的行数 = 178
快照中的 claude-sonnet-5 input = 3
（模拟新版 repair 把 DB 改成 99）
第二次启动后 DB 中的 claude-sonnet-5 input = 3   ← 修正被回滚

关掉功能、断网、从不打开面板的用户同样中招，所以确认弹窗对它无效。

upsert_pricing 的 ON CONFLICT DO UPDATE ... WHERE <任一列不同> 只用于统计 changed，不构成保护——它不比较新旧值的时间先后。

修法（两处删减，现有 8 个测试不受影响）：

- load_or_create_file_unlocked 初始改为 models: Vec::new()
- 删掉 sync_local_model_pricing 末尾把 query_all_pricing 全表回写进 file.models 的那段

即：本地文件只存用户覆盖，不存全表快照。

---
必修 2：tombstone 导致每次启动反复 seed → 删除 → 全量回填

三处拼起来构成一个不收敛的循环：

// commands/usage.rs:126  每次调用都重新 seed，INSERT OR IGNORE 会把被删的内置行插回来
state.db.ensure_model_pricing_seeded()?;
crate::services::model_pricing::sync_local_model_pricing(&state.db)?;

// services/model_pricing.rs:302  删除计入 changed
for model_id in &file.deleted_model_ids {
    changed += transaction.execute("DELETE FROM model_pricing WHERE model_id = ?1", ...)?;
}

// services/model_pricing.rs:333  changed > 0 → 全表回填
if changed > 0 { db.backfill_missing_usage_costs() }

用户只要在 UI 上删过一条内置模型定价，就进入这个循环。3 轮复现：

第 1 轮: seed 后重新出现=1  sync 返回 changed=1  sync 后=0
第 2 轮: seed 后重新出现=1  sync 返回 changed=1  sync 后=0
第 3 轮: seed 后重新出现=1  sync 返回 changed=1  sync 后=0

每一轮都返回 1，没有自愈路径。触发入口有两个：每次启动（database/mod.rs:146），以及每次打开定价页面（get_model_pricing 命令）——后者频率更高。

backfill_missing_usage_costs_on_conn 扫的是所有 total_cost_usd <= 0 且有 token 的 proxy_request_logs 行并逐行重算，重度用户是几万行。同样不受 autoSyncEnabled 控制。

根因是 changed 被同时用作两种语义：upsert_pricing 里它是"价格真的变了"（WHERE 任一列不同 会把无变化的 upsert 归零），DELETE 分支里它是"执行了一次删除"——后者永远为真，因为被删的行每轮都被 seed 插回来。

修法：

- 最小修复：apply_file_to_database 返回 (upserted, deleted) 两个计数，只用 upserted > 0 触发回填。删除定价本来也不该触发回填——删了价的行回填也查不到价。
- 更彻底：让 seed 尊重 tombstone，从源头断掉空转。但 ensure_model_pricing_seeded 在 schema.rs，直接读文件会引入 schema → services 的反向依赖，需要把 tombstone 提前读出来传进去，可以放到后续。

---
建议：轻量模式重建 renderer 会重复触发启动同步

let startupSync: Promise<ModelsDevSyncResult> | null = null;
/** Run at most once per renderer lifetime, including React StrictMode. */
export function syncModelsDevPricingOnStartup() {
  startupSync ??= syncModelsDevPricing();
}

注释描述准确，但 lightweight.rs 的进出是真销毁真重建：enter_lightweight_mode 调 window.destroy()，exit_lightweight_mode 用 WebviewWindowBuilder::from_config(...).build() 全新创建。新 WebView = 新 JS 上下文 = main.tsx 重新执行 = startupSync 重置为 null。轻量模式是托盘常驻场景的常用功能，进出频率可能高于冷启动。

建议直接用已有的 lastSyncAt 做最小间隔（例如 6 小时）——它落在文件里跨进程有效，顺带把删掉 no-store 之后剩下的频率问题一起收了。后端 OnceLock 只能防本进程，冷启动仍是每次一拉。

YUZHEthefool:
Addressed the follow-up review in [`eae7792a`](https://github.com/YUZHEthefool/cc-switch/commit/eae7792a):

### Required 1: keep built-in pricing repairable

- New `model-pricing.json` files now start with an empty `models` override list instead of exporting the seeded pricing table.
- Removed the full database-to-file writeback from `sync_local_model_pricing`.
- The local file now contains only explicit user/models.dev overrides and deletion tombstones, so later `repair_current_model_pricing` updates are not rolled back.
- Added a regression test that simulates a built-in price correction and verifies local synchronization preserves it and leaves the override file empty.

### Required 2: avoid tombstone-triggered full backfills

- `apply_file_to_database` now returns separate `(upserted, deleted)` counts.
- Full historical usage-cost backfill runs only when pricing is inserted or changed, never for deletion-only synchronization.
- Added a regression test for the repeated seed -> tombstone delete cycle and verified unrelated zero-cost usage remains untouched.

### Lightweight renderer / sync frequency suggestion

- Successful startup synchronization is now limited by persisted `lastSyncAt` to at most once every 6 hours, so destroying and rebuilding the WebView does not repeatedly fetch models.dev.
- Manual **Sync now** remains forced and bypasses both the interval and the automatic-sync switch.
- Updated all four locale descriptions and confirmation messages to describe the interval.

Validation completed:

- TypeScript: `tsc --noEmit`
- Frontend: 81 test files / 538 tests passed
- Rust library: 2186 passed / 2 ignored / 0 failed
- Focused model-pricing Rust tests: 7 passed
- `cargo check` passed
- Tauri release executable and MSI were generated for local verification

farion1231 (approved):

PR #5763, opened 2026-07-25
fix(usage): fall back to top-level apiKey when env token is empty

Qfbxz:
## Problem
Usage/balance queries for providers (e.g. DeepSeek) incorrectly report "missing API Key", even though a valid key is configured.

## Root Cause
The provider form stores the real key at the top-level `apiKey`/`api_key` field, while `env.ANTHROPIC_AUTH_TOKEN` is frequently an empty placeholder. Usage lookup only read env keys, so it missed the real key and reported it as missing.

## Fix
- **`provider.rs`**: when the env-derived `api_key` is empty, fall back to the top-level `apiKey`/`api_key`.
- **`UsageScriptModal.tsx`**: apply the same fallback for `GOOGLE_API_KEY`; also drop `grokbuild` from the app-id heuristic — its official state is TOML-parsed server-side (`grok_config::is_official_live_config`) and cannot be faithfully mirrored by regex (quoted keys / inline tables / invalid TOML all false-positive as official, which made the component discard already-saved non-official scripts during state init).

## Compatibility
Backward compatible — the fallback only triggers when the env key is empty.

### DIFF
--- a/src-tauri/src/provider.rs
+++ b/src-tauri/src/provider.rs
@@ -210,7 +210,7 @@ impl Provider {
             AppType::Claude | AppType::ClaudeDesktop => {
                 let env = settings.get("env");
                 let base_url = str_at(env.and_then(|e| e.get("ANTHROPIC_BASE_URL")));
-                let api_key = first_non_empty(
+                let mut api_key = first_non_empty(
                     env,
                     &[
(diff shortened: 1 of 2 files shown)


farion1231:
@codex review

PR #5764, opened 2026-07-25
fix(codex): trim MCP command and accept string args

Qfbxz:
## Problem
codex MCP servers fail to start in two common cases:
1. `command` carries trailing whitespace → codex cannot find the executable.
2. `args` is stored as a string (e.g. `"serve --mcp"`) instead of an array → MCP starts without its arguments.

## Root Cause
In `json_server_to_toml_table`:
- `command` is read verbatim, no `trim()`.
- `args` accepts only `Value::Array`. When `server_config` stores args as a string, `as_array()` returns `None` and the whole args block is silently dropped.

## Fix
- `command`: `.trim()` trailing whitespace.
- `args`: accept array **or** string (string split on whitespace); TOML output remains an array.

## Compatibility
Backward compatible — existing array args are unaffected; string args now work; trimmed command is strictly more lenient.

### DIFF
--- a/src-tauri/src/mcp/codex.rs
+++ b/src-tauri/src/mcp/codex.rs
@@ -652,17 +652,34 @@ pub(super) fn json_server_to_toml_table(spec: &Value) -> Result<toml_edit::Table
     // 1. 处理核心字段（强类型）
     match typ {
         "stdio" => {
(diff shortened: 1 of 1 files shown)


farion1231:
@codex review

PR #5765, opened 2026-07-25
fix(codex): tolerate invalid escape in tool-call arguments

Qfbxz:
## Problem
When Codex runs through the Anthropic transport against a third-party model (e.g. GLM), a single tool-call whose arguments contain an invalid JSON escape **permanently dead-loops the session**: every subsequent turn returns HTTP 400 `Invalid function_call arguments for 'update_plan': invalid escape at line 1 column N`.

## Root Cause — asymmetric strictness
- **Response path** (`json_canonical::canonicalize_json_string_if_parseable`): on parse failure it did `.unwrap_or_else(|_| value.to_string())`, passing the invalid JSON through **verbatim** to Codex, which stores it in session history.
- **Request path** (`transform_codex_anthropic::convert_input_to_messages`): next turn, Codex replays history; strict `serde_json::from_str` fails on the now-poisoned arguments → `ProxyError::InvalidRequest` → HTTP 400.
- `InvalidRequest` is categorized `NonRetryable` (`forwarder.rs`), so Codex does not retry — the session is stuck every turn until manually repaired.

Real-world trigger observed: GLM-5.2 emits `update_plan` arguments with `\'` (backslash-escaped single quote) inside plan step text. JSON permits only `\" \\ \/ \b \f \n \r \t \uXXXX`, so `\'` is invalid.

## Fix (both directions)
- **`json_canonical.rs`**: best-effort `repair_invalid_escapes` — a byte-level string-state machine that doubles every backslash not starting a valid escape (only inside string values; structural JSON untouched). `\'` becomes `\\'`, which parses as a literal `\'`. This is the response-path canonicalize entry, so `transform_codex_chat`, `streaming_codex_anthropic`, and `streaming_codex_chat` all benefit automatically.
- **`transform_codex_anthropic.rs:540`**: degrade unparseable historical arguments to `{}` + `warn!` instead of 400-ing — defense in depth so pre-existing poisoned history also recovers.

## Verification (TDD, all green on a clean upstream-main checkout)
- `canonicalize_tool_arguments_str_repairs_invalid_escape` — `\'` repaired to parseable JSON
- `canonicalize_tool_arguments_str_passthrough_when_unrepairable` — structural errors still passed through verbatim (no silent rewrite of non-escape failures)
- `test_request_function_call_with_invalid_escape_arguments_does_not_error` — request path tolerates poisoned history
- Updated `test_request_invalid_or_non_object_arguments_error` → split: unparseable degrades to `{}` (ok), non-object still rejected
- Full regression: `json_canonical` 9, `transform_codex_anthropic` 69, `transform_responses` 74, codex suite 533 — all pass

## Compatibility
Backward compatible. Valid JSON is unchanged (first parse succeeds, repair never runs). Only invalid-escape inputs are normalized, and only as a last-resort fallback.

### DIFF
--- a/src-tauri/src/proxy/json_canonical.rs
+++ b/src-tauri/src/proxy/json_canonical.rs
@@ -54,9 +54,67 @@ pub(crate) fn canonicalize_json_string_if_parseable(value: &str) -> String {
         return value.to_string();
     }
 
-    serde_json::from_str::<Value>(trimmed)
-        .map(|parsed| canonical_json_string(&parsed))
-        .unwrap_or_else(|_| value.to_string())
+    if let Ok(parsed) = serde_json::from_str::<Value>(trimmed) {
+        return canonical_json_string(&parsed);
+    }
+
+    // Third-party tool-call arguments sometimes contain invalid `\x` escapes
+    // inside string values (e.g. GLM emits `\'`, which JSON does not permit).
+    // Passing such text through verbatim poisons the session: the arguments
+    // land in the client's history and 400 the next request when the strict
+    // request path tries to parse them. As a best-effort repair, double every
+    // backslash that does not start a valid escape so it becomes a literal
+    // backslash; structural JSON outside string values is left untouched.
+    if let Ok(parsed) = serde_json::from_str::<Value>(&repair_invalid_escapes(trimmed)) {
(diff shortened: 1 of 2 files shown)


allenxu09:
@codex review

PR #5767, opened 2026-07-25
fix(ui): reduce idle GPU usage for route status

### LINKED ISSUE #5760: [Bug] 开启路由服务总开关后 GPU 占用持续偏高（WebView2 GPU 进程 ~17%），导致风扇持续高速运转
### Self Checks / 自检

### CC Switch Version / 版本号

v3.18.0

### Operating System / 操作系统

Windows

### Related App / 涉及应用

Claude Code

### Steps to Reproduce / 重现步骤

1. 打开 ccSwitch
2. 进入「设置 → 路由 → 本地路由」，开启路由总开关（并启用某个应用路由，如 Claude）
3. 不进行任何 API 调用，保持 ccSwitch 窗口打开（或最小化）
4. 打开 Windows 任务管理器，观察 ccSwitch 关联的 WebView2 GPU 进程

Image

### Expected Behavior / 期望行为

路由服务空闲时 GPU 占用应接近 0%（路由服务本身是 Rust 后端的 HTTP 代理，
理论上不产生 GPU 负载）。 #

### Actual Behavior / 实际行为

空闲状态下 WebView2 GPU 进程持续占用 10%–20% GPU，风扇持续运转。
怀疑是路由面板/用量统计界面的实时刷新（如统计数据、日志流、动画效果）
在持续触发 WebView2 重新渲染所致。
- 关闭路由总开关 → GPU 占用恢复正常，可确认是路由功能引起

### Additional Context / 补充信息

任务管理器显示 WebView2 GPU 进程占用 16.6% GPU。

allenxu09:
## Summary

- replace continuously running route and failover status pulses with a short heartbeat every three seconds
- pause the cosmetic heartbeat when the CC Switch window loses focus
- guard the Tauri focus listener so browser-only renderer development still starts normally
- leave TanStack Query polling unchanged when a visible window is unfocused
- respect the user's reduced-motion preference

## Root cause

The active route indicators used Tailwind's infinite `animate-pulse`. WebView2 kept ticking and compositing that opacity animation even while the route service was idle. WebView2 also does not reliably update page visibility when its host window is minimized, so the heartbeat uses host window focus events when running in Tauri.

This keeps the active-state feedback while allowing the WebView2 GPU process to become idle between heartbeats and whenever the window is unfocused or minimized, without freezing background polling for visible windows.

Fixes #5760

## Validation

- `pnpm typecheck`
- `pnpm exec prettier --check src/lib/windowActivity.ts src/main.tsx`
- `git diff --check`

### DIFF
--- a/src/components/proxy/ClaudeDesktopRouteToggle.tsx
+++ b/src/components/proxy/ClaudeDesktopRouteToggle.tsx
@@ -84,7 +84,7 @@ export function ClaudeDesktopRouteToggle({
           className={cn(
             "h-4 w-4 transition-colors",
             isRunning
-              ? "text-emerald-500 animate-pulse"
+              ? "text-emerald-500 status-heartbeat"
               : "text-muted-foreground",
           )}
         />
--- a/src/components/proxy/FailoverToggle.tsx
+++ b/src/components/proxy/FailoverToggle.tsx
@@ -72,7 +72,7 @@ export function FailoverToggle({ className, activeApp }: FailoverToggleProps) {
           className={cn(
             "h-4 w-4 transition-colors",
             isEnabled
-              ? "text-emerald-500 animate-pulse"
+              ? "text-emerald-500 status-heartbeat"
               : "text-muted-foreground",
           )}
         />
--- a/src/components/proxy/ProxyToggle.tsx
+++ b/src/components/proxy/ProxyToggle.tsx
@@ -75,7 +75,7 @@ export function ProxyToggle({ className, activeApp }: ProxyToggleProps) {
           className={cn(
             "h-4 w-4 transition-colors",
             takeoverEnabled
-              ? "text-emerald-500 animate-pulse"
+              ? "text-emerald-500 status-heartbeat"
               : "text-muted-foreground",
           )}
         />
--- a/src/components/settings/ProxyTabContent.tsx
+++ b/src/components/settings/ProxyTabContent.tsx
@@ -118,7 +118,7 @@ export function ProxyTabContent({
                 className="gap-1.5 h-6 ml-auto mr-2"
               >
                 <Activity
-                  className={`h-3 w-3 ${isRunning ? "animate-pulse" : ""}`}
+                  className={`h-3 w-3 ${isRunning ? "status-heartbeat" : ""}`}
                 />
                 {isRunning
                   ? t("settings.advanced.proxy.running")
--- a/src/index.css
+++ b/src/index.css
@@ -145,6 +145,30 @@ html.dark {
   color-scheme: dark;
 }
 
+.status-heartbeat {
+  opacity: 1;
+  transition-property: color, opacity;
+  transition-duration: 150ms, 300ms;
+  transition-timing-function: ease, ease-out;
+}
+
+html[data-window-active="true"][data-status-heartbeat="true"]
+  .status-heartbeat {
+  opacity: 0.5;
+}
+
+html[data-window-active="false"] .status-heartbeat {
+  opacity: 1;
+  transition-duration: 0s;
+}
+
+@media (prefers-reduced-motion: reduce) {
+  .status-heartbeat {
+    opacity: 1;
+    transition-property: color;
+  }
+}
+
 ::-webkit-scrollbar {
   display: none;
 }
--- a/src/index.css
+++ b/src/index.css
@@ -167,6 +167,11 @@ html[data-window-active="false"] .status-heartbeat {
     opacity: 1;
     transition-property: color;
   }
+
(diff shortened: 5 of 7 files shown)


farion1231:
@codex review

allenxu09:
@farion1231 ptal!

farion1231 (commented):
Thanks — the approach here is right, and I verified the key pieces: driving a short dim via a `data-` attribute + CSS transitions really does let the compositor go idle between beats (unlike an infinite `animate-pulse`, which keeps sampling every frame), the `.status-heartbeat` rule correctly overrides Tailwind's `transition-colors` (declared after `@tailwind utilities`, same specificity, so the later rule wins), and the reduced-motion override fully disables the opacity dim. The heartbeat lifecycle is also clean — `startHeartbeat()` re-entering through `stopHeartbeat()` means the duplicate events from the browser + Tauri listeners can't leak timers.

Two things need to change before merge, plus a rebase.

**[P2] Non-Tauri renderers crash at bootstrap** — [windowActivity.ts#L55-L59](https://github.com/farion1231/cc-switch/blob/1d537c17613e8c26d607b0c4c0bbe23408bffdc8/src/lib/windowActivity.ts#L55-L59)

`getCurrentWindow()` dereferences `window.__TAURI_INTERNALS__.metadata` synchronously, so in any non-Tauri environment it throws `TypeError: Cannot read properties of undefined (reading 'metadata')` before the `.catch()` ever attaches — `.catch` only handles promise rejections, not a synchronous throw. Since `initializeWindowActivity()` runs inside `bootstrap()` before `createRoot().render()` ([main.tsx#L114](https://github.com/farion1231/cc-switch/blob/1d537c17613e8c26d607b0c4c0bbe23408bffdc8/src/main.tsx#L114)) and `void bootstrap()` has no fallback catch, `pnpm dev:renderer` white-screens. This also defeats the browser-fallback comment on L51 — the fallback listeners never get a chance to matter. Note the existing convention: every `getCurrentWindow()` call in `App.tsx` is wrapped in try/catch for exactly this reason.

Minimal fix — guard the call:

```ts
import { isTauri } from "@tauri-apps/api/core";

if (isTauri()) {
  void getCurrentWindow()
    .onFocusChanged(({ payload }) => setWindowActive(payload))
    .catch((error) => {
      console.error("Failed to observe window focus changes", error);
    });
}
```

(or wrap it in try/catch, matching the `App.tsx` pattern). A small jsdom test asserting `initializeWindowActivity()` doesn't throw without `__TAURI_INTERNALS__` would lock this in.

**[P2] `focusManager.setFocused(...)` pauses polling for visible-but-unfocused windows, not just minimized ones** — [windowActivity.ts#L36](https://github.com/farion1231/cc-switch/blob/1d537c17613e8c26d607b0c4c0bbe23408bffdc8/src/lib/windowActivity.ts#L36)

Once `focusManager.setFocused(boolean)` is called, it permanently overrides TanStack Query's default `visibilitychange`-based detection (`isFocused()` prefers the manual value). Every query with the default `refetchIntervalInBackground: false` then stops polling whenever the window merely loses focus: the usage request log / trend charts (30s, `src/lib/query/usage.ts`), proxy status (2s, `src/hooks/useProxyStatus.ts`), failover state (5s, `src/lib/query/failover.ts`). Keeping the usage dashboard visible on a second monitor while working in an editor is a core workflow for this app — after this change those panels freeze until the window is clicked again (global `refetchOnWindowFocus` catches the data up at that point, but nothing updates while unfocused).

I get why focus is the signal — WebView2 doesn't reliably report visibility when minimized, so focus is the proxy that's available. But it's a stronger condition than the problem needs. Either of these keeps the GPU win without the behavior change:

1. Keep the cosmetic heartbeat focus-driven (that part is fine) and drop the `focusManager` wiring from this PR, or
2. On focus loss, only call `focusManager.setFocused(false)` when the window is actually minimized (`getCurrentWindow().isMinimized()`), so visible-but-unfocused windows keep polling.

**Rebase**: the PR currently conflicts with `main`, and the conflict is only in `src/main.tsx` — `main` gained the models.dev startup-sync block in the same bootstrap region. Keep both that and the new `initializeWindowActivity()` call.

nits (non-blocking):

- The heartbeat interval runs unconditionally, even when no `.status-heartbeat` element exists (no route/failover active) — an attribute flip + style recalc every 3s for nothing. Could start it only while something is actually pulsing.
- `AgentsPanel`'s `animate-pulse-slow` is the one remaining persistent animation with the same idle-GPU characteristics — fine as a follow-up, out of scope here.

Fix the crash guard, pick a direction on the polling scope, and rebase — then this is good to go. Happy to re-review quickly.

farion1231 (approved):

PR #5773, opened 2026-07-26
fix: preserve pins and extend failover grouping

senvev47:
## Summary
- Preserve session pin state across refreshes and add cleanup coverage for pinned sessions.
- Keep provider groups stable when the active provider moves within failover queues, including collapsed group state and group-level queue actions.
- Allow selected upstream 4xx failures to fail over when alternate providers are available, and add tests for the behavior.

## Tests
- cargo test --lib
- npm test -- --run tests/components/SessionManagerPage.test.tsx tests/components/ProviderList.grouping.test.tsx
- npm run lint
- npm run type-check
- Applied rustfmt and Prettier formatting fixes for CI.


farion1231:
@codex review

PR #5790, opened 2026-07-27
fix(grokbuild): sync image endpoint and credentials

skyswordw:
## Summary

Make Grok Build provider switches configure Grok Imagine alongside chat, so `/imagine` and `image_gen` use the selected relay without relying on environment variables at runtime.

- normalize the image endpoint and generation/editing feature overrides in the Rust live-write path
- merge the selected inline API key, or the resolved configured `env_key`, into `~/.grok/auth.json` under the `xai::api_key` scope
- preserve and restore a pre-existing user-owned `xai::api_key` when returning to Grok Official
- apply the same normalization and credential update during proxy hot switches while keeping image requests on the real upstream
- roll back Grok live files and logical current-provider state when a switch cannot complete

There is no new GUI control. Image synchronization is automatic for custom providers when their key is available.

## Root cause

Grok Build uses separate configuration paths for chat and image generation. Chat reads `[model.*].base_url` and `[model.*].api_key`, while image generation reads `[endpoints].xai_api_base_url` and the `xai::api_key` credential scope from `auth.json`. CC Switch previously updated only the chat model table.

## Behavior

| Scenario | Chat config | Image endpoint | Image credential |
| --- | --- | --- | --- |
| Add/switch custom relay | selected model table | same upstream base URL | selected inline key |
| Custom provider with available `env_key` | keeps `env_key` | same upstream base URL | resolved environment value |
| Custom provider with unavailable `env_key` | selected model table | image sync disabled | prior user-owned scope restored |
| Malformed credential or ownership state file | selected model table | image sync disabled | existing files left unchanged |
| Switch to Grok Official | exact official snapshot | official snapshot behavior | prior user-owned scope restored when readable |
| Local proxy takeover | local proxy URL | selected real upstream URL | selected upstream key |

## Safety

- the previous `xai::api_key` scope is stored in `~/.grok/.cc-switch-image-auth.json` with mode `0600` while CC Switch owns the live scope
- the active managed key is stored in the ownership state only as a SHA-256 fingerprint, avoiding a second plaintext copy
- restoration verifies both the CC Switch ownership marker and the managed-key fingerprint, so a later manual user edit is not overwritten
- unrelated TOML sections, MCP configuration, and unrelated `auth.json` scopes are preserved
- malformed credential files disable image synchronization without blocking the selected chat provider or rewriting the malformed files
- one Grok live snapshot covers `config.toml`, `auth.json`, the ownership state file, and logical current-provider rollback

## Verification

- `pnpm typecheck`
- `pnpm format:check`
- `pnpm test:unit` (79 files, 522 tests)
- `cargo clippy --all-targets --all-features -- -D warnings`
- `cargo test` (2,200 library tests passed, 2 ignored; all integration suites passed)

### DIFF
--- a/src-tauri/src/grok_config.rs
+++ b/src-tauri/src/grok_config.rs
@@ -1,6 +1,9 @@
-use serde_json::{json, Value};
+use chrono::{SecondsFormat, Utc};
+use serde::{Deserialize, Serialize};
+use serde_json::{json, Map, Value};
+use sha2::{Digest, Sha256};
 use std::fs;
-use std::path::PathBuf;
+use std::path::{Path, PathBuf};
 
 use crate::config::{get_home_dir, write_text_file};
 use crate::error::AppError;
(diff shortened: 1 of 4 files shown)


farion1231:
@codex review

PR #5795, opened 2026-07-27
feat: one-click provider share via ccswitch:// deeplink

gcdong:
## Summary / 概述

Adds a one-click "copy share link" action on provider cards. It generates a `ccswitch://v1/import?...` deeplink carrying the provider's full `settings_config` (base64 JSON), so the receiver imports a provider that is byte-for-byte identical to the sender's — including plugin/extra fields that the template-based import path would drop.

在供应商卡片上新增"复制分享链接"按钮，生成携带完整 `settings_config`（base64 JSON）的 `ccswitch://v1/import` 深链接，接收方导入后配置与发送方逐字一致（含 Key 与插件等额外字段）。

Three logical commits:

1. **Import passthrough** — when a deeplink carries a native-shaped config (Claude/Gemini: `env` object; Codex/GrokBuild: TOML string; OpenCode: `options` object; OpenClaw: `api` string; Hermes: `api_mode` string), the config is preserved verbatim instead of being rebuilt from extracted fields; falls back to the existing template path otherwise. URL params still override where supported.
2. **`generate_provider_deeplink` command** — builds the share URL, then re-parses it through the real import pipeline and rejects at generation time any link that would fail or mutate on the receiving side (no extractable key/endpoint, no inferable homepage, or non-verbatim `settings_config` round-trip). Claude Desktop is excluded.
3. **Frontend** — Share button on provider cards (hidden for Claude Desktop), clipboard copy with success/failure toasts; the success toast warns that the link contains the API key. i18n for zh/en/ja/ru.

Note: the link intentionally embeds the API key (that is the point of sharing a working provider); generation is gated so that OAuth/official providers without an exportable key cannot produce a link, and the toast explicitly warns the user.

## Related Issue / 关联 Issue

Fixes #

## Screenshots / 截图

## Checklist / 检查清单

Tests: 70 deeplink tests (incl. per-app passthrough/fallback/URL-override and share-gate round-trip fidelity cases) + frontend unit tests, all green.


farion1231:
@codex review

gcdong:
@codex review

Pushed a fix for the four findings (commit 4201ea33):
- P1: `meta.apiFormat` now carried via `apiFormat` param
- P2a: full `UsageScript` carried via `usageScriptConfig` base64-JSON blob (templateType / coding-plan creds preserved)
- P2b: `ANTHROPIC_API_KEY` extraction + credential field preserved on round-trip
- P2c: explicit apiKey/endpoint/model overrides now applied to native GrokBuild TOML
- added a generation-time guard rejecting Claude providers that depend on un-carried routing meta

Fidelity gate extended to `meta.api_format` + `meta.usage_script`. 74 deeplink tests + full lib suite (2232) green.

gcdong:
Fixed all `cargo fmt` formatting issues. All local checks pass now:

- `pnpm typecheck` ✅
- `pnpm format:check` ✅
- `cargo fmt --check` ✅
- `cargo clippy` ✅
- `pnpm test:unit` ✅ (526 tests)
- `cargo test` ✅

Could you please approve the CI workflow run?

gcdong:
All checks are green: https://github.com/farion1231/cc-switch/actions/runs/32326915660

A note on the one non-feature commit in this branch, `ci(wsl2): run prebuilt lib test binary instead of cargo`: `backend-windows-wsl2` points `TEMP` at `\\wsl.localhost\...` and then ran `cargo test` there. On a cold cargo cache cargo relinks in that step, and `mt.exe` cannot write its manifest temp file to a UNC path, so it fails with `LNK1327`. This was not introduced here — unrelated branches such as `integrate/cursor-main` and `fix/issue-6624-pi-import` hit the same failure, and since `actions/cache` only saves on success, the cache was never rewritten and every later run started cold and failed too.

The fix captures the test binary path while `TEMP` is still native and runs that binary directly after `TEMP` is redirected, so no cargo, compilation or `mt.exe` runs against the UNC share. Test arguments and `CC_SWITCH_TEST_HOME` / `std::env::temp_dir()` are unchanged. The run above happened to be a cold-cache run, so the fix is verified under the condition that used to fail.

Happy to split the CI fix into a separate PR if you'd prefer.

PR #5811, opened 2026-07-27
fix(Security): zip-slip on skill install, two credential leaks, and three panic paths

zayokami:
## Summary

Seven fixes in the v3.18.0 code: two credential-leak paths, an arbitrary-file-write on skill install, a path-traversal in the skill backup/sync handling, and three panic paths that fire on malformed upstream data or a user-edited config. Every fix has a regression test, and each test was **verified against the vulnerable code first** — reverting the guard makes the test fail, so none of these are speculative.

> **Note:** the credential-related fixes (1, 2, 5) were reported to you privately by email first, per the project's security policy — `/security/advisories/new` returns 404, so Private Vulnerability Reporting appears to be off for the repo (you can enable it under *Settings → Code security and analysis*). This description deliberately omits step-by-step exploit detail; the email has it.

## 1. `services/skill.rs` — zip-slip: arbitrary file write on skill install

`download_and_extract` built the output path from the **raw** ZIP entry name, stripped only the leading `<repo>-<branch>/` component, then `dest.join(relative_path)` → `fs::File::create`. An entry like `repo-main/../../../evil.sh` escaped the destination directory.

The archive is third-party controlled: it comes from `https://github.com/<owner>/<name>/archive/refs/heads/<branch>.zip`, and a skill repo (`owner`/`name`) can be added through an untrusted `ccswitch://` deeplink (`deeplink/skill.rs`), so an attacker controls the archive contents.

Fixed by resolving each entry through `enclosed_name()` (rejects `..` and absolute paths) before stripping the root and joining. This matches the two sibling extractors that were **already** correct (`extract_local_zip`, `webdav_sync/archive.rs`) — this call site was the one that had been missed. The loop is split into `extract_repo_archive()` so the guard is testable without network access.

*Reverse-proof:* with the old `file.name()` behaviour the new test fails (`zip-slip entry must not escape dest`).

## 2. `grok_config.rs` — remove the unconditional `XAI_API_KEY` fallback

`extract_credentials` now resolves a credential **only** from an explicit inline `api_key` or the process env var named by `env_key`. It previously also fell back unconditionally to `XAI_API_KEY` — silently substituting a different account's key precisely in the broken-config case (declared `env_key` var unset), paired with whatever `base_url` the config declares. An unset declared credential now yields `None` so callers fail loudly instead of transmitting the wrong secret.

*Reverse-proof:* with the fallback restored the test fails, returning `Some(("https://attacker.example/v1", "xai-secret-should-not-leak"))`.

## 3. `deeplink/provider.rs` — don't resolve env vars into plaintext on import

`merge_grokbuild_config` no longer calls `extract_credentials` to turn an env var into a plaintext `api_key`. A `ccswitch://` deeplink is untrusted input; resolving + inlining would persist the user's environment secret into the imported provider's `config.toml`. Only an explicitly-declared inline `api_key` carries through — `env_key` stays an indirection (a name), never a resolved secret.

## 4. Proxy transforms — stop panicking on malformed upstream data

serde_json's `value["key"] = …` panics when the receiver is a non-object `Value`; both sites index-assigned into data taken verbatim from an upstream response.

- **`transform_codex_anthropic.rs::anthropic_sse_to_message_value`** — only store a `message` / `content_block` when it is a JSON object. A malformed upstream could send a scalar or array there, and the later `["content"]` / `["text"]` / `["signature"]` assignments would panic. Reachable from `handlers.rs`, which falls back to SSE aggregation when an upstream returns an unmarked-content-type SSE body (the case #2234 describes).
- **`streaming_codex_anthropic.rs::responses_sse_events_from_anthropic_message`** — emit a `response.failed` event when the buffered body is a top-level JSON array/scalar (a gateway that ignores `stream:true` and returns HTTP 200 JSON), instead of index-assigning into a non-object.

*Reverse-proof:* a standalone repro confirms serde_json aborts on both shapes (`string-message panics: true`, `array-block panics: true`).

## 5. `services/provider/mod.rs` — strip all credentials from the shared Gemini snippet

`extract_gemini_common_config` skipped only `GOOGLE_GEMINI_BASE_URL` and `GEMINI_API_KEY`, but **`GOOGLE_API_KEY` is a first-class Gemini credential** (`provider.rs` resolves it via `first_non_empty(env, &["GEMINI_API_KEY", "GOOGLE_API_KEY"])`), so it was never stripped.

The snippet is not inert: `apply_common_config_to_settings` deep-merges it into the `env` of *every other* Gemini provider using the common config, and extraction runs automatically on startup, import, and switch. Net effect: account A's key lands in provider B and is sent to B's `GOOGLE_GEMINI_BASE_URL`, potentially a third-party relay. Anything else in `env` (`GOOGLE_APPLICATION_CREDENTIALS`, a proxy `*_AUTH_TOKEN`, …) leaked the same way.

Fixed by also skipping `is_sensitive_config_key(key)` — the pattern matcher the Claude extractor already relies on for exactly this reason (its comment notes a fixed enumeration "will always miss the next `*_API_KEY`"). Shareable non-secret config (e.g. `GEMINI_TIMEOUT_MS`) is preserved.

*Reverse-proof:* against the old two-key filter the test fails with `credential GOOGLE_API_KEY must not leak into the shared Gemini snippet`.

I checked the sibling extractors while here: OpenCode / OpenClaw / GrokBuild / Hermes / ClaudeDesktop are **not** affected in practice, because `apply_common_config_to_settings` returns their settings unchanged (no merge-back path), so nothing can leak provider-to-provider. Worth hardening later for defence-in-depth, but left alone here to keep this PR focused.

## 6. MCP config writers — normalize a non-table `mcp_servers`

`~/.grok/config.toml` and `~/.codex/config.toml` are user-editable. Both writers guarded only with `contains_key("mcp_servers")`, so a config where the key **exists but is not a table** (`mcp_servers = "x"` / `[]` / `42`) skipped the rebuild and then hit `doc["mcp_servers"][id] = …`, which panics in toml_edit's `IndexMut`. In the provider-switch flow that panic also fires *after* the DB and live write have committed, leaving a half-applied switch. The Codex path has the wider blast radius of the two.

For Codex the normalize-then-insert step is extracted into `upsert_mcp_server_table()` so the previously panic-prone logic is unit-testable without touching the real `~/.codex/config.toml`. Audited the siblings: `mcp/claude.rs`, `gemini.rs`, `hermes.rs`, `opencode.rs` don't have this pattern (their `contains_key` uses are read-only), and the other index-assignments in `codex.rs` operate on freshly built `Table::new()` values, which are safe.

*Reverse-proof:* a standalone repro walking the original code path prints `panic: true`.

## 7. `services/skill.rs` — validate `directory` from backups and sync-imported DB rows

The install pipeline sanitizes the on-disk skill directory name (`sanitize_install_name`), but two entry points bypassed it:

- **`restore_from_backup`** joined `metadata.skill.directory` — raw `meta.json` content — into the SSOT dir with no validation, then `copy_dir_recursive` into it and persisted the poisoned value to the DB. A crafted backup means arbitrary directory write plus a poisoned DB row.
- **Sync import** (WebDAV/S3) loads the remote database dump verbatim (`import_sql_string_for_sync`), and `save_skill` does not validate either, so a malicious or compromised sync snapshot can plant a `skills` row whose `directory` contains `../`.

Once poisoned, every raw join is attacker-controlled: `uninstall` / `remove_from_app` run `remove_dir_all` on it (**arbitrary directory deletion** — `directory = "../.."` resolves to the user's home), and `sync_to_app_dir` writes/symlinks outside the managed dirs.

Fixed with `require_valid_directory()` (built on the existing `sanitize_install_name`) enforced at all four sinks: `restore_from_backup`, `uninstall`, `sync_to_app_dir`, `remove_from_app`. Invalid values now fail loudly before any filesystem operation.

*Reverse-proof:* with the guard disabled, all four new tests fail — the restore test completes the traversal write and the uninstall test completes the traversal delete.

## Verification

- New regression tests: `extract_repo_archive_rejects_path_traversal_entries`, `does_not_fall_back_to_xai_api_key_when_declared_env_key_is_unset`, `extract_gemini_common_config_strips_credentials_keeps_shareable`, `test_anthropic_sse_aggregation_non_object_content_block_does_not_panic`, `test_anthropic_sse_aggregation_non_object_message_errors_not_panic`, `test_json_non_object_body_returns_failed_event_not_panic`, `upsert_normalizes_non_table_mcp_servers_without_panicking`, `upsert_preserves_existing_servers_in_a_valid_table`, `restore_from_backup_rejects_traversal_directory_in_metadata`, `remove_from_app_rejects_traversal_directory`, `uninstall_rejects_traversal_directory_from_db_row`, `require_valid_directory_accepts_single_segment_names_only`.
- `cargo test --lib` → **2191 passed / 0 failed** (merge-base baseline was 2179)
- `cargo clippy --lib -- -D warnings` → clean
- Happy to split this into separate PRs (credentials / zip-slip / panics) if you'd rather review them apart.

@farion1231 Please review this PR！Thank You！


farion1231 (approved):

PR #5818, opened 2026-07-28
feat: sanitize Chat message content for strict upstream APIs

Ustinian-ss:
## Summary

Add a Chat Completions message `content` field sanitization pass to prevent HTTP 400 errors from strict third-party OpenAI-compatible upstream APIs.

## Problem

When Codex sends requests through CC Switch to a third-party OpenAI-compatible gateway (e.g., reverse-proxy aggregators like `api.ltzy.top`), some messages may have `content` fields that are JSON `Object`, `Number`, `Boolean`, or `null` — types that are valid in Codex's Responses format but rejected by strict Chat Completions parsers:

```
HTTP 400: messages[13] 的 content 类型不合法，应为字符串或数组
```

This happens because `transform_codex_chat::responses_to_chat_completions_with_reasoning` faithfully converts between formats but doesn't normalize edge-case content types.

### New module: `transform_codex_chat_sanitize.rs`

A post-transform normalization pass that iterates through all `messages` and ensures each `content` field is either a `String` or `Array`:
- `String` / `Array` → unchanged (already valid)
- `null` → `""` (empty string)
- `Object` → JSON-serialized string
- `Number` / `Boolean` → `to_string()`

The function is idempotent and returns whether any changes were made.

### Integration point

The sanitizer should be called immediately after `responses_to_chat_completions_with_reasoning()` and before `inject_codex_chat_prompt_cache_key()` in `forwarder.rs`. See `forwarder.patch` for the exact diff (2 lines added at line 1440).

## Design

- Follows the same pattern as the existing `transform_codex_responses_xai_sanitize` module
- Deterministic and idempotent — running twice produces the same result
- Comprehensive test coverage

## Related

Fixes: "messages[N] content 类型不合法" errors with providers like `aqua` using models through `api.ltzy.top`.

### DIFF
--- a/forwarder.patch
+++ b/forwarder.patch
@@ -1,0 +1,11 @@
+--- a/src-tauri/src/proxy/forwarder.rs
++++ b/src-tauri/src/proxy/forwarder.rs
+@@ -1437,6 +1437,8 @@
+                 mapped_body,
+                 reasoning_config.as_ref(),
+             )?;
++            // Normalize message content for strict upstream Chat parsers
++            super::providers::transform_codex_chat_sanitize::sanitize_chat_message_content(&mut chat_body);
+             super::providers::inject_codex_chat_prompt_cache_key(
+                 provider,
+                 &mut chat_body,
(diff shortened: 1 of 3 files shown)


farion1231:
@codex review

PR #5826, opened 2026-07-28
fix(codex): avoid replaying synthetic reasoning items

### LINKED ISSUE #5825: [Bug] Synthetic Codex reasoning IDs fail after switching to native Responses
### Self Checks / 自检

- [x] I have read the FAQ in README.
- [x] I searched existing open and closed issues/PRs.

### CC Switch Version / 版本号

v3.18.0; also present on `main` at `ccda04bfa6d8397ec5a1a1c9b4cbf09458299d3c`.

### Operating System / 操作系统

Windows

### Related App / 涉及应用

Codex

### Steps to Reproduce / 重现步骤

1. Route a Codex task through a Chat Completions provider that returns `reasoning_content`, `reasoning`, or an inline `<think>` block.
2. Let CC Switch convert the response back to Responses format. It synthesizes a `reasoning` output item with an ID such as `rs_resp_chatcmpl_...`.
3. Continue the same task after switching to an official/native Responses provider where `store` is false.
4. Codex replays the synthetic item ID to the native provider.

### Expected Behavior / 期望行为

Chat-to-Responses conversion must not advertise a persisted Responses reasoning item unless such an item actually exists upstream. Reasoning text may still be retained internally and attached to converted tool calls for same-provider continuation.

### Actual Behavior / 实际行为

The native Responses endpoint rejects the replayed synthetic ID:

```text
unexpected status 404 Not Found: Item with id 'rs_resp_...' not found.
Items are not persisted when `store` is set to false. Try again with `store` set to true, or remove this item from your input.
```

Starting a new task works because its history does not contain the synthetic `rs_*` item.

### Root Cause / 根因

Both Chat-to-Responses conversion paths manufacture IDs that were never persisted by OpenAI:

- `transform_codex_chat.rs`: adds a synthetic `reasoning` output item with `id = rs_{response_id}`.
- `streaming_codex_chat.rs`: emits reasoning item/summary SSE events with the same synthetic ID.

Codex treats those events as replayable Responses items, so switching providers later turns the local synthetic ID into an upstream item lookup.

### Related Work / 相关工作

PR #5740 fixes invalid `resp_*_msg` message IDs and legacy message replay. It intentionally preserves reasoning/message pairs, so it does not address this separate unpersisted `rs_*` lookup failure.

tomfhcfhnb:
## Summary / 概述

- Stop Chat Completions to Responses conversion from emitting synthetic `reasoning` output items with locally generated `rs_*` IDs.
- Suppress the equivalent streaming reasoning item/summary events while still accumulating reasoning text for converted tool-call continuation.
- Prevent Codex from replaying an ID that never existed at the native Responses upstream after switching providers with `store=false`.

## Root Cause / 根因

The Chat bridge currently makes `rs_{response_id}` look like a persisted Responses item in both non-streaming JSON and streaming SSE. Codex records that output and may later replay it as an item reference. After switching back to an official/native Responses provider, the upstream tries to resolve the local synthetic ID and returns HTTP 404 because `store` is false.

This change keeps final text, tool calls, usage, and tool-call `reasoning_content`, but no longer advertises Chat-only reasoning as a replayable Responses item.

## Related Issue / 关联 Issue

Fixes #5825

Related: #5740 handles the separate invalid `resp_*_msg` message-ID problem. This PR deliberately leaves message-ID generation to that broader fix to avoid duplicate implementation.

## Validation / 验证

Focused validation on the same `ccda04bf` base commit before separating the overlapping #5740 message-ID changes:

- `cargo fmt --check`: passed
- non-streaming converter tests: 7 passed
- streaming converter tests: 19 passed
- verifies no reasoning item/events are emitted
- verifies reasoning remains attached to tool calls
- verifies inline `<think>` text does not leak and final text/usage remain intact
- `git diff --check`: passed on this focused branch

## Screenshots / 截图

Not applicable; this is a proxy protocol fix with no UI changes.

## Checklist / 检查清单

- [ ] `pnpm typecheck` passes (not applicable; no frontend changes)
- [ ] `pnpm format:check` passes (not applicable; no frontend changes)
- [ ] `cargo clippy` passes (pending CI/toolchain availability; Draft PR)
- [x] Updated i18n files if user-facing text changed (not applicable; no user-facing text changes)

### DIFF
--- a/src-tauri/src/proxy/providers/streaming_codex_chat.rs
+++ b/src-tauri/src/proxy/providers/streaming_codex_chat.rs
@@ -279,28 +279,10 @@ impl ChatToResponsesState {
     }
 
     fn push_reasoning_delta(&mut self, delta: &str) -> Vec<Bytes> {
-        let mut events = Vec::new();
-
-        if !self.reasoning.added {
-            let output_index = self.next_output_index();
-            let item_id = format!("rs_{}", self.response_id);
-            self.reasoning.output_index = Some(output_index);
-            self.reasoning.item_id = item_id.clone();
-            self.reasoning.added = true;
-
-            events.push(sse::reasoning_item_added(output_index, &item_id));
-            events.push(sse::reasoning_summary_part_added(output_index, &item_id));
-        }
-
+        // Keep reasoning for same-provider tool-call continuation, but do not
+        // expose a synthetic Responses item whose ID was never persisted.
         self.reasoning.text.push_str(delta);
-        let output_index = self.reasoning.output_index.unwrap_or(0);
-        events.push(sse::reasoning_summary_text_delta(
-            output_index,
-            &self.reasoning.item_id,
-            delta,
-        ));
-
-        events
+        Vec::new()
     }
 
     fn push_text_delta(&mut self, delta: &str) -> Vec<Bytes> {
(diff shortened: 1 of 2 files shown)


farion1231:
@codex review

PR #5838, opened 2026-07-28
fix(kimi): support K3 1M context in Claude Code

### LINKED ISSUE #5836: fix(kimi): derive Claude Code context window from the selected Coding Plan model
## Problem

CC Switch 3.18.0 treats every Claude Code provider whose `ANTHROPIC_BASE_URL` is `https://api.kimi.com/coding` as a 256K provider when the two context env keys are absent. `apply_kimi_for_coding_context_defaults` injects both `CLAUDE_CODE_MAX_CONTEXT_TOKENS` and `CLAUDE_CODE_AUTO_COMPACT_WINDOW` as `262144` without inspecting the selected model.

Kimi Coding Plan now exposes models with different limits on the same endpoint: K2.7 Code and `k3-256k` use 256K, while Allegretto-and-above accounts can select K3 with a 1M window. Kimi's Claude Code documentation uses the explicit `k3[1m]` env selector and sets both context env keys to `1048576`.

There is also an editor round-trip issue: `EditProviderDialog` initializes the current provider from live settings, so backend-injected context defaults can be shown as provider-owned values and saved back into the database.

## Expected behavior

- Derive the Kimi live default from the provider's model: `k3[1m]` gets `1048576`; existing K2.7/256K configurations retain `262144`.
- Preserve explicit provider values over defaults.
- Keep injection and switch-away stripping symmetric.
- Do not let live-only injected context values harden into the stored provider config when editing the current provider.
- Offer a Claude Code Kimi K3 1M preset without changing the existing 256K preset for lower plans/models.

## Environment

- CC Switch 3.18.0
- macOS
- Claude Code + `https://api.kimi.com/coding/`
- Kimi Allegretto, K3 1M

Official reference: https://www.kimi.com/code/docs/en/third-party-tools/claude-code.html

sparklesea:
## Summary

- derive the Kimi Coding context defaults from every configured Claude Code model route
- keep the existing 256K behavior for `kimi-for-coding`, `k3-256k`, and mixed model routes
- add a `Kimi K3 (1M)` preset using Kimi's documented `k3[1m]` selector and 1,048,576-token context values
- prevent injected live-only Kimi defaults from being serialized back into the provider record while preserving explicit provider overrides

## Root cause

The backend recognized `https://api.kimi.com/coding` only at the endpoint level and unconditionally injected `262144` for both context environment variables. When the active provider was edited, the frontend preferred the effective live settings over the database record, so those generated values appeared in the editor and were persisted again on save.

Kimi's current Claude Code documentation distinguishes the plan/model combinations: K2.7 Code and K3-256K use 256K, while Allegretto and above can request K3's 1M window with `k3[1m]`.

## User impact

Allegretto and higher-plan users can select the new K3 1M preset or explicitly store the 1M values without CC Switch rewriting them to 256K. Lower-plan and 256K model configurations retain the existing defaults. Because the context environment variables are global, a provider with mixed routes stays conservatively at 256K.

## Validation

- `tsc --noEmit`
- Prettier check for frontend sources
- Vitest: 79 files, 526 tests passed
- `cargo fmt --check`
- `cargo clippy --lib --tests -- -D warnings`
- `cargo test --lib`: 2247 passed, 2 ignored

Closes #5836

### DIFF
--- a/src-tauri/src/services/provider/live.rs
+++ b/src-tauri/src/services/provider/live.rs
@@ -29,13 +29,13 @@ use super::normalize_claude_models_in_value;
 /// usable context.
 const CODEX_OAUTH_CLAUDE_MAX_CONTEXT_TOKENS: &str = "372000";
 const CODEX_OAUTH_CLAUDE_AUTO_COMPACT_WINDOW: &str = "372000";
-const KIMI_FOR_CODING_CONTEXT_TOKENS: &str = "262144";
+const KIMI_FOR_CODING_256K_CONTEXT_TOKENS: &str = "262144";
+const KIMI_K3_1M_CONTEXT_TOKENS: &str = "1048576";
 
-/// Model env keys Claude Code may route requests through. The defaults above
-/// are calibrated against gpt-5.6's Codex catalog, so every configured model
-/// must belong to that family before they are injected — gpt-5.5's upstream
-/// catalog oscillates between 272K and 372K and must not inherit them.
-const CODEX_OAUTH_MODEL_ENV_KEYS: [&str; 6] = [
+/// Model env keys Claude Code may route requests through. Context-window envs
+/// are global, so every configured route must support a model-specific default
+/// before that default may be injected.
+const CLAUDE_MODEL_ENV_KEYS: [&str; 6] = [
     "ANTHROPIC_MODEL",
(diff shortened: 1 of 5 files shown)


farion1231:
@codex review

PR #5845, opened 2026-07-28
fix: strip namespace from input items for native Responses passthrough

N0TPO3T:
## Problem

When cc-switch takes over Codex (by setting a custom `model_provider` in config.toml), the Codex client sends `namespace` on `function_call` items in the `input[]` history array. The official ChatGPT backend (`chatgpt.com/backend-api/codex`) rejects this field with:

```
[ObjectParam] [input[115].namespace] [unknown_parameter] Unknown parameter: 'input[115].namespace'
```

This makes the Codex official subscription unusable through cc-switch, regardless of whether the "unify Codex session history" setting is enabled.

## Root Cause

When Codex uses a custom provider (which cc-switch sets up), it preserves `namespace` on `function_call` items in the `input[]` history. The official ChatGPT backend accepts `namespace`-typed tools in the `tools[]` array, but **rejects** `namespace` fields on individual `input[]` items.

The native Responses passthrough path (used by the official provider) does not strip this field, so the request is forwarded verbatim and the upstream rejects it.

## Fix

- Added `strip_namespace_from_input_items()` in `transform_codex_responses_namespace.rs`
  - Removes `namespace` from every object in the `input[]` array
  - The `namespace` for a function call can be inferred from the `tools[]` array, so stripping is safe
- Called in the native Responses passthrough path in `forwarder.rs`
  - Gated on: native passthrough (not Chat/Anthropic transform) + NOT xAI OAuth (which already handles namespace via `flatten_request_namespaces`)
- Added unit tests for the new function

## Testing

- `strip_namespace_from_input_items_removes_namespace_fields`: verifies namespace removal from input items
- `strip_namespace_noop_when_no_input_or_no_namespace`: verifies no-op on bodies without input array or without namespace fields
- Existing namespace flatten/restore tests continue to pass (unchanged)

### DIFF
--- a/src-tauri/src/proxy/forwarder.rs
+++ b/src-tauri/src/proxy/forwarder.rs
@@ -1563,6 +1563,27 @@ impl RequestForwarder {
             );
         }
 
+        // Native Responses passthrough: strip `namespace` from `input[]` items.
+        // Codex 0.142+ sends `namespace` on `function_call` items in the `input`
+        // history when the client is configured with a custom `model_provider`.
+        // The official ChatGPT backend (`chatgpt.com/backend-api/codex`) rejects
+        // this redundant field with `unknown_parameter: input[N].namespace`.
+        // xAI is excluded because its namespace flatten (above) already rewrites
+        // namespace-qualified calls, removing the `namespace` field.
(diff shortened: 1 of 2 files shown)


farion1231:
@codex review

PR #5881, opened 2026-07-29
fix(pricing): correct MiniMax M3 rates

### LINKED ISSUE #5841: [Bug] MiniMax-M3 built-in pricing uses the undiscounted rates
### Summary

The built-in `minimax-m3` pricing row currently uses these USD rates per million tokens:

- Input: `$0.60`
- Output: `$2.40`
- Cache read: `$0.12`
- Cache write: `$0`

MiniMax's current pay-as-you-go page marks the 50% discount as permanent. For requests with up to 512K input tokens, the effective rates are `$0.30`, `$1.20`, and `$0.06`. The current row therefore makes usage cost estimates twice the billed amount for the standard context tier.

Official pricing: https://platform.minimax.io/subscribe/token-plan?tab=api-enterprise

Current row: https://github.com/farion1231/cc-switch/blob/ff3bc242cc8c51c7d7f5540c867e0472cb2da5a8/src-tauri/src/database/schema.rs#L2168

The 512K–1M tier uses `$0.60`, `$2.40`, and `$0.12`, but the current pricing table stores only one rate per model. Using the standard ≤512K tier here is consistent with how the catalog handles other models without context-tiered pricing.

### Suggested fix

1. Change the seed row to `0.30 / 1.20 / 0.06 / 0`.
2. Add a guarded entry to `repair_current_model_pricing` that updates existing rows only when all fields still match the old built-in values. This preserves user-customized prices.
3. Add tests for a fresh database and for repairing an untouched old row while preserving a customized row.

No database schema change is needed.

nightcityblade:
## Summary / 概述

Use MiniMax M3's permanent discounted standard-tier rates for new databases, and repair existing rows only when all pricing fields still match the old built-in values. User-customized pricing remains untouched.

## Related Issue / 关联 Issue

Fixes #5841

## Screenshots / 截图

N/A — backend-only pricing data change.

## Testing / 测试

- `cargo test minimax_m3_pricing_repairs_old_but_preserves_custom_prices --lib` — passed
- `cargo test database::tests::model_pricing --lib` — passed
- `cargo clippy --lib` — passed
- `rustfmt --check --edition 2021 src/database/schema.rs src/database/tests.rs` — passed

## Checklist / 检查清单

- [ ] `pnpm typecheck` passes / 通过 TypeScript 类型检查 — N/A, no TypeScript changes
- [ ] `pnpm format:check` passes / 通过代码格式检查 — N/A, no frontend changes
- [x] Updated i18n files if user-facing text changed / 如修改了用户可见文本，已更新国际化文件 — N/A, no user-facing text changed

### DIFF
--- a/src-tauri/src/database/schema.rs
+++ b/src-tauri/src/database/schema.rs
@@ -2165,7 +2165,7 @@ impl Database {
                 "0.06",
                 "0.375",
             ),
-            ("minimax-m3", "MiniMax M3", "0.60", "2.40", "0.12", "0"),
+            ("minimax-m3", "MiniMax M3", "0.30", "1.20", "0.06", "0"),
             // GLM (智谱)
             ("glm-4.7", "GLM-4.7", "0.6", "2.2", "0.11", "0"),
             ("glm-4.6", "GLM-4.6", "0.6", "2.2", "0.11", "0"),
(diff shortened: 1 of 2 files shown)


farion1231:
@codex review

nightcityblade:
Closing as superseded by upstream commit f42534ed2648cea47bf9edd072f462672698e7e2, which now contains both the corrected MiniMax M3 seed price and the guarded repair for existing databases.

PR #5890, opened 2026-07-30
fix(session-manager): show forked Claude sessions under their project, with a title

### LINKED ISSUE #5888: Forked Claude Code sessions appear under "unknown directory" with no title in Session Manager
### Self Checks / 自检

### CC Switch Version / 版本号

3.18.0

### Operating System / 操作系统

Linux

### Related App / 涉及应用

Claude Code

### Steps to Reproduce / 重现步骤

1. In Claude Code, work in a session that touches many files (file-history snapshots accumulate).
2. Fork that session (or continue it via an agents-created fork).
3. Open cc-switch → Session Manager.
4. Look for the forked session under its project directory group.

### Expected Behavior / 期望行为

A forked session should appear in Session Manager exactly like a normal session:
listed under its project directory group (e.g. "spider_hub_workflow"), with a readable title and its metadata populated — same as the parent session it was forked from.

### Actual Behavior / 实际行为

The forked session is listed under the "unknown directory / 未知目录" group, titled only by its 8-char session-id prefix (e.g. "91cf3f79"), with no project
path and no creation date. Since it is not under the project group where the user looks for it, it effectively appears to be missing from session history.

The parent forked session shows the same symptom; normal (non-forked) sessions in the same directory display correctly.

### Additional Context / 补充信息

Traced in source: forked Claude Code sessions inherit the parent's metadata preamble at the top of the transcript (`ai-title` / `agent-name` / `mode` /
`permission-mode` + one `file-history-snapshot` per tracked file). On two real forked sessions this preamble is 104-105 lines, so the first record carrying `cwd` / `timestamp` sits at line 105/106 — while `parse_session` in src-tauri/src/session_manager/providers/claude.rs only scans the first 10 lines (`read_head_tail_lines(path, 10, 30)`), leaving project_dir / created_at / title all None.

Normal sessions are unaffected (`cwd` appears on line 1-4).

I have a fix ready and will open a PR referencing this issue: recover `cwd` from the tail records (already read — zero extra IO) and fall back to Claude
Code's own `ai-title` record (line 1) for the title. Deliberately not widening the head window: `file-history-snapshot` records are several KB each, and a 256-line head measured 35x more IO per scan on a real session directory.

Image

dreamingtech:
## Summary / 概述

Sessions forked in Claude Code (and agent-created forks) inherit the parent's metadata
preamble: `ai-title` / `agent-name` / `mode` / `permission-mode` plus one
`file-history-snapshot` per tracked file. In real forks this preamble runs 100+ lines
(measured on two real forked sessions: first record carrying `cwd` at line 105 / 106),
so nothing inside `parse_session`'s 10-line head window carries `cwd`, `timestamp`, or a
user message. The resulting meta has `project_dir = None` and `title = None`, and the
session renders under the "unknown directory" group titled only by its 8-char session-id
prefix — which users read as "my forked session is missing from session history".

Fix — without reading any extra bytes:

- **`project_dir`**: fall back to `cwd` from the tail records. The tail is already read
  for `last_active_at` / `summary` / `custom-title`, and tail records carry the same
  `cwd`, so the session returns to its project group for free. The tail loop's
  early-break now also waits for `project_dir`, so a trailing `custom-title` record
  cannot cut the scan short before the fallback runs.
- **title**: fall back to Claude Code's own `ai-title` record (line 1 of the transcript,
  already inside the head window). Ranked *below* the first user message, so existing
  sessions keep the title they have today.

Deliberately **not** widening the head window: `file-history-snapshot` records are
several KB each, so a 256-line head measured **35× more IO** across a real session
directory (205 KB → 7.07 MB per scan), and `scan_sessions` runs on every history
refresh. This version reads byte-for-byte what upstream reads.

Tests: 4 new cases (forked session recovers `project_dir` from tail; forked session
falls back to `ai-title`; normal session still prefers its first user message over
`ai-title`; tail early-break waits for `project_dir`). Fixture reproduces the real
preamble shape (4 metadata records + 100 `file-history-snapshot` records). Validated
against 7 real session files: all 7 resolve `project_dir`; the two forked ones get real
titles instead of ID prefixes.

Notes for reviewers:
- `created_at` stays `None` for forked sessions — list ordering is unaffected because
  sorting prefers `last_active_at` (read from the tail). Recovering it would require the
  expensive read this PR avoids.
- `codex.rs:300` / `openclaw.rs:187` use the same 10-line head window. Left untouched:
  I had no Codex/OpenClaw session data to verify whether their formats grow a similar
  preamble, and did not want to change parsers I could not test. If they do, the same
  two zero-IO fallbacks should port cleanly.

## Related Issue / 关联 Issue

Fixes #5888

## Checklist / 检查清单

- [x] `pnpm typecheck` passes / 通过 TypeScript 类型检查 (no frontend changes)
- [x] `pnpm format:check` passes / 通过代码格式检查 (no frontend changes)
- [x] `cargo clippy` passes (if Rust code changed)
- [x] Updated i18n files if user-facing text changed (no user-facing text changed)

### DIFF
--- a/src-tauri/src/session_manager/providers/claude.rs
+++ b/src-tauri/src/session_manager/providers/claude.rs
@@ -131,6 +131,12 @@ fn parse_session(path: &Path) -> Option<SessionMeta> {
     let mut project_dir: Option<String> = None;
     let mut created_at: Option<i64> = None;
     let mut first_user_message: Option<String> = None;
+    // Claude Code's own generated session title (`ai-title` record, written near the top).
+    // Forked sessions inherit a long metadata preamble (`ai-title` / `agent-name` / `mode` /
+    // `permission-mode` + one `file-history-snapshot` per tracked file — 100+ lines, each
+    // several KB), which pushes the first `cwd` / `timestamp` / user message far out of the
+    // head window. `ai-title` is the one title source that stays cheap to reach.
+    let mut ai_title: Option<String> = None;
 
     // Extract metadata and first user message from head lines
     for line in &head {
@@ -152,6 +158,13 @@ fn parse_session(path: &Path) -> Option<SessionMeta> {
         }
         if created_at.is_none() {
             created_at = value.get("timestamp").and_then(parse_timestamp_to_ms);
+        }
+        if ai_title.is_none() && value.get("type").and_then(Value::as_str) == Some("ai-title") {
+            ai_title = value
+                .get("aiTitle")
+                .and_then(Value::as_str)
+                .map(|s| s.trim().to_string())
+                .filter(|s| !s.is_empty());
         }
         // Extract first real user message as title candidate
         // Skip system-injected caveats and slash commands (e.g. /clear, /compact)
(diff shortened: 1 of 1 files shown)


farion1231:
@codex review

PR #5919, opened 2026-07-30
fix(security): cap scripts, file reads, and proxy bodies; surface deeplink usage fields

zayokami:
## Summary

Six security / robustness fixes, every one with a regression test, plus one CI reliability fix:

1. **`usage_script.rs` — cap untrusted JS execution** (security fix)
2. **`session_usage_grokbuild.rs` — skip oversized session log files** (robustness hardening)
3. **`session_usage_grokbuild.rs` — bound directory traversal, don't follow symlinks** (robustness hardening)
4. **`DeepLinkImportDialog.tsx` — show hidden usage access token / user id** (transparency fix)
5. **`codex_config.rs` — `model_catalog_json` arbitrary file read** (security fix)
6. **Proxy — unbounded response bodies / decompression bombs** (security fix)
7. **CI — set up pnpm via Corepack instead of `pnpm/action-setup` v6** (CI reliability fix)

Each code fix was verified against the vulnerable behavior first: reverting the guard makes the corresponding test fail. The CI change has no unit test; it is exercised by this PR's own workflow run.

---

## 1. Usage script: prevent infinite-loop DoS

`execute_usage_script` creates a QuickJS `Runtime` to run scripts that configure how to query a provider's remaining quota. These scripts can arrive from **untrusted input**:

- a `ccswitch://` deeplink can carry a Base64 `usageScript` parameter
- a WebDAV/S3 sync snapshot can import a provider row whose `usage_script` is enabled

The old code used a plain `Runtime::new()` with **no execution time limit, no memory limit, and no stack limit**. A malicious or simply buggy script containing `while (true) {}` would hang the backend thread until the process was killed.

### Fix

Add `create_script_runtime()` that configures the runtime **before** any `eval`:

- `set_interrupt_handler` with a 5-second deadline — QuickJS raises an uncatchable exception when time is up
- `set_memory_limit(16 * 1024 * 1024)` — caps JS heap growth
- `set_max_stack_size(256 * 1024)` — explicit stack budget

### Verification

Regression test `infinite_loop_usage_script_is_interrupted_before_blocking_the_backend` runs a `while(true)` script through the real `execute_usage_script` and asserts it returns an error in well under 15 seconds. A standalone repro using the original `Runtime::new()` pattern ran until killed by `timeout -s KILL 12` (exit code 137), confirming the issue is real and the test is not a tautology.

---

## 2. Grok session log: skip oversized files

`sync_single_grok_file` read the entire `updates.jsonl` with `fs::read_to_string` and had no size limit. A runaway CLI or a maliciously planted multi-gigabyte file under `~/.grok/sessions` could exhaust process memory during the periodic sync.

### Fix

Check `metadata.len()` before reading; skip files larger than 50 MiB with a warning. 50 MiB is far above normal usage (a single line is a few KiB) while still being safe to read as a string.

### Verification

Regression test `oversized_updates_jsonl_is_skipped_without_reading_into_memory` creates a file of `MAX_GROK_FILE_BYTES + 1` and asserts `sync_single_grok_file` returns zero imported rows without failing.

---

## 3. Grok session log: bound directory traversal

`collect_files_named` recursed with `path.is_dir()` and no depth limit. `is_dir()` follows symlinks, so a symlink cycle under `~/.grok/sessions` caused unbounded recursion and a stack-overflow abort.

### Fix

- Add a 16-level depth limit
- Use `entry.metadata()` and skip symlinks entirely instead of following them

### Verification

Regression test `symlink_cycle_does_not_cause_stack_overflow` creates `sessions/enc-project/sub/cycle -> ../..` and asserts the collector returns only the real `updates.jsonl` without crashing. A standalone repro using the original recursive logic produced `thread 'main' has overflowed its stack`.

---

## 4. Deeplink provider import: show hidden usage fields

The provider import confirmation dialog parsed `usageAccessToken` and `usageUserId` from `ccswitch://` URLs and persisted them into `ProviderMeta.usage_script`, but never rendered them. A malicious or overbroad deeplink could smuggle credentials into the imported provider without the user seeing them before clicking **Import**.

### Fix

- Add both fields to the usage-script section of `DeepLinkImportDialog`:
  - `usageAccessToken` is masked (first 4 chars + `************`), consistent with `usageApiKey`.
  - `usageUserId` is shown in full with `break-all` so long identifiers are not truncated by CSS.
- New i18n keys added for `en`, `zh`, `zh-TW`, and `ja`.

### Verification

Regression test `renders masked usage access token and user id for provider imports` emits a `deeplink-import` event and asserts the dialog displays the masked token and the user id.

---

## 5. Codex `model_catalog_json` arbitrary file read

`resolve_cc_switch_catalog_path` in `src-tauri/src/codex_config.rs` resolved the `model_catalog_json` pointer from the live `~/.codex/config.toml`. It only verified that the file name equals `cc-switch-model-catalog.json`; if the value was an absolute path, it was returned verbatim without checking it lives under `~/.codex`.

That meant a user (or a malicious sync/profile import) could set:

```toml
model_catalog_json = "/tmp/secret/cc-switch-model-catalog.json"
```

and CC Switch would read the file during live-settings backfill and, when the proxy is running, serve it from `/v1/models`. Pointing it at a huge file also exhausted memory because both readers used `read_to_string` without a size cap.

### Fix

- Add `path_is_within(base, path)` in `config.rs` — a lexical containment check that normalizes paths without hitting the filesystem, so it is safe for non-existent paths and avoids symlink tricks.
- Treat Unix-style absolute paths (`/…`) as absolute even on Windows.
- Resolve relative paths under the Codex config directory, not directly to the generated filename.
- Cap catalog reads at **32 MiB** (`read_codex_model_catalog_text`).

### Verification

New regression tests:

- `resolve_catalog_rejects_absolute_path_outside_config_dir`
- `resolve_catalog_accepts_absolute_path_inside_config_dir`
- `resolve_catalog_rejects_traversal_to_parent_directory`
- `read_limited_string_rejects_oversized_file`

Reverting the containment check makes the first test fail.

---

## 6. Unbounded proxy response bodies / decompression bombs

The proxy's non-streaming and error paths collected upstream response bodies with `response.bytes().await` and, when `content-encoding` was present, fully decompressed them with no output limit. A malicious or misbehaving upstream could therefore send an arbitrarily large body, or a tiny gzip/zstd/br body that expands to gigabytes, and exhaust the backend's memory.

### Fix

- Add `ProxyResponse::bytes_with_limit(max_bytes)` that aborts collection when the accumulated body exceeds **128 MiB**.
- Use it in `read_decoded_body` for every non-streaming path.
- After decompression, reject payloads that exceed the same ceiling with the new `ProxyError::ResponseBodyTooLarge` variant.
- Map the new error into the Codex proxy error response builder so it is surfaced correctly to clients.

### Verification

New regression tests:

- `bytes_with_limit_rejects_oversized_buffered_response`
- `bytes_with_limit_rejects_oversized_streamed_response`

Reverting the limit makes both tests fail.

---

## 7. CI: replace `pnpm/action-setup` v6 with Corepack

`pnpm/action-setup` v6 no longer downloads the requested pnpm directly. It bootstraps the *latest* pnpm (11.7.0 at the time of writing) through an npm-driven self-installer, and only then downloads the pinned version (10.12.3) from the npm registry to "switch". That is three registry round-trips before any build step runs, and the second one is fragile: when the `@pnpm/linux-x64-10.12.3.tgz` fetch failed on a degraded runner network (`error (23)`, two retries), the self-installer still reported `done`, then crashed with `ENOENT ... @pnpm/linux-x64/package.json` and exit code 254, failing the CI setup step. The floating `@v6` tag also means upstream changes to this installer land in our CI without notice — several regressions are currently open against v6.0.x.

### Fix

- Drop `pnpm/action-setup` from `ci.yml` and `release.yml`; set up pnpm with Node 20's bundled **Corepack** (`corepack enable && corepack install`) — a single, exact-version download with no floating bootstrap.
- Pin the version once via `"packageManager": "pnpm@10.12.3"` in `package.json`, which Corepack reads both in CI and locally (single source of truth).
- Set `COREPACK_ENABLE_DOWNLOAD_PROMPT=0` so the step can never block on an interactive prompt.
- `release.yml` no longer needs its Windows ARM64 special case (Corepack is a pure-JS shim with no per-arch binaries), so all five release runners now share one setup step.

### Verification

Both workflow files parse cleanly. Because GitHub runs `pull_request` workflows from the PR branch, this PR's own CI exercises the new Corepack-based `Setup pnpm` step end to end.

---

## Test / lint status

Backend:
- `cargo test --lib -- usage_script session_usage_grokbuild` → **27 passed / 0 failed**
- `cargo test --lib proxy` → **1250 passed / 0 failed**
- `cargo test --lib codex_config` → **73 passed / 0 failed**
- `cargo test --lib config` → **295 passed / 0 failed**
- `cargo clippy --lib -- -D warnings` → clean
- `cargo fmt --check` → clean

Frontend:
- `pnpm typecheck` → clean
- `pnpm format:check` → clean
- `pnpm vitest run tests/components/DeepLinkImportDialog.test.tsx` → **1 passed**

CI:
- `ci.yml` / `release.yml` YAML validated; the `Setup pnpm` step now installs pnpm 10.12.3 via Corepack (reads the version from `package.json`'s `packageManager` field)

Note: `services::model_pricing` tests already fail on `main` (unrelated, environment-dependent). I verified they fail identically with and without these changes.

---

## Round 2 — review feedback addressed

Thanks for the close read — all three P1s were real gaps in fix 6's coverage, and both P2s were holes in fixes 4/5. Everything below landed in `b0f8d7ae`, same methodology: each fix ships a regression test that fails when the guard is reverted.

### [P1] The four missed forwarder paths — fixed

`forwarder.rs` L2311 (non-2xx error body), L2349 (failover pre-read), L2371/L2396 (2xx error-envelope validators) now all read via `bytes_with_limit(MAX_RESPONSE_BODY_BYTES)` and decode via `decompress_body_with_limit`. The error-path fallback (`_ => raw.to_vec()`) now falls back to bytes that are themselves already capped, so no path propagates an unbounded body. `ProxyResponse::bytes()` is gone entirely — no caller can accidentally go back to an unbounded read.

### [P1] `bytes_with_limit` now truly bounds every variant — fixed

Hyper/Reqwest/Streamed share one chunk-by-chunk accumulation loop (`bytes_stream()`) that errors and drops the connection the moment the budget is exceeded; only the already-in-memory `Buffered` variant keeps the post-hoc check. I used a manual accumulation loop rather than `http_body_util::Limited` because the loop can raise the precise `ProxyError::ResponseBodyTooLarge(size)` (mapped to 502) instead of a generic body error that would be misclassified as a failover-able forward failure.

Tests: a real TCP upstream that streams a 16 MiB body while counting bytes actually written. New assertions: client errors with `ResponseBodyTooLarge` **and** the server wrote less than half the body — the second assertion fails under the old collect-then-compare code, which drains all 16 MiB before erroring. Covered for both the Hyper and Reqwest variants.

### [P1] Decompression budget moved into the decoders — fixed

New `decompress_body_with_limit` in `content_encoding.rs`: gzip/deflate/zstd decode through `Read::take(limit + 1)`, brotli through its `Read` adapter with the same wrapper, erroring `DecompressError::TooLarge` the moment the output budget is exhausted — nothing expands in full first. Stacked encodings (`gzip, zstd`) are bounded at every stage, including intermediates. `TooLarge` is a distinct error from corrupt-data `Io`, so `read_decoded_body` maps it to 502 (`ResponseBodyTooLarge`) while keeping the fail-open-warn behavior for genuinely broken streams; the old unbounded `decompress_body` remains only for the request side, which has its own size constraints.

Tests: a gzip bomb truncated mid-stream must error `TooLarge`, not `UnexpectedEof` — that fails under any decode-then-compare implementation; zstd/br bombs from a few-KiB input are rejected; stacked-encoding intermediates are rejected; exactly-at-limit payloads still pass; plus an end-to-end `read_decoded_body` test where a ~130 KiB gzip expands past 128 MiB.

### [P2] Usage credentials visible without `usageScript` — fixed

The dialog's usage section gate now mirrors `build_provider_meta` 1:1 — any of `usageScript` / `usageEnabled` / `usageApiKey` / `usageBaseUrl` / `usageAccessToken` / `usageUserId` / `usageAutoInterval` opens the section, so a deeplink carrying only `usageAccessToken`/`usageUserId` shows those rows before Import. The script code block and execution warning stay gated on `usageScript` (they describe the script); the enabled badge shows whenever `usageScript` or `usageEnabled` is present. Regression test: script-less deeplink renders the masked token + user id rows and no script warning — reverting the gate makes it fail.

### [P2] `path_is_within` symlink escape — fixed

`resolve_cc_switch_catalog_path` now, after the lexical check, canonicalizes the existing file and re-verifies containment against the canonicalized `~/.codex` (prefix-consistent on Windows `\\?\` and macOS `/private`), then returns the canonical path so the read never traverses the symlink component. Non-existent files keep the lexical result (callers already handle ENOENT). A base that fails to canonicalize falls back to the lexical base, which can only over-reject (degrades to "no catalog"), never over-accept. Tests: `link/cc-switch-model-catalog.json` with `~/.codex/link` pointing outside is rejected; a real file inside is still accepted. The `path_is_within` doc comment now says explicitly it is not a symlink defense.

### Nits — all three addressed

- `session_usage_grokbuild.rs`: comment rewritten to match the code (every symlink is skipped unconditionally, including symlinked `updates.jsonl`), and each skip logs at info level so a user whose sessions live behind a symlink can diagnose silently missing usage data.
- `codex_config.rs`: the WSL-style-absolute-path behavior change is now documented at the `is_unix_absolute` site as an intentional decision, including the self-heal-on-next-switch rationale.
- `CONTRIBUTING.md`: added a note (EN + ZH) that the `packageManager` pin auto-switches Corepack users to pnpm 10.12.3 (one-time download prompt), non-Corepack users are unaffected (`package-manager-strict-version` defaults off), and pnpm upgrades now happen by editing that field rather than via Dependabot.

### Round-2 test / lint status

Backend:
- `cargo clippy --all-targets -- -D warnings` → clean
- `cargo fmt --check` → clean
- `cargo test --lib` → **2262 passed / 4 failed** — the 4 failures are the same pre-existing environment-dependent `services::model_pricing` tests noted in round 1; they fail identically without these changes.

Frontend:
- `pnpm typecheck` → clean
- `pnpm format:check` → clean
- `pnpm test:unit` → **566 passed / 3 failed** — the 3 failures are pre-existing flaky `tests/integration/App.test.tsx` timeouts under full-suite load; I verified they fail identically on clean `main` (and pass in isolation). The two dialog tests pass, including the new script-less-credentials regression test.

One thing I deliberately left alone: `xai_oauth_auth.rs` has the same collect-then-compare shape against its own `MAX_OAUTH_RESPONSE_BYTES` on the reqwest path. It talks to a fixed first-party OAuth endpoint, so it's out of this PR's threat model, but the same streaming-accumulation pattern could be applied there as a follow-up if you want consistency.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -24,10 +24,12 @@ jobs:
           node-version: "20"
 
       - name: Setup pnpm
-        uses: pnpm/action-setup@v6
-        with:
-          version: 10.12.3
-          run_install: false
+        env:
+          COREPACK_ENABLE_DOWNLOAD_PROMPT: "0"
+        run: |
+          corepack enable
+          corepack install
+          pnpm --version
 
       - name: Get pnpm store directory
         id: pnpm-store
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -83,19 +83,11 @@ jobs:
             || sudo apt-get install -y --no-install-recommends libsoup2.4-dev
 
       - name: Setup pnpm
-        if: runner.os != 'Windows' || matrix.arch != 'arm64'
-        uses: pnpm/action-setup@v6
-        with:
-          version: 10.12.3
-          run_install: false
-
-      - name: Setup pnpm (Windows ARM64)
-        if: runner.os == 'Windows' && matrix.arch == 'arm64'
-        shell: pwsh
+        env:
+          COREPACK_ENABLE_DOWNLOAD_PROMPT: "0"
         run: |
-          $ErrorActionPreference = 'Stop'
           corepack enable
-          corepack prepare pnpm@10.12.3 --activate
+          corepack install
           node --version
           pnpm --version
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -24,6 +24,14 @@ There are many ways to contribute:
 - Rust 1.85+ and Cargo
 - [Tauri 2.0 prerequisites](https://v2.tauri.app/start/prerequisites/)
 
+> **pnpm version**: the exact version is pinned via the `packageManager` field
+> in `package.json` (and used by CI, which sets pnpm up through Corepack).
+> With Corepack enabled locally (`corepack enable`), your pnpm shim switches
+> to the pinned version automatically, with a one-time download on first use;
+> without Corepack, any pnpm 8+ keeps working (`package-manager-strict-version`
+> defaults to off). To upgrade pnpm, edit the `packageManager` field — Dependabot
+> bumps no longer cover the version.
+
 ### Quick Start
 
 ```bash
@@ -151,6 +159,12 @@ CC Switch supports three languages. When modifying user-facing text:
 - Node.js 18+ 和 pnpm 8+
 - Rust 1.85+ 和 Cargo
 - [Tauri 2.0 开发环境](https://v2.tauri.app/start/prerequisites/)
+
+> **pnpm 版本**：确切版本由 `package.json` 的 `packageManager` 字段固定
+> （CI 通过 Corepack 按此字段安装）。本地启用 Corepack（`corepack enable`）
+> 后，pnpm shim 会自动切到该版本（首次使用需一次性下载）；未启用 Corepack
+> 时任意 pnpm 8+ 均可正常使用（`package-manager-strict-version` 默认关闭）。
+> 升级 pnpm 请直接修改 `packageManager` 字段，不再由 Dependabot 代劳。
 
 ### 快速开始
 
--- a/package.json
+++ b/package.json
@@ -18,6 +18,7 @@
   "keywords": [],
   "author": "Jason Young",
   "license": "MIT",
+  "packageManager": "pnpm@10.12.3",
   "devDependencies": {
     "@tauri-apps/cli": "^2.8.0",
     "@testing-library/jest-dom": "^6.6.3",
--- a/src-tauri/src/codex_config.rs
+++ b/src-tauri/src/codex_config.rs
@@ -1200,6 +1200,11 @@ pub(crate) fn resolve_cc_switch_catalog_path(
         return None;
     }
 
+    // 注意（有意的行为变更）：Windows 上 `/…` 形式的旧 WSL 风格 Linux 路径也会
+    // 被视为绝对路径，从而在下方的包含性校验中失败——此前这类路径会因无法匹配
+    // 生成文件名而回退为按文件名解析、碰巧能工作。可接受：下一次切换供应商时
+    // 写入侧会重新落一个裸文件名，配置自愈（见
+    // `set_catalog_json_none_removes_cc_switch_owned_by_filename` 的场景注释）。
     let is_unix_absolute = catalog_path_str.starts_with('/');
     let resolved = if referenced_path.is_absolute() || is_unix_absolute {
         referenced_path.to_path_buf()
@@ -1214,6 +1219,37 @@ pub(crate) fn resolve_cc_switch_catalog_path(
             base_dir.display()
         );
         return None;
+    }
+
+    // 词法包含不等于运行时包含：配置目录内的符号链接（如 ~/.codex/link ->
+    // /etc）能让 `link/cc-switch-model-catalog.json` 通过上面的检查，读取却
+    // 落到目录外。文件存在时把真实路径 canonicalize 出来再校验一次，并把
+    // canonical 路径返回给调用方——后续读取不再经过 symlink 组件。
+    if resolved.exists() {
+        let canonical = match fs::canonicalize(&resolved) {
+            Ok(path) => path,
+            Err(error) => {
+                log::warn!(
+                    "Codex model_catalog_json canonicalize 失败: {}: {error}",
+                    resolved.display()
+                );
+                return None;
+            }
+        };
+        // base 同样 canonicalize，保证两侧前缀一致（Windows \\?\、
+        // macOS /tmp -> /private/tmp）；base 失败时退回词法 base——
+        // 词法 base 与 canonical 路径比较只会误拒（退化为不读），不会误放。
+        let canonical_base = fs::canonicalize(base_dir).unwrap_or_else(|_| base_dir.to_path_buf());
+        if !path_is_within(&canonical_base, &canonical) {
+            log::warn!(
+                "Codex model_catalog_json 经符号链接解析到配置目录外: {} -> {}（允许目录: {}）",
+                resolved.display(),
+                canonical.display(),
+                canonical_base.display()
+            );
+            return None;
+        }
+        return Some(canonical);
     }
 
     Some(resolved)
@@ -3808,6 +3844,52 @@ model_catalog_json = "cc-switch-model-catalog.json"
         assert_eq!(
             result, None,
             "relative traversal outside ~/.codex must not be accepted"
+        );
+    }
+
+    #[test]
+    fn resolve_catalog_rejects_symlink_escaping_config_dir() {
+        // 词法包含可被符号链接绕过：~/.codex/link -> 外部目录，
+        // "link/cc-switch-model-catalog.json" 词法上在 base 内，真实读取却落到
+        // base 外。canonicalize 之后的二次校验必须拒绝。
+        let temp = tempfile::tempdir().expect("tempdir");
+        let base_dir = temp.path().join("codex");
+        let outside_dir = temp.path().join("outside");
+        fs::create_dir_all(&base_dir).expect("create base");
+        fs::create_dir_all(&outside_dir).expect("create outside");
+        let escaped_file = outside_dir.join(CC_SWITCH_CODEX_MODEL_CATALOG_FILENAME);
+        fs::write(&escaped_file, r#"{"models":[]}"#).expect("write escaped catalog");
+
+        #[cfg(unix)]
+        std::os::unix::fs::symlink(&outside_dir, base_dir.join("link")).expect("symlink");
+        #[cfg(windows)]
+        std::os::windows::fs::symlink_dir(&outside_dir, base_dir.join("link")).expect("symlink");
+
+        let config_text = r#"model_catalog_json = "link/cc-switch-model-catalog.json"
+"#;
+        let result = resolve_cc_switch_catalog_path(config_text, &base_dir);
+        assert_eq!(
+            result, None,
+            "symlink escaping the config dir must be rejected after canonicalization"
+        );
+    }
+
+    #[test]
+    fn resolve_catalog_accepts_real_file_inside_config_dir() {
+        // 存在于 base 内的真实文件：canonical 校验通过后仍应接受
+        let temp = tempfile::tempdir().expect("tempdir");
+        let base_dir = temp.path().join("codex");
+        fs::create_dir_all(&base_dir).expect("create base");
+        let catalog_file = base_dir.join(CC_SWITCH_CODEX_MODEL_CATALOG_FILENAME);
+        fs::write(&catalog_file, r#"{"models":[]}"#).expect("write catalog");
+
+        let config_text = r#"model_catalog_json = "cc-switch-model-catalog.json"
+"#;
+        let result = resolve_cc_switch_catalog_path(config_text, &base_dir);
+        let resolved = result.expect("real file inside config dir should be accepted");
+        assert_eq!(
+            resolved.file_name().and_then(|n| n.to_str()),
+            Some(CC_SWITCH_CODEX_MODEL_CATALOG_FILENAME)
         );
     }
 
--- a/src-tauri/src/codex_config.rs
+++ b/src-tauri/src/codex_config.rs
@@ -2,8 +2,8 @@ use std::collections::HashSet;
 use std::path::{Path, PathBuf};
 
 use crate::config::{
-    atomic_write, delete_file, get_home_dir, read_json_file, sanitize_provider_name,
-    write_json_file, write_text_file,
+    atomic_write, delete_file, get_home_dir, path_is_within, read_json_file,
+    sanitize_provider_name, write_json_file, write_text_file,
 };
 use crate::error::AppError;
 use crate::model_capabilities::{image_input_capability_from_modalities, ImageInputCapability};
@@ -1128,17 +1128,28 @@ pub fn prepare_codex_config_text_with_model_catalog(
 /// All failure modes (missing file, parse error, no `model_catalog_json`,
 /// entries without `slug`) collapse to `Ok(None)` so callers can treat this
 /// as best-effort enrichment without making `read_live_settings` brittle.
+/// 模型目录文件读取上限（32 MiB）。目录 JSON 正常只有几百 KiB；超过则视为异常，
+/// 避免指向外部大文件时耗尽内存。
+const MAX_CODEX_CATALOG_BYTES: u64 = 32 * 1024 * 1024;
+
 pub fn read_codex_model_catalog_simplified_from_live() -> Result<Option<Value>, AppError> {
     let config_text = read_codex_config_text()?;
-    let generated_path = get_codex_model_catalog_path();
-    let Some(catalog_path) = resolve_cc_switch_catalog_path(&config_text, &generated_path) else {
+    let config_dir = get_codex_config_dir();
+    let Some(catalog_path) = resolve_cc_switch_catalog_path(&config_text, &config_dir) else {
         return Ok(None);
     };
     if !catalog_path.exists() {
         return Ok(None);
     }
-    let Ok(catalog_text) = fs::read_to_string(&catalog_path) else {
-        return Ok(None);
+    let catalog_text = match read_limited_string(&catalog_path, MAX_CODEX_CATALOG_BYTES) {
+        Ok(text) => text,
+        Err(error) => {
+            log::warn!(
+                "拒绝读取越界或过大的 Codex 模型目录 {}: {error}",
+                catalog_path.display()
+            );
+            return Ok(None);
+        }
     };
     Ok(build_simplified_catalog_from_texts(
         &config_text,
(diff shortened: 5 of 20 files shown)


zayokami:
@farion1231 麻烦尽快审核一下，谢谢啦🫡🌹

farion1231:
@codex review

farion1231 (commented):
Thanks — the direction of all six fixes is right, and the methodology here is better than most security PRs we receive: every fix ships a regression test verified against the reverted guard, `path_is_within`'s lexical normalization genuinely collapses `..` components, and the two `Runtime::new()` sites you capped are indeed the only QuickJS entry points in the tree. That said, three of the fixes don't fully close the holes they describe, so a few things need to land before this can merge.

**[P1] The proxy body cap misses four forwarder paths — including the error path the description claims is covered.**
[`forwarder.rs` L2311, L2349, L2371, L2396](https://github.com/farion1231/cc-switch/blob/2dc6603c16a934592ce4c7fe26e01f2772202a28/src-tauri/src/proxy/forwarder.rs#L2305-L2403) still call unbounded `response.bytes()`: the non-2xx error path (L2311), the failover non-streaming pre-read in `prepare_success_response_for_failover` (L2349), and the 2xx error-envelope validators (L2371, L2396). Three of them then feed upstream-controlled bytes into `decompress_body` with no cap. Same threat model as fix 6, unfixed. Minimal fix: switch these to `bytes_with_limit(MAX_RESPONSE_BODY_BYTES)` and apply the post-decompress ceiling there too.

**[P1] `bytes_with_limit` only truly bounds the `Streamed` variant.**
The [Hyper arm collects the full body before checking (L195) and the Reqwest arm does the same (L205)](https://github.com/farion1231/cc-switch/blob/2dc6603c16a934592ce4c7fe26e01f2772202a28/src-tauri/src/proxy/hyper_client.rs#L192-L215). Real upstream responses arrive as exactly those two variants, so an oversized plaintext body is still fully buffered — the limit prevents propagation, not the OOM itself. Notably, the two new tests cover only `Buffered`/`Streamed`, the variants where the limit does abort. Minimal fix: wrap the hyper body in `http_body_util::Limited::new(body, max_bytes)` before collecting, and accumulate reqwest via `bytes_stream()` with the cap. (The existing inline comment flags this too.)

**[P1] The decompression-bomb check runs after full expansion.**
[`response_processor.rs` L115–L118](https://github.com/farion1231/cc-switch/blob/2dc6603c16a934592ce4c7fe26e01f2772202a28/src-tauri/src/proxy/response_processor.rs#L113-L121) checks the size only after `decompress_body` returns, and [`decompress_single`](https://github.com/farion1231/cc-switch/blob/2dc6603c16a934592ce4c7fe26e01f2772202a28/src-tauri/src/proxy/content_encoding.rs#L31-L68) uses unbounded `read_to_end` / `zstd::decode_all` / `BrotliDecompress`. A ≤128 MiB zstd/br input can still balloon in memory before the check ever sees it. Minimal fix: bound the decoders themselves — `Read::take(limit + 1)` on the flate2/zstd readers, a bounded output for brotli — and error when the budget is exhausted. (Also flagged by the existing inline comment.)

**[P2] Usage credentials are still imported silently when no `usageScript` is present.**
The dialog's entire usage section is gated on [`request.usageScript` (L645)](https://github.com/farion1231/cc-switch/blob/2dc6603c16a934592ce4c7fe26e01f2772202a28/src/components/DeepLinkImportDialog.tsx#L645), and your two new rows (L737+) live inside that gate. But [`build_provider_meta` (L235–L289)](https://github.com/farion1231/cc-switch/blob/2dc6603c16a934592ce4c7fe26e01f2772202a28/src-tauri/src/deeplink/provider.rs#L235-L289) builds and persists a `UsageScript` carrying `access_token`/`user_id` whenever *any* usage field is present, script or not. So a deeplink with only `usageAccessToken`/`usageUserId` (the pre-existing `usageApiKey`/`usageBaseUrl` are affected the same way) persists credentials with nothing shown to the user — the exact gap fix 4 set out to close, and the backend comment you added calls the dialog display load-bearing for consent. Minimal fix: widen the section gate to "any usage field present"; the per-field conditions can stay as they are.

**[P2] `path_is_within` can be escaped through symlinks.**
The check is lexical by design, but the read path then follows links: `exists()` → `fs::metadata` → `read_to_string`. `model_catalog_json = "link/cc-switch-model-catalog.json"` with `~/.codex/link` pointing elsewhere [passes the containment check (L1210)](https://github.com/farion1231/cc-switch/blob/2dc6603c16a934592ce4c7fe26e01f2772202a28/src-tauri/src/codex_config.rs#L1203-L1216) and reads outside the config dir. Impact is bounded — planting a symlink in `~/.codex` already requires local write access as the user, and the exposure is localhost `/v1/models` — so I read this as defense-in-depth, but the doc comment's "avoids symlink-related surprises" currently oversells it. Minimal fix: after the lexical check, `canonicalize()` the existing file and re-verify containment before reading.

**nit:** [`session_usage_grokbuild.rs` L159–L163](https://github.com/farion1231/cc-switch/blob/2dc6603c16a934592ce4c7fe26e01f2772202a28/src-tauri/src/services/session_usage_grokbuild.rs#L156-L167) — the comment says symlinked dirs are "tolerated once at the current depth", but the code unconditionally skips every symlink, including symlinked `updates.jsonl` files that were followed before. The conservative behavior is fine; please fix the comment and add a log line on skip, so a user whose sessions live behind a symlink can diagnose silently missing usage data.

**nit:** [`codex_config.rs` L1203](https://github.com/farion1231/cc-switch/blob/2dc6603c16a934592ce4c7fe26e01f2772202a28/src-tauri/src/codex_config.rs#L1203-L1204) — treating `/…` as absolute on Windows means a legacy WSL-style Linux path (the scenario your own `set_catalog_json_none_removes_cc_switch_owned_by_filename` test comment mentions) now fails containment where it previously fell back to the generated path and worked. It self-heals on the next provider switch since the writer emits the bare filename, so this is acceptable — just noting the behavior change so it's a decision rather than an accident.

**nit:** the `packageManager` pin means contributors with corepack-enabled pnpm shims get auto-switched to 10.12.3 (with a one-time download prompt) inside this repo; non-corepack users are unaffected since `package-manager-strict-version` defaults to false. Worth a line in the dev docs, and note pnpm upgrades now happen by editing this field rather than via dependabot bumping the action.

To merge: the three P1s plus the P2 dialog gate (it's a small change, same PR). The symlink P2 and the nits can be follow-ups if you prefer. Happy to re-review quickly once pushed — the test discipline in this PR made verification fast, and I'd like to see this land.

zayokami:
SGTM! Thanks a lot for the incredibly detailed and thorough review. You caught some really good edge cases (especially regarding the decompression bomb and stream buffering). I'm working on the P1/P2 fixes and will push an update soon!

zayokami:
@farion1231 Thanks again for the thorough review! I've addressed all the P1s, P2s, and nits in the latest commit:
Fixed the missing body caps and stream buffering limit for Hyper/Reqwest.
Added bounded readers for decompression.
Widened the usage section gate in the UI.
Fixed symlink handling in path_is_within with canonicalize().
Updated the comments and docs as noted in the nits.
Please take another look when you have time!

zayokami:
Thanks again for the excellent and precise feedback! I have addressed all points (3 P1s, 2 P2s, and 3 Nits) in commit `b0f8d7ae`.

Here is a detailed itemized breakdown of the updates:

### P1: Proxy Body Caps & Early Aborts

1. **Missed Forwarder Paths (`forwarder.rs` L2311, L2349, L2371, L2396)**
   - Replaced all four unbounded `response.bytes()` calls with `bytes_with_limit` followed by bounded decompression.
   - Completely deleted `ProxyResponse::bytes()` from the codebase so unbounded reads cannot be accidentally reintroduced in the future.

2. **True Early Abort for Hyper / Reqwest / Streamed**
   - Implemented chunk-level accumulation so oversized bodies abort immediately during streaming.
   - *Design Choice:* Instead of `http_body_util::Limited`, I used a manual accumulation loop to explicitly yield a distinct `ResponseBodyTooLarge` error (mapped to HTTP 502). This prevents the cap breach from being misclassified as a generic/failoverable network error.
   - *Testing:* Added an integration test spinning up a real TCP upstream sending a 16 MiB payload, asserting that the client errors out early and the upstream writes less than half the body.

3. **Decoder-Level Decompression Limits**
   - Pushed size limits down into the decoders (`decompress_body_with_limit`). Used `Read::take(limit + 1)` on `flate2`/`zstd` readers and `Decompressor + take` for Brotli to prevent memory expansion during decoding.
   - Distinguishes `TooLarge` (mapped to 502) from data corruption / IO errors (which retain warn/fallback behavior).
   - *Testing:* Added a test ensuring truncated/high-ratio gzip streams trigger `TooLarge` rather than `UnexpectedEof`.

---

### P2: Consent Gate & Symlinks

4. **Dialog Consent Gate Alignment (`build_provider_meta`)**
   - Widened the dialog's usage section gate to render if *any* usage field is present (matching backend persistence logic).
   - Script code display and execution warnings remain strictly gated on `request.usageScript`.
   - *Testing:* Added a regression test asserting masked credentials are shown for scriptless deeplinks containing only `usageAccessToken`/`usageUserId`.

5. **Symlink Escape in `path_is_within`**
   - When the file exists, `path_is_within` now resolves the canonical path via `canonicalize()` and re-verifies containment against the canonical base directory.
   - Updated the doc comments to explicitly clarify the symlink defense behavior.

---

### Nits & Maintenance

- **`session_usage_grokbuild.rs`**: Aligned comment with the code (unconditionally skipping all symlinks) and added an `info!` log line when skipping.
- **`codex_config.rs`**: Added a comment noting that treating `/...` as absolute on Windows is an intentional trade-off (self-heals on the next provider switch).
- **`CONTRIBUTING.md`**: Updated both English and Chinese docs regarding the `packageManager` pnpm pin and Corepack auto-switch behavior.

---

### Verification & Follow-up

- `cargo clippy --all-targets -D warnings` & `cargo fmt --check` pass cleanly.
- `cargo test --lib` passed (4 unrelated failures checked and confirmed to be pre-existing environment-dependent pricing tests).
- `pnpm typecheck`, `format:check`, and `test:unit` all pass.

> **Proactive Note on Legacy Code:**
> Documented Round 2 in `PR_security_all.md`. As part of a self-audit, I noted a legacy site in `xai_oauth_auth.rs` that reads the full response before checking. Since this communicates strictly with a fixed xAI OAuth endpoint, it falls outside this threat model, but I've logged it for potential follow-up.

Ready for another look!

farion1231 (approved):
Thank you for your contributions！

PR #5932, opened 2026-07-30
stop skipping Claude MCP sync when config dir doesn't exist

xdfqgg:
The guard function should_sync_claude_mcp() was used to skip both sync and remove operations when ~/.claude/ or ~/.claude.json don't exist yet. This meant users who configured CC Switch before ever launching Claude Code never had their MCP servers deployed — the sync returned Ok without writing anything.

Fix: remove the guard from sync_enabled_to_claude and sync_single_server_to_claude. The underlying write_json_value already handles creating parent directories. Keep the guard only on remove_server_from_claude (nothing to remove from a nonexistent file).

Updated the corresponding integration test to verify that the config file IS created when toggling a server onto a fresh Claude install.

## Related Issue / 关联 Issue

Fixes #

## Screenshots / 截图

## Checklist / 检查清单

### DIFF
--- a/src-tauri/src/mcp/claude.rs
+++ b/src-tauri/src/mcp/claude.rs
@@ -8,9 +8,11 @@ use crate::error::AppError;
 
 use super::validation::{extract_server_spec, validate_server_spec};
 
+/// Check if Claude's config locations exist on disk.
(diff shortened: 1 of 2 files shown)


farion1231:
@codex review

xdfqgg (commented):
1

PR #5950, opened 2026-07-31
fix(hermes): refresh model defaults on provider switch

### LINKED ISSUE #5933: Bug: apply_switch_defaults preserves stale model.context_length/base_url when switching Hermes providers (Kimi K3 256k vs 1M)
### Summary
When switching Hermes providers via cc-switch, `model.default` and `model.provider` are updated correctly, but `model.context_length`, `model.base_url`, and `model.max_tokens` retain values from the **previous** provider. This causes a config mismatch when the two providers declare different context lengths or base URLs.

### Concrete case (Kimi K3 Coding Plan)
Kimi's Coding Plan exposes two variants of the same model:
- `k3-256k` — context 256,000
- `k3` (1M variant) — context 1,024,000

These are registered in cc-switch as two separate Hermes providers (`kimik3` and `kimik3-copy`), each with its own `models[0].context_length`.

**Steps:**
1. Activate `kimik3-copy` (1M variant) — `model.context_length` correctly becomes `1048576`.
2. Switch to `kimik3` (256k variant).

**Expected `config.yaml` after step 2:**
```yaml
model:
  default: k3-256k
  provider: kimik3
  context_length: 256000       # ← should follow the new provider's models[0].context_length
```

**Actual `config.yaml` after step 2:**
```yaml
model:
  default: k3-256k
  provider: kimik3
  context_length: 1048576      # ← stale value from the previous provider
  base_url: https:https://api.kimi.com/coding/   # ← also stale (and shows a separate double-scheme concatenation bug)
```

Hermes then runs the 256k model while believing it has a 1M context window, which silently breaks context budgeting / auto-compaction behavior.

### Root cause (source reference)
`src-tauri/src/hermes_config.rs` L927–947, `apply_switch_defaults`:

```rust
pub fn apply_switch_defaults(
    provider_id: &str,
    settings_config: &serde_json::Value,
) -> Result<HermesWriteOutcome, AppError> {
    let first_model_id = settings_config
        .get("models")
        .and_then(|v| v.as_array())
        .and_then(|arr| arr.first())
        .and_then(|m| m.get("id"))
        .and_then(|id| id.as_str())
        .map(|s| s.trim().to_string())
        .filter(|s| !s.is_empty());

    let current = get_model_config()?.unwrap_or_default();
    let merged = HermesModelConfig {
        default: first_model_id.or(current.default.clone()),
        provider: Some(provider_id.to_string()),
        ..current                       // ← L944: carries over stale base_url / context_length / max_tokens / extra
    };
    set_model_config(&merged)
}
```

The doc comment at L915–926 explicitly states this is intentional:
> Existing fields in `model:` (`context_length` / `max_tokens` / `base_url` / `extra`) are preserved via struct-update.

That design works when the user has only one provider, or when all providers share the same context/base_url. It breaks down when two providers are variants of the **same** model with different context lengths (e.g. Kimi K3 256k vs 1M) — the user expects switching the provider to also switch the per-model metadata declared in `settings_config.models[0]`.

### Suggested fix
In `apply_switch_defaults`, when `settings_config.models[0]` declares `context_length` (and optionally `max_tokens` / `base_url`), overwrite the corresponding fields in `HermesModelConfig` instead of falling back to `..current`. Only preserve `..current` for fields the new provider does **not** declare.

Sketch:

```rust
let first = settings_config
    .get("models")
    .and_then(|v| v.as_array())
    .and_then(|arr| arr.first())
    .cloned();

let first_model_id = first.as_ref()
    .and_then(|m| m.get("id"))
    .and_then(|id| id.as_str())
    .map(|s| s.trim().to_string())
    .filter(|s| !s.is_empty());

let new_ctx = first.as_ref()
    .and_then(|m| m.get("context_length"))
    .and_then(|v| v.as_u64());

let current = get_model_config()?.unwrap_or_default();
let merged = HermesModelConfig {
    default: first_model_id.or(current.default.clone()),
    provider: Some(provider_id.to_string()),
    context_length: new_ctx.or(current.context_length),
    ..current
};
set_model_config(&merged)
```

(Same treatment could be applied to `base_url` if the new provider declares one, which would also fix the secondary `https:https://…` concatenation issue visible in the example above.)

### Environment
- cc-switch: latest Windows MSI (installed 2026-07)
- OS: Windows 10
- Hermes config path: `C:\Hermes\config.yaml` (via `HERMES_HOME`)
- Provider settings in cc-switch DB (sanitized):
  - `kimik3` → `models[0] = { id: "k3-256k", context_length: 256000 }`
  - `kimik3-copy` → `models[0] = { id: "k3", context_length: 1024000 }`

Happy to test a fix or provide the full sanitized `config.yaml` if useful.

Komikawayi:
## Summary

- refresh Hermes top-level `model` defaults when switching providers
- update `base_url`, `context_length`, and `max_tokens` only when their current values still match the previous provider declaration
- preserve explicit user overrides and unknown model fields
- ignore blank URLs/model ids and non-positive numeric defaults

## Scope

This fixes cross-provider switches such as the Kimi K3 1M to 256k case. Re-activating an edited provider without changing provider id is unchanged. The malformed double-scheme URL normalization mentioned in the issue is a separate bug and is not handled here.

This supersedes the base-url-only approaches in #2333 and #2760.

## Testing

- `cargo test --manifest-path src-tauri/Cargo.toml --lib hermes_config::tests::apply_switch_defaults -- --nocapture` (5 passed)
- `cargo test --manifest-path src-tauri/Cargo.toml --lib` (2224 passed, 2 ignored)
- `git diff --check`

Fixes #5933

### DIFF
--- a/src-tauri/src/hermes_config.rs
+++ b/src-tauri/src/hermes_config.rs
@@ -911,6 +911,55 @@ pub fn set_model_config(model: &HermesModelConfig) -> Result<HermesWriteOutcome,
     write_yaml_section_to_config("model", &yaml_val)
 }
 
+#[derive(Default)]
+struct HermesSwitchDefaults {
+    default: Option<String>,
+    base_url: Option<String>,
+    context_length: Option<u64>,
+    max_tokens: Option<u64>,
+}
+
+fn hermes_switch_defaults(settings_config: &serde_json::Value) -> HermesSwitchDefaults {
+    let first_model = settings_config
+        .get("models")
+        .and_then(|v| v.as_array())
+        .and_then(|arr| arr.first());
+    HermesSwitchDefaults {
+        default: first_model
+            .and_then(|model| model.get("id"))
+            .and_then(|id| id.as_str())
+            .map(str::trim)
+            .filter(|id| !id.is_empty())
+            .map(str::to_string),
+        base_url: settings_config
+            .get("base_url")
+            .and_then(|value| value.as_str())
+            .map(str::trim)
+            .filter(|value| !value.is_empty())
+            .map(str::to_string),
+        context_length: first_model
+            .and_then(|model| model.get("context_length"))
+            .and_then(|value| value.as_u64())
+            .filter(|value| *value > 0),
+        max_tokens: first_model
+            .and_then(|model| model.get("max_tokens"))
+            .and_then(|value| value.as_u64())
+            .filter(|value| *value > 0),
+    }
+}
+
+fn refresh_provider_derived_value<T: PartialEq>(
+    current: Option<T>,
+    previous: Option<T>,
+    next: Option<T>,
+) -> Option<T> {
+    match (current, previous) {
+        (None, _) => next,
+        (Some(current), Some(previous)) if current == previous => next,
+        (current, _) => current,
+    }
+}
+
 /// Apply the top-level `model:` defaults when switching to a Hermes provider.
 ///
 /// `model.provider` is **always** updated to the new provider id — without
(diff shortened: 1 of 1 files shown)


farion1231:
@codex review

PR #5958, opened 2026-07-31
feat(claude-desktop): add safe 1P to 3P data migration

jieshu666:
> **Status: Draft** — backend migration engine + tests are complete and validated; the UI flow is implemented but still needs on-device manual verification (see below). Feedback on the safety model and scope is very welcome.
>
> **Contribution note:** I read `CONTRIBUTING.md` — I know this is a larger feature PR and that the project prefers an issue first. I couldn't find an existing issue that covers 1P→3P *data migration* specifically (#3061 is adjacent — it's about `Claude-3p` session visibility, which this fixes as a side effect), so I've opened this as a **Draft** to propose the safety model and scope for discussion. Very happy to open a dedicated tracking issue, or split this into smaller stacked PRs (engine → commands → UI), if that's easier to review.

> **Codex review (all 6 findings) — addressed in commits `02798e57` (backend) and `e17bec68` (frontend):**
>
> - **P1 · Return the ledger when a later component fails** — `apply` now returns a partial result carrying the persisted ledger; the UI surfaces it and Undo targets that exact ledger.
> - **P1 · Check Cowork metadata before deleting it** — metadata gets its own fingerprint; Undo removes a task only when both metadata and session dir still match what was installed, so a task you updated after migrating is never orphaned.
> - **P1 · Hash directory contents before destructive restore** — `dir_fingerprint` now includes per-file SHA-256, so a same-size edit after migration is detected and restore keeps the directory.
> - **P1 · Retain registry entries that lack an ID** — Undo now removes only IDs this migration installed; id-less / malformed entries are preserved.
> - **P2 · Add controls for ambiguous account roots** — the plan step now lists candidate roots and lets you pick one (pinned as an explicit override), instead of dead-ending on the blocked Continue.
> - **P2 · Verify only the components that were applied** — verification is scoped to the applied components, so a successful partial migration verifies cleanly.
>
> **CI:** the previous backend failures were a rust 1.95 clippy lint (`unnecessary_sort_by`) fixed in `02798e57`; the re-run needs the maintainer-approval gate (`action_required`) — frontend job already passed.

## Problem

On macOS, Claude Desktop's **official (1P)** mode and **third-party (3P)** mode use two completely separate data roots:

- 1P: `~/Library/Application Support/Claude`
- 3P: `~/Library/Application Support/Claude-3p`

When a user switches from the official provider to a third-party provider, CC Switch rewrites the sign-in method / profile configuration but **does not move any data**. Their existing Claude Desktop **Code sessions** and **Cowork tasks** are not deleted — they remain in the 1P root — but the 3P profile reads from the 3P root, so to the user the data looks like it **disappeared**.

Related to #3061 (user noticed CC Switch does not surface the data under `Claude-3p/local-agent-mode-sessions/`).

## Root cause

Provider switching only writes `deploymentMode` / profile / meta. The data sub-directories (`claude-code-sessions/…`, `local-agent-mode-sessions/…`) live under the per-mode app root and are never carried over, so each mode sees only its own root.

## Scope (this PR)

An **opt-in, safe, reversible** local migration that copies data from the 1P root into the 3P root, with a guided UI:

- **Read-only discovery** of the macOS 1P / 3P roots (no guessing, no writing).
- **Account-root candidate identification** for Code & Cowork; **refuses to guess** when there are multiple candidates and surfaces a blocking issue instead.
- **Pre-migration Audit / Plan** (read-only): per-component new/conflict counts, missing folders, missing transcripts, estimated copy/backup sizes.
- **Code** index non-overwrite migration (shared transcripts are left in place — they are already shared).
- **Cowork** `local_*.json` + same-name session dir **paired** migration (stage → hash-verify → atomic install; metadata written last).
- **Folder-binding repair**: rewrites only structural path fields (`cwd` / `originCwd` / `userSelectedFolders` / `fsDetectedFiles[].hostPath`), rebuilds `resolvedFolderKinds` **only for folders that genuinely exist**, and supports explicit OLD→NEW prefix mappings for relocated folders. **No global string replace** on chat bodies.
- **Auto-backup** of the target roots before any write; **stage + verify + atomic install**; **idempotent** (re-running installs nothing).
- **Migration ledger** + **Verify** + **Restore** (ledger-driven; removes only what the migration installed, never post-migration sessions).
- **Frontend**: audit → plan → explicit second confirmation → apply → verify / undo, with progress and per-component breakdown. i18n in en / zh / ja / zh-TW.
- **Optional-in-v1** (all additive, skip-on-conflict): Scheduled task definitions (imported `enabled: false`), Project cache copy (missing only), Artifact registry merge by ID.

### Explicitly NOT in scope

Cloud Chat history, Memory export/import, Skills/Plugins packaging, Connectors/OAuth, **Cookie/Token/Keychain/OAuth/API Key**, caches (Cache/Code Cache/GPUCache), `vm_bundles`, `claude-code-vm`, and any whole-app-dir brute copy. The migration never reads or moves credentials.

## Migration flow

```
audit (read-only)  →  plan (read-only, blocking issues gate)  →  confirm (explicit consent + quit Claude Desktop)
      →  backup target roots  →  stage + hash-verify  →  atomic install (no overwrite, skip existing)
      →  write metadata last  →  ledger  →  verify  →  (optional) restore
```

## Safety guarantees

- **Source (1P) is strictly read-only** — verified by a source-fingerprint check in tests.
- **Never overwrite** any existing 3P record; duplicates are skipped.
- **Refuses to run while Claude Desktop is running** (pgrep gate at the command layer).
- **Never guesses** among multiple candidate roots — surfaces a blocking issue.
- **Never generates or guesses device/account UUIDs.**
- `audit.jsonl` / `uploads/` / `outputs/` / transcripts are copied **byte-for-byte** (content-hash multiset verified); only structural path fields are transformed.
- `resolvedFolderKinds` is rebuilt **only** for folders that exist on disk; relocated folders go through an explicit OLD→NEW mapping.
- **Backup & ledger contain no chat bodies and no credentials** — the ledger records only component, ids, target paths, and content fingerprints.

## Backup + restore behavior

- Before any write, the **target** account roots are snapshotted to `backups/claude-desktop-data-migration/claude-3p-before-migration-<ts>-<nonce>/`.
- A `migration-ledger.json` + `plan.json` are written into the same backup dir.
- **Restore** is surgical and ledger-driven: it removes only ledger-installed records (fingerprint-verified before deleting anything) and reverts registry merges (scheduled-tasks / artifacts) by removing only the IDs the migration added. **Records created after the migration are never touched.**

## Tests

Backend (`src-tauri/src/claude_desktop_data_migration.rs`, 35 tests, all synthetic temp-dir fixtures — no real user data):

- unique account root resolution; **refuse to guess on multiple roots**; missing-target-seed reporting
- code index with/without shared transcript; **no transcript duplication**
- normal Cowork pair; scheduled Cowork; **missing session dir → failed, not silently repaired**
- target conflict → skip without overwrite; **double-run idempotency**
- folder exists / relocated (path map) / missing (kept, kind dropped); **no prose rewrite**
- project cache missing-only; artifact missing-entity skip; **scheduled stays disabled**
- **no staging leftovers**; stale staging dir handled; **source never modified**
- **restore removes only installed records and preserves new sessions**; restore without ledger
- spaces / apostrophes / non-ASCII paths; verify pass/fail; audit read-only; path-map validation; prefix-boundary safety

Validation run locally:

| Check | Result |
| ----- | ------ |
| `cargo test --lib claude_desktop_data_migration` | **29 passed / 0 failed** |
| `cargo fmt --check` | pass |
| `cargo clippy --all-targets -- -D warnings` | pass (0 warnings) |
| `cargo test` (full suite) | 2267 passed, **1 pre-existing env failure** (see below) |
| `tsc --noEmit` | pass |
| `prettier --check` | pass |
| `vitest run` | 573 passed, **2 pre-existing flaky** (see below) |

**Pre-existing failures (not caused by this PR — verified against a clean tree):**

- `cargo test` full run: `services::provider::…proxy_takeover_is_active` failed with `Address already in use (os error 48)`. **Root cause (verified, unrelated to this PR):** the test starts the real proxy service, which binds the fixed default port `127.0.0.1:15721` (`src/proxy/types.rs:46`, schema `DEFAULT 15721`). On the machine used for validation, the *installed* CC Switch app was running and already bound to `127.0.0.1:15721` (confirmed via `lsof`), so the bind fails. It fails identically in a single-test isolation run (2269 other tests filtered out) and does not reference this PR's code (`src/services/provider/mod.rs` has no dependency on the migration module). It should pass on any machine/CI runner where that port is free. I deliberately did **not** kill the running app to force it green.
- `vitest`: `tests/integration/App.test.tsx` (2 MSW timing tests) fail identically on the pristine base tree (confirmed via `git stash`); `App.test.tsx` mocks `ProviderList`, and the dedicated `ProviderList.test.tsx` (which exercises the changed component) passes.

## Manual verification (still needed — help wanted)

I could not exercise the live UI against real data, so the following need an on-device pass before this leaves Draft:

1. Claude Desktop panel → "Migrate official (1P) data…" card → wizard opens and audit renders.
2. Plan step shows correct per-component counts; blocking issues disable Continue.
3. Apply is refused while Claude Desktop is running; succeeds after quitting.
4. Post-apply: 3P Claude Desktop shows the migrated Code sessions & Cowork tasks; scheduled tasks remain disabled.
5. Verify passes; Undo removes only the migrated records.

## Screenshots

_Not attached yet._ Capturing real screenshots requires running the app against real Claude Desktop data, which I deliberately did not do. I will add synthetic-fixture screenshots during the manual pass above (or a maintainer can grab them). The UI is a standard `Dialog` wizard matching the existing settings dialogs.

## Known limitations

- **macOS only** (root discovery is macOS-specific; the engine itself is platform-agnostic and fully tested).
- A Cowork metadata file whose session directory is missing is marked **failed** (not auto-repaired) — a deliberate, conservative choice so the user sees it rather than silently skipping.
- Restore relies on the ledger; if the user deletes the backup dir, surgical undo is unavailable (target data is still never overwritten, so nothing is lost).

## Follow-up PRs

- Windows 3P root discovery (engine is already platform-agnostic).
- Optional: surface migrated-session management UI (closer to what #3061 literally asks for).
- Optional: Memory official export/import once a safe path is designed.


farion1231:
@codex review

PR #5986, opened 2026-08-01
feat(codex): support per-model reasoning effort levels in the model catalog

Chixuer:
﻿## What changed

Adds an optional per-model `supportedReasoningLevels` field to the Codex model catalog mapping. When a row declares it, catalog generation replaces the template's neutral `[none, high]` `supported_reasoning_levels` with the declared levels, so the Codex desktop UI only offers effort options the upstream actually validates.

- `src/types.ts`: new `supportedReasoningLevels?: string[]` on `CodexCatalogModel`
- `src/components/providers/forms/CodexFormFields.tsx`: the hidden field survives load→save round-trips
- `src-tauri/src/codex_config.rs`: parse the field, override `supported_reasoning_levels` (descriptions match Codex's `models_cache.json` wording), keep `default_reasoning_level` consistent, and round-trip the field through the simplified reverse-parse

## Why

The generated Codex catalog currently clones a neutral template with only `none` and `high`, regardless of what the configured gateway actually supports. For third-party `/responses` gateways this hides real effort levels (e.g. `low`/`medium`/`xhigh`/`max`) and can even offer `ultra`, which some upstreams reject with `invalid_enum_value`. With this change, providers can declare exactly which levels each model accepts.

## Impact

- Users of third-party Codex providers can see and select the effort levels their gateway supports.
- Providers whose models only support `[none, high]` are unaffected: the field is optional and the template default is kept when absent.
- DB/live round-trips preserve the field; opening and saving a provider in the UI will not drop it.

## Validation

- `tsc --noEmit` passes
- `vite build` passes
- `cargo check` passes
- New unit test `per_model_supported_reasoning_levels_override_template_and_round_trip` passes (generation override + reverse-parse round-trip)

### DIFF
--- a/src-tauri/src/codex_config.rs
+++ b/src-tauri/src/codex_config.rs
@@ -573,6 +573,42 @@ fn codex_catalog_model_entry(
     entry_obj.insert("availability_nux".to_string(), Value::Null);
     entry_obj.insert("upgrade".to_string(), Value::Null);
 
+    // Per-model reasoning effort override. The template only carries a neutral
+    // [none, high] default; when the model mapping declares the levels the
+    // upstream actually validates, replace the list so the Codex UI only
+    // offers those levels.
+    if let Some(levels) = &spec.supported_reasoning_levels {
+        let levels_json: Vec<Value> = levels
+            .iter()
+            .map(|effort| {
+                json!({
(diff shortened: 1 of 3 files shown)


farion1231:
@codex review

PR #5991, opened 2026-08-01
fix(codex): correct DeepSeek catalog reasoning effort to xhigh

NorthPointer:
## What changed

`src-tauri/src/resources/codex_deepseek_catalog_template.json` — the bundled DeepSeek model catalog used for native Responses providers — declared the top reasoning effort as `"effort": "max"` for both `deepseek-v4-flash` and `deepseek-v4-pro`. Both entries now use `"effort": "xhigh"` instead.

Supporting updates:

- `src-tauri/src/codex_config.rs`: the mirror test now asserts `["low", "high", "xhigh"]`; comments note the correction.
- `docs/guides/codex-deepseek-routing-guide-{en,zh,ja}.md` and `src/config/codexProviderPresets.ts`: reference `low/high/xhigh` instead of `low/high/max`.

## Why

When a provider uses DeepSeek's native Responses API, CC Switch writes `model_catalog_json = "cc-switch-model-catalog.json"` into the Codex config and generates the catalog from this template. Codex's model catalog parser only accepts `xhigh` as the top reasoning effort — `max` is not a valid value — so the "Maximum reasoning depth for the hardest problems" level was silently missing from the reasoning-level selector. Manually changing the value to `xhigh` makes the option appear and work.

The template mirrored DeepSeek's official `models.json` verbatim (introduced in 8ae1ce8), and that upstream file declares `max`; the bundled copy now carries the one correction Codex needs.

## Validation

- Template JSON parses; both models expose `["low", "high", "xhigh"]`.
- The mirror test assertion was updated to match; no Rust toolchain was available locally to run `cargo test`, so CI verifies it.
- The diff is limited to the two template values plus directly related comments/docs.


farion1231:
@codex review

PR #5997, opened 2026-08-01
fix(usage): prevent Codex session sync gaps

puppnn:
## Summary

- persist the processed Codex rollout file size alongside mtime and line offset, so appended usage is synced even when Windows keeps the open file's mtime unchanged
- allow a child fork to sync after the parent has emitted an explicit terminal lifecycle event, while still deferring children of active or incomplete parents
- add the v16 to v17 schema migration and regression coverage for both cases

## Root cause

Codex session sync skipped a rollout whenever its mtime had not advanced. On Windows, an actively appended JSONL file can grow while retaining the same observed mtime, leaving later usage events permanently absent from CC Switch.

Fork replay alignment also required the parent's maximum timestamp to reach the exact child fork timestamp. A normal gap after `task_complete` therefore kept a valid child rollout deferred forever. One observed reproduction had a roughly 21-second gap between the parent's final completion event and the child fork.

## Behavior and safety

- unchanged: token delta calculation, replay-prefix deduplication, request IDs, pricing, and proxy request ingestion
- file size is an additional change cursor; same-mtime files are skipped only when their size is also unchanged
- only explicit `task_complete`, `turn_complete`, or `turn_aborted` lifecycle events make a parent timeline terminal
- active parents still remain deferred until enough history is available
- existing sync rows migrate with `last_file_size = 0`, causing one safe rescan before the new cursor is established

## Validation

- `cargo fmt -- --check`
- `cargo clippy --locked --lib`
- `cargo test --locked session_usage_codex --lib` (36 passed)
- `cargo test --locked migrate_v16_to_v17_adds_session_file_size_cursor --lib` (1 passed)
- `cargo test --locked --lib` (2242 passed, 2 ignored)

## Related issue

N/A; this is a focused fix for locally reproduced Windows session-sync data loss.

### DIFF
--- a/src-tauri/src/database/schema.rs
+++ b/src-tauri/src/database/schema.rs
@@ -302,6 +302,7 @@ impl Database {
                 file_path TEXT PRIMARY KEY,
                 last_modified INTEGER NOT NULL,
                 last_line_offset INTEGER NOT NULL DEFAULT 0,
+                last_file_size INTEGER NOT NULL DEFAULT 0,
                 last_synced_at INTEGER NOT NULL
             )",
             [],
@@ -510,6 +511,11 @@ impl Database {
                         log::info!("迁移数据库从 v15 到 v16（重建 Codex 会话用量）");
                         Self::migrate_v15_to_v16(conn)?;
                         Self::set_user_version(conn, 16)?;
+                    }
+                    16 => {
+                        log::info!("迁移数据库从 v16 到 v17（记录会话日志文件大小）");
+                        Self::migrate_v16_to_v17(conn)?;
+                        Self::set_user_version(conn, 17)?;
                     }
                     _ => {
                         return Err(AppError::Database(format!(
@@ -1521,6 +1527,18 @@ impl Database {
     fn migrate_v15_to_v16(conn: &Connection) -> Result<(), AppError> {
         let codex_dir = crate::codex_config::get_codex_config_dir();
         crate::services::session_usage_codex::reset_codex_usage_on_conn(conn, &codex_dir)
+    }
+
+    fn migrate_v16_to_v17(conn: &Connection) -> Result<(), AppError> {
+        if Self::table_exists(conn, "session_log_sync")? {
+            Self::add_column_if_missing(
+                conn,
+                "session_log_sync",
+                "last_file_size",
+                "INTEGER NOT NULL DEFAULT 0",
+            )?;
+        }
+        Ok(())
     }
 
     /// 插入默认模型定价数据
@@ -3222,7 +3240,12 @@ mod tests {
 
         Database::apply_schema_migrations_on_conn(&conn)?;
 
-        assert_eq!(Database::get_user_version(&conn)?, 16);
+        assert_eq!(Database::get_user_version(&conn)?, SCHEMA_VERSION);
+        assert!(Database::has_column(
+            &conn,
+            "session_log_sync",
+            "last_file_size"
+        )?);
         let counts: (i64, i64, i64, i64) = conn.query_row(
             "SELECT
                 (SELECT COUNT(*) FROM proxy_request_logs WHERE data_source = 'codex_session'),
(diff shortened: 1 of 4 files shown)


src-tauri/src/services/session_usage.rs
```diff
@@ -440,6 +440,28 @@ pub(crate) fn get_sync_state(db: &Database, file_path: &str) -> Result<(i64, i64
...
+        |row| {
+            let size = row.get::<_, i64>(2)?;
+            Ok((
+                row.get::<_, i64>(0)?,
+                row.get::<_, i64>(1)?,
+                size.max(0) as u64,
+            ))
+        },
+    );
+    Ok(result.unwrap_or((0, 0, 0)))
```
allenxu09: [bug · medium] This treats every SQLite read failure as a missing sync state. Errors such as a schema mismatch, corrupt value, or database failure will silently reset the cursor to zero, potentially triggering a full replay and then overwriting the existing state. Only `QueryReturnedNoRows` should map to `(0, 0, 0)`; propagate all other errors as `AppError::Database` (or use `OptionalExtension::optional()` and transpose the result).

```suggestion
    match result {
        Ok(state) => Ok(state),
        Err(rusqlite::Error::QueryReturnedNoRows) => Ok((0, 0, 0)),
        Err(e) => Err(AppError::Database(format!("读取同步状态失败: {e}"))),
    }
```

src-tauri/src/database/schema.rs
```diff
@@ -1523,6 +1529,18 @@ impl Database {
         crate::services::session_usage_codex::reset_codex_usage_on_conn(conn, &codex_dir)
     }
 
+    fn migrate_v16_to_v17(conn: &Connection) -> Result<(), AppError> {
+        if Self::table_exists(conn, "session_log_sync")? {
+            Self::add_column_if_missing(
+                conn,
+                "session_log_sync",
```
allenxu09, line 1536: [bug · medium] Existing cursors are migrated with `last_file_size = 0`, but the Codex sync logic treats that as a real observed size. If a previously processed rollout is empty/truncated when the first v17 sync runs and its mtime is unchanged or older, `(file_modified <= last_modified && file_size == last_file_size)` skips it and preserves the old nonzero line offset. A later append can then skip new records up to that stale offset, creating the gap this change is intended to prevent. Represent the migrated size as unknown, or invalidate existing cursor mtimes/offsets so the first v17 pass must establish a consistent size/offset pair; add a migration-to-first-sync test for this case.

For example, reset the existing cursor after adding the column:

```rust
conn.execute(
    "UPDATE session_log_sync SET last_modified = 0, last_line_offset = 0",
    [],
)
.map_err(|e| AppError::Database(format!("重置会话日志同步游标失败: {e}")))?;
```

allenxu09 (commented):
I found two correctness issues in the new sync cursor handling.

src-tauri/src/services/session_usage.rs
puppnn: Addressed in the follow-up commit. `get_sync_state` and `get_sync_state_with_size` now map only `QueryReturnedNoRows` to an empty cursor; all other rusqlite errors are propagated as `AppError::Database`. I also added `sync_state_lookup_defaults_only_when_row_is_missing` to verify both paths fail closed on database errors. CI is green on the current head.

src-tauri/src/database/schema.rs
puppnn: Addressed in the follow-up commit. The v16 -> v17 migration now calls `reset_codex_sync_cursors_on_conn`, which resets `last_modified`, `last_line_offset`, and `last_file_size` for Codex rollout cursors before the first v17 sync; non-Codex cursors remain unchanged. The migration regression test verifies the reset and preserves non-Codex state.

farion1231:
@codex review

puppnn:
Superseded by #6246, which rebases the fix onto current main and integrates it with the new CodexSyncPass cursor preload and batched transaction path. The original correctness fixes and regression coverage are preserved there.

PR #6011, opened 2026-08-02
fix(opencode): support unified OMO config

### LINKED ISSUE #5945: OpenCode OMO 配置管理未适配 ~/.omo/omo.jsonc 统一配置
### Self Checks / 自检

### CC Switch Version / 版本号

3.19.0

### Operating System / 操作系统

macOS

### Related App / 涉及应用

OpenCode

## 问题

CC Switch 的 OpenCode OMO 配置管理（`src-tauri/src/services/omo.rs`）仅识别以下文件：

```
oh-my-openagent.jsonc
oh-my-openagent.json
oh-my-opencode.jsonc
oh-my-opencode.json
```

这组候选文件位于 `get_opencode_dir()` 即 `~/.config/opencode/` 下。代码中**没有任何对 `~/.omo/omo.jsonc` 的引用**。

但 OMO v4.19.3（2026-07-28）的 breaking change 明确声明：

> *"the legacy `~/.omo/config.jsonc` loader is gone — `omo.jsonc` is now the only config surface"*

升级后，OMO 将 OpenCode 的旧 `~/.config/opencode/oh-my-openagent.jsonc` 内容迁移到 `~/.omo/omo.jsonc` 的 `[opencode]` 分区下，此后 **OpenCode OMO 插件从该文件读取配置，不再读取旧路径**。

## 现状

| 文件                                       | CC Switch 是否感知 | OMO v4.19.3 是否读取 |
| ------------------------------------------ | ------------------ | -------------------- |
| `~/.config/opencode/oh-my-openagent.jsonc` | ✅ 读写             | ❌ 不再读取           |
| `~/.omo/omo.jsonc`（`[opencode]` 分区）    | ❌ 完全不知         | ✅ 唯一数据源         |

**结果**：用户在 CC Switch 中修改 OpenCode 的 agent/category 模型分配后，配置被写入 `~/.config/opencode/oh-my-openagent.jsonc`，但 OMO v4.19.3 忽略该文件，实际生效的始终是 `~/.omo/omo.jsonc` 中迁移时留下的旧内容。用户在 GUI 中的配置修改完全无效。

## 建议修复

1. **路径**：识别并读写 `~/.omo/omo.jsonc`（当 `[opencode]` 分区存在时）
2. **格式**：`build_config()` 生成的 key 可能需要包在 `[opencode]` 分区下（如 `{ "[opencode]": { "agents": {...}, "categories": {...} } }`）
3. **向后兼容**：如果 `~/.omo/omo.jsonc` 不存在或没有 `[opencode]` 分区，回退到现有的 `oh-my-openagent.jsonc` 路径

## 环境

- CC Switch: v3.19.0
- oh-my-openagent: v4.19.3（最新）
- OS: macOS

### Expected Behavior / 期望行为

CC Switch 需要适配 OMO v4.19.3 的配置变更，确保修改 OpenCode 的 agent/category 模型映射时能正确写入 OMO 当前读取的配置文件和格式。

### Actual Behavior / 实际行为

_No response_

### Additional Context / 补充信息

_No response_

allenxu09:
## Summary

- read and write standard OMO profiles from `~/.omo/omo.jsonc` when it contains an `[opencode]` block
- preserve unrelated shared settings, harness blocks, comments, and formatting while switching or disabling OMO profiles
- retain the legacy OpenCode config fallback when the unified block is absent, without changing OMO Slim behavior

## Root cause

CC Switch only searched the legacy `~/.config/opencode/oh-my-openagent.json[c]` files. Current OMO versions use the `[opencode]` block in `~/.omo/omo.jsonc`, so changes made in CC Switch were written to files OMO no longer reads.

Closes #5945

### DIFF
--- a/src-tauri/src/config.rs
+++ b/src-tauri/src/config.rs
@@ -270,8 +270,11 @@ fn sort_json_keys(value: &Value) -> Value {
     }
 }
 
-/// 写入 JSON 配置文件（键按字母排序，确保确定性输出）
-pub fn write_json_file<T: Serialize>(path: &Path, data: &T) -> Result<(), AppError> {
+/// 写入 JSON 配置文件并返回实际写入的字节。
+pub fn write_json_file_with_contents<T: Serialize>(
+    path: &Path,
+    data: &T,
+) -> Result<Vec<u8>, AppError> {
     // 确保目录存在
     if let Some(parent) = path.parent() {
         fs::create_dir_all(parent).map_err(|e| AppError::io(parent, e))?;
@@ -282,7 +285,14 @@ pub fn write_json_file<T: Serialize>(path: &Path, data: &T) -> Result<(), AppErr
     let json = serde_json::to_string_pretty(&sorted_value)
         .map_err(|e| AppError::JsonSerialize { source: e })?;
 
-    atomic_write(path, json.as_bytes())
+    let contents = json.into_bytes();
+    atomic_write(path, &contents)?;
+    Ok(contents)
+}
+
+/// 写入 JSON 配置文件（键按字母排序，确保确定性输出）
+pub fn write_json_file<T: Serialize>(path: &Path, data: &T) -> Result<(), AppError> {
+    write_json_file_with_contents(path, data).map(|_| ())
 }
 
 /// 原子写入文本文件（用于 TOML/纯文本）
@@ -302,7 +312,6 @@ pub fn atomic_write(path: &Path, data: &[u8]) -> Result<(), AppError> {
     let parent = path
         .parent()
         .ok_or_else(|| AppError::Config("无效的路径".to_string()))?;
-    let mut tmp = parent.to_path_buf();
     let file_name = path
         .file_name()
         .ok_or_else(|| AppError::Config("无效的文件名".to_string()))?
@@ -312,13 +321,38 @@ pub fn atomic_write(path: &Path, data: &[u8]) -> Result<(), AppError> {
         .duration_since(std::time::UNIX_EPOCH)
         .unwrap_or_default()
         .as_nanos();
-    tmp.push(format!("{file_name}.tmp.{ts}"));
+    static TEMP_COUNTER: std::sync::atomic::AtomicU64 = std::sync::atomic::AtomicU64::new(0);
+    let (tmp, mut file) = (|| -> Result<(PathBuf, fs::File), AppError> {
+        let mut last_collision = None;
+        for _ in 0..16 {
+            let counter = TEMP_COUNTER.fetch_add(1, std::sync::atomic::Ordering::Relaxed);
+            let candidate = parent.join(format!(
+                "{file_name}.tmp.{}.{ts}.{counter}",
+                std::process::id()
+            ));
+            match fs::OpenOptions::new()
+                .write(true)
+                .create_new(true)
+                .open(&candidate)
+            {
+                Ok(file) => return Ok((candidate, file)),
+                Err(source) if source.kind() == std::io::ErrorKind::AlreadyExists => {
+                    last_collision = Some((candidate, source));
+                }
+                Err(source) => return Err(AppError::io(&candidate, source)),
+            }
+        }
 
-    {
-        let mut f = fs::File::create(&tmp).map_err(|e| AppError::io(&tmp, e))?;
-        f.write_all(data).map_err(|e| AppError::io(&tmp, e))?;
-        f.flush().map_err(|e| AppError::io(&tmp, e))?;
+        let (candidate, source) = last_collision.expect("temporary filename loop must run");
+        Err(AppError::io(&candidate, source))
+    })()?;
+
+    if let Err(source) = file.write_all(data).and_then(|_| file.flush()) {
+        drop(file);
+        let _ = fs::remove_file(&tmp);
+        return Err(AppError::io(&tmp, source));
     }
+    drop(file);
 
     #[cfg(unix)]
     {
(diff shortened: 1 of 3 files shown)


farion1231:
@codex review

farion1231 (commented):
Thanks — the direction here is right, and I verified the core premise against upstream v4.19.4 (source + running the released CLI against scratch homes): since 4.19.3 the OpenCode plugin only reads the unified config chain, so targeting `~/.omo/omo.jsonc` is correct. The round-trip whitespace handling in `remove_opencode_section` also held up well under edge-case testing (first/middle/last/only-pair removal, trailing commas, CRLF), and the full test suite is green locally.

That said, I found two P1s that need fixing before this can merge, plus two smaller items.

**[P1] `json-five` 0.3.1 corrupts block comments on serialize — writes invalid JSONC**

Every `/* ... */` in the document comes back from `text.to_string()` ([`save()`, omo.rs#L156-L167](https://github.com/farion1231/cc-switch/blob/fb98403200fc153ebf96b0d1a34aaa0d65aa0856/src-tauri/src/services/omo.rs#L156-L167)) with the trailing `/` dropped:

```rust
let text = json_five::rt::parser::from_str("{ /* keep */ \"a\": 1 }").unwrap();
assert_eq!(text.to_string(), "{ /* keep * \"a\": 1 }"); // invalid JSONC
```

Root cause is in the crate's tokenizer: the block-comment token's end is recorded as the index of the final `/`, and the exclusive slice drops it (line comments are unaffected because they end at the newline, which belongs to the next whitespace token). I reproduced this with 7 comment shapes — all corrupted. So if the user's `omo.jsonc` contains any block comment, switching or disabling OMO successfully writes a file OMO can no longer parse at all — taking `[codex]` and every other section down with it — and cc-switch can't read it back either. 0.3.1 is the latest release, so there's no crate upgrade to reach for.

Minimal fix: in `save()`, re-parse the serialized output (`rt_from_str` + `json5::from_str`) and refuse to write if either fails. That converts silent corruption into a loud error. A real fix means patching/vendoring the crate — fine as a follow-up. Note `openclaw_config.rs` has the same `to_string()` → `atomic_write` sink, so this bug already affects shipped OpenClaw writes; I'll open a separate issue for that rather than growing this PR.

**[P1] disable → re-enable round trip writes to a file OMO never reads**

`delete_config_file` removes the `[opencode]` section but (correctly) leaves the root `_migrations` marker in place. On the next enable, [`find_unified_config_path`](https://github.com/farion1231/cc-switch/blob/fb98403200fc153ebf96b0d1a34aaa0d65aa0856/src-tauri/src/services/omo.rs#L302-L321) sees the file exists but the section is missing, returns `None`, and the write falls back to the legacy path. But upstream ≥ 4.19.3 never reads the legacy path at runtime, and the startup migration is gated on the `_migrations` marker ([predicate.ts#L6-L12](https://github.com/code-yeongyu/oh-my-openagent/blob/v4.19.4/packages/omo-config-core/src/migration/predicate.ts#L6-L12)) — already present, so it never re-runs. Net effect: after one disable, every subsequent profile switch writes config that OMO silently ignores — the exact #5945 symptom this PR fixes.

Minimal fix: treat "unified file exists" as unified mode, and make [`set_opencode_section`](https://github.com/farion1231/cc-switch/blob/fb98403200fc153ebf96b0d1a34aaa0d65aa0856/src-tauri/src/services/omo.rs#L85-L107) insert the pair when it's missing instead of erroring. `"[opencode]"` is a declared key in the upstream schema, so inserting it is safe — but be aware the root schema is `.strict()` ([config.ts#L37-L51](https://github.com/code-yeongyu/oh-my-openagent/blob/v4.19.4/packages/omo-config-core/src/schema/config.ts#L37-L51)): writing any *other* unknown root key makes OMO reject the entire file, so everything cc-switch writes must stay inside the `"[opencode]"` value.

**[P2] `omo.json` fallback not handled**

Upstream resolves `~/.omo/omo.jsonc` first, then falls back to `~/.omo/omo.json` ([paths.ts#L35-L40](https://github.com/code-yeongyu/oh-my-openagent/blob/v4.19.4/packages/omo-config-core/src/loader/paths.ts#L35-L40)). `find_unified_config_path` only probes `omo.jsonc`, so a valid `omo.json` setup still gets legacy writes. Mirror the two-step probe.

**[P2] `expected_contents` captured by re-reading after write**

[opencode_config.rs#L312](https://github.com/farion1231/cc-switch/blob/fb98403200fc153ebf96b0d1a34aaa0d65aa0856/src-tauri/src/opencode_config.rs#L312) and the legacy branch at [omo.rs#L477-L478](https://github.com/farion1231/cc-switch/blob/fb98403200fc153ebf96b0d1a34aaa0d65aa0856/src-tauri/src/services/omo.rs#L477-L478) re-read the file after writing to capture `expected_contents`. An external write landing in that window gets mistaken for ours, and a later rollback would overwrite it. The unified branch already does this right — `save()` returns the exact bytes it wrote. Make the other two write paths return their serialized bytes instead of re-reading.

nit: an end-to-end unified test through `config_location` (via `CC_SWITCH_TEST_HOME`) writing and reading back a commented document would have caught the round-trip gap — worth adding.

With the two P1s fixed (ideally the P2s too), I'd like to land this — the overall shape is right and the concurrency guards are a real improvement over what's there today. Happy to re-review quickly once it's updated.

farion1231 (commented):
Re-checked f414093d — this round resolves everything I flagged on the previous head: disable → re-enable now stays on the unified config (`find_unified_config_path` keys off file existence and `set_opencode_section` can insert a missing section, with a regression test covering the exact #5945 revival chain), the `omo.json` fallback matches upstream's lookup order, rollback snapshots are coupled to the actual written bytes, and `save()` now refuses output that no longer parses. Implementation quality is high; the full backend suite passes on a merge with current `main` on my machine, and CI is green on all three platforms.

One data-loss hole remains in the new validation, so this still needs a small fix before merge:

**[P1] `save()` validates parseability only, not semantic equality — corrupted output can still pass both parsers and silently destroy other sections.**

[`omo.rs#L217-L229`](https://github.com/farion1231/cc-switch/blob/f414093d5dff34f2e589e02d3324fd59c2e5c15c/src-tauri/src/services/omo.rs#L217-L229) reparses `next_source` with both parsers and discards the results. json-five 0.3.1 still truncates every block comment to `/* … *` (trailing `/` lost). When the runaway comment gets re-closed by a later `*/` — e.g. a line comment ending with `*/` — everything in between is swallowed into the comment and the result is still grammatically valid JSON5. Verified end-to-end on this head with:

```jsonc
{
  /* block */
  "[codex]": {"agents": {"reviewer": {}}},
  // tail */
  "[opencode]": {"agents": {}}
}
```

After `set_opencode_section` + `save()`, `save()` returns `Ok` and the file on disk reads:

```jsonc
{
  /* block *
  "[codex]": {"agents": {"reviewer": {}}},
  // tail */
  "[opencode]": {"agents": {"new": {}}}
}
```

The `[codex]` section now sits inside the comment — silently gone for OMO and for cc-switch's own reader.

Minimal fix: you already keep `self.semantic` up to date in `set_opencode_section` / `remove_opencode_section`, so compare the reparse against it instead of discarding it:

```rust
let reparsed: Value = json5::from_str(&next_source)
    .map_err(|e| AppError::Config(format!(/* … */)))?;
if reparsed != self.semantic {
    return Err(AppError::Config(
        "Refusing to write OMO config: serialized output does not match the intended state".into(),
    ));
}
```

I checked the obvious false-positive risk: the json5 crate parses integers into the same `serde_json::Value` representation as `json!()` literals (`json5::from_str("{a:1}") == json!({"a":1})`), so the comparison won't reject legitimate writes over number representation. Please add a regression test for the scenario above.

**[P3] Duplicate keys inside `[opencode]` make the merge keep the stale shadowed value.**

JSON5 duplicate keys are last-wins, but the equality short-circuit in `merge_rt_value` ([`omo.rs#L522-L524`](https://github.com/farion1231/cc-switch/blob/f414093d5dff34f2e589e02d3324fd59c2e5c15c/src-tauri/src/services/omo.rs#L522-L524)) compares the semantic (last-occurrence) value against the desired one; when equal it skips rewriting the **first** rt occurrence — which holds the old shadowed value — and the later duplicate holding the winning value is then dropped as a dup ([`omo.rs#L551-L554`](https://github.com/farion1231/cc-switch/blob/f414093d5dff34f2e589e02d3324fd59c2e5c15c/src-tauri/src/services/omo.rs#L551-L554)). Verified: `{"model": "loser", "model": "winner"}` + desired `{"model": "winner", "extra": true}` writes `model: "loser"`. Trigger conditions are narrow (hand-edited duplicates whose shadowed value differs), it self-heals on the next apply, and the P1 fix above turns it from a silent wrong write into an explicit refusal — so I'm fine with this as a follow-up. A direct fix would be a dedup pre-pass that keeps the last occurrence (the semantic winner) before merging.

Side note, non-blocking: the `atomic_write` rework goes beyond this PR's scope, but I verified it's a genuine improvement — the old Windows remove-then-rename had a window where a failure after the remove lost the destination file, and `ReplaceFileW` closes it; the `windows-sys` feature is already declared and the new Windows-only test runs on CI's windows-latest job.

Fix the P1 (with its regression test) and this is good to merge from my side — happy to do a quick re-review.

farion1231 (approved):

PR #6029, opened 2026-08-02
fix(hermes): persist default api_mode for UI-created custom providers

### LINKED ISSUE #4011: Bug: Missing api_mode field when creating custom Hermes provider via UI
### Environment
- **cc-switch version**: Latest (cloned 2026-06-10)
- **OS**: Linux (Ubuntu 24.04)
- **App**: Hermes Agent

### Description
When creating a custom Hermes provider through the UI without using a preset template, the `api_mode` field is missing from the saved configuration, causing connection tests to fail with error: **"Hermes provider is missing the `api_mode` field"**.

### Steps to Reproduce
1. Open cc-switch and navigate to Hermes app
2. Click "Add Provider" → "Custom Provider" (not using any preset)
3. Fill in:
   - Name: `selfuse`
   - Base URL: `https://example.com/v1`
   - API Key: `sk-xxx`
   - Models: Add at least one model
   - **Do NOT touch the "API Mode" dropdown** (leave it at the default "chat_completions" shown in UI)
4. Click Save
5. Click "Test" on the newly created provider

### Expected Behavior
- The `api_mode` field should default to `"chat_completions"` in the saved configuration
- Connection test should succeed

### Actual Behavior
- The `api_mode` field is **missing** from `~/.cc-switch/cc-switch.db` → `providers.settings_config`
- Connection test fails with: "Hermes provider is missing the `api_mode` field"

#### Frontend Issue
`src/components/providers/forms/hooks/useHermesFormState.ts` lines 104-112 initialize the state with default value but don't write it to JSON:

```typescript
const [hermesApiMode, setHermesApiMode] = useState<HermesApiMode>(() => {
  const stored = parseHermesField(initialData, "api_mode", "");
  return stored || HERMES_DEFAULT_API_MODE;  // UI state only
});
```

The field is only written when user manually changes the dropdown (lines 159-167).

#### Backend Issue
`src-tauri/src/hermes_config.rs` function `set_provider` (lines 665-734) does not validate required fields or inject defaults.

### Why Preset Templates Work
All presets in `src/config/hermesProviderPresets.ts` explicitly include `api_mode`, so providers from presets work correctly.

### Proposed Fix

**Backend (Recommended)**: Add validation in `hermes_config.rs` after line 682:
```rust
if let Some(obj) = normalized.as_object_mut() {
    if !obj.contains_key("api_mode") {
        obj.insert(
            "api_mode".to_string(),
            serde_json::Value::String("chat_completions".to_string()),
        );
    }
}
```

### Workaround
Always manually click the API Mode dropdown and select the value before saving (even if it's already showing the default).

Dicarbene:
## Summary

Fixes #4011: when creating a custom Hermes provider via the UI, if the user does not manually touch the API Mode dropdown, the `api_mode` field is missing from the saved config, causing connection tests to fail.

## Changes

1. **Frontend** — `src/components/providers/forms/hooks/useHermesFormState.ts`
   - Added `api_mode: "chat_completions"` to `HERMES_DEFAULT_CONFIG_OBJ`

2. **Backend** — `src-tauri/src/hermes_config.rs`
   - In `set_provider`, after `sanitize_hermes_provider_keys`, inject `api_mode = "chat_completions"` as a fallback when the key is missing

## Test plan

- Create a custom Hermes provider via UI without touching API Mode → saved config contains `api_mode: chat_completions`
- Connection test succeeds
- Existing Hermes providers with explicit `api_mode` are unaffected

### DIFF
--- a/src-tauri/src/hermes_config.rs
+++ b/src-tauri/src/hermes_config.rs
@@ -808,6 +808,16 @@ pub fn set_provider(
     let mut normalized = provider_config;
     sanitize_hermes_provider_keys(&mut normalized);
 
+    // Inject default api_mode when missing (e.g. UI custom provider w/o explicit selection).
+    if let Some(obj) = normalized.as_object_mut() {
+        if !obj.contains_key("api_mode") {
+            obj.insert(
+                "api_mode".to_string(),
+                serde_json::Value::String("chat_completions".to_string()),
+            );
+        }
+    }
+
     // Normalize `models` from UI array to Hermes YAML dict before serializing.
     normalize_provider_models_for_write(&mut normalized);
 
--- a/src/components/providers/forms/hooks/useHermesFormState.ts
+++ b/src/components/providers/forms/hooks/useHermesFormState.ts
@@ -22,6 +22,7 @@ const HERMES_DEFAULT_CONFIG_OBJ = {
   name: "",
   base_url: "",
   api_key: "",
+  api_mode: "chat_completions",
 } as const;
 
 export const HERMES_DEFAULT_CONFIG = JSON.stringify(


farion1231:
@codex review

Issue #3289, opened 2026-05-29
Codex OAuth as Claude provider can write direct settings and trigger /login


Fortytwoo:
### Self Checks / ??

- [x] I have searched existing issues, including closed ones.
- [x] I found #1997, but this is still reproducible on 3.15.0 when the local proxy process is running while Claude takeover is not enabled.

### CC Switch Version / ???

3.15.0

### Operating System / ????

Windows 11

### Related App / ????

Claude Code + Codex OAuth provider

### Steps to Reproduce / ????

1. In CC Switch, complete Codex OAuth authorization.
2. Add/select the `Codex` provider under Claude Code.
3. Keep the CC Switch local proxy process running, but do not enable Claude live takeover / ensure `~/.claude/settings.json` is not pointing at `http://127.0.0.1:15721`.
4. Launch Claude Code.

### Expected Behavior / ????

For managed-account Claude providers such as Codex OAuth and GitHub Copilot, CC Switch should not write a direct upstream Claude live config. It should either enable Claude local-proxy takeover automatically or block the switch with a clear error.

The resulting `~/.claude/settings.json` should point to the local proxy, for example:

```json
{
  "env": {
    "ANTHROPIC_BASE_URL": "http://127.0.0.1:15721",
    "ANTHROPIC_API_KEY": "PROXY_MANAGED"
  }
}
```

### Actual Behavior / ????

CC Switch can write a direct Codex backend URL to Claude Code without any Claude-visible auth placeholder:

```json
{
  "env": {
    "ANTHROPIC_BASE_URL": "https://chatgpt.com/backend-api/codex",
    "ANTHROPIC_MODEL": "gpt-5.5"
  }
}
```

Claude Code then fails before requests are routed through CC Switch and shows:

```text
Not logged in ? Please run /login
```

### Additional Context / ????

This looks related to #1997, but that issue was closed with a manual workaround: enable local proxy + Claude takeover. The bug here is that switching to a managed-account provider still permits writing a direct live config path that cannot work for Claude Code.

A local manual fix was to set Claude Code to the CC Switch local proxy and add a placeholder auth value; after that `claude -p "??? OK"` returned `OK` and CC Switch logs showed forwarding to `https://chatgpt.com/backend-api/codex/responses`.

JavaBoy6:
I reproduced this on CC Switch v3.19.1 on Windows 11, specifically through the provider card's **Open Terminal** action.

### Additional reproduction evidence

[...]

PR #6039, opened 2026-08-02
fix(claude): guard routed provider terminals

grantyxy:
## Summary / 概述

- Prevent provider-specific **Open Terminal** from writing standalone Claude settings for providers that require local routing.
- Enforce the restriction in both the frontend and the Tauri backend while preserving official and direct Anthropic providers.
- Cover managed OAuth providers, transformed API formats, full-URL mode, and legacy managed-provider/format settings; show actionable guidance in all maintained locales.

## Related Issue / 关联 Issue

Addresses the provider-specific **Open Terminal** reproduction in #3289.

This intentionally does not add a provider-scoped proxy route or change provider-switching/takeover semantics. The original non-takeover switching path remains tracked in #3289, so this PR does not auto-close the issue.

## Screenshots / 截图

Not included. The visible change is a localized error toast shown before directory selection, and the behavior is covered by regression tests.

## Testing / 测试

- `pnpm typecheck`
- `pnpm format:check`
- `pnpm test:unit` — 86 files, 590 tests passed
- `cargo fmt --check`
- `cargo test --no-fail-fast` — 2,364 passed, 2 ignored

## Checklist / 检查清单


farion1231:
@codex review

PR #6053, opened 2026-08-03
feat(pricing): seed Qwen3.8 Max built-in model pricing

mhy1227:
## Summary
- Add built-in pricing for `qwen3.8-max` only
- Prices: 2 / 6 / 0.20 / 0 USD per Mtok (input / output / cache read / cache write)
- Insert-if-absent only; no schema bump; other model rows untouched

## Test plan
- [ ] Launch app on existing DB: Usage pricing list includes `qwen3.8-max`
- [ ] Confirm hand-edited prices for other models remain unchanged


farion1231 (approved):

PR #6080, opened 2026-08-03
fix(codex): sync active rollouts when mtime is unchanged

### LINKED ISSUE #6060: Codex usage sync misses active Windows rollouts when mtime stays unchanged
### Self Checks / 自检

### CC Switch Version / 版本号

3.19.1

### Operating System / 操作系统

Windows

### Related App / 涉及应用

Codex

### Steps to Reproduce / 重现步骤

1. Start a long-running Codex Desktop session on Windows.
2. Let CC Switch perform its initial background import for the active rollout JSONL.
3. Continue using the same Codex session so additional `event_msg/token_count` lines are appended while Codex keeps the file handle open.
4. Observe that the JSONL file size and line count increase, but its filesystem modification timestamp can remain unchanged until the writer closes the file.
5. Wait for subsequent CC Switch background sync cycles and open Usage Statistics.

### Expected Behavior / 期望行为

CC Switch should detect that the active rollout grew and import the newly appended token usage, even when the filesystem modification timestamp does not advance.

### Actual Behavior / 实际行为

The Codex sync cursor remains at the previous line offset and new usage is not imported while the session is active.

The current implementation reads both metadata values, but only uses the timestamp for the early return:

```rust
let file_modified = metadata_modified_nanos(&metadata);
let file_size = metadata.len();

if file_modified <= last_modified {
    return Ok(CodexFileSyncResult::default());
}
```

On Windows, an open rollout can grow while `LastWriteTime` remains unchanged, so the file is incorrectly treated as unchanged. Closing Codex or manually invalidating the sync cursor can make the missing usage appear later.

### Additional Context / 补充信息

- The affected JSONL remained valid and contained normal `turn_context.model` and `event_msg/token_count` records.
- This is separate from missing-parent/deferred fork history.
- A local fix persists the last observed file size in `session_log_sync` and skips a file only when both mtime and size are unchanged.
- The fix was validated against a real active Windows rollout: the cursor advanced across consecutive background scans while mtime stayed fixed.
- A regression test simulates a growing rollout with a non-advancing mtime.
- No private paths, session IDs, provider details, or usage totals are included here.

LimiChan-2026:
﻿## Summary / 概述

Fix Codex session usage getting stuck for active rollout JSONL files on Windows.

- Persist the last observed file size alongside the Codex sync cursor.
- Treat a rollout as unchanged only when both its modification timestamp and file size are unchanged.
- Preserve the stored file size when archived rollouts inherit an existing cursor.
- Add a regression test that appends token usage while the stored mtime is deliberately newer than the file mtime.

### Root cause

`sync_single_codex_file` already read `metadata.len()`, but its early-return condition only compared mtime. On Windows, Codex can keep an active rollout handle open and append new JSONL lines without advancing `LastWriteTime`. CC Switch then repeatedly skipped the growing file and left the usage cursor stale.

### User impact

Active Codex usage now appears on subsequent background syncs instead of waiting for the session/file handle to close. Historical deduplication and parent/fork replay behavior are unchanged.

## Related Issue / 关联 Issue

Fixes #6060

## Screenshots / 截图

Not applicable; this is a backend session-import fix.

## Validation / 验证

- [x] `cargo fmt --all -- --check`
- [x] `cargo clippy --lib`
- [x] `cargo test services::session_usage_codex::tests --lib` — 34 passed
- [x] `pnpm typecheck`
- [x] `pnpm format:check`
- [x] Windows release build with `pnpm tauri build --no-bundle`
- [x] Real active-rollout verification across consecutive background scans while mtime remained unchanged

## Checklist / 检查清单

- [x] `cargo clippy` passes (Rust changed) / 通过 Clippy 检查
- [x] No user-facing text changed; i18n updates are not required / 未修改用户可见文本，无需更新国际化文件

This contribution was developed with AI assistance and reviewed and validated against the reported Windows behavior before submission.

---
> **Note / 说明**: Original PR #6062 was auto-closed by GitHub when this fork was briefly set to private, and GitHub does not allow reopening it. This PR is resubmitted from the exact same branch and commit (no code changes).

### DIFF
--- a/src-tauri/src/database/schema.rs
+++ b/src-tauri/src/database/schema.rs
@@ -302,11 +302,20 @@ impl Database {
                 file_path TEXT PRIMARY KEY,
                 last_modified INTEGER NOT NULL,
                 last_line_offset INTEGER NOT NULL DEFAULT 0,
+                last_file_size INTEGER NOT NULL DEFAULT 0,
                 last_synced_at INTEGER NOT NULL
             )",
             [],
         )
         .map_err(|e| AppError::Database(e.to_string()))?;
+        // Compatibility migration kept outside user_version so the patched build can
+        // still be rolled back to an official v3.19.1 binary if needed.
+        Self::add_column_if_missing(
+            conn,
+            "session_log_sync",
+            "last_file_size",
+            "INTEGER NOT NULL DEFAULT 0",
+        )?;
 
         // 19. Profiles 表（全应用共享的项目实体，payload 按 app 分槽快照
         //     供应商/MCP/Skills/Prompt；各应用分组的 current 标记在 settings 表）
@@ -1212,6 +1221,7 @@ impl Database {
                 file_path TEXT PRIMARY KEY,
                 last_modified INTEGER NOT NULL,
                 last_line_offset INTEGER NOT NULL DEFAULT 0,
+                last_file_size INTEGER NOT NULL DEFAULT 0,
                 last_synced_at INTEGER NOT NULL
             )",
             [],
--- a/src-tauri/src/services/session_usage.rs
+++ b/src-tauri/src/services/session_usage.rs
@@ -440,6 +440,29 @@ pub(crate) fn get_sync_state(db: &Database, file_path: &str) -> Result<(i64, i64
     Ok(result.unwrap_or((0, 0)))
 }
 
+/// Return sync progress together with the file size observed at the last sync.
+/// Codex needs the size because an open JSONL file can grow on Windows while its
+/// modification timestamp remains unchanged.
+pub(crate) fn get_sync_state_with_size(
+    db: &Database,
+    file_path: &str,
+) -> Result<(i64, i64, i64), AppError> {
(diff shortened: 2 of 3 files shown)


farion1231:
@codex review

woshimaxfive:
Nice catch — same root cause as #6023 (reported 08-02, Windows). Codex appends to an open rollout without advancing LastWriteTime, so the old cursor check (mtime only) skips the growing file; I've reproduced it three times here (cursor stuck at line 13/17/20).

I opened a similar fix at #6027 (same mtime+size double-check, plus handling for truncated/replaced files and partial last lines). Both PRs converged on the same approach, which is good validation. Leaving the call to the maintainers.

farion1231:
Thanks for digging into this one. The same root cause has now shipped in v3.20.3 via PR #7219 (persisted byte cursor + mtime/size gate), which was rebased onto the current schema; this PR conflicts with that change and is superseded, so I'm closing it. Your analysis helped confirm the fix direction — much appreciated.

PR #6083, opened 2026-08-03
feat(skills): allow inactive installation

### LINKED ISSUE #6082: feat(skills): allow install without enabling a consumer
## Problem

Repository Skill installation currently requires `currentApp`, creates `SkillApps::only(current_app)`, and enables the current consumer even when the goal is only to curate a managed inactive pool. Re-installing an existing same-source Skill also implicitly enables the current app.

That makes "install, then disable" the only UI path. It introduces a transient consumer projection and is unsuitable for reviewing third-party Skills before activation.

## Proposed focused change

- Keep the current install-and-enable behavior as the default.
- Add an explicit UI option to install without enabling the current app.
- Let the service accept an explicit initial `SkillApps` set.
- When that set is empty, persist the Skill in the manager SSOT/database and skip all consumer projection.
- For an existing same-source row, merge only explicitly requested apps; an empty set must not enable anything.

## Out of scope

This issue does **not** claim cohort-wide atomic acquisition, dependency/digest verification, transaction journaling, or crash recovery. Those require a separate design and failure matrix. This is the smallest upstream primitive needed to remove transient activation from single-Skill installation.

## Acceptance checks

- Existing callers that omit the new option still enable the current app.
- An inactive install stores all app flags as false.
- No Claude or Codex consumer directory is created for an inactive install.
- Re-installing the same source in inactive mode preserves its disabled state.
- Frontend typecheck/format/build and Rust format/clippy/build pass.

yiheng8023:
## Summary / 概述

- keep the existing install-and-enable flow as the default
- add an explicit "Enable for current app after install" option to the Skills discovery page
- let `SkillService` install with an explicit initial `SkillApps` set
- persist empty app sets without projecting into any consumer directory
- avoid implicitly enabling an existing same-source Skill when the requested app set is empty

This removes the transient consumer exposure caused by an install-then-disable workflow while preserving compatibility for existing callers.

## Related Issue / 关联 Issue

Fixes #6082

## Scope boundary

This focused PR provides the single-Skill inactive-install primitive only. It does not claim cohort-wide atomic acquisition, dependency/digest verification, transaction journaling, or crash recovery; those need a separate design and failure matrix.

## Screenshots / 截图

Not attached yet; the PR is opened as a draft. The new checkbox is covered by an interaction test.

## Checklist / 检查清单

- [x] Prettier format check passes
- [x] Rust format check passes
- [x] `cargo clippy --all-targets` passes (one unrelated pre-existing `items_after_test_module` warning remains)
- [x] Updated all four locale files for user-facing text

## Verification / 验证

- `vitest run tests/components/SkillsPageInstall.test.tsx`: 8 passed
- `vitest run tests/integration/App.test.tsx`: 4 passed in isolation
- full `vitest run`: the Skills tests pass; two existing order-sensitive failures occur in `tests/integration/App.test.tsx` (timeout and duplicate `provider-list`), while that module passes when isolated
- `vite build`: passed
- `cargo test install_with_empty_apps --lib`: 2 passed
- `cargo build`: passed
- full `cargo test --lib`: 2245 passed, 1 failed, 2 ignored; the sole failure is `services::model_pricing::tests::reloads_manual_file_edits_and_deletion_tombstones`, and the identical failure was reproduced in a detached worktree at unchanged upstream `eb356e15bd898a434fde7ca74e5e3a2aec6c90e4`

### DIFF
--- a/src-tauri/src/commands/skill.rs
+++ b/src-tauri/src/commands/skill.rs
@@ -4,7 +4,7 @@
 //! - 支持三应用开关（Claude/Codex/Gemini）
 //! - SSOT 存储在 ~/.cc-switch/skills/
 
-use crate::app_config::{AppType, InstalledSkill, UnmanagedSkill};
+use crate::app_config::{AppType, InstalledSkill, SkillApps, UnmanagedSkill};
 use crate::error::format_skill_error;
 use crate::services::skill::{
     DiscoverableSkill, ImportSkillSelection, MigrationResult, Skill, SkillBackupEntry, SkillRepo,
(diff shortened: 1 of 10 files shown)


yiheng8023:
Superseded by #6086, which rebases the inactive-install primitive onto the newer upstream base and extends it with the bounded atomic cohort transaction and recovery boundary. Keeping this earlier draft closed avoids duplicate review; #6082 remains the shared issue context.

PR #6087, opened 2026-08-03
fix(codex): respect user-owned model_catalog_json when generating catalog

misaka-myu:
## Problem

In the Codex provider configuration, `model_catalog_json` is always overwritten with `cc-switch-model-catalog.json` in the generated `config.toml`, regardless of whether the user specified a custom catalog filename or full path. This means any user-managed external catalog file is silently discarded on every config write.

## Root Cause

`set_codex_model_catalog_json_field` in `src-tauri/src/codex_config.rs` has two arms:

- **`None` arm**: already checks ownership — it only removes the pointer when it points at the cc-switch-owned file, preserving user-managed external catalogs.
- **`Some` arm**: unconditionally writes the cc-switch-owned filename, clobbering any user-provided value.

The `Some` arm lacks the ownership check that the `None` arm already implements, causing the inconsistency.

## Fix

Mirror the `None` arm's ownership rule in the `Some` arm: only claim the `model_catalog_json` pointer when it is **absent** or **already cc-switch-owned**. A user-managed external catalog file (custom filename or path) is left untouched.

This is consistent with the existing design documented in `resolve_cc_switch_catalog_path`, which explicitly distinguishes cc-switch-owned vs user-owned catalog pointers.

## Testing

- [x] `pnpm typecheck` passes
- [x] `pnpm test:unit` — all existing catalog tests pass (no regression)
- [x] Added two tests covering the `Some` arm with user-owned catalogs:
  - `set_catalog_json_some_preserves_user_owned_catalog` (full path)
  - `set_catalog_json_some_preserves_user_owned_relative_filename` (bare filename)
- [x] Verified manually: a custom `model_catalog_json` value is now preserved in `config.toml` after switching providers

```
running 8 tests
test codex_config::tests::set_catalog_json_none_removes_relative_path ... ok
test codex_config::tests::set_catalog_json_none_preserves_user_owned_catalog ... ok
test codex_config::tests::set_catalog_json_field_writes_filename_ignoring_unc_path ... ok
test codex_config::tests::set_catalog_json_none_removes_cc_switch_owned_by_filename ... ok
test codex_config::tests::model_catalog_json_field_writes_relative_filename ... ok
test codex_config::tests::set_catalog_json_field_writes_filename_for_any_path ... ok
test codex_config::tests::set_catalog_json_some_preserves_user_owned_catalog ... ok
test codex_config::tests::set_catalog_json_some_preserves_user_owned_relative_filename ... ok

test result: ok. 8 passed; 0 failed; 0 ignored; 0 measured
```

### DIFF
--- a/src-tauri/src/codex_config.rs
+++ b/src-tauri/src/codex_config.rs
@@ -1248,7 +1248,22 @@ fn set_codex_model_catalog_json_field(
 
     match catalog_path {
         Some(_) => {
-            doc["model_catalog_json"] = toml_edit::value(CC_SWITCH_CODEX_MODEL_CATALOG_FILENAME);
+            // Only claim the pointer when it is absent or already cc-switch-owned.
+            // A user-managed external catalog file (custom filename or path) is
+            // left untouched, mirroring the None arm's ownership rule that
+            // `resolve_cc_switch_catalog_path` relies on.
+            let is_cc_switch_owned = doc
+                .get("model_catalog_json")
+                .and_then(|item| item.as_str())
+                .map(|path| {
+                    Path::new(path).file_name().and_then(|name| name.to_str())
+                        == Some(CC_SWITCH_CODEX_MODEL_CATALOG_FILENAME)
+                })
+                .unwrap_or(true);
+            if is_cc_switch_owned {
+                doc["model_catalog_json"] =
+                    toml_edit::value(CC_SWITCH_CODEX_MODEL_CATALOG_FILENAME);
+            }
         }
         None => {
             let should_remove = doc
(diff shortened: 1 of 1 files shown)


farion1231:
@codex review

STEVENTAN100:
是的，怎么一言不合就把我自己的model catalog给覆盖了，而且推理强度只剩下high #6181

farion1231 (approved):

farion1231:
Merged and shipped in v3.20.0 — thank you @misaka-myu! Sorry the thanks comes late.

I really appreciated the shape of this fix: the ownership guard on the `Some` arm uses the same filename rule the `None` arm and the read side already follow, so all paths now agree on who owns the `model_catalog_json` pointer — and the regression tests pin it in both directions (user-owned paths survive, cc-switch-owned ones are still reclaimed). Minimal, surgical, and exactly the behavior proposed in #3302.

One note for anyone landing here from #3302 / #6181: if your pointer was already overwritten before this fix, it now points at the cc-switch file and will keep being managed as such — re-point it at your own catalog once, and it will be respected from then on.

Thanks again for the contribution!

PR #6122, opened 2026-08-04
perf(backup): batch dump INSERTs and restore preserved tables in one transaction

zayokami:
**Title:** `perf(backup): batch dump INSERTs and restore preserved tables in one transaction`

## Summary

Backup import was pathologically slow once the database accumulates log rows (`proxy_request_logs`, `stream_check_logs`, `usage_daily_rollups` — which grow unbounded with proxy usage and are part of every full backup, and are preserved across WebDAV/S3 syncs).

Measured on a database with **26k log rows** (Windows, debug build — release numbers differ in absolute terms but the scaling structure is the same):

| Path | Before | After | Speedup |
|---|---|---|---|
| `import_sql_string` (local file import) | **27.96 s** | **0.39 s** | ~71× |
| `import_sql_string_for_sync` (WebDAV/S3 apply) | **118.9 s** | **0.33 s** | ~360× |
| `export_sql_string` (full) | 367 ms / **15.4 MB** | 193 ms / **3.9 MB** | ~2× faster, ~4× smaller |

The sync-import figure matters beyond the manual "restore" button: WebDAV/S3 **auto-sync** applies remote snapshots through the same path, so affected users were eating multi-minute stalls on a timer.

## Root cause analysis

Two independent bottlenecks, isolated with a phase-level benchmark (both perf harnesses are committed as `#[ignore]`d tests — `cargo test --lib perf_backup / perf_import_phases -- --ignored --nocapture`):

### 1. One `INSERT` per row → import cost scales with *statement count*, not data size

`dump_sql` emitted one single-row `INSERT` statement per row. On import, `execute_batch` parses / prepares / finalizes **every statement individually** — 20k rows ≈ 21 s in the staging step alone.

Evidence chain (controls in `perf_import_phases`):

- Same script on an **in-memory** database: equally slow (24.3 s) → CPU-bound parsing, **not** disk I/O
- `journal_mode=MEMORY` + `synchronous=OFF` on the staging DB: **no difference** → not journaling either
- Same data as **multi-row `VALUES` batches of 200**: **84.6 ms** (~280×)

### 2. `restore_tables`: one implicit transaction per row = one fsync per row

The WebDAV/S3 path preserves local-only tables by re-inserting them into the disk-backed staging DB. Each row was its own autocommit transaction → ~4.6 ms fsync per row on Windows → 119 s for 26k rows. This also meant a mid-restore failure left a half-restored table.

## Changes

- **`dump_sql`**: emits multi-row `VALUES (...), (...), ...` batches of 200 rows (single-row `INSERT`s were ~200× more statements). Also hoists the column-list string out of the row loop.
- **`restore_tables`**: wraps the whole restore in **one transaction** with the `INSERT` prepared once per table — one fsync total, and failures now roll back atomically.

## Compatibility

- **Old dumps → new app**: single-row dumps import through the same generic `execute_batch` — unchanged.
- **New dumps → old app**: multi-row `VALUES` requires SQLite ≥ 3.7.11 (2012); every shipped rusqlite bundle is far newer, and old builds import via the same code path. Dumps remain cross-version compatible in both directions.
- No schema changes; sync wire format (`manifest.json` hashes, artifacts) unchanged — just smaller `db.sql` payloads.

## Tests

- `dump_sql_batches_rows_into_multi_row_inserts` — pins the batch shape: 450 rows ⇒ exactly 3 `INSERT` statements. **Reverting to per-row export fails this test.**
- `multi_row_dump_round_trips_special_values` — round-trips single quotes, newlines, commas (the `VALUES` separator), CJK, emoji, BLOB storage class (`typeof()` preserved), and `NULL` through export+import.
- Existing coverage still green: authorizer rejection of cross-file statements, genuine-export acceptance, sync preserve-tables correctness.
- Ignored perf harnesses reproduce the before/after numbers above.

### DIFF
--- a/src-tauri/src/database/backup.rs
+++ b/src-tauri/src/database/backup.rs
@@ -15,6 +15,11 @@ use tempfile::NamedTempFile;
 
 const CC_SWITCH_SQL_EXPORT_HEADER: &str = "-- CC Switch SQLite 导出";
 
+/// Bound combined INSERT batches while still amortizing statement parsing.
+/// A row larger than this cap is emitted alone because it cannot be split.
+const INSERT_BATCH_MAX_ROWS: usize = 200;
+const INSERT_BATCH_MAX_BYTES: usize = 1024 * 1024;
+
 /// `dump_sql` 会写出的 PRAGMA。其余 PRAGMA 一律拒绝——`temp_store_directory`
 /// 能把临时文件重定向到任意目录，`writable_schema` 能绕过 schema 完整性检查。
 const IMPORT_ALLOWED_PRAGMAS: &[&str] = &["foreign_keys", "user_version"];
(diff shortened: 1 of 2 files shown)


farion1231 (approved):

PR #6144, opened 2026-08-04
fix: preserve provider entries when merging config arrays

### LINKED ISSUE #6141: 应用通用配置时，供应商配置数组字段被整体覆盖 "permissions.deny" 原有条目丢失
### Self Checks / 自检

### CC Switch Version / 版本号

3.19.1

### Operating System / 操作系统

macOS Ventura

### Related App / 涉及应用

Claude Code

### Steps to Reproduce / 重现步骤

1. 供应商配置含自己的 deny 规则:

```json
{ "permissions": { "deny": ["WebSearch"] } }
```

2. 通用配置片段也含 deny 规则:

```json
{ "permissions": { "deny": ["Read(~/.ssh/**)"] } }
```

3. 勾选"应用通用配置",查看合并后的配置

### Expected Behavior / 期望行为

数组做并集合并(或至少提示冲突,不静默丢弃):

```json
{ "permissions": { "deny": ["WebSearch", "Read(~/.ssh/**)"] } }
```

### Actual Behavior / 实际行为

通用配置的数组整体替换了原数组,"WebSearch" 被静默丢弃,且取消勾选后不会还原:

```json
{ "permissions": { "deny": ["Read(~/.ssh/**)"] } }
```

### Additional Context / 补充信息

原因:deepMerge 只对 object 递归合并,数组走整体替换分支:

https://github.com/farion1231/cc-switch/blob/59a2bd10407707282dcefe85b290f0ddaf4d0a74/src/utils/providerConfigUtils.ts#L60-L68

若改为数组并集,deepRemove (取消勾选时的剥离)和 isSubset (开关状态检测,要求数组完全相等)需同步调整,否则勾选状态会失灵。

allenxu09:
## Summary

- merge common-config arrays as a deduplicated union instead of replacing provider entries
- treat common-config arrays as subsets when detecting whether a snippet is applied
- remove only matching common-config array entries when disabling the snippet

## Root cause

`deepMerge` replaced arrays wholesale, while `isSubset` required equal length and order and `deepRemove` deleted the whole matching array. Applying a shared `permissions.deny` array therefore discarded provider-specific deny rules and disabling it could not restore them.

## Validation

- `pnpm exec vitest run src/utils/providerConfigUtils.test.ts` (22 tests passed)
- `pnpm typecheck`
- `pnpm exec prettier --check src/utils/providerConfigUtils.ts src/utils/providerConfigUtils.test.ts`
- `git diff --check`
- `ocr review --audience agent ...` (0 findings)

Closes #6141

### DIFF
--- a/src-tauri/src/services/provider/live.rs
+++ b/src-tauri/src/services/provider/live.rs
@@ -224,9 +224,7 @@ fn json_array_contains_subset(target_arr: &[Value], source_arr: &[Value]) -> boo
         matched_source_by_target: &mut [Option<usize>],
     ) -> bool {
         for (target_index, target_item) in target_arr.iter().enumerate() {
-            if seen[target_index]
-                || !json_is_subset(target_item, &source_arr[source_index])
-            {
+            if seen[target_index] || !json_is_subset(target_item, &source_arr[source_index]) {
                 continue;
             }
 
@@ -262,13 +260,9 @@ fn json_array_contains_subset(target_arr: &[Value], source_arr: &[Value]) -> boo
 
 fn json_remove_array_items(target_arr: &mut Vec<Value>, source_arr: &[Value]) {
     for source_item in source_arr {
-        if let Some(index) = target_arr
-            .iter()
-            .position(|target_item| {
-                json_is_subset(target_item, source_item)
-                    && json_is_subset(source_item, target_item)
-            })
-        {
+        if let Some(index) = target_arr.iter().position(|target_item| {
+            json_is_subset(target_item, source_item) && json_is_subset(source_item, target_item)
+        }) {
             target_arr.remove(index);
         }
     }
--- a/src-tauri/src/services/provider/live.rs
+++ b/src-tauri/src/services/provider/live.rs
@@ -216,17 +216,47 @@ fn json_is_subset(target: &Value, source: &Value) -> bool {
 }
 
 fn json_array_contains_subset(target_arr: &[Value], source_arr: &[Value]) -> bool {
-    let mut matched = vec![false; target_arr.len()];
+    fn try_match(
+        target_arr: &[Value],
+        source_arr: &[Value],
+        source_index: usize,
+        seen: &mut [bool],
+        matched_source_by_target: &mut [Option<usize>],
+    ) -> bool {
+        for (target_index, target_item) in target_arr.iter().enumerate() {
+            if seen[target_index]
+                || !json_is_subset(target_item, &source_arr[source_index])
+            {
+                continue;
+            }
 
-    source_arr.iter().all(|source_item| {
-        if let Some((index, _)) = target_arr.iter().enumerate().find(|(index, target_item)| {
-            !matched[*index] && json_is_subset(target_item, source_item)
-        }) {
-            matched[index] = true;
-            true
-        } else {
-            false
+            seen[target_index] = true;
+            let matched_source = matched_source_by_target[target_index];
+            if matched_source.is_none_or(|matched_source| {
+                try_match(
+                    target_arr,
+                    source_arr,
+                    matched_source,
+                    seen,
+                    matched_source_by_target,
+                )
+            }) {
+                matched_source_by_target[target_index] = Some(source_index);
+                return true;
+            }
         }
+        false
+    }
+
+    let mut matched_source_by_target = vec![None; target_arr.len()];
+    source_arr.iter().enumerate().all(|(source_index, _)| {
+        try_match(
+            target_arr,
+            source_arr,
+            source_index,
+            &mut vec![false; target_arr.len()],
+            &mut matched_source_by_target,
+        )
     })
 }
 
@@ -234,7 +264,10 @@ fn json_remove_array_items(target_arr: &mut Vec<Value>, source_arr: &[Value]) {
     for source_item in source_arr {
         if let Some(index) = target_arr
             .iter()
-            .position(|target_item| json_is_subset(target_item, source_item))
+            .position(|target_item| {
+                json_is_subset(target_item, source_item)
+                    && json_is_subset(source_item, target_item)
+            })
         {
             target_arr.remove(index);
         }
@@ -250,6 +283,37 @@ fn json_deep_merge(target: &mut Value, source: &Value) {
                     None => {
                         target_map.insert(key.clone(), source_value.clone());
                     }
+                }
+            }
+        }
+        (target_value, source_value) => {
+            *target_value = source_value.clone();
+        }
+    }
+}
+
+fn json_deep_merge_with_array_union(target: &mut Value, source: &Value) {
+    match (target, source) {
+        (Value::Object(target_map), Value::Object(source_map)) => {
+            for (key, source_value) in source_map {
+                match target_map.get_mut(key) {
+                    Some(target_value) => {
+                        json_deep_merge_with_array_union(target_value, source_value)
+                    }
+                    None => {
+                        target_map.insert(key.clone(), source_value.clone());
+                    }
+                }
+            }
+        }
+        (Value::Array(target_arr), Value::Array(source_arr)) => {
+            for source_item in source_arr {
+                let exists = target_arr.iter().any(|target_item| {
+                    json_is_subset(target_item, source_item)
+                        && json_is_subset(source_item, target_item)
+                });
+                if !exists {
+                    target_arr.push(source_item.clone());
                 }
             }
         }
@@ -278,6 +342,90 @@ fn json_deep_remove(target: &mut Value, source: &Value) {
                 remove_key = target_arr.is_empty();
             } else if json_is_subset(target_value, source_value) {
                 remove_key = true;
+            }
+        }
+
+        if remove_key {
+            target_map.remove(key);
+        }
+    }
+}
+
+fn json_deep_remove_preserving_original_arrays(
+    target: &mut Value,
+    source: &Value,
+    original: Option<&Value>,
+) {
+    let (Some(target_map), Some(source_map)) = (target.as_object_mut(), source.as_object()) else {
+        return;
+    };
+    let original_map = original.and_then(Value::as_object);
+
+    for (key, source_value) in source_map {
+        let mut remove_key = false;
+        let original_value = original_map.and_then(|map| map.get(key));
+
+        if let Some(target_value) = target_map.get_mut(key) {
+            if source_value.is_object() && target_value.is_object() {
+                json_deep_remove_preserving_original_arrays(
+                    target_value,
+                    source_value,
+                    original_value,
+                );
+                remove_key = original_value.is_none()
+                    && target_value.as_object().is_some_and(|obj| obj.is_empty());
+            } else if let (Some(target_arr), Some(source_arr)) =
+                (target_value.as_array_mut(), source_value.as_array())
+            {
+                let original_arr = original_value
+                    .and_then(Value::as_array)
+                    .map(Vec::as_slice)
+                    .unwrap_or(&[]);
+                let mut processed_source_items: Vec<&Value> = Vec::new();
+                for source_item in source_arr {
+                    let already_processed = processed_source_items.iter().any(|processed_item| {
+                        json_is_subset(processed_item, source_item)
+                            && json_is_subset(source_item, processed_item)
+                    });
+                    if already_processed {
+                        continue;
+                    }
+                    processed_source_items.push(source_item);
+
+                    let original_count = original_arr
+                        .iter()
+                        .filter(|original_item| {
+                            json_is_subset(original_item, source_item)
+                                && json_is_subset(source_item, original_item)
+                        })
+                        .count();
+                    let injection_budget = usize::from(original_count == 0);
+                    let target_count = target_arr
+                        .iter()
+                        .filter(|target_item| {
+                            json_is_subset(target_item, source_item)
+                                && json_is_subset(source_item, target_item)
+                        })
+                        .count();
+                    let remove_count = target_count
+                        .saturating_sub(original_count)
+                        .min(injection_budget);
+                    for _ in 0..remove_count {
+                        if let Some(index) = target_arr.iter().rposition(|target_item| {
+                            json_is_subset(target_item, source_item)
+                                && json_is_subset(source_item, target_item)
+                        }) {
+                            target_arr.remove(index);
+                        }
+                    }
+                }
+                remove_key = original_value.is_none() && target_arr.is_empty();
+            } else if json_is_subset(target_value, source_value) {
+                if let Some(original_value) = original_value {
+                    *target_value = original_value.clone();
+                } else {
+                    remove_key = true;
+                }
             }
         }
 
@@ -610,6 +758,28 @@ pub(crate) fn remove_common_config_from_settings(
     }
 }
 
+fn remove_common_config_from_live_settings(
+    app_type: &AppType,
+    settings: &Value,
+    snippet: &str,
+    original_settings: &Value,
+) -> Result<Value, AppError> {
+    if !matches!(app_type, AppType::Claude) {
+        return remove_common_config_from_settings(app_type, settings, snippet);
+    }
+
+    let trimmed = snippet.trim();
+    if trimmed.is_empty() {
+        return Ok(settings.clone());
+    }
+
+    let source = serde_json::from_str::<Value>(trimmed)
+        .map_err(|e| AppError::Message(format!("Invalid Claude common config: {e}")))?;
+    let mut result = settings.clone();
+    json_deep_remove_preserving_original_arrays(&mut result, &source, Some(original_settings));
+    Ok(result)
+}
+
 fn apply_common_config_to_settings(
     app_type: &AppType,
     settings: &Value,
@@ -625,7 +795,7 @@ fn apply_common_config_to_settings(
             let source = serde_json::from_str::<Value>(trimmed)
                 .map_err(|e| AppError::Message(format!("Invalid Claude common config: {e}")))?;
             let mut result = settings.clone();
-            json_deep_merge(&mut result, &source);
+            json_deep_merge_with_array_union(&mut result, &source);
             Ok(result)
         }
         AppType::Codex => {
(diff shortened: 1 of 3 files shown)


farion1231 (commented):
Thanks for this fix — the root-cause analysis is spot on, and updating all three
operators (`deepMerge` / `deepRemove` / `isSubset`) together with tests is exactly
the right shape for the frontend half. With this PR, provider-specific array entries
are no longer permanently lost from the saved provider snapshot, which is a real
improvement on its own.

However, there is a second implementation of this merge logic on the Rust side that
the issue didn't mention, and it's the one that actually produces the live config.
As it stands, the fix resolves the snapshot/editor half of #6141 but not the half
users observe in `~/.claude/settings.json`.

## [P1] The backend overlay still replaces arrays wholesale in the live config

Since the runtime-overlay change (6d078e7f), common-config snippets are stripped
from provider snapshots on save and re-applied dynamically when writing live files:
`write_live_with_common_config` → `apply_common_config_to_settings` →
[`json_deep_merge`](https://github.com/farion1231/cc-switch/blob/290b65c03bc72787f907a788e4d49673cc604142/src-tauri/src/services/provider/live.rs#L241-L257).
That function only recurses on `(Object, Object)`; arrays fall through to the
catch-all clone and are **replaced**, not unioned.

Verified with a probe test on the merge tree — feeding the exact #6141 scenario
into `apply_common_config_to_settings`:

```
snapshot: {"permissions": {"deny": ["WebSearch"]}}
snippet:  {"permissions": {"deny": ["Read(~/.ssh/**)"]}}
result:   deny = ["Read(~/.ssh/**)"]        // WebSearch still dropped
```

So after this PR, the editor preview shows the union (`["WebSearch", "Read(~/.ssh/**)"]`)
while the file Claude Code actually reads still loses `WebSearch` — arguably more
misleading than before, since the two used to at least agree. Notably, the sibling
operators in the same file (`json_remove_array_items`, `json_array_contains_subset`)
are already array-aware; merge is the only one missing the array branch.

Suggested fix — mirror this PR's union semantics in `json_deep_merge`:

```rust
(Value::Array(target_arr), Value::Array(source_arr)) => {
    for source_item in source_arr {
        let exists = target_arr.iter().any(|target_item| {
            json_is_subset(target_item, source_item)
                && json_is_subset(source_item, target_item)
        });
        if !exists {
            target_arr.push(source_item.clone());
        }
    }
}
```

Plus a roundtrip test in `live.rs` covering *overlapping* array fields (the existing
roundtrip tests only cover non-overlapping fields, which is why this went unnoticed).

## [P2] Frontend subset check can match one target element twice

The new `isSubset` array branch
([`providerConfigUtils.ts#L134-L139`](https://github.com/farion1231/cc-switch/blob/cc02cc4279e00626001e3db0b814d6d80957191e/src/utils/providerConfigUtils.ts#L134-L139))
uses `source.every(... target.some(...))`, so a single target element can satisfy
multiple source elements. With target `[{a:1,b:2}]` and snippet `[{a:1},{a:1,b:2}]`
it returns `true`, while the backend's `json_array_contains_subset` (one-to-one
matching via a `matched` index) returns `false`. Consequence: in edit mode the form
infers "already applied" and skips the auto re-merge
([`useCommonConfigSnippet.ts#L137`](https://github.com/farion1231/cc-switch/blob/290b65c03bc72787f907a788e4d49673cc604142/src/components/providers/forms/hooks/useCommonConfigSnippet.ts#L137)),
so the missing `{a:1}` entry never gets added. Only bites object-element arrays
(hooks-like structures), but since you're touching this function anyway, mirroring
the backend's consumed-match logic is cheap:

```ts
if (Array.isArray(source)) {
  if (!Array.isArray(target)) return false;
  const matched = new Array(target.length).fill(false);
  return source.every((sourceItem) => {
    const index = target.findIndex(
      (targetItem, i) => !matched[i] && isSubset(targetItem, sourceItem),
    );
    if (index === -1) return false;
    matched[index] = true;
    return true;
  });
}
```

## [P2] Codex TOML side has the same replace-on-merge gap (fine as a follow-up)

`merge_toml_item`
([live.rs#L375-L384](https://github.com/farion1231/cc-switch/blob/290b65c03bc72787f907a788e4d49673cc604142/src-tauri/src/services/provider/live.rs#L375-L384))
also clones arrays wholesale, and it's shared by both the Codex form materialization
(`update_toml_common_config_snippet`) and the Codex live overlay. Since #6141 only
reports Claude, I'd be fine with a follow-up issue rather than growing this PR —
but it shouldn't be forgotten, or the Codex half of the same bug survives.

---

**To merge**: the P1 backend union (with an overlapping-field test) is required —
without it `Closes #6141` doesn't hold for the live config. The first P2 is a small
diff in a function you're already editing, so ideally in this PR; the Codex P2 can
be a follow-up issue. Happy to re-review quickly once updated.

farion1231:
@codex review

SaladDay (commented):
Thanks for this PR. The core algorithm (bipartite matching, array union merge, injection_budget logic) is correct and well-tested.

Two issues were found during review that should be addressed before merge:

**1. Type-mismatch bug in `json_deep_remove_preserving_original_arrays`**

When a common config snippet changes a key'\''s type (e.g., scalar to object, or array to object), the backfill loses the original value instead of restoring it.

Example:
- Provider original: `{"deny": "WebSearch"}`
- Snippet: `{"deny": {"rule": "x"}}`
- After apply (live): `{"deny": {"rule": "x"}}`
- After backfill: `{"deny": {}}` (expected `{"deny": "WebSearch"}`)

Root cause: the object-recurse branch fires when both `source_value` and `target_value` are objects, but it passes `original_value` into the recursion without checking that `original_value` is also an object. When `original_value` is a scalar/array, `original_map` becomes `None` inside the recursion, so all inner keys are treated as "original absent" and removed.

Suggested fix: guard the recursion by checking `original_value`'\''s type. If `original_value` is `Some` but not the same type as `target_value`/`source_value`, restore `*target_value = original_value.clone()` wholesale instead of recursing.

**2. Rust/TS inconsistency in array removal**

Rust `json_remove_array_items` removes **one** matching target per source item (first match via `position`), but TS `deepRemove` uses `filter` to remove **all** matching items. This causes different behavior when duplicate array entries exist: the frontend preview would show all duplicates removed, while the backend preserves one.

Suggested fix: change TS `deepRemove` to mirror Rust by using a splice-one-per-source-item loop instead of `filter`.

**Minor:** `json_remove_array_items` uses `position` (first match) while `json_deep_remove_preserving_original_arrays` uses `rposition` (last match). Since matched items are mutual-subset (deep-equal), the result is the same, but consistency would be cleaner.

SaladDay:
Review superseded. We are taking over this fix directly on the main branch.

SaladDay:
Pushed the type-mismatch fix and Rust/TS consistency improvements on top of the existing commits:

1. **Type-mismatch guard for array branch**: `json_deep_remove_preserving_original_arrays` now guards the array branch the same way as the object branch — when the original value's type differs from the snippet's array type, restore the original wholesale instead of leaving an empty array.
2. **Simplified `remove_count`**: `target_count.min(injection_budget)` — the `saturating_sub` was redundant.
3. **Added tests**: scalar→array, object→array, array→object type-change round-trips.
4. **TS `deepRemove` alignment**: changed from `filter` (removes all matching) to splice-one-per-source-item, matching Rust's `json_remove_array_items`.
5. **Removed dead code**: unreachable `else { remove_key = true; }` in the object branch guard.
6. **Formatting**: rustfmt + prettier.

allenxu09:
@farion1231

farion1231 (commented):
Thanks for the thorough follow-up — this round addresses everything from the previous review. The backend live path now union-merges arrays with the #6141 scenario locked by tests, the Rust/TS operators are aligned on the same one-to-one matching semantics (nice touch replacing greedy matching with augmenting-path matching on both sides), and the original-aware backfill in `json_deep_remove_preserving_original_arrays` correctly keeps provider-owned entries intact when stripping the snippet from live. I also merged this branch onto current `main` locally and ran the full backend + frontend suites: all green.

Two items before merge:

**[P2] Array elements are merged without sanitization, breaking the forbidden-key contract**

The new array branch in `deepMerge` pushes source elements as-is ([providerConfigUtils.ts#L60-L70](https://github.com/farion1231/cc-switch/blob/34f1da7eb0cc2d584f736c3c62b604056e398071/src/utils/providerConfigUtils.ts#L60-L70)). `FORBIDDEN_MERGE_KEYS` only guards object *keys* during traversal, so an element that itself contains a forbidden key (e.g. `{"hooks":[{"__proto__":{}}]}`) flows through raw. Reproduced consequences:

- the raw element (with `__proto__` as an own property) is written into the config text — no runtime prototype pollution occurs, but the write-side invariant documented at [#L223-L225](https://github.com/farion1231/cc-switch/blob/34f1da7eb0cc2d584f736c3c62b604056e398071/src/utils/providerConfigUtils.ts#L223-L225) ("sanitizing is a byte-identical no-op on the write side") no longer holds;
- re-enabling the toggle appends a duplicate every time: the dedup check uses bidirectional `isSubset`, which by design vetoes any value containing a forbidden key, so `exists` is always false (3 merges → 3 copies);
- disabling can't remove it either — the same veto makes `findIndex` in [`deepRemove`](https://github.com/farion1231/cc-switch/blob/34f1da7eb0cc2d584f736c3c62b604056e398071/src/utils/providerConfigUtils.ts#L94-L108) never match.

This matters because the snippet is remote-reachable input per the threat model documented above `FORBIDDEN_MERGE_KEYS` (WebDAV/S3 sync can overwrite it without XSS).

Minimal fix — sanitize once at the entry instead of patching all three traversal sites, and update the now-stale comment:

```ts
// updateCommonConfigSnippet, ~L225
const snippet = sanitizeSnippet(
  JSON.parse(snippetString),
) as Record<string, any>;
```

With elements cleaned on write, dedup and removal compare cleaned-vs-cleaned and behave correctly, and the read side ([`hasCommonConfigSnippet` already sanitizes at #L252](https://github.com/farion1231/cc-switch/blob/34f1da7eb0cc2d584f736c3c62b604056e398071/src/utils/providerConfigUtils.ts#L250-L252)) ends up comparing exactly what the write side produces. A regression test with a forbidden-key element (merge twice → length stays 1; disable → removed) would lock it in.

**[P2] Rebase needed — conflicts with current `main`**

`src-tauri/src/services/provider/live.rs` now conflicts: both this branch and `main` inserted new functions right after `write_live_with_common_config` (main's Codex OAuth helpers vs this PR's `remove_common_config_from_live_settings`). It's a pure insertion-position conflict — keeping both blocks resolves it; that's exactly the resolution I tested green locally. CI will need a re-run after the rebase since the current green run predates those `main` changes.

**Follow-up (non-blocking, separate issue/PR welcome)**

Now that the live-backfill path is original-aware, the save-time normalization in [`normalize_provider_common_config_for_storage`](https://github.com/farion1231/cc-switch/blob/34f1da7eb0cc2d584f736c3c62b604056e398071/src-tauri/src/services/provider/live.rs#L1080-L1108) still strips with the plain `remove_common_config_from_settings` (no original reference), so a provider-owned array entry that coincides with a snippet entry (e.g. the provider's own `"WebSearch"` while the snippet also lists it) is consumed from the stored snapshot on save. The stored row is available at that point, so it could reuse `json_deep_remove_preserving_original_arrays` the same way the backfill does. Pre-existing behavior was strictly worse (the whole array was lost), so no need to grow this PR — flagging it for symmetry. The Codex TOML side (`merge_toml_item` still replacing arrays wholesale) from the previous review also remains open as a follow-up.

Fix the sanitization item and rebase, and this is ready to merge — I'll re-review quickly once pushed.

allenxu09:
All blocking review items are addressed at 48ea2192: the frontend write path sanitizes forbidden keys inside array elements, regression coverage includes repeated merge and removal, and the branch is mergeable with current main. All frontend and backend CI checks are green. Ready for re-review.

farion1231 (commented):
Re-checked `48ea219` merged onto current `main` (`f3b18df`): clean merge, Rust 2825/0 and frontend 1091/0 green, and the two blocking items from last round are resolved. Sanitizing the whole snippet once at the `updateCommonConfigSnippet` entry is a cleaner fix than the three-site one I suggested, and the rebase landed the new live.rs function exactly where it should.

One more pass with a degenerate input — a snippet array that contains **duplicate entries** — surfaced two regressions against `main`. Both are small to fix, and they share a single root cause: the three operators no longer agree on what a duplicate means. Merge de-duplicates, but detection and the backfill budget still count copies.

Duplicates are not only a hand-typing mistake. When a provider is switched away, `sync_common_config_snippet_from_live` re-extracts the snippet from the live file via [`extract_claude_common_config`](https://github.com/farion1231/cc-switch/blob/48ea21924c811b5417af84deb8e8f9a020d420bc/src-tauri/src/services/provider/mod.rs#L5775), which clones the settings verbatim and only strips excluded keys — so a duplicated `deny` rule in `~/.claude/settings.json` flows straight into the snippet.

### [P2] Merged config no longer passes the "applied" check when the snippet array has duplicates

[`deepMerge`](https://github.com/farion1231/cc-switch/blob/48ea21924c811b5417af84deb8e8f9a020d420bc/src/utils/providerConfigUtils.ts#L60-L70) de-duplicates (`["A","A"]` → `["A"]`), but the [`isSubset` array branch](https://github.com/farion1231/cc-switch/blob/48ea21924c811b5417af84deb8e8f9a020d420bc/src/utils/providerConfigUtils.ts#L137-L162) is one-to-one matching: two source elements need two distinct target elements, so the config we just merged is reported as *not applied*.

User-visible chain: the effect in [`useCommonConfigSnippet.ts#L311-L322`](https://github.com/farion1231/cc-switch/blob/48ea21924c811b5417af84deb8e8f9a020d420bc/src/components/providers/forms/hooks/useCommonConfigSnippet.ts#L311-L322) re-derives the toggle on any `settingsConfig` change (e.g. editing the model field), flips it off, and [`ProviderForm.tsx#L1716`](https://github.com/farion1231/cc-switch/blob/48ea21924c811b5417af84deb8e8f9a020d420bc/src/components/providers/forms/ProviderForm.tsx#L1716-L1718) persists that as `commonConfigEnabled: false`. The entries stay baked into the config, but later snippet edits stop propagating to that provider and the switch-away backfill is skipped.

Repro (fails on `48ea219`, passes on `main` where arrays were copied wholesale):

```ts
const snippet = JSON.stringify({ permissions: { deny: ["Read(~/.ssh/**)", "Read(~/.ssh/**)"] } });
const merged = updateCommonConfigSnippet("{}", snippet, true).updatedConfig;
// merged => {"permissions":{"deny":["Read(~/.ssh/**)"]}}
expect(hasCommonConfigSnippet(merged, snippet)).toBe(true); // got false
```

Same result when the provider already owns one copy (`deny: ["WebSearch"]` + snippet `["WebSearch","WebSearch"]`). `json_is_subset` on the Rust side returns `false` for the same input after a union merge; that only matters for providers without an explicit `commonConfigEnabled` flag ([`provider_uses_common_config`](https://github.com/farion1231/cc-switch/blob/48ea21924c811b5417af84deb8e8f9a020d420bc/src-tauri/src/services/provider/live.rs#L682-L697) infers from the settings in that case), so it is a smaller blast radius than the frontend one.

### [P2] Backfill budget misses duplicates injected by the new-key branch

When the provider has no `permissions.deny` at all, [`json_deep_merge`'s `None` branch](https://github.com/farion1231/cc-switch/blob/48ea21924c811b5417af84deb8e8f9a020d420bc/src-tauri/src/services/provider/live.rs#L279-L281) inserts `source_value.clone()` verbatim — two copies land in the live file. The backfill in [`json_deep_remove_preserving_original_arrays`](https://github.com/farion1231/cc-switch/blob/48ea21924c811b5417af84deb8e8f9a020d420bc/src-tauri/src/services/provider/live.rs#L402-L407) caps removal at `injection_budget = 1`, so one copy survives into the provider snapshot. On the next cycle the snapshot already owns that entry (`original_count == 1` → budget 0), and the rule is permanently hardened into the provider: clearing or editing the common snippet no longer removes it. That is exactly the "snapshots never absorb injected entries" guarantee this PR's backfill is meant to provide.

Repro (fails on `48ea219`; on `main` the old `json_remove_array_items` strips one target per source item, so both copies go and the result is `{}`):

```rust
let settings = json!({});
let snippet = r#"{"permissions":{"deny":["Read(~/.ssh/**)","Read(~/.ssh/**)"]}}"#;
let applied = apply_common_config_to_settings(&AppType::Claude, &settings, snippet).unwrap();
// applied.permissions.deny => ["Read(~/.ssh/**)", "Read(~/.ssh/**)"]
let backfilled = remove_common_config_from_live_settings(&AppType::Claude, &applied, snippet, &settings).unwrap();
assert_eq!(backfilled, settings); // got {"permissions":{"deny":["Read(~/.ssh/**)"]}}
```

### Suggested fix: treat snippet arrays as sets, on both sides

Once merge de-duplicates, detection and removal have to see the same de-duplicated shape. The smallest way to get there is to normalize at the points where the snippet enters the operators:

TS — de-duplicate in [`sanitizeSnippet`](https://github.com/farion1231/cc-switch/blob/48ea21924c811b5417af84deb8e8f9a020d420bc/src/utils/providerConfigUtils.ts#L41-L51), which is already the single choke point for the JSON write path and both read paths:

```ts
if (Array.isArray(value)) {
  const items = value.map(sanitizeSnippet);
  return items.filter(
    (item, i) => items.findIndex((o) => isSubset(o, item) && isSubset(item, o)) === i,
  );
}
```

Rust — make the new-key branch inject at most one copy (this also aligns it with the `(Array, Array)` union branch and the budget of 1):

```rust
None => {
    if source_value.is_array() {
        let mut fresh = Value::Array(Vec::new());
        json_deep_merge(&mut fresh, source_value);
        target_map.insert(key.clone(), fresh);
    } else {
        target_map.insert(key.clone(), source_value.clone());
    }
}
```

and de-duplicate `source_arr` (by `json_arrays_equal`) before matching in [`json_array_contains_subset`](https://github.com/farion1231/cc-switch/blob/48ea21924c811b5417af84deb8e8f9a020d420bc/src-tauri/src/services/provider/live.rs#L218) and before iterating in [`json_remove_array_items`](https://github.com/farion1231/cc-switch/blob/48ea21924c811b5417af84deb8e8f9a020d420bc/src-tauri/src/services/provider/live.rs#L262-L271). Roughly 20 lines total. The two repros above (plus their "provider already owns one copy" variants) make good regression tests — the invariant to lock is `detect(merge(cfg, s), s) == true` and `remove(merge(cfg, s), s) == cfg` even when `s` contains duplicates.

For the record: my earlier note calling the Rust-keeps / TS-de-duplicates asymmetry cosmetic was wrong — it is the root of the second item.

The two follow-ups from last round (save-path normalization without the original snapshot; Codex TOML arrays) are unchanged and still not blocking.

Fix the two P2s above and this is good to merge. Happy to re-review quickly once they land.

Issue #6129, opened 2026-08-04
[Bug] Backup and sync follow-up audit finds fidelity and atomicity gaps


YUZHEthefool:
### Self Checks / 自检

- [x] I have read the FAQ.
- [x] I searched existing open and closed issues.

### CC Switch Version / 版本号

3.19.1 (`main` at `59a2bd10`), follow-up audit based on PR #6122

### Operating System / 操作系统

All platforms; deterministic regression tests were run on Windows.

### Related App / 涉及应用

Claude Code, Codex, Gemini CLI, OpenCode, OpenClaw, Other / database backup and cloud sync

### Summary / 问题概述

A broader audit of SQL backup and WebDAV/S3 restore paths found a related set of fidelity and atomicity gaps. They will be reviewed as two Draft stacked PRs with small conventional commits, instead of many separate PRs:

1. **Backup fidelity layer** (depends on #6122)
   - preserve incremental `auto_vacuum` during staged import;
   - preserve NUL/non-UTF-8 TEXT and REAL storage/value semantics;
   - preserve `sqlite_sequence` AUTOINCREMENT high-water marks;
   - accept genuine schema-complete empty exports while rejecting truncated/header-only SQL.

2. **Sync consistency and reconciliation layer** (stacked on the backup layer)
   - keep `session_log_sync` cursors device-local while preserving ordinary SQL backup behavior;
   - preserve local-only writes that arrive during staging;
   - serialize WebDAV, S3, manual SQL import, and `.db` restore through one process lock;
   - keep post-import live projection inside that lock;
   - coordinate Skills DB rows with the filesystem SSOT through a fixed lock order;
   - trigger auto-sync for Profile changes;
   - rebuild Prompt/provider/MCP/Skill live projections, re-apply the local pricing sidecar, refresh settings/log state, and invalidate the real process UsageCache after restore.

### Expected Behavior / 期望行为

Backups round-trip supported SQLite state, and every restore produces one coherent database/Skills/live configuration state without losing concurrent device-local writes.

### Actual Behavior / 实际行为

The current paths can lose SQLite metadata/value fidelity, reject a valid empty export, mix database and Skills snapshots from different moments, overwrite device-local cursors/writes, allow restore operations to overlap, and leave managed live files/process caches stale after a successful database replacement.

### Test Plan / 测试计划

- legacy single-row and current batched SQL imports;
- file/String SQL API round-trips and failed-import atomicity;
[...]

PR #6146, opened 2026-08-04
fix(backup): preserve SQL fidelity and recovery safety

YUZHEthefool:
## Summary / 概述

This is the first layer of the follow-up audit after #6122.

- Preserve SQLite `auto_vacuum`, TEXT byte/storage semantics, REAL values, and `sqlite_sequence` AUTOINCREMENT high-water marks across SQL export/import.
- Accept genuine schema-complete empty exports while rejecting header-only, truncated-transaction, cross-file, and otherwise unsafe SQL without touching the live database.
- Publish database snapshots through hidden temporary files with no-clobber retry, keep restore source/safety snapshots protected during retention, reject corrupt/future-schema snapshots before replacement, and serialize backup-directory operations inside the process.
- Add focused regression coverage for legacy/current SQL formats, failure atomicity, filename collisions, retention, rename/delete races, and exact pre-restore safety snapshots.

Backup-directory coordination is process-local and relies on CC Switch's existing single-instance runtime invariant. Concurrent destructive operations from external tools or another process remain out of scope.

## Related Issue / 关联 Issue

Refs #6129

Follow-up to #6122. Dependent Draft: #6147.

## Screenshots / 截图

Not applicable — Rust backend and tests only.

## Verification / 验证

- `cargo fmt --check`
- `cargo test --lib database::backup::tests -- --test-threads=1` — 34 passed, 2 ignored performance harnesses
- `cargo test --test import_export_sync -- --test-threads=1` — 26 passed

## Checklist / 检查清单

- [x] `pnpm typecheck` not applicable (Rust-only diff)
- [x] `pnpm format:check` not applicable (Rust-only diff)
- [x] `cargo clippy` passes (Rust code changed)

### DIFF
--- a/src-tauri/src/database/backup.rs
+++ b/src-tauri/src/database/backup.rs
@@ -11,7 +11,8 @@ use rusqlite::types::ValueRef;
 use rusqlite::Connection;
 use std::fs;
 use std::path::{Path, PathBuf};
-use tempfile::NamedTempFile;
+use std::sync::{Mutex, MutexGuard};
+use tempfile::{Builder, NamedTempFile};
 
 const CC_SWITCH_SQL_EXPORT_HEADER: &str = "-- CC Switch SQLite 导出";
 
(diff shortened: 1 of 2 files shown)


farion1231:
@codex review

farion1231:
Merged — thank you for this one. The restore-path rework is a real data-safety win: while reviewing we independently reproduced the old failure chain where retention cleanup could delete the very backup being restored, letting `Connection::open` recreate it as an empty database that then silently replaced the live DB. Read-only open + staged validation + protected retention closes that for good, and the test coverage across collisions, races and retention edges is much appreciated.

Two small follow-ups if you're interested (neither is a blocker; they'd fold naturally into the #6147 stack):

- `validate_imported_schema` only probes table names, so a hand-edited current-version backup (table present, column dropped) still restores and only fails later with `no such column`. A cheap column-level probe after migrations (e.g. `SELECT <key columns> … LIMIT 0` per required table) would close it.
- On the restore/import staging path only, `integrity_check` + `foreign_key_check` would catch index/UNIQUE corruption and dangling FKs that `quick_check` passes. The periodic-backup path should keep `quick_check` for speed.

#6147 can now rebase onto main. Thanks again!

PR #6147, opened 2026-08-04
fix(sync): harden sync and restore consistency

### LINKED ISSUE #6129: [Bug] Backup and sync follow-up audit finds fidelity and atomicity gaps

YUZHEthefool:
## Summary / 概述

This is the second and final layer of the #6129 follow-up audit.

- Keep `session_log_sync` cursors device-local in remote payloads while preserving their existing behavior in ordinary SQL backup/export.
- Apply remote SQL only after staging is complete, then hold the backup-file and live-database guards across the exact safety snapshot, preserved-table read, and final replacement so late local writes are not lost.
- Serialize WebDAV, S3, manual SQL import, and binary restore through one process-wide sync operation lock; keep download suppression, post-import projection, status/error persistence, and upload triggering inside the correct guarded lifecycle.
- Coordinate Skills database snapshots and filesystem SSOT with a fixed lock order, and reconcile Prompt, provider/MCP live files, Skills, the local pricing sidecar, settings/log state, and the real process `UsageCache` after restore.
- Trigger auto-sync for Profile changes and add deterministic concurrency/late-write regression tests for both in-memory and file-backed databases.

## Stack / 依赖关系

Depends on #6146 and must be merged after it.

Both Draft PRs target `main`; until #6146 is merged, GitHub shows the parent commits here as well. Review the second-layer delta at:
https://github.com/YUZHEthefool/cc-switch/compare/fix/backup-followup-fidelity...fix/sync-followup-consistency

## Related Issue / 关联 Issue

Closes #6129

## Screenshots / 截图

Not applicable — Rust backend and tests only.

## Verification / 验证

- `cargo fmt --check`
- Backup/SQL unit tests — 34 passed, 2 ignored performance harnesses
- Sync-named unit tests — 94 passed
- Skills unit tests — 40 passed
- Manual restore, Prompt projection, provider live, and UsageCache tests — 28 passed
- `cargo test --test import_export_sync -- --test-threads=1` — 26 passed
- `cargo test --test skill_sync -- --test-threads=1` — 7 passed

## Checklist / 检查清单

- [x] `pnpm typecheck` not applicable (Rust-only diff)
- [x] `pnpm format:check` not applicable (Rust-only diff)
- [x] `cargo clippy` passes (Rust code changed)


farion1231:
@codex review

farion1231:
Merged — thank you for this two-layer audit. It was a pleasure to review: every commit maps cleanly onto one invariant, the staging-hook injection is exactly the right tool for pinning the late-write windows deterministically, and moving the post-import projection onto the real process state fixed a cache-invalidation bug nobody had noticed. The ~60% test share across the stack is what made the concurrency claims verifiable at all.

Merge note: since #6146 landed as a squash, I resolved the expected `backup.rs` overlap by taking this branch's side (verified byte-identical to layer 1 + layer 2) and ran the full suite on the merged tree before merging.

A few cheap follow-ups if you're interested — none were blocking:
1. The two new global-lock tests (`webdav_and_s3_operations_share_one_sync_mutex`, `manual_restore_starts_blocking_work_after_global_lock_acquisition`) could use `#[serial]`: tokio's mutex is fair, so on `drop` the lock hands off to a queued waiter and the `try_lock().is_ok()` assertion fails deterministically if test scheduling ever overlaps with the other lock tests.
2. `install`/`update_skill` resolve `get_ssot_dir()` before the download `await` and keep using the cached `dest` after re-acquiring the write guard; re-resolving under the guard (~10 lines) would close the narrow race with a concurrent storage-location migration — the update variant is the sneaky one, since `content_hash` gets written for the new version while the files land in the old location.
3. The follow-ups invited on #6146 (column-level schema probe, FK/integrity re-check on the restore path) remain open too.

Thanks again — this pair meaningfully hardens the sync/restore paths.

Issue #6101, opened 2026-08-04
[Bug] Databricks GPT-5.6: fresh Codex task fails when automation_update is flattened into Chat tools


zxy719:
## Summary

CC Switch 3.19.1 fails deterministically when a fresh Codex Desktop task sends the built-in `automation_update` tool directly through the Responses -> Chat Completions compatibility route to a Databricks-hosted GPT-5.6 deployment.

A pre-existing task using the same machine, provider, model, and project continues to work because its desktop tools remain grouped in a deferred `codex_app` namespace. This provides a useful A/B comparison showing that the failure depends on the per-task tool registration shape, not the prompt, credentials, model availability, or network.

Related issues: #5829 and #5839.

## Environment

- Windows
- CC Switch: `3.19.1`
- Codex Desktop: `26.727.51351`
- Codex runtime: `0.146.0-alpha.9.2`
- CC Switch route: Codex Responses -> OpenAI Chat Completions
- Upstream: Databricks serving endpoint backed by OpenAI
- Model: `gsds-gpt-56-sol`

Credentials, workspace hostname, usernames, local paths, and conversation content are intentionally redacted.

## Error

```text
CC Switch local proxy failed while handling Codex endpoint /responses.
Provider: Databricks GPT-5.6 Sol;
model: gsds-gpt-56-sol;
upstream_status: HTTP 400;
cause: {
  "external_model_provider": "openai",
  "external_model_error": {
    "error": {
      "message": "Invalid schema for function 'automation_update': schema must have type 'object' and not have 'oneOf'/'anyOf'/'allOf'/'enum'/'const'/'not' at the top level.",
      "type": "invalid_request_error",
      "param": "tools[13].function.parameters",
      "code": "invalid_function_parameters"
    }
  }
}
```

The same task failed five consecutive times before any model output was produced.

### Working task

The task's session metadata consistently registers two deferred namespaces:

```text
dynamic_tools:
  - codex_app
  - plugin_management
```

Inside `codex_app`, `automation_update` has:

```json
{
  "name": "automation_update",
  "deferLoading": true,
  "inputSchema": {
    "$schema": "...",
    "oneOf": ["..."],
    "$defs": {"...": "..."}
  }
}
```

The task uses `tool_search` to load only the tools it needs. It never loads `automation_update`, so that schema is not exposed to the strict Chat Completions upstream and the task works normally.

### Failing fresh task

The fresh task registers 17 desktop tools directly instead of two deferred namespaces:

```text
[...]

PR #6162, opened 2026-08-05
fix(proxy): flatten Responses tool schemas for strict Chat gateways

KDB-Wind:
# PR draft: #6101

## Title

fix(proxy): flatten Responses tool schemas for strict Chat gateways

## Summary

Strict Chat Completions gateways can reject Responses tool schemas that still contain root `oneOf`, `$defs`, or other unsupported JSON Schema metadata. The transform expands local `$ref` definitions, flattens union branches into the tool's object properties, keeps fields required by every union branch, and removes unsupported root keywords. It also avoids walking an already expanded schema a second time.

## Validation

- `cargo test --manifest-path src-tauri/Cargo.toml --lib transform_codex_chat`

### DIFF
--- a/PR_DRAFT_01_6101.md
+++ b/PR_DRAFT_01_6101.md
@@ -1,0 +1,17 @@
+# PR draft: #6101
+
+## Title
+
+fix(proxy): flatten Responses tool schemas for strict Chat gateways
+
+## Summary
+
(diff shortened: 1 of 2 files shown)


farion1231:
@codex review

PR #6163, opened 2026-08-05
fix(proxy): keep zero cache details in Responses usage

KDB-Wind:
# PR draft: #6140

## Title

fix(proxy): keep zero cache details in Responses usage

## Summary

Responses usage always includes `input_tokens_details.cached_tokens`. The value is `0` when the upstream reports zero or omits `prompt_tokens_details`; `cache_write_tokens` is included only when it is positive. This keeps the response shape valid for clients that treat `input_tokens_details` as required.

## Validation

- `cargo test --manifest-path src-tauri/Cargo.toml --lib transform_codex_chat`

### DIFF
--- a/PR_DRAFT_02_6140.md
+++ b/PR_DRAFT_02_6140.md
@@ -1,0 +1,17 @@
+# PR draft: #6140
+
+## Title
+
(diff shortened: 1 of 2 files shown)


farion1231:
@codex review

PR #6164, opened 2026-08-05
fix(tray): preserve automatic failover when switching providers

KDB-Wind:
# PR draft: #6022

## Title

fix(tray): preserve automatic failover when switching providers

## Summary

Switching providers from the tray now preserves the stored `auto_failover_enabled` value. The flag pair is written back before the switch, which also refreshes its `updated_at` timestamp. The tray event reports the stored value instead of a false disabled state. Manual provider switching and automatic failover remain separate operations.

## Validation

- Rust library tests covering proxy and tray behavior

### DIFF
--- a/PR_DRAFT_03_6022.md
+++ b/PR_DRAFT_03_6022.md
@@ -1,0 +1,20 @@
+# PR draft: #6022
+
+## Title
+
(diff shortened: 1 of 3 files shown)


allenxu09:
@codex review

Issue #6017, opened 2026-08-02
[Bug] 代理接管（异常退出恢复路径）重写 config.toml 时不合并「应用通用配置」，导致 sandbox_mode/web_search 等设置丢失


GengchenXU:
**环境**：CC Switch v3.19.1 · macOS · Codex 桌面应用（Rust 版 config.toml）

**现象**：开启本地路由（代理接管）+ 供应商勾选「应用通用配置」后，若发生异常退出再启动，config.toml 会被从旧的 Live 备份重写，**通用配置片段没有被合并进去**，导致 sandbox_mode、web_search、[sandbox_workspace_write]、[mcp_servers] 等与供应商无关的设置丢失/回到默认。

**复现步骤**：
1. 设置 → 开启本地路由，Codex 进入接管（config.toml 指向 http://127.0.0.1:15721/v1）
2. Codex 供应商勾选「应用通用配置」，片段里含 sandbox_mode / web_search / [sandbox_workspace_write]
3. 让 cc-switch 异常退出（留下接管残留），再次启动
4. 日志出现「检测到上次异常退出（存在接管残留），正在恢复 Live 配置…」→「codex Live 配置已从备份恢复」→「正在重新接管并补齐 Live」
5. 结果：config.toml 只包含供应商 config + 接管字段，通用配置片段未合并

**预期**：接管重写时同样合并「应用通用配置」片段（开关已开启）
**实际**：未合并，无关配置被丢弃

**日志证据**：
[2026-08-02][13:41:27][WARN] 检测到上次异常退出（存在接管残留），正在恢复 Live 配置...
[2026-08-02][13:41:27][INFO] codex Live 配置已从备份恢复
[2026-08-02][13:41:27][WARN] codex 标记为已接管，但 backup=false live_matches_current_proxy=false，正在重新接管并补齐 Live
[2026-08-02][13:41:28][INFO] Codex Live 配置已接管，代理地址: http://127.0.0.1:15721/v1

**备注**：v3.12.2 曾修复 "Common Config loss during proxy takeover"，此问题疑似该场景在「异常退出恢复 / 重新接管」路径上的回归或遗漏。

crazyzhang277:
Additional reproduction from Windows

Environment:
- CC Switch v3.19.2
- Windows
- Codex Desktop
- Codex local proxy takeover enabled
- Custom Codex provider
- MCP servers configured through CC Switch MCP management:
  - `codex-agy-bridge`
  - `codegraph`

[...]

PR #6168, opened 2026-08-05
fix(recovery): merge common config into the existing Live backup

KDB-Wind:
# PR draft: #6017

## Title

fix(recovery): merge common config into the existing Live backup

## Summary

Crash recovery now applies enabled common configuration to the saved Live snapshot instead of rebuilding the file from the provider record. The backup's authentication, model, MCP, and other user settings remain intact. If the optional snippet is malformed, or the current provider cannot be found, recovery keeps the original backup and logs the reason.

## Validation

- `cargo test --manifest-path src-tauri/Cargo.toml --lib crash_restore_merges_common_config_into_existing_codex_backup`

### DIFF
--- a/PR_DRAFT_05_6017.md
+++ b/PR_DRAFT_05_6017.md
@@ -1,0 +1,17 @@
+# PR draft: #6017
+
+## Title
+
+fix(recovery): merge common config into the existing Live backup
+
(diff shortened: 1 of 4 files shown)


farion1231:
@codex review

PR #6187, opened 2026-08-06
fix(provider): make editing a provider reach the live config file

BingZi-233:
## Summary

Editing the currently-active provider could leave the live config file at its pre-edit value while the UI reported success — a changed request base URL never took effect. This PR fixes the three independent paths that caused it, plus a related frontend bug where ProviderForm's config editors overwrote each other.

## Root causes & fixes

**1. Live ownership decision (backend)** — The gate `if has_live_backup || live_taken_over` was duplicated in three places (`ProviderService::update`, `sync_current_provider_for_app`, `sync_current_provider_for_app_respecting_takeover`) and treated the mere presence of a `proxy_live_backup` row as proof that proxy takeover owns the live file. That row survives a crash or a failed restore, and once it lingers, every later save updates only the backup — permanently, and without any warning.

- Collapse the three copies into `sync_live_for_provider_respecting_takeover`.
- Require takeover to actually be in effect: a live placeholder, a committed `proxy_config.enabled` flag, or a backup row **while the proxy process is running** (the activation window, since `set_takeover_for_app` starts the proxy before it stamps the live file).
- Per-app ownership: replace `ProxyService::is_running()` (a single global flag) with `SwitchLockManager::is_locked_for_app`. Previously, with takeover active for app A, a stale backup row for app B was still read as "takeover owns B's live file", and saving a B provider went back to writing only the backup.
- A stale row with the proxy stopped now writes the live file and refreshes the leftover backup, so a later restore cannot resurrect the old URL.
- Add the missing GrokBuild arm to the takeover refresh, mirroring the switch path.
- Log the previously-swallowed errors when the takeover flag / live-backup row cannot be read.

**2. Universal providers (backend)** — `sync_universal_to_apps` only wrote DB rows, so a unified provider that was the active one for Claude/Codex/Gemini never reached any live file. Now re-project the live config when the generated child is the effective current provider. Per-app failures are collected and reported (naming the apps whose file was not written) instead of logging a failed projection and still returning `Ok(true)` — one app cannot abort the other two.

**3. Form editors overwriting each other (frontend)** — In JsonEditor, the effect that pushes the `value` prop into CodeMirror dispatches a full document replacement; that dispatch counts as `docChanged`, and the updateListener echoed it straight back through `onChange`. Two editor instances holding different documents echoed each other forever — opening the edit dialog flickered endlessly, oscillating between the DB-stored config and the live-file config.

- Mark the sync transaction with an `ExternalSync` annotation and skip it in the listener. Use `every`, not `some`: CodeMirror coalesces transactions into one ViewUpdate, so a keystroke batched with a sync dispatch must still be reported.
- Read `onChange` through a ref — the editor-creation effect deliberately excludes `onChange` from its deps and would otherwise freeze the mount-time closure.
- With the echo fixed, the forced re-render is gone. Snapshot staleness is instead solved by reading the current value at event time: ProviderForm exposes `getSettingsConfig()`, threaded through `useBaseUrlState`, `useApiKeyState`, `useModelState`, `useTemplateValues` and `useCommonConfigSnippet`, each of which previously rebuilt the whole JSON from a render-time snapshot and clobbered whatever else had been written since.

## Tests

- Two Rust regression tests (proxy never enabled; stale backup row).
- Two frontend tests for the snapshot invariant (`JsonEditorExternalSync.test.tsx`, `ProviderFormSettingsConfigSnapshot.test.tsx`). The pre-existing edit-writes-live tests all forced takeover on and started the proxy, so deleting the normal branch entirely still passed.


farion1231:
@codex review

PR #6198, opened 2026-08-06
fix(grokbuild): restore provider form background

SaladDay:
## Summary

- apply the shared glass form container styles to the Grok Build provider form
- match the background, border, radius, and padding used by the other provider forms

## Verification

- `pnpm exec prettier --check src/components/providers/forms/GrokBuildProviderForm.tsx`
- `pnpm typecheck`
- `pnpm build:renderer`

### DIFF
--- a/src/components/providers/forms/GrokBuildProviderForm.tsx
+++ b/src/components/providers/forms/GrokBuildProviderForm.tsx
@@ -435,7 +435,7 @@ export function GrokBuildProviderForm({
       <form
         id="provider-form"
         onSubmit={form.handleSubmit(handleSubmit)}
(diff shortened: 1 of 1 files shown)


SaladDay:
BEFORE：
image
AFTER：
image

farion1231:
@codex review

farion1231:
Thanks for the fix — and for the before/after screenshots, they made this a quick review. Verified the class string matches the generic ProviderForm exactly, so all provider forms now share the same card styling.

Small note for a possible follow-up: ClaudeDesktopProviderForm has the same missing container, so the same one-liner there would be welcome if you're up for it. Merging!

PR #6199, opened 2026-08-06
fix(ui): reduce provider editor empty space

SaladDay:
## Summary

- reduce the default JSON editor minimum height from 12 rows to 3
- use the compact minimum for inline Claude, Codex, Gemini, Grok, OpenCode, OpenClaw, Hermes, and OMO provider editors
- keep full-screen, fill-height, script, and fixed-height preview editors unchanged

## Cause

`JsonEditor` converts `rows` into a CSS `min-height`. Inline provider editors requested 6-14 rows, so short three-line configurations retained a large empty area below their content.

## Validation

- `pnpm exec prettier --check` on changed files
- `pnpm typecheck`
- `pnpm build:renderer`

No tests added; this only changes the minimum visible height of compact editor instances.

### DIFF
--- a/src/components/JsonEditor.tsx
+++ b/src/components/JsonEditor.tsx
@@ -29,7 +29,7 @@ const JsonEditor: React.FC<JsonEditorProps> = ({
   onChange,
   placeholder: placeholderText = "",
   darkMode = false,
-  rows = 12,
+  rows = 3,
   showValidation = true,
   language = "json",
   height,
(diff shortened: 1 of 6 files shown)


SaladDay:
BEFORE：
image

AFTER：
image

farion1231:
@codex review

farion1231:
Thanks for the fix! Verified locally: since `rows` only sets the editor's min-height and the editor grows with its content, this trims the dead space under short configs without affecting larger ones — the before/after screenshots show it nicely. Also confirmed all existing `JsonEditor` call sites pass explicit `rows`/`height`, so the default change is safe. Merging.

Issue #5366, opened 2026-07-14
openclaw里面切换gpt-5.6-sol为gpt-5.6-terra的时候无效,无法再这三个智剪切换


myjr2015:
### Self Checks / 自检

- [x] I have read the [FAQ](https://github.com/farion1231/cc-switch#faq) section in README.
我已阅读 README 中的[常见问题](https://github.com/farion1231/cc-switch#常见问题)。

- [x] I have searched for [existing issues](https://github.com/farion1231/cc-switch/issues), including closed ones.
我已搜索过[已有的 Issue](https://github.com/farion1231/cc-switch/issues)，包括已关闭的。

### CC Switch Version / 版本号

3.17.0

### Operating System / 操作系统

Windows

### Related App / 涉及应用

OpenClaw

### Steps to Reproduce / 重现步骤

# Bug: OpenClaw default-model button stays disabled after editing the active provider's model ID

## Suggested issue title

`[Bug] OpenClaw "Set as default" remains disabled after changing the current provider model ID`

## Environment

- CC Switch: 3.17.0
- OpenClaw: 2026.7.1 (`2d2ddc4`)
- Node.js: 24.18.0
- OS: Windows 10.0.26200.0 x64
- OpenClaw installation: global npm package
- Provider type: custom OpenAI Responses-compatible provider

No API endpoint, API key, or account identifier is required to reproduce this issue.

## Summary

When an OpenClaw provider is already the default and its configured model ID is changed, CC Switch updates the provider model catalog but does not update OpenClaw's default model. The provider's **Set as default** button remains disabled, so the user cannot directly apply the newly configured model as the default.

The button becomes enabled only after setting a different provider/model as default first. The user can then return and set the original provider's new model as default.

## Steps to reproduce

1. Open the **OpenClaw** page in CC Switch.
2. Configure a provider with model ID `gpt-5.6-luna`.
3. Set `provider-id/gpt-5.6-luna` as the OpenClaw default model.
4. Open the same provider's **Configure** dialog.
5. Change its model ID from `gpt-5.6-luna` to `gpt-5.6-terra` and save.
6. Return to the OpenClaw provider/model page.
7. Observe the provider's **Set as default** button.
8. Restart OpenClaw or inspect its configuration and actual model requests.

## Actual behavior

- The provider model catalog is updated to `gpt-5.6-terra`.
- OpenClaw's generated agent model registry is also updated to `gpt-5.6-terra`.
[...]

PR #6201, opened 2026-08-06
refactor(openclaw): align provider model editor with family UI

SaladDay:
## Summary

- align the OpenClaw provider model editor with the flat, expandable model-row pattern used across CC Switch
- reuse the shared model picker and keep native reasoning, input, token-limit, and cost fields in progressive details
- remove inferred primary/fallback badges from provider model rows
- select an exact model when setting an OpenClaw provider as default while preserving the existing fallback chain and unknown default-model fields

## Scope

Frontend and existing OpenClaw default-model flow only. No backend or database schema changes.

## Validation

- pnpm typecheck
- focused OpenClaw tests: 34 passed
- full frontend suite: 102 test files, 700 tests passed


SaladDay:
BEFORE：
image
AFTER：
image

SaladDay:
A note on removing the Default Model / Fallback Model labels from the provider editor:

OpenClaw treats a provider models array as a model catalog. The authoritative default model and fallback chain are stored separately in agents.defaults.model.primary and agents.defaults.model.fallbacks; the order of models within a provider does not define either role.

The previous UI inferred that the first model was the default and every remaining model was a fallback. That could show state that disagreed with OpenClaw, make harmless model reordering appear semantic, and introduce a second source of truth.

This PR removes only those inferred labels and ordering semantics, not OpenClaw default-model support. The provider form now manages the provider model catalog only. Set as Default still works from the provider card: for providers with multiple models, CC Switch asks for the exact model, updates the authoritative primary value, preserves the existing fallback chain and unknown fields, and does not automatically turn the other provider models into fallbacks.

farion1231:
@codex review

farion1231:
Merged — thanks! The write path is the real win here: "Set as default" used to fabricate an entire fallback chain and overwrite hand-edited `agents.defaults` wholesale, and the read-modify-write rework (with tests pinning the preserved fields) turns that into a safe operation. Also nice that your rewrite had already adopted the shared `ModelDropdown` — when #6353 landed on main it collided in `OpenClawFormFields.tsx`, but both sides had converged on identical code, so I merged main into your branch and the resolution was simply keeping your version.

Two follow-ups if you're interested (no obligation):

1. **#5366 quick win**: the provider that's already default still can't switch its model from this menu (the `!isDefaultModel` gate). Dropping the gate and marking the current default with a checkmark would close that issue — just take care not to disturb the Hermes branch that shares this menu code.
2. **Stale `agents.defaults` after provider deletion**: keeping cross-provider fallbacks on switch is the right call (better than wiping user config each time), but deleting a provider still leaves its entries behind — cleanup-on-delete would fix the root cause.

Heads-up we'll note in the release notes: setting a default no longer auto-generates a fallback chain, and upgraders may still carry a previously generated one.

PR #6203, opened 2026-08-06
fix(ui): unify checkbox styles

SaladDay:
## Summary

- render the shared Checkbox component with the same native checkbox control used by the provider config options
- unify the 1M support controls and other shared checkbox usages with the Hide AI Attribution style
- preserve controlled, disabled, and indeterminate behavior

## Validation

- pnpm exec prettier --check src/components/ui/checkbox.tsx
- pnpm typecheck
- pnpm vitest run tests/components/CommonConfigEditor.test.tsx tests/components/McpFormModal.test.tsx tests/components/ClaudeFormFields.test.tsx tests/components/SessionManagerPage.test.tsx
- visually reviewed in the macOS app before commit

### DIFF
--- a/src/components/ui/checkbox.tsx
+++ b/src/components/ui/checkbox.tsx
@@ -1,33 +1,62 @@
 import * as React from "react";
-import * as CheckboxPrimitive from "@radix-ui/react-checkbox";
-import { Check, Minus } from "lucide-react";
 
 import { cn } from "@/lib/utils";
 
-const Checkbox = React.forwardRef<
-  React.ElementRef<typeof CheckboxPrimitive.Root>,
-  React.ComponentPropsWithoutRef<typeof CheckboxPrimitive.Root>
->(({ className, ...props }, ref) => (
-  <CheckboxPrimitive.Root
-    ref={ref}
-    className={cn(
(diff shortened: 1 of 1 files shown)


farion1231:
@codex review

farion1231:
Merged — thanks! Unifying all the shared checkboxes onto the same native control the provider options already use is a nice consistency win, and keeping the controlled / disabled / indeterminate contract intact while swapping the internals made this an easy review.

A few small follow-ups if you're interested (no obligation):

1. **Three utility groups got lost in the style swap**: the old implementation had `shrink-0` (prevents the box from being squeezed in flex rows with long labels), `disabled:cursor-not-allowed disabled:opacity-50` (right now a disabled checkbox looks identical to an enabled one), and used `focus-visible:` rather than `focus:` (ring only on keyboard navigation, not mouse clicks). Re-adding these three to the class string restores parity.
2. **Controlled `indeterminate` can drift on click**: the browser natively clears `HTMLInputElement.indeterminate` on activation, and React only restores `checked` — so if a consumer keeps `checked="indeterminate"` after the change callback (or sets the same value, which bails out of re-render), the DOM ends up showing an unchecked box while `aria-checked="mixed"` still announces "partially checked". All current call sites are safe (the session manager group checkbox always transitions state), so this is contract hardening rather than a live bug. Three-line fix at the end of the onChange handler: `if (isIndeterminate) event.target.indeterminate = true;` — plus a small component test to pin it.
3. `@radix-ui/react-checkbox` is now unused and can be dropped from package.json.

PR #6232, opened 2026-08-07
fix(windows): fall back when WSL rejects atomic replace

allenxu09:
## Summary

- fall back to `std::fs::rename` when WSL UNC paths reject `ReplaceFileW` with `ERROR_NOT_SUPPORTED`
- preserve the existing `ReplaceFileW` behavior for local Windows filesystems and all other errors
- reuse the existing temporary-file cleanup and retry handling without deleting the destination first

## Root cause

v3.19.2 changed Windows atomic writes to use `ReplaceFileW`. The WSL UNC filesystem returns OS error 50 for that API, and the code only entered its rename fallback for `NotFound`, so Codex and Claude configuration writes failed before the temporary file could be installed.

Fixes #6224.
Also addresses #6219 and #6188 and #6247.

## Validation

- `cargo fmt --manifest-path src-tauri/Cargo.toml -- --check`
- `git diff --check upstream/main...HEAD`
- focused static review of the final branch diff: no findings

Automated tests were not run. Reproducing OS error 50 requires a Windows host with a configured WSL UNC path.

Issue #6224, opened 2026-08-07
更新3.19.2 版本后 wsl 切换失败


avyou:
### Self Checks / 自检

- [x] I have read the [FAQ](https://github.com/farion1231/cc-switch#faq) section in README.
我已阅读 README 中的[常见问题](https://github.com/farion1231/cc-switch#常见问题)。

- [x] I have searched for [existing issues](https://github.com/farion1231/cc-switch/issues), including closed ones.
我已搜索过[已有的 Issue](https://github.com/farion1231/cc-switch/issues)，包括已关闭的。

### Related App / 涉及应用

Codex

### Your Question / 你的问题

更新3.19.2 版本后，在windows 上使用cc switch 切换到wsl 路径失败，求解决：

切换失败：热切换失败: 写入 Codex 配置失败: 原子替换失败: \\wsl.localhost\Ubuntu-24.04\home\xxxx\.codex\cc-switch-model-catalog.json.tmp.50344.1786083969121366900.21 -> \\wsl.localhost\Ubuntu-24.04\home\xxx\.codex\cc-switch-model-catalog.json: 不支持该请求。 (os error 50)

RebornForPower:
+1 我在用Claude也遇到这个问题了

dpp814:
+1啊 原子替换失败: \\wsl.localhost\Ubuntu\home\bihu\.claude\settings.json.tmp.29368.1786084725157187600.1 -> \\wsl.localhost\Ubuntu\home\bihu\.claude\settings.json: 不支持该请求。 (os error 50)

lijiaqi-reolink:
+1 同样，在修改过配置后就无法替换WSL文件了

QingshuangZhu:
同样的问题

allenxu09:
可以关注 pr #6232.

zhongjun1018-hue:
+1

blsnt8586:
+1

Waveshare-wulu:
Thanks for the fix! I saw that #6232 has already been merged.

Do you have an estimate for when a new release including this fix will be published? I'm currently affected by this issue when using CC Switch with Codex in WSL.

Thanks!

PR #6233, opened 2026-08-07
ci: run Windows backend tests against WSL2 filesystems

allenxu09:
## Summary

- add a Windows 2025 CI job that installs Ubuntu 24.04 under WSL2 while keeping Rust and cc-switch tests running as native Windows processes
- point `CC_SWITCH_TEST_HOME`, `TEMP`, and `TMP` at directories under a real `\\wsl.localhost\...` root
- run the complete existing backend test suite serially against that WSL2-backed filesystem
- add an atomic-write contract that replaces an existing file, verifies its contents, and checks that no temporary files remain
- explicitly verify that the WSL test root, resolved cc-switch home, and Windows temporary directory are all UNC paths under the configured WSL2 root

## Why

The existing OS matrix tests native Linux, Windows, and macOS filesystems, but it does not test a Windows cc-switch process writing into a WSL2 filesystem. That gap allowed the `ReplaceFileW` regression reported in #6224 to reach v3.19.2.

This job also covers future tests without requiring authors to write WSL-specific copies. Tests using `tempfile`, `std::env::temp_dir()`, or cc-switch's test-home override automatically execute once on NTFS in the normal Windows job and once on the WSL2 filesystem in this job.

## Regression coverage

The dedicated ignored contract creates a unique child directory on the WSL2 filesystem, writes an existing file, calls `atomic_write`, verifies replacement, and checks cleanup. CI confirms the exact ignored test is discovered before running it so a renamed or missing test cannot silently pass.

## Dependency

This PR is ready for review but depends on #6232. Against the current unpatched `main`, the WSL2 atomic-write contract is expected to fail with Windows error 50. After #6232 lands and this branch is rebased, this job becomes the regression gate for that behavior.

## Validation

- `pnpm exec prettier --check .github/workflows/ci.yml`
- `git diff --check`
- `cargo check --tests --manifest-path src-tauri/Cargo.toml`
- `open-code-review`: 0 findings

The revised WSL2 job itself is validated by GitHub Actions rather than the local Windows checkout.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -105,3 +105,81 @@ jobs:
 
       - name: Run tests
         run: cargo test --manifest-path src-tauri/Cargo.toml
+
+  backend-windows-wsl2:
+    name: Backend Checks (Windows + WSL2 home)
+    runs-on: windows-2025
+    timeout-minutes: 45
+    permissions:
+      contents: read
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v6
+
+      - name: Setup Rust
+        uses: dtolnay/rust-toolchain@4360b52568e2003a75bf9bc1d59f33a8e3fc893c # stable
+        with:
+          components: rustfmt, clippy
+
+      - name: Setup WSL2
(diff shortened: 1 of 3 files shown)


farion1231:
@codex review

farion1231 (approved):

PR #6239, opened 2026-08-07
feat(provider): add PPIO support

### LINKED ISSUE #6226: feat: add PPIO provider support
## Summary

Add first-class PPIO LLM provider presets across the applications supported by CC Switch.

## Scope

- Configure native Anthropic endpoints for Claude Code and Claude Desktop.
- Configure OpenAI-compatible Chat Completions endpoints for Codex, OpenCode, OpenClaw, and Hermes.
- Preset only `deepseek/deepseek-v4-flash-0731` on every application surface.
- Add PPIO provider metadata and its brand icon.
- Add unit coverage for authentication, endpoint, model, reasoning, default-model, and icon settings.

## Validation

- Verify the PPIO model-list endpoint with a real API key.
- Verify OpenAI Chat Completions and Anthropic Messages requests.
- Run the repository type check, formatting check, and complete unit-test suite.

hu-miao:
## Summary / 概述

- Add first-class PPIO provider presets for Claude Code, Claude Desktop, Codex, OpenCode, OpenClaw, and Hermes.
- Configure PPIO Anthropic and OpenAI-compatible endpoints.
- Preset `deepseek/deepseek-v4-flash-0731` with verified model metadata.
- Register the PPIO brand icon and add focused regression coverage.

## Related Issue / 关联 Issue

Fixes #6226

## Screenshots / 截图

| ----------------- | --------------- |
| N/A | N/A |

No visual UI changes.

## Validation

- [x] `pnpm typecheck`
- [x] `pnpm format:check`
- [x] `pnpm test:unit` (714 tests)
- [x] `cargo fmt --check`
- [x] `cargo clippy -- -D warnings`
- [x] PPIO model-list endpoint verified with a real API key.
- [x] OpenAI Chat Completions and Anthropic Messages requests verified with a real API key.

## Checklist / 检查清单

- [x] Updated i18n files if user-facing text changed / 如修改了用户可见文本，已更新国际化文件 (not applicable; no user-facing text changed)


farion1231:
@codex review

farion1231 (commented):
Thanks for the thorough preset work — I cross-checked the metadata against PPIO's
public `/openai/v1/models` endpoint (context window 1048576, max output 393216,
text-only modalities, pricing ratios) and it all matches, and the Anthropic base
URL + `ANTHROPIC_AUTH_TOKEN` setup matches PPIO's official Claude Code guide.

**[P2] The tests pin PPIO to the tail of every preset array**

[These assertions](https://github.com/farion1231/cc-switch/blob/ca237d7cee8899c7ae64a4874e32786b3201544c/src/config/ppioProviderPresets.test.ts#L47-L53)
require PPIO to be the last entry of all six preset arrays, and
[the icon test](https://github.com/farion1231/cc-switch/blob/ca237d7cee8899c7ae64a4874e32786b3201544c/src/config/ppioProviderPresets.test.ts#L209-L210)
requires `ppio` to be the last key of both icon registries.

File position carries no meaning for non-sponsor presets — the selector sorts
them by display name at render time (`sortPresetEntries` in
`ProviderPresetSelector.tsx`) — and the convention for new entries is to append
at the end. So the next PR that adds any provider or icon will break this test
file for unrelated reasons. The existing `grokBuildProviderPresets.test.ts`
deliberately avoids position assertions for the same reason.

Minimal fix: drop the `"%s places PPIO at the end of the preset list"` `it.each`
block and the two `slice(-1)` assertions on `icons` / `iconMetadata`, keeping
the "registers exactly one" checks and all the field-level assertions (those are
great).

Happy to take another look once the ordering assertions are dropped.

hu-miao:
> Thanks for the thorough preset work — I cross-checked the metadata against PPIO's public `/openai/v1/models` endpoint (context window 1048576, max output 393216, text-only modalities, pricing ratios) and it all matches, and the Anthropic base URL + `ANTHROPIC_AUTH_TOKEN` setup matches PPIO's official Claude Code guide.
> 
> **[P2] The tests pin PPIO to the tail of every preset array**
> 
> [These assertions](https://github.com/farion1231/cc-switch/blob/ca237d7cee8899c7ae64a4874e32786b3201544c/src/config/ppioProviderPresets.test.ts#L47-L53) require PPIO to be the last entry of all six preset arrays, and [the icon test](https://github.com/farion1231/cc-switch/blob/ca237d7cee8899c7ae64a4874e32786b3201544c/src/config/ppioProviderPresets.test.ts#L209-L210) requires `ppio` to be the last key of both icon registries.
> 
> File position carries no meaning for non-sponsor presets — the selector sorts them by display name at render time (`sortPresetEntries` in `ProviderPresetSelector.tsx`) — and the convention for new entries is to append at the end. So the next PR that adds any provider or icon will break this test file for unrelated reasons. The existing `grokBuildProviderPresets.test.ts` deliberately avoids position assertions for the same reason.
> 
> Minimal fix: drop the `"%s places PPIO at the end of the preset list"` `it.each` block and the two `slice(-1)` assertions on `icons` / `iconMetadata`, keeping the "registers exactly one" checks and all the field-level assertions (those are great).
> 
> Happy to take another look once the ordering assertions are dropped.

@farion1231 Thanks for the feedback, and I've removed these ordering assertions, please take another look.

farion1231:
Merged — thanks @hu-miao for the solid, well-verified preset work and the quick turnaround on the ordering assertions. Welcome aboard, PPIO! 🎉

PR #6243, opened 2026-08-07
tools(codex): add model picker fix utility for custom providers

ly2601620664:
### Summary

Adds a workaround utility for Codex Desktop's empty model picker when using CC Switch custom providers.

**Root cause:** Codex Desktop applies a Statsig allowlist (107580212: use_hidden_models=true, vailable_models only contains stock GPT ids) in the renderer, which strips every custom-provider model returned by model/list. See openai/codex#19694 and cc-switch#5797.

**What this adds:** 	ools/codex-model-picker-fix/ with:
- ix-codex-model-picker.ps1 - one-click fix: backs up the Desktop Local Storage LevelDB, sets use_hidden_models=false, adds the custom model IDs to vailable_models, pushes the cache timestamp forward, and blocks b.chatgpt.com in hosts so the remote allowlist cannot refresh back.
- patch-statsig.js - the LevelDB patcher (validated on Windows with Codex Desktop 26.803).
- README with usage/rollback instructions.

This is complementary to the code-level fix in #5265: it works with the current shipped Desktop build without repackaging the MSIX.

### DIFF
--- a/tools/codex-model-picker-fix/.gitignore
+++ b/tools/codex-model-picker-fix/.gitignore
@@ -1,0 +1,3 @@
+node_modules/
+*.bak-*/
+package-lock.json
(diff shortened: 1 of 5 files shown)


farion1231:
@codex review

PR #6246, opened 2026-08-07
fix(usage): prevent Codex session sync gaps after sync-pass refactor

puppnn:
## Summary

- persist each processed Codex rollout file size alongside mtime and line offset, so appended usage is still synced when Windows leaves an open JSONL file's mtime unchanged
- allow a child fork to sync after its parent has emitted an explicit terminal lifecycle event, while continuing to defer children of active or incomplete parents
- integrate the size-aware cursor with the current `CodexSyncPass` preload and batched insert transactions, keeping the final data batch and cursor advance atomic
- add the v16 to v17 schema migration and reset only Codex rollout cursors once so existing installations establish a consistent `(mtime, offset, size)` state

## Why a new PR

This supersedes #5997. Since that PR was opened, `main` gained the batched Codex insert and preloaded-cursor refactor in `baf07a2701`. This branch is rebuilt on the current `main` and preserves those performance and crash-consistency guarantees instead of layering the old implementation over them.

## Behavior and safety

- unchanged: token delta calculation, replay-prefix deduplication, request IDs, pricing, proxy request ingestion, and non-Codex session sync
- same-mtime files are skipped only when their size is also unchanged
- only explicit `task_complete`, `turn_complete`, or `turn_aborted` events make a parent timeline terminal
- active parents remain deferred until sufficient history is available
- SQLite lookup failures are propagated; only a genuinely missing cursor defaults to zero
- migration resets Codex rollout cursors only; Gemini and other session cursors are preserved

## Validation

- `cargo fmt -- --check`
- `cargo clippy --locked --lib -- -D warnings`
- `cargo test --locked session_usage_codex --lib` (48 passed, 1 ignored)
- `cargo test --locked migrate_v16_to_v17_adds_session_file_size_cursor --lib`
- `cargo test --locked sync_state_lookup_defaults_only_when_row_is_missing --lib`
- `cargo test --locked --lib` (2342 passed, 5 ignored)

### DIFF
--- a/src-tauri/src/database/mod.rs
+++ b/src-tauri/src/database/mod.rs
@@ -53,7 +53,7 @@ use std::sync::Mutex;
 
 /// 当前 Schema 版本号
 /// 每次修改表结构时递增，并在 schema.rs 中添加相应的迁移逻辑
-pub(crate) const SCHEMA_VERSION: i32 = 16;
+pub(crate) const SCHEMA_VERSION: i32 = 17;
 
 /// 安全地序列化 JSON，避免 unwrap panic
 pub(crate) fn to_json_string<T: Serialize>(value: &T) -> Result<String, AppError> {
--- a/src-tauri/src/database/schema.rs
+++ b/src-tauri/src/database/schema.rs
@@ -1537,6 +1537,10 @@ impl Database {
                 "last_file_size",
                 "INTEGER NOT NULL DEFAULT 0",
             )?;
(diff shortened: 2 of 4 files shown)


allenxu09:
@codex review

farion1231:
Thanks for digging into this one. The same root cause has now shipped in v3.20.3 via PR #7219 (persisted byte cursor + mtime/size gate), which was rebased onto the current schema; this PR conflicts with that change and is superseded, so I'm closing it. Your analysis helped confirm the fix direction — much appreciated.

PR #6252, opened 2026-08-07
fix(windows): eliminate startup white-black flash (FOUC)

### LINKED ISSUE #6182: windows X64便携版，应用启动时出现闪屏，有突然的白黑闪屏
### Self Checks / 自检

### CC Switch Version / 版本号

3.19.1

### Operating System / 操作系统

Windows

### Related App / 涉及应用

Other / 其他

### Steps to Reproduce / 重现步骤

https://github.com/user-attachments/assets/0f5d6968-8ecc-4af9-ad74-a81626c20484

https://github.com/user-attachments/assets/dde5a045-06eb-4cd4-aa5b-5da4c129e164

### Expected Behavior / 期望行为

我期望不出现闪屏情况

### Actual Behavior / 实际行为

已安装EdgeWebView151.0.4129.59版本，启动闪屏

### Additional Context / 补充信息

_No response_

SaladDay:
Fixes #6182

### DIFF
--- a/src-tauri/src/lib.rs
+++ b/src-tauri/src/lib.rs
@@ -70,6 +70,8 @@ pub use store::AppState;
 use tauri_plugin_deep_link::DeepLinkExt;
 use tauri_plugin_dialog::{DialogExt, MessageDialogButtons, MessageDialogKind};
 
+#[cfg(target_os = "windows")]
+use std::sync::atomic::{AtomicBool, Ordering};
 use std::{fmt, sync::Arc};
 #[cfg(target_os = "macos")]
 use tauri::image::Image;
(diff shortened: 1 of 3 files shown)


farion1231:
@codex review

SaladDay:
Addressed in e0206ba8. @farion1231, PTAL.

farion1231:
Merged — thanks @SaladDay! Great root-cause writeup on #6182, and switching from the backgroundColor approach to deferred show after the bot feedback was exactly the right call (the Windows override's `app.windows` array replacement would have swallowed that property anyway).

Verified before merge: the CSP hash matches the inline script byte-for-byte (with tauri-codegen's build-time auto-hashing as a second safety net), `tauri.windows.conf.json` keeps `visible: false`, silent startup is respected on both paths, and the tray "Open main window" entry remains an unconditional fallback if a broken WebView2 never fires `Finished`.

Two optional follow-ups, neither urgent:
1. The inline script treats any unexpected stored value as light, while ThemeProvider falls back to system. Aligning them is actually a shorter script — `t==="dark"||(t!=="light"&&matchMedia(...).matches)` — just remember to regenerate the CSP hash.
2. Longer term, a frontend "ready" signal (after first React commit + native title-bar theme sync) could close the remaining few-ms gap where a themed-but-empty page shows. It would need its own timeout fallback so a failed signal can't leave the window hidden.

Once the next build is out, a quick confirmation on a real Windows machine (ideally the reporter's scenario) would be great to rule out any residual single-frame flash.

bigmanBass666:
我的眼睛再也不会被闪瞎了.

PR #6253, opened 2026-08-07
fix(codex): declare 1M context for DeepSeek Responses providers with empty modelCatalog

### LINKED ISSUE #6192: Codex app使用cc switch的deepseek v4 flash通用Responses配置异常导致实际费用翻倍
### Self Checks / 自检

### CC Switch Version / 版本号

v3.19.2

### Operating System / 操作系统

Windows

### Related App / 涉及应用

Codex

### Steps to Reproduce / 重现步骤

cc switch对deepseek v4 flash的通用配置存在重大bug。测试使用opencode go套餐：
1. cc switch 上游格式选择Chat Completions 协议，消耗的金额比opencode中多十余倍，会出无法显示模型名称、现对话中断且上下文长度128k。
2. cc switch 上游格式选择Responses协议，消耗的金额是opencode中的2倍，会出现无法显示模型名称、对话卡顿且上下文长度128k。
3. 按照官网配置，消耗的金额和opencode中相同，上下文长度恢复到1m。
 https://api-docs.deepseek.com/zh-cn/quick_start/agent_integrations/codex

### Expected Behavior / 期望行为

修改cc switch现有Responses协议，按照deepseek官方配置

### Actual Behavior / 实际行为

_No response_

### Additional Context / 补充信息

_No response_

SaladDay:
## Problem

DeepSeek official native Responses providers with an empty `modelCatalog` fall back to Codex's 128k default context window. That can trigger avoidable mid-session compaction instead of using DeepSeek's published 1M catalog entry.

This is related to #6192, but it does not cover the reporter's third-party OpenCode Go gateway or the Chat-protocol case.

## Fix

When the stored `modelCatalog` is empty, synthesize one spec from the active model. The existing official-catalog path then mirrors DeepSeek's published entry, including its 1M context window and tool contract.

The host check parses the URL and matches only `deepseek.com` or its subdomains. Active profile model/provider overrides are respected without leaking the previous provider's URL.

## Scope

One file: `src-tauri/src/codex_config.rs`.

Non-DeepSeek hosts, third-party gateways, ProxyChat, and Anthropic profiles keep their existing behavior. This PR does not change proxy routing or protocol conversion.

## Verification

- `cargo fmt --check`
- `cargo test --lib codex_config`: 115 passed, 0 failed

Refs #6192

### DIFF
--- a/src-tauri/src/codex_config.rs
+++ b/src-tauri/src/codex_config.rs
@@ -86,6 +86,29 @@ fn codex_top_level_model(config_text: &str) -> Option<String> {
         .filter(|value| !value.is_empty())
 }
 
+fn codex_active_profile_value<'a>(doc: &'a toml::Value, key: &str) -> Option<&'a str> {
+    let profile = doc.get("profile").and_then(|value| value.as_str())?;
+    doc.get("profiles")
+        .and_then(|profiles| profiles.get(profile))
+        .and_then(|profile| profile.get(key))
+        .and_then(|value| value.as_str())
+        .map(str::trim)
+        .filter(|value| !value.is_empty())
+}
+
+/// Model selected by the active profile, falling back to the top-level model.
+fn codex_active_model(config_text: &str) -> Option<String> {
+    let doc = config_text.parse::<toml::Value>().ok()?;
+    codex_active_profile_value(&doc, "model")
+        .or_else(|| {
+            doc.get("model")
+                .and_then(|value| value.as_str())
+                .map(str::trim)
+                .filter(|value| !value.is_empty())
+        })
+        .map(str::to_string)
+}
+
 /// Whether a native `/responses` provider's gateway is known to reject the Codex
 /// `web_search` hosted tool — by `base_url` host OR by the active model's brand
 /// (so an aggregator fronting a reject vendor's model is caught too). Driven by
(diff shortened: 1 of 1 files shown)


farion1231:
@codex review

SaladDay:
Addressed in a343f205. @farion1231, PTAL.

farion1231 (commented):
Thanks — this is a clean fix. It reuses the existing official-catalog mirror machinery end to end (no new subsystem, nothing on the forwarding path), and both hardenings in a343f205 are the right call: the hostname suffix match closes the `contains("deepseek.com")` spoof/path-embedding hole (with negative tests for both fake-host shapes, plus the absolute-FQDN case), and the profile-aware resolution means an official-docs-shaped config pasted into the generic template gets the grant — and, per the negative test, loses it when the active profile switches providers. The failure chain also checks out against #6192: empty catalog → no `model_catalog_json` → Codex's 128k default → mid-session auto-compact churns DeepSeek's disk-cache prefix.

Two items before merge:

**[P1] Won't compile against current main — #6228 added two fields to `CodexCatalogModelSpec`.**
main now has `reasoning_levels` / `default_reasoning_level` (merged after this branch was cut), so the struct literal in `synthesize_official_catalog_specs` fails with E0063 on a merge build:
https://github.com/farion1231/cc-switch/blob/a343f2056ec54aaac5345f234e65937b77640682/src-tauri/src/codex_config.rs#L1138-L1145

Minimal fix:
```rust
        base_instructions: None,
        reasoning_levels: None,
        default_reasoning_level: None,
    }]
```
`None` is the correct semantics here: on the vendor-mirror path the official entry's own `supported_reasoning_levels`/default survive untouched. Verified locally: with just these two lines added, a merge onto current main passes the full suite (`cargo test --lib`: 2561 passed, 0 failed), fmt/clippy clean. Note GitHub's MERGEABLE state and the green CI here are both computed against the pre-#6228 base, so neither catches this.

**[P2] `Closes #6192` likely over-claims — suggest `Refs #6192`, or confirm the reporter's `base_url` first.**
The reporter tested on the OpenCode Go plan (per the issue), i.e. a third-party gateway host. The deepseek.com host gate is intentionally narrow — the official catalog is a capability grant (freeform `apply_patch` + official harness), which a gateway hosting the same model doesn't necessarily honor — so this PR correctly does not fire for that setup, and the Chat-protocol 10x case (scenario 1 in the issue) is gated out by design as well. Merging as-is would auto-close the issue while the reporter's own configuration may still sit on the 128k default. Suggest downgrading the footer to `Refs #6192` and asking the reporter for their `base_url`; for gateway users the existing remedy is adding a modelCatalog row with contextWindow 1048576, which declares the window through the neutral template without the official capability grant. Whether to extend anything to known reseller hosts is a separate design decision — out of scope for this PR.

**nit:** the `display_name` backfill at
https://github.com/farion1231/cc-switch/blob/a343f2056ec54aaac5345f234e65937b77640682/src-tauri/src/codex_config.rs#L1125-L1134
is redundant — when the slug matches, the mirrored official entry already carries that exact `display_name`, and the override writes back the identical value; the unknown-slug path yields `None` anyway. Dropping it would make the "no explicit overrides" comment literally true.

**Follow-up (not this PR):** the web_search reject gate still resolves the base URL non-profile-aware (`extract_codex_base_url`), now asymmetric with the catalog gate for profile-based configs. Pre-existing on main; worth a small separate issue.

With the two-line P1 fix in (and the Closes footer settled either way), this is good to merge from my side — happy to re-review quickly.

SaladDay:
Rebased onto current main and addressed the current `CodexCatalogModelSpec` fields. I also tightened the active-profile URL fallback so selecting the same provider can keep its top-level URL, while switching providers cannot inherit the previous provider’s URL. The PR and commit now use `Refs #6192` because the reported third-party OpenCode Go / Chat paths remain out of scope. Verification: `cargo fmt --check`; `cargo test --lib codex_config` (115 passed).

PR #6272, opened 2026-08-08
fix(grokbuild): align PackyCode endpoint config

### LINKED ISSUE #6271: Grok Build PackyCode preset injects unsupported x_search and rejects models_base_url
### Self Checks / 自检

- [x] I have searched existing issues and pull requests, including closed ones, for PackyCode, Grok Build, x_search, models_base_url, and supports_backend_search.

### CC Switch Version / 版本号

3.19.2

### Operating System / 操作系统

macOS

### Related App / 涉及应用

Other / 其他 — Grok Build

### Steps to Reproduce / 重现步骤

1. Configure and activate the built-in Grok Build PackyCode provider with `grok-4.5` and the Responses backend.
2. Run a normal Grok Build prompt without explicitly disallowing `x_search`.
3. Observe that Packy returns HTTP 403 because the built-in Grok model capability enables the server-side `x_search` tool.
4. Update the provider TOML to Packy's current documented shape using `[endpoints].models_base_url` and `[model."grok-4.5"].supports_backend_search = false`.
5. Save the provider in CC Switch.

### Expected Behavior / 期望行为

- The PackyCode preset should use Packy's current `[endpoints].models_base_url` configuration.
- The PackyCode model should explicitly set `supports_backend_search = false` so Grok Build does not inject an unsupported hosted tool.
- Frontend and backend validation should accept either the current endpoint-level URL or the legacy model-level `base_url`, with the model-level value taking precedence when both exist.

### Actual Behavior / 实际行为

- The 3.19.2 PackyCode preset writes a model-level `base_url` and inherits backend-search support from Grok Build's built-in model.
- A normal request fails with `403 Forbidden: server-side tool "x_search" is disabled on this platform`.
- If the config is migrated to `[endpoints].models_base_url`, CC Switch rejects the save with `Grok Build configuration is missing a valid base_url field` because the Rust validation/extraction path only reads the model-level field.

### Additional Context / 补充信息

- `--disable-web-search` does not prevent this independently injected hosted tool.
- `--disallowed-tools x_search` succeeds with the same API key, endpoint, and model, confirming the credential and base inference route are valid.
- An isolated config using Packy's documented endpoint format also succeeds.
- Packy guide: https://docs.packyapi.ai/docs/cli/6-grok-build.html
- xAI X Search documentation: https://docs.x.ai/developers/tools/x-search
- A tested fix is ready and includes frontend, Rust validation/extraction, migration, proxy-takeover, and provider-update regression coverage.

LosEcher:
## Summary / 概述

- update the Grok Build PackyCode preset to use Packy's documented `[endpoints].models_base_url` format and explicitly disable unsupported backend search injection
- accept endpoint-level and legacy model-level base URLs across frontend and Rust validation/extraction, with model-level precedence for proxy takeover compatibility
- migrate existing PackyCode providers on save while preserving the selected endpoint and credential, with frontend and backend regression coverage

## Related Issue / 关联 Issue

Fixes #6271

## Root Cause / 根因

The existing PackyCode preset used a model-level `base_url` and inherited Grok Build's backend-search capability. Grok Build therefore injected the hosted `x_search` tool, which Packy does not support and rejects with HTTP 403. Packy's current Grok Build template instead uses `[endpoints].models_base_url`, but CC Switch's Rust provider validation previously required `[model.*].base_url` and rejected that documented shape.

## Testing / 测试

- `pnpm typecheck`
- `pnpm format:check`
- `pnpm test:unit` — 100 files, 699 tests passed
- `cargo fmt --check`
- `cargo clippy`
- full `cargo test` passed once locally — 2357 passed, 5 ignored, all integration test binaries passed
- final Rust rerun hit the existing macOS loopback timing test `proxy::hyper_client::tests::bytes_with_limit_aborts_reqwest_response_before_full_body_arrives`; rerunning the complete suite with only that test skipped passed — 2356 passed, 5 ignored, 1 filtered, all integration test binaries passed
- built and installed the app locally; saving an existing PackyCode provider produced the endpoint-level config and a real `grok-4.5` request succeeded without `--disallowed-tools x_search`

## Screenshots / 截图

N/A — configuration and provider-routing behavior only.

## Checklist / 检查清单

- [x] Updated i18n files if user-facing text changed / 未新增或修改用户可见文案，无需更新 i18n

### DIFF
--- a/src-tauri/src/grok_config.rs
+++ b/src-tauri/src/grok_config.rs
@@ -59,6 +59,20 @@ fn optional_non_empty_string(table: &toml::value::Table, key: &str) -> Option<St
         .map(ToString::to_string)
 }
 
+fn resolve_base_url(
+    root: &toml::value::Table,
+    selected_model: &toml::value::Table,
+) -> Option<String> {
+    optional_non_empty_string(selected_model, "base_url")
+        .or_else(|| {
+            root.get("endpoints")
+                .and_then(toml::Value::as_table)
+                .and_then(|endpoints| optional_non_empty_string(endpoints, "models_base_url"))
+        })
+        .map(|base_url| base_url.trim_end_matches('/').to_string())
+        .filter(|base_url| !base_url.is_empty())
+}
+
 /// Syntax-only validation for a Grok Build config document (empty allowed).
 ///
 /// 官方条目走 Grok CLI 自带的 xAI OAuth 登录，config.toml 不需要（通常也没有）
@@ -145,7 +159,13 @@ pub fn validate_config_toml(config_toml: &str) -> Result<(), AppError> {
         })?;
 
     required_non_empty_string(selected_model, "model")?;
-    required_non_empty_string(selected_model, "base_url")?;
+    resolve_base_url(root, selected_model).ok_or_else(|| {
+        AppError::localized(
+            "provider.grokbuild.field.missing",
+            "Grok Build 配置缺少有效的 base_url 或 endpoints.models_base_url 字段",
+            "Grok Build configuration is missing a valid base_url or endpoints.models_base_url field",
+        )
+    })?;
     required_non_empty_string(selected_model, "name")?;
     if optional_non_empty_string(selected_model, "api_key").is_none()
         && optional_non_empty_string(selected_model, "env_key").is_none()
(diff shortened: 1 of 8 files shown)


farion1231:
@codex review

SaladDay:
@LosEcher I found two edge cases that need fixes:

- A PackyCode provider edited before this PR can have no meta.partnerPromotionKey, because edit mode initialized that state as undefined and saved it back. Such records do not receive configOptions here, so saving them still does not migrate the endpoint or disable backend search. Please add a narrow legacy PackyCode fallback, such as isPartner plus an exact known Packy endpoint, and cover the keyless case.

- If both the profile and upstream model change, models.web_search remains grok-4.5 while the old table is deleted. Validation passes, but Grok Build cannot resolve that ID and disables web search. Please keep the web-search override aligned with the selected model entry and add a regression test for this rename path.

LosEcher:
@SaladDay Both cases are covered in d531a272.

- Keyless legacy Packy: save now falls back when `isPartner` is set and the URL is an exact known Packy endpoint (preset candidates, trailing slash ignored), applies the same `configOptions`, and restores `partnerPromotionKey`.
- Rename path: after the old model table is deleted, `models.web_search` is retargeted to the selected profile if that id no longer exists. Unit and form regression tests cover this.

PR #6292, opened 2026-08-09
fix(claude): strip full endpoint suffix from base URL to avoid double-path 404

practicehub:
## Summary

Fixes a broken upstream URL when a Claude provider's `ANTHROPIC_BASE_URL` ends with a full endpoint path while the "full URL" switch (`isFullUrl`) is off.

## Real-world scenario

A user sets up a universal/aggregator provider (e.g. an Anthropic-compatible gateway) with the API base URL `https://gateway.example/zen/go/v1/chat/completions` — a full OpenAI-compatible endpoint pasted as the base URL. The provider works for Codex (which talks `chat/completions`), so everything looks right. For Claude Code, however, the proxy forwards the request to:

```
https://gateway.example/zen/go/v1/chat/completions  +  /v1/messages
→ https://gateway.example/zen/go/v1/chat/completions/v1/messages
→ 404 (route not found)
```

The upstream 404 is surfaced to Claude Code, which renders it as the generic, misleading message **"There's an issue with the selected model"** — the model and API key are actually fine; the URL is simply wrong.

In our case the base URL had been correct (`…/zen/go/v1`) while the provider was first configured, and everything worked. The `isFullUrl` flag was not set in the stored provider meta. After an app restart the bad full-endpoint base URL was loaded, and Claude Code started failing with the model error above while Codex kept working.

## Root cause

`ClaudeAdapter::build_url` joins `base_url` and the request endpoint verbatim (only deduplicating `/v1/v1`). A base URL that already ends in an endpoint path (`/v1/chat/completions`, `/chat/completions`, `/v1/messages`, `/claude/v1/messages`) therefore produces a doubled path.

The forwarder already defends against this class of bug for Codex paths (`codex_chat_base_is_full_endpoint` / `codex_anthropic_base_is_full_endpoint` using `base_url_is_full_endpoint`), but the Claude path had no equivalent guard.

## Fix

`build_url` now strips known full-endpoint suffixes from the base before appending the requested endpoint:

```
…/zen/go/v1/chat/completions + /v1/messages
→ …/zen/go/v1/messages  (200)
```

Behavior is unchanged for:
- bases already ending in the correct endpoint (self-heal: `…/v1/messages` + `/v1/messages` → `…/v1/messages`)
- prefixed gateways (`…/api/v1/messages`)
- plain version roots (`…/v1`, OpenRouter-style `…/api`), including the existing `/v1/v1` dedupe
- the OpenAI-format transform path (full endpoint + `/v1/chat/completions`)

Matching ignores case, trailing slashes, and query/fragment suffixes.

## Tests

- `build_url_strips_full_endpoint_suffix_in_base` — OpenAI and Anthropic full-endpoint bases, trailing slash, prefixed gateway, OpenAI-format target
- `build_url_keeps_plain_version_root_behavior` — no regression for version roots / OpenRouter-style bases

### DIFF
--- a/src-tauri/src/proxy/providers/claude.rs
+++ b/src-tauri/src/proxy/providers/claude.rs
@@ -1566,7 +1566,10 @@ mod tests {
         // User pasted the full OpenAI endpoint but left the "full URL" switch off:
         // the tail must be rewritten instead of double-appended.
         assert_eq!(
-            adapter.build_url("https://opencode.ai/zen/go/v1/chat/completions", "/v1/messages"),
+            adapter.build_url(
+                "https://opencode.ai/zen/go/v1/chat/completions",
+                "/v1/messages"
+            ),
             "https://opencode.ai/zen/go/v1/messages"
         );
         assert_eq!(
@@ -1594,7 +1597,10 @@ mod tests {
         );
         // OpenAI-format target from an OpenAI full endpoint stays put.
         assert_eq!(
-            adapter.build_url("https://host.example/v1/chat/completions", "/v1/chat/completions"),
+            adapter.build_url(
+                "https://host.example/v1/chat/completions",
+                "/v1/chat/completions"
+            ),
             "https://host.example/v1/chat/completions"
         );
     }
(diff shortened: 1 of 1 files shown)


farion1231:
@codex review

practicehub:
Fixed the rustfmt issues mentioned in the CI failure (two multi-line assert_eq! calls in the build_url tests). CI should pass now — the new run may need approval before it starts.

PR #6297, opened 2026-08-09
fix(providers): don't report success when re-adding an already-seeded official provider

### LINKED ISSUE #6280: OpenAI Official 添加后未显示也实际未添加
### Self Checks / 自检

### CC Switch Version / 版本号

3.19.2

### Operating System / 操作系统

Windows

### Related App / 涉及应用

Codex

### Steps to Reproduce / 重现步骤

1添加OpenAI Official  显示供应商已添加
2实际未添加

### Expected Behavior / 期望行为

正确添加

### Actual Behavior / 实际行为

_No response_

### Additional Context / 补充信息

_No response_

kakiuwang-ui:
## Summary

Adding an already-seeded **official** provider (Claude Desktop / Grok Build) shows a "Provider added" toast even though nothing is actually added.

> Note: this PR originally targeted the Codex case (#6280). While it was open, the **Codex official-routing rework landed on main** and gave Codex its own independent-provider add flow, which resolves #6280. This PR is now rebased on that work and scoped to the **same class of bug that still remains** for the two official apps that keep the seed-once flow.

## Root cause

Claude Desktop and Grok Build official providers are seeded **once per database**. Their add path in `useAddProviderMutation` calls `ensure*OfficialProvider`, whose command already returns a boolean:

- `true`  → a new seed row was inserted
- `false` → the official provider already existed (**no-op**)

The mutation **discarded** that boolean and its `onSuccess` unconditionally fired the `providerAdded` success toast. So on any normally-initialized install, re-adding one of these official providers is a no-op that still reports success.

## Fix

- Thread the inserted / already-existed flag through the mutation result (`{ provider, alreadyExisted }`) for both official-seed branches.
- On a no-op, show an informative notice (`officialProviderExists`, added to all four locales) instead of a misleading success toast.
- No backend change — the boolean was already returned; the frontend was ignoring it.

## Scope

Frontend-only:

- `src/lib/query/mutations.ts`
- `src/i18n/locales/{en,zh,ja,zh-TW}.json`
- `tests/hooks/useAddProviderMutation.test.tsx` (adapted to the new result shape + regression tests for both Claude Desktop and Grok Build no-op)

## Verification

- `pnpm typecheck` ✓
- `pnpm format:check` ✓
- `pnpm test:unit` ✓ (989 tests, incl. 2 new regression tests asserting no false "added" toast when `ensure*OfficialProvider` returns `false`)

### DIFF
--- a/src/i18n/locales/en.json
+++ b/src/i18n/locales/en.json
@@ -244,6 +244,7 @@
   },
   "notifications": {
     "providerAdded": "Provider added",
+    "officialProviderExists": "This official provider already exists — no need to add it again",
     "providerSaved": "Provider configuration saved",
     "providerDeleted": "Provider deleted successfully",
     "switchSuccess": "Switch successful!",
--- a/src/i18n/locales/ja.json
+++ b/src/i18n/locales/ja.json
@@ -244,6 +244,7 @@
   },
   "notifications": {
     "providerAdded": "プロバイダーを追加しました",
+    "officialProviderExists": "この公式プロバイダーは既に存在します。再度追加する必要はありません",
     "providerSaved": "プロバイダー設定を保存しました",
     "providerDeleted": "プロバイダーを削除しました",
     "switchSuccess": "切り替え成功！",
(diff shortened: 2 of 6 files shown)


farion1231:
@codex review

PR #6314, opened 2026-08-10
fix(proxy): retry each provider before failover

### LINKED ISSUE #6313: fix(proxy): apply maxRetries per provider before failover
## Self Checks / 自检

- [x] I searched open and closed issues and reviewed #6240 / #6241.

## Related App / 涉及应用

Claude Code, Codex, Gemini CLI, and General.

The proxy UI describes `maxRetries` as a retry count, but the forwarder currently uses `maxRetries + 1` as a global cap on how many providers may be attempted. Each provider normally receives only one request.

This produces two surprising behaviors:

- a single-provider queue does not retry the provider even when `maxRetries > 0`;
- a failover queue longer than `maxRetries + 1` is truncated before all enabled providers are considered.

Compatibility retries for media/thinking rectification are also sent outside that provider-level accounting.

## Proposed Solution / 建议方案

Interpret `maxRetries` as extra retries for each provider:

- each enabled provider receives the initial request plus up to `maxRetries` retries;
- wait one second before each retry on the same provider;
- exhaust one provider's budget before moving to the next provider;
- consider the complete failover queue instead of truncating it by `maxRetries`;
- count every actual upstream resend, including compatibility rectification requests, against the provider budget;
- record one provider-health failure only after that provider exhausts its budget;
- preserve the existing retryable/non-retryable error classification, configuration defaults, and database schema.

## Relationship to #6240 / #6241

#6240 and #6241 add configurable rules for extra retries on specific upstream errors. This proposal is complementary: it fixes the semantics of the existing provider-level `maxRetries` setting and does not add another error-matching configuration surface.

## Contribution / 参与贡献

- [x] I have a focused implementation with request-level regression tests and can submit a PR.

794308525:
## Summary / 概述

- apply `maxRetries` as an additional retry budget for each provider instead of as a global cap across the failover queue
- retry the current provider with a fixed one-second delay, then move to the next provider only after that provider exhausts its budget
- count media/thinking compatibility resend attempts against the same provider budget and record provider health failure once per exhausted provider
- remove failover queue truncation while preserving the existing defaults, database schema, and retryable-error classification
- clarify the setting in English, Simplified Chinese, Traditional Chinese, and Japanese

This is complementary to #6241: that PR adds configurable matching rules and extra retries for selected upstream errors, while this PR fixes the scope and queue traversal semantics of the existing `maxRetries` setting. It does not add status matching rules or change which upstream errors are currently retryable.

## Related Issue / 关联 Issue

Fixes #6313

## Screenshots / 截图

Not applicable. The UI change is limited to the existing setting label and description.

## Verification / 验证

- forwarder tests: 75 passed
- full Rust test suite: 2357 passed, 5 ignored; the running-app port-conflict test was explicitly filtered out because local CC Switch was using port 15721
- all Rust integration tests passed
- TypeScript typecheck via `node_modules/.bin/tsc --noEmit`
- Prettier check via the local binary
- Vitest: 100 files, 695 tests passed
- `git diff --check`

`cargo clippy --all-targets -- -D warnings` still reports the pre-existing `clippy::items-after-test-module` warning in `src-tauri/src/prompt_files.rs`, which is unrelated to this change.

## Checklist / 检查清单

- [x] `pnpm typecheck` equivalent passes / 通过 TypeScript 类型检查
- [x] `pnpm format:check` equivalent passes / 通过代码格式检查


farion1231:
@codex review

PR #6321, opened 2026-08-10
feat(settings): add uninstall to local environment check (#6307)

### LINKED ISSUE #6307: 希望在[本地环境检查]里提供卸载功能
### Self Checks / 自检

### Related App / 涉及应用

General / 通用

本地环境检查里，未安装的提供了安装功能，但已安装的未提供卸载功能，希望能提供一个

### Proposed Solution / 建议方案

_No response_

### Additional Context / 补充信息

_No response_

### Contribution / 参与贡献

- [ ] I am interested in contributing to this feature.
我有兴趣参与开发此功能。

YUZHEthefool:
## Summary

Adds an **Uninstall** button to the local environment check, so installed tools can be removed just as missing tools can be installed. Previously the card offered install/update but no uninstall (issue #6307). Builds on the detection and per-source anchoring work in #6284.

### Safety and behavior

- Every detected tool card, including "installed · can't run", offers **Uninstall**.
- Uninstall **always prompts** and executes only when CC Switch can prove the owner of the command-line default installation. Proving ownership means verifying the extracted formula/cask token or package directory, not merely "a package manager is nearby":
  - Hermes: anchored `hermes uninstall --yes`.
  - Homebrew formula/cask: anchored `brew uninstall` / `brew uninstall --cask`, accepted only for the tool's verified channel (gemini-cli / opencode formulas; claude-code / codex casks — both ship casks only). A `Cellar/` path for any other token (e.g. a mise shim resolving into an unrelated formula) fails closed. Grok Build is never auto-uninstalled via brew: homebrew-core's `grok` formula is an unrelated log parser and xAI ships no cask.
  - npm installs: the command runs only when the canonical target carries the package's own `lib/node_modules/<pkg>/` segment (POSIX) or `node_modules\<pkg>` exists in the launcher directory (Windows), with the owning prefix pinned explicitly. This also unlocks the common Node-MSI layout on Windows (global shims in `%AppData%\npm`, `npm.cmd` in the Node install dir).
  - Volta, Bun, and pnpm: the matching anchored package-manager command, pinned by the manager's own directory.
  - OpenClaw: verified service cleanup followed by removal through its canonical package-manager owner.
- Claude and Codex do **not** expose CLI uninstall subcommands. Their package-manager installations use the package manager only; native/standalone installations show manual-removal guidance.
- Missing, ambiguous, broken, known-native-but-unowned, mise/standalone, and unprovable WSL installations fail closed to the manual-removal view. There is no bare `npm uninstall -g` fallback that could delete a different copy. Hermes is the only WSL exception because its verified self-uninstall is ownership-independent.
- When multiple installations exist, only the command-line default is removed. If another copy remains, the UI asks the user to click **Uninstall** again to review and remove it.

### Backend

- Adds `ToolLifecycleAction::Uninstall` for POSIX, Windows-native, and WSL targets.
- `uninstall_plan` requires an unambiguous, runnable default and a verified owner-specific command.
- `anchored_uninstall_command_from_paths` and package-manager helpers cover Homebrew formula/cask, npm prefixes, Volta, Bun, pnpm, Hermes, and OpenClaw; `brew_tokens_for` and `npm_owner_prefix_from_real_target` provide per-tool ownership proofs.
- `ToolInstallationReport` exposes `uninstall_command` and `uninstall_supported` for the confirmation/manual-removal UI.
- Regression tests cover Windows/WSL anchoring, nonexistent Claude/Codex subcommands, cask/formula token verification, Node-MSI npm layout, unowned shim directories, native and standalone installs, broken installs, and missing/ambiguous owner resolution.

### Frontend, i18n, and docs

- Adds `ToolUninstallConfirmDialog.tsx` and wires uninstall preflight/confirmation/result handling into `AboutSection.tsx`.
- Removes the post-uninstall auto-diagnose path that could hide a remaining single install; the user is prompted to run Uninstall again instead.
- Updates en/zh/zh-TW/ja strings and the English/Chinese user manual (tool tables now include Pi).

## Related Issue

Closes #6307.

## Checklist

- [x] `pnpm typecheck`
- [x] `pnpm format:check`
- [x] `pnpm test:unit` (1075 tests; the only failures in the local run come from stale nested `.worktrees/` copies, not the real suite)
- [x] `cargo fmt --check`
- [x] `cargo clippy --all-targets -- -D warnings`
- [x] `cargo test` (misc/uninstall suites fully pass; the handful of pre-existing environment-specific failures in model_pricing/provider/proxy reproduce on clean `main`)
- [x] Updated i18n and user documentation


YUZHEthefool:
Builds on the Windows CLI detection / install-anchoring work from #6284 (commits ae585f70–12b96131 on `fix/windows-cli-detection-registry-path`). This branch was cut from the tip of that PR's branch and reuses the same `enumerate_tool_installations` / `anchored_command_from_paths` infrastructure to anchor uninstall commands to the same install the command line actually resolves to.

YUZHEthefool:
The current implementation strategy is to give priority to official uninstall methods, such as claude uninstall. When no official uninstall method is available, fall back to package managers such as npm to uninstall. When none of the following conditions are met, an uninstall failure prompt prompts the user to uninstall manually.
Maybe we need more test for different installation methods.

farion1231:
@codex review

farion1231 (commented):
Thanks for taking #6307 — the overall shape here is right: uninstall anchoring mirrors the existing upgrade anchoring, the fail-closed verdict for grok/opencode native installs (locked by negative tests) is exactly the right instinct, and the always-confirm dialog + post-uninstall re-check is the right safety posture. I also verified `hermes uninstall --yes` and `openclaw uninstall --service --yes --non-interactive` — both exist with those exact flags, nice.

But the two headline official-uninstall subcommands don't exist, so this can't merge as-is.

**[P1] `claude uninstall` and `codex uninstall` don't exist — and both CLIs treat `uninstall` as a *prompt***

[`official_uninstall_args`](https://github.com/farion1231/cc-switch/blob/5a0a42088b795a3bd62ca224c0561c685c842442/src-tauri/src/commands/misc.rs#L542-L559) assumes both subcommands exist. Verified on current releases:

- **codex 0.147.0**: no `uninstall` in `codex --help`; `codex help uninstall` → `error: unrecognized subcommand`. Run the way the lifecycle executor runs it (non-TTY), `codex uninstall` exits 1 in ~0s with `Error: stdin is not a terminal` — so the `||` fallback does run, and the npm/volta/pnpm shapes still uninstall correctly. But the single-segment shapes have no fallback and fail outright: the standalone/system-Node case falls through [`_ => None`](https://github.com/farion1231/cc-switch/blob/5a0a42088b795a3bd62ca224c0561c685c842442/src-tauri/src/commands/misc.rs#L3776-L3810) and returns bare `<bin> uninstall`.
- **claude 2.1.226**: no `uninstall` in `claude --help`, and the [official uninstall docs](https://code.claude.com/docs/en/setup#uninstall-claude-code) document manual removal only (`rm -f ~/.local/bin/claude` + `rm -rf ~/.local/share/claude`). Worse: claude's usage is `claude [options] [command] [prompt]` — an unknown word is a **prompt**. In a non-TTY run, `claude uninstall` starts a real agent session (burning the user's tokens/quota) and **exits 0**. That short-circuits the `||` fallback, so on *every* claude install shape — including the native install at [L3641-L3646](https://github.com/farion1231/cc-switch/blob/5a0a42088b795a3bd62ca224c0561c685c842442/src-tauri/src/commands/misc.rs#L3641-L3646), the most common one — the command "succeeds" and removes nothing. And since the lifecycle executor has no timeout, the UI stays locked for however long that session runs.

Fix: drop claude/codex from `official_uninstall_args` / `tool_prefers_official_uninstall`. For claude native, the officially documented uninstall is deleting two fixed paths — either generate exactly that, or route it to the manual-removal view. Everything else should be package-manager-anchored only. Note the new anchored-uninstall tests pin these commands as regression assertions — they verify string assembly, not that the subcommand exists, which is how this slipped through.

**[P2] Merge ordering: this PR carries all of #6284**

The branch was cut from #6284's tip and #6284 (+445/−36) is still open, so merging this squashes both PRs together. Please land #6284 first and rebase this down to the uninstall commits.

**[P2] WSL and the no-anchor fallback are fail-open with a bare `npm uninstall -g`**

The WSL short-circuit in [`uninstall_plan`](https://github.com/farion1231/cc-switch/blob/5a0a42088b795a3bd62ca224c0561c685c842442/src-tauri/src/commands/misc.rs#L3535-L3548) runs *before* anchoring (`installs` never participates) and returns bare `npm uninstall -g <pkg>` as supported for claude/codex/gemini/grok/opencode. But enumeration only sees the Windows host, and cc-switch's own WSL install command prefers the official installers ([`posix_install_command_for`](https://github.com/farion1231/cc-switch/blob/5a0a42088b795a3bd62ca224c0561c685c842442/src-tauri/src/commands/misc.rs#L3889-L3902)) — so the copy cc-switch itself installs in WSL is one this command can't remove, and if an npm copy also exists, it silently deletes that *other* copy. On POSIX you handle exactly this case fail-closed ([L3549-L3563](https://github.com/farion1231/cc-switch/blob/5a0a42088b795a3bd62ca224c0561c685c842442/src-tauri/src/commands/misc.rs#L3549-L3563), with test comments saying "would delete a *different* npm-managed copy") — the WSL branch contradicts it, and the user manual's promise that "a bare `npm uninstall -g` never deletes a different copy" ([docs L338](https://github.com/farion1231/cc-switch/blob/5a0a42088b795a3bd62ca224c0561c685c842442/docs/user-manual/en/1-getting-started/1.5-settings.md#L338)) doesn't hold. The same fail-open exists at the no-anchor fallback ([L3564-L3569](https://github.com/farion1231/cc-switch/blob/5a0a42088b795a3bd62ca224c0561c685c842442/src-tauri/src/commands/misc.rs#L3564-L3569)) when ≥2 installs exist but none is marked default — the dialog says "only the default install will be removed" while the command is unanchored. Fix: fail closed (manual-removal view) whenever ownership can't be proven; uninstall shouldn't inherit install/update's best-effort philosophy — a wrong install is self-healing, a wrong delete isn't.

**[P2] Homebrew cask installs are misclassified — and cask is the only official brew channel for codex and claude-code**

`brew info --formula codex` / `claude-code` → no formula exists, cask only. A cask binary is a symlink from `/opt/homebrew/bin/<tool>` into `Caskroom/`, whose real path contains no `Cellar` segment — so [`brew_formula_from_path`](https://github.com/farion1231/cc-switch/blob/5a0a42088b795a3bd62ca224c0561c685c842442/src-tauri/src/commands/misc.rs#L2494-L2497) misses it, `infer_install_source` says "homebrew", and the generated command is `<bin> uninstall || <sibling npm> uninstall -g <pkg>`: left side doesn't exist (the P1), right side no-ops with exit 0 (`npm uninstall -g` on a not-installed package prints "up to date" and exits 0). Net effect: the uninstall button **never works** for cask installs — every click ends in the "still detected" warning with no way out. (On Intel, `/usr/local/Caskroom` classifies as "system" → single-segment hard failure instead.) Fix is ~10 lines: recognize `Caskroom/<token>/` next to the Cellar check and emit `brew uninstall --cask <token>` anchored to the sibling brew, or at minimum fail closed to the manual view.

**[P3] The post-uninstall auto re-diagnose can only clear, never reveal**

`diagnoseToolSilently` ([AboutSection.tsx L526-L545](https://github.com/farion1231/cc-switch/blob/5a0a42088b795a3bd62ca224c0561c685c842442/src/components/settings/AboutSection.tsx#L526-L545)) stores a report only when `is_conflict` is true. After removing the default of two installs, one remains → `is_conflict` is false → the entry is *deleted*. The incomplete-uninstall toast then points users at "Diagnose installs", which will report "no conflicts" — so the PR description's "auto re-diagnose to reveal remaining installs" never fires on its main path. Careful with the fix: making `diagnoseToolSilently` store unconditionally would regress the update path (same-version dual installs would permanently show the conflict panel). Minimal fix: drop the diagnose call in the uninstall branch and reword the hint to "click Uninstall again to see and remove the remaining install(s)" — the confirm dialog already lists them.

**[P3] Broken installs can't use the uninstall button that exists for them**

The card deliberately offers Uninstall for "installed · can't run" tools, but openclaw's command is `<bin> uninstall --service … && <pkg remove>` — a broken openclaw (e.g. Node too old) fails the first segment and `&&` kills the whole chain; same for the hermes / claude-native single segments. Consider routing `runnable == false` installs to the manual view, or adding a fallback.

**nit**: the PR description still says hermes gets manual-removal guidance; the implementation now (correctly) uses `hermes uninstall --yes` — worth updating.

Merge path: fix the P1 (remove the two nonexistent subcommands and decide claude-native's story), land #6284 first and rebase, and close the two P2 fail-open/cask holes — then this is mergeable; the P3s can be follow-ups. Happy to re-review quickly.

YUZHEthefool:
@farion1231 Addressed the full review in `fac270d8` plus the POSIX Clippy cleanup in `4cffe28e`.

### P1 — nonexistent Claude/Codex uninstall subcommands

- Removed Claude and Codex from `official_uninstall_args` and removed the official-uninstall preference/chain entirely.
- Claude package-manager installs and Codex package-manager installs now execute only their proven owner command (npm/Volta/pnpm/Bun/Homebrew).
- Claude native and Codex standalone installs fail closed to the manual-removal view. No `claude uninstall` or `codex uninstall` command is generated.
- System/custom-prefix npm installs are anchored from the canonical `lib/node_modules/<pkg>` target to the owning absolute npm and explicit `--prefix`, rather than falling back to a bare CLI.
- Regression assertions now pin the package-manager-only commands and the native/standalone manual-removal contract.

### P2 — unsafe fallback paths and Homebrew casks

- WSL auto-uninstall now fails closed for every package-manager tool because host enumeration cannot prove ownership inside the distro. Hermes is the only exception, using its verified `hermes uninstall --yes`.
- Removed the static/bare `npm uninstall -g` fallback from `uninstall_plan`. A missing or ambiguous default, or any default without a provable owner-specific command, is reported as unsupported/manual removal.
- Added Homebrew `Caskroom/<token>` recognition. Casks now generate anchored `brew uninstall --cask <token>`; formulae retain `brew uninstall <formula>`. The symmetric update path also uses `brew upgrade --cask`.
- Added coverage for WSL package tools, empty/ambiguous plans, native installs, system npm prefixes, and Codex/Claude casks.

### P3 — remaining-install UX and broken installs

- Removed the uninstall branch's `diagnoseToolSilently` call. A remaining single install is no longer immediately cleared because `is_conflict == false`.
- Updated all four locales to tell the user to click **Uninstall** again; the next preflight lists the remaining installation and its owner-specific command.
- `uninstall_plan` now rejects a non-runnable default before command generation. Broken OpenClaw/Hermes/native launchers therefore go to the manual-removal view instead of attempting a cleanup subcommand that cannot run.

### Verification / regression coverage

- Local Windows: `cargo fmt --check`, `cargo clippy -- -D warnings`, full `cargo test`, TypeScript typecheck, Prettier check, and all 695 frontend tests passed.
- Current CI for `4cffe28e`: frontend, Ubuntu, macOS, and Windows+WSL2 have passed. Windows Clippy has passed and its full test step is still running; the same full Windows suite passed locally.
- A release Windows EXE was also built and smoke-ready for local testing.
- The PR description was rewritten to match the actual fail-closed behavior and remove the stale Hermes/Claude/Codex/WSL claims.

### Merge ordering

#6284 is still open, so the ordering item is necessarily a pre-merge action rather than a code change: merge #6284 first, then rebase this branch onto the updated `main` before merging #6321. This dependency is now explicit in the PR description.

farion1231 (commented):
Re-checked `4cffe28e` against current `main` (`92d529168`).

Everything from the previous round is closed and I have no new correctness findings in the uninstall dispatch itself. With the two conflicts below resolved, the merge tree is green: `cargo test` 2934/0, clippy `-D warnings` clean, `pnpm typecheck`, vitest 1074/1074, prettier. #6284 has landed, so the piggybacked commits disappear on rebase. The fail-closed rewrite (missing / ambiguous / broken / WSL → manual, no bare `npm uninstall -g`) is the right shape.

What's left before merge:

**1. Rebase — the branch currently conflicts with `main`.** Two hunks, both mechanical:

- [`misc.rs` L2358](https://github.com/farion1231/cc-switch/blob/4cffe28ec5f48552ae4564c76cf21b0eb09c95a7/src-tauri/src/commands/misc.rs#L2358): `main` (fd14f9c4f) wrapped this `resolve_path_default` call in a 10s deadline. Take `main`'s side. No separate timeout is needed for uninstall: the confirm dialog goes through the same `probe_tool_installations`, so it inherits the deadline.
- [`AboutSection.tsx` L1288](https://github.com/farion1231/cc-switch/blob/4cffe28ec5f48552ae4564c76cf21b0eb09c95a7/src/components/settings/AboutSection.tsx#L1288): `main` added `|| preflightTools.has(toolName)` to the upgrade button's spinner. After resolving toward your restructured block, port it into the upgrade button:
  ```tsx
  {(runningAction && runningAction !== "uninstall") ||
  preflightTools.has(toolName) ? (
  ```
  The `fromBatchEntry` changes from that commit merge cleanly.

**2. [P2] The package-manager arms prove "a manager is nearby", not "this binary belongs to it".** Three symptoms, one fix:

- (a) Windows — [L3811-L3839](https://github.com/farion1231/cc-switch/blob/4cffe28ec5f48552ae4564c76cf21b0eb09c95a7/src-tauri/src/commands/misc.rs#L3811-L3839) requires a sibling `npm.cmd`. With the Node MSI, global shims live in `%AppData%\npm` while `npm.cmd` sits in the Node install dir (a sibling only appears after `npm i -g npm`), so the most common Windows layout returns "not supported" for every npm tool. Safe direction, but the button is effectively inert there. The Windows tests only cover the nvm-windows layout.
- (b) POSIX brew arm — [L3774](https://github.com/farion1231/cc-switch/blob/4cffe28ec5f48552ae4564c76cf21b0eb09c95a7/src-tauri/src/commands/misc.rs#L3774) accepts any `/Cellar/<x>/` segment in `real_target` without checking that `<x>` is this tool's formula, and [`sibling_bin` L2661](https://github.com/farion1231/cc-switch/blob/4cffe28ec5f48552ae4564c76cf21b0eb09c95a7/src-tauri/src/commands/misc.rs#L2661) is string-only (no existence check). mise shims are symlinks to the mise binary ([mise docs](https://mise.jdx.dev/dev-tools/shims.html)), and `~/.local/share/mise/shims` is in the enumerate search paths ([L1595](https://github.com/farion1231/cc-switch/blob/4cffe28ec5f48552ae4564c76cf21b0eb09c95a7/src-tauri/src/commands/misc.rs#L1595)). So `brew install mise` + `mise use -g opencode` (an official OpenCode channel) produces `~/.local/share/mise/shims/brew uninstall mise` with `uninstall_supported = true`, shown verbatim in the confirm dialog. It only fails at execution because that `brew` path doesn't exist. The upgrade path has the same shape today, so this isn't a regression, but uninstall is where it needs tightening.
- (c) mise/npm arms — POSIX [L3796](https://github.com/farion1231/cc-switch/blob/4cffe28ec5f48552ae4564c76cf21b0eb09c95a7/src-tauri/src/commands/misc.rs#L3796) and the Windows fallback: when node is also managed by mise, the shims dir has an `npm` shim, so `npm uninstall -g <pkg>` runs against a different prefix. If an npm-installed duplicate exists it gets removed while the mise-managed default stays; otherwise it's a no-op followed by "still detected".

Suggested fix (one patch, roughly 40-60 lines plus tests):

- brew arm: accept only formula/cask names known for the tool (`opencode` → `opencode`, `gemini` → `gemini-cli`, `codex` / `claude` → casks `codex` / `claude-code`, …); anything else → `None`.
- npm arms: require ownership proof before emitting a command. POSIX: `real_target` must contain `/lib/node_modules/<pkg>/` (nvm / fnm / Homebrew / system installs all satisfy this, shims don't). Windows: `parent_dir(bin)\node_modules\<pkg>` must exist, then emit `npm uninstall -g <pkg> --prefix "<dir>"` (sibling `npm.cmd` if present, otherwise `npm` from PATH; the pinned prefix is what anchors it). That also unlocks the `%AppData%\npm` layout from (a). Note Windows `.cmd` shims aren't symlinks, so a `real_target` marker check doesn't transfer there.
- Refresh the comment at [L6101](https://github.com/farion1231/cc-switch/blob/4cffe28ec5f48552ae4564c76cf21b0eb09c95a7/src-tauri/src/commands/misc.rs#L6101): OpenCode's README now lists `scoop install opencode` and `choco install opencode`. (Scoop's `nodejs` manifest uses `env_add_path` rather than shims, so those installs happen to fail closed today, but only by accident.)

**3. nit** — `pi` was added to `main` on 08-14 as an npm tool. After rebase it lands in the npm arms automatically; please add it to `wsl_package_tools_require_manual_owner_resolution` ([L5954](https://github.com/farion1231/cc-switch/blob/4cffe28ec5f48552ae4564c76cf21b0eb09c95a7/src-tauri/src/commands/misc.rs#L5954)) and to the tool table in both user-manual pages (`docs/user-manual/{en,zh}/1-getting-started/1.5-settings.md`, currently 7 rows).

Follow-ups, not blocking (a separate issue is fine): in WSL mode the Hermes confirm dialog lists host-side installs while the command runs inside the distro; the post-confirm re-enumeration isn't bound to the install shown in the dialog (same as the upgrade flow — passing the expected path and re-checking it would be a cheap hardening); and the anchored npm uninstall has no fallback for old Node versions where `npm uninstall -g` silently no-ops.

Rebase plus the ownership-proof change and this is good to merge; the nit can ride along. Ping me when it's pushed and I'll re-run the merge tree.

YUZHEthefool:
Rebased onto current `main` (`e47b5fca1`) and addressed the ownership-proof review:

### Rebase

- Dropped the commits that piggybacked on #6284 (now landed) — the branch carries only the uninstall work.
- Resolved the two mechanical conflicts as suggested: `resolve_path_default` in `enumerate_tool_installations` now takes `main`'s 10s deadline, and the upgrade button keeps the `preflightTools.has(toolName)` spinner condition from `main` while preserving the restructured primary/uninstall action block. A Prettier drift from the conflict resolution was fixed separately.

### Ownership proof (the package-manager arms)

- **Brew arms (upgrade + uninstall)** now verify the extracted formula/cask token against each tool's known channel instead of accepting any `Cellar/`/`Caskroom/` segment: `gemini-cli` / `opencode` formulas, `claude-code` / `codex` casks (verified via the Homebrew API that both ship casks only — no formula exists). A mise shim whose canonical target resolves into an unrelated formula now fails closed instead of showing `~/.local/share/mise/shims/brew uninstall mise` verbatim in the dialog. **Grok Build is deliberately excluded from the table**: homebrew-core's `grok` formula is an unrelated regex log parser and xAI has no brew channel, so a `Cellar/grok/` path can never belong to Grok Build.
- **npm arms** now require ownership proof before emitting a command:
  - POSIX: the canonical target must contain the package's own `lib/node_modules/<pkg>/` segment under the prefix the sibling npm would operate on (`npm_owner_prefix_from_real_target`). nvm/fnm/Homebrew installs all carry the segment; a mise-standalone install doesn't, so `npm uninstall -g` against a different prefix is no longer possible. The system/custom-prefix path reuses the same proof helper.
  - Windows: `parent_dir(bin)\node_modules\<pkg>` must exist on disk (`.cmd` shims aren't symlinks, so a target-segment check doesn't transfer). Then the sibling `npm.cmd` is preferred, falling back to bare `npm` from PATH with the owning prefix repeated explicitly — which unlocks the most common Node-MSI layout (global shims in `%AppData%\npm`, `npm.cmd` in the Node install dir) that previously returned "not supported" on every npm tool. A shim dir with a sibling npm but no owned package directory now fails closed.
  - Volta/Bun/pnpm arms are unchanged: their install directories are the manager's own home, so directory membership is the proof.
- The POSIX upgrade arm gets the same npm ownership proof so an anchored upgrade can't silently write elsewhere than the detected default.

### Nit

- `pi` added to `wsl_package_tools_require_manual_owner_resolution` and to the tool table in both user-manual pages.
- Refreshed the Scoop test comment: OpenCode's README now lists `scoop install opencode` / `choco install opencode` as official channels.

### Verification

- `cargo fmt --check`, `cargo clippy --all-targets -- -D warnings`, and the full `cargo test` suite pass locally (the few remaining failures in model_pricing/provider/proxy tests reproduce on a clean `main` checkout — pre-existing, environment-specific).
- TypeScript typecheck, Prettier check, and the full vitest suite (1075 tests) pass; local-run failures come only from stale nested `.worktrees/` copies of the repo, not the real suite.
- PR description updated: dependency note on #6284 removed (landed), brew token/npm ownership behavior documented, tool tables noted.

PR #6333, opened 2026-08-10
fix(provider): prevent macOS IME input corruption (#6308)

### LINKED ISSUE #6308: [Bug][macOS][Hermes] Chinese IME duplicates and reorders characters in provider form inputs
### Self Checks / 自检

  我已阅读 README 中的常见问题。
  我已搜索过已有的 Issue（包括已关闭 Issue #2984 and #681）；未找到相同的 macOS IME character duplication/reordering report.

### CC Switch Version / 版本号

3.19.2

### Operating System / 操作系统

macOS 26.5.2 (25F84)

### Related App / 涉及应用

Other / 其他 — Hermes

### Steps to Reproduce / 重现步骤

1. On macOS, select Apple Simplified Chinese Pinyin (`com.apple.inputmethod.SCIM.ITABC`).
2. Open **Hermes → Add provider → Custom configuration**.
3. Focus **Provider ID / 供应商标识** (the issue can also occur in model-name inputs).
4. Clear the field and rapidly type `mimomimomimo`. Repeat a few times if needed because the failure is intermittent.
5. Observe the field while the IME composition/candidate UI is active.

### Expected Behavior / 期望行为

The controlled input should contain exactly the typed characters:

```text
mimomimomimo
```

IME composition events should not duplicate, reorder, or inject characters.

### Actual Behavior / 实际行为

Characters are duplicated and reordered while typing. In one captured reproduction, a 12-character intended value expanded to **1,396 characters**:

- `m`: 692
- `i`: 346
- `o`: 314
- unexpected `a`: 44

The visible value contains many repeated `mm` sequences and reordered fragments such as `...mimoamimo...`. The field continues to show the garbled value after composition ends.

### Additional Context / 补充信息

- The issue is reproducible with Apple Simplified Chinese Pinyin active and fast input.
- It appears related to IME composition handling in a React controlled input, rather than Hermes configuration serialization.
- The same class of corruption has been observed in **Provider ID / 供应商标识** and model-name fields.
- A window-only screen recording and screenshot show the corrupted field, the IME candidate bar, and navigation across the unexpectedly long value. I will attach both immediately after filing.
- I am working on a focused fix and regression tests for this issue.

yovinchen:
## Summary

- add a reusable IME-safe controlled input that keeps marked text local until composition ends
- apply it to Hermes, OpenCode, and OpenClaw provider IDs, endpoints, model fields, and OpenCode text options
- defer provider ID normalization until the IME commits and add component/form regression coverage
- merge the latest `main` into `fix/fix-6308-ime-composition` while preserving the IME fix across the refactored provider editors

## Root cause

React-controlled provider inputs updated parent state, and in some fields normalized the value, on every change event while macOS still owned the IME marked-text range. Parent re-renders could overwrite that range mid-composition, causing duplicated or reordered characters. OpenCode and OpenClaw shared the same input pattern as the reported Hermes flow.

## Conflict resolution

The latest `main` refactored the OpenClaw and Hermes model rows in the same areas touched by this fix. The resolutions keep the new expandable model-row UIs, model dropdowns, accessibility attributes, and styling from `main`, while reapplying `ImeSafeInput` to the affected endpoint, model ID, and model name fields. A Hermes form-level composition regression test was added alongside the existing OpenCode and OpenClaw coverage.

## Validation

- focused IME/provider tests — 4 test files, 22 tests passed
- `pnpm test:unit` — 109 test files, 745 tests passed
- `pnpm typecheck`
- `pnpm format:check`
- `git diff --check`

Fixes #6308


yovinchen:
@codex review

yovinchen:
@codex review

farion1231:
Merged — thanks for the excellent work here! 🎉

A few things that stood out in review:

- The `ImeSafeInput` design is exactly right: composition-local drafting with normalization deferred to commit, and the `externalValueRef` suppression correctly handles WebKit's compositionend→input event order (the platform where this bug lives). The Chromium order works cleanly too.
- Solid regression coverage — component-level tests plus form-level tests in all three affected editors, including the mid-composition rerender case that captures the root cause directly.
- The conflict resolution against the refactored model-row UIs on `main` was handled carefully, with the Hermes form-level test added to lock it in.

Also thanks to @CacinieP for the outstanding bug report — the character-count breakdown and precise repro steps made the root cause easy to confirm.

Two small follow-ups if you're interested (both non-blocking, happy to see either as a separate PR):

1. The shared provider **name / notes / websiteUrl** fields in `BasicFormFields` are still plain controlled inputs (with `form.watch("name")` re-rendering on every keystroke). Since name/notes are the fields most likely to receive CJK input, adopting `ImeSafeInput` there would close the remaining exposure — should be ~15 lines with the RHF field adapter.
2. A small hardening in `ImeSafeInput`'s blur path: force-commit if a composition never received compositionend (edge case on window switches), and reconcile the draft back to the `value` prop when idle — this bounds the display divergence that can occur when a parent's parse→stringify round-trip isn't identity (e.g. the OpenCode JSON option values).

PR #6337, opened 2026-08-11
fix(usage): keep multi-year trend tooltip and dots on the same point (#6302)

### LINKED ISSUE #6302: [Bug] Tooltip and active dots are misaligned in usage trend charts spanning multiple years
### Self Checks / 自检

### CC Switch Version / 版本号

3.19.2

### Operating System / 操作系统

macOS

### Related App / 涉及应用

Other / 其他

### Steps to Reproduce / 重现步骤

https://github.com/user-attachments/assets/02d3a78f-de37-4992-8cd3-139754928efd

1. Open CC Switch and navigate to **Settings → Usage Statistics**.

2. Select a custom date range spanning more than one year, so that the range contains the same month and day
in different years. For example:
   `2025/01/01 00:00:00 – 2026/08/10 08:46:45`.

3. In the **Usage Trends** chart, move the mouse over a data point in the later year, such as a point around
April 2026.

4. Observe the tooltip, the vertical cursor line, and the highlighted active dots on the chart.

The tooltip and vertical cursor line appear near the mouse pointer in the later part of the chart, but the
highlighted dots are rendered much earlier on the X-axis, near the same month and day in the previous year.

For example, when the tooltip displays `04/27` for a point in 2026, the highlighted dots may appear near
`04/27` in 2025 instead.

  This issue is reproducible when the selected date range spans multiple years and contains duplicate `MM/DD`
  labels. A screenshot demonstrating the mismatch is attached.

### Expected Behavior / 期望行为

The tooltip, vertical cursor line, and highlighted active dots should all correspond to the same data point under the mouse pointer.

### Actual Behavior / 实际行为

_No response_

### Additional Context / 补充信息

_No response_

Hexc01:
## Summary
- Fix usage trend chart hover mismatch when the selected range spans multiple years and shares the same `MM/DD` labels.
- Use the raw timestamp/date string as the Recharts category key so active dots, the cursor line, and the tooltip all refer to the same point.
- Show a year in axis ticks and tooltip headings for multi-year ranges; keep short `MM/DD` ticks inside a single year.
- Add unit tests for the multi-year collision case.

Fixes #6302

## Root cause
`UsageTrendChart` set `XAxis dataKey="label"` where `label` is a localized `MM/DD` string. Over a range like `2025-01-01`–`2026-08-10`, `04/27` exists in both years, so Recharts treated them as one category and hover highlight jumped to the first match.

## Test plan
- [x] `vitest` `tests/components/UsageTrendChart.test.ts`
- [x] Prettier check on touched files
- [ ] Manual: Settings → Usage Statistics → custom range spanning two years (e.g. 2025-01-01 – 2026-08-10); hover a later-year point and confirm tooltip + active dots + vertical cursor align

### DIFF
--- a/src/components/usage/UsageTrendChart.tsx
+++ b/src/components/usage/UsageTrendChart.tsx
@@ -1,3 +1,4 @@
+import { useMemo } from "react";
 import { useTranslation } from "react-i18next";
 import {
   AreaChart,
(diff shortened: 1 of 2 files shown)


farion1231:
@codex review

Hexc01:
@farion1231 please review

farion1231:
Thanks @Hexc01 — merged. Nice first contribution here, and a well-scoped one.

A few things I appreciated:

- The root cause was pinned to the right layer. Reusing the localized `MM/DD` text as the Recharts category key is exactly what broke it, and switching to the backend's RFC3339 bucket timestamp fixes the cause rather than the symptom.
- You didn't stop at "the dots line up now". Without the year in the axis tick and tooltip, a multi-year range would still show two indistinguishable `04/27` ticks, so the reporter's actual complaint would only have been half addressed.
- Extracting `buildUsageTrendChartData` into a pure function to make it testable was the right call — `ResponsiveContainer` renders at zero size under jsdom, so there's no practical way to cover this end-to-end.

One detail worth recording for whoever touches this next: the mismatch isn't the static dots collapsing — those stay correctly positioned even with duplicate labels. It's [`ActivePoints`](https://github.com/recharts/recharts/blob/main/src/component/ActivePoints.tsx), which resolves the highlighted point with

```js
var activePoint = points.find(p => activeDataPoints.includes(p.payload));
```

`activeDataPoints` is matched by axis category value, so a duplicate `04/27` matches both years' rows and `find` returns the first. The tooltip and cursor take an index-based path instead. That divergence is why the tooltip stayed correct and only the dot jumped.

Verified before merging: merge tree against `main` is clean, `typecheck` plus all 975 frontend tests green, and all four UI locales produce compact numeric ticks (`26/04/27` for zh/ja, `04/27/26` for en) so the axis doesn't get crowded.

Three small non-blocking cleanups if you'd like a follow-up — entirely optional:

1. `allowDuplicatedCategory={false}` on the `XAxis` is a no-op now that the keys are unique; I diffed the rendered output with and without it and it's identical. It's also backwards as a safety net — if keys ever did collide it would silently re-collapse them and bring this bug straight back. Safe to drop.
2. The tooltip now always carries a year, including for 24h and 7-day ranges. That might read better gated on `spansMultipleYears`, the same way the axis tick already is.
3. `rawDate` is now an exact duplicate of `xKey` and `hour` is unused (both predate this PR), and `UsageTrendStatLike` restates `DailyStats` from [`src/types/usage.ts`](https://github.com/farion1231/cc-switch/blob/7eb68f2dc75acbad336c30f07753e3e897bf3f4d/src/types/usage.ts#L103-L112) — a `Pick<DailyStats, ...>` would keep the two in sync if those backend fields are ever renamed.

PR #6285, opened 2026-08-09
feat(provider): add fuzzy search to model mapping dropdown

YUZHEthefool:
## Summary

The model mapping dropdown lists every fetched model grouped by vendor, with no way to filter. For providers that expose hundreds of models, finding the right one means scrolling a long flat list — as raised in #6275.

This adds a fuzzy search box to the top of that dropdown, so you can type part of a model id (or vendor name) and the list narrows instantly.

### What changed

- **`ModelDropdown`** (shared by the Claude Desktop, Codex, and OpenCode model-mapping forms) is converted from a plain `DropdownMenu` to a `Popover` + `Command` combobox — the same pattern already used by `OmoFormFields`. Powered by `cmdk`'s built-in fuzzy filter:
  - groups with no matches auto-hide; an empty state shows when nothing matches
  - the vendor name is indexed as a keyword, so models can be matched by vendor (e.g. typing `openai`) as well as by model id
  - the `{ models, onSelect }` API is unchanged, so all 5 call sites keep working
- **Consolidated two near-identical inline copies** of the same grouped dropdown onto `ModelDropdown`:
  - `ModelInputWithFetch` (Claude / Gemini / Codex-OAuth / xAI-OAuth model input)
  - the Copilot branch in `ClaudeFormFields` (maps `CopilotModel` → `FetchedModel`)
  
  Now every model picker gains the search and the duplicated grouping code is removed (net −112 / +94).
- **i18n**: added `searchModelPlaceholder`, `searchModelEmpty`, `searchModelAriaLabel` for `en`, `zh`, `zh-TW`, `ja`.

### Scope note

This is an enhancement to existing UI behavior, so the screenshots below are from the provider model-mapping section. Open a provider with model mapping enabled, fetch models, then click the chevron next to a model field and start typing to filter.

## Related Issue

Fixes #6275

## Screenshots

| Before | After |
|--------|-------|
| Flat vendor-grouped list, no search; long providers require scrolling the whole list | Search box at top of the dropdown; typing filters models (and vendors) instantly, empty groups hide |
image
image

## Checklist

- [x] `pnpm typecheck` passes
- [x] `pnpm format:check` passes
- [x] `cargo clippy` passes (no Rust code changed)
- [x] Updated i18n files (en / zh / zh-TW / ja) for the new user-visible strings
- [x] `pnpm test:unit` — 695 tests pass

## Verification

- `npx tsc --noEmit` — clean
- `npx vitest run` — 695 tests, 100 files, all passing
[...]

PR #6353, opened 2026-08-11
fix(provider): finish model dropdown follow-ups

YUZHEthefool:
## Summary

- Reuse the shared searchable ModelDropdown in Hermes and OpenClaw, removing their duplicated vendor-grouped dropdown implementations.
- Give the cmdk search inputs explicit accessible labels in ModelDropdown, OmoFormFields, and ProfileSwitcher.
- Add a regression test covering vendor filtering and model selection.

This follows the two follow-up items noted after #6285.

## Testing

- pnpm typecheck
- pnpm test:unit (102 files, 709 tests)
- Prettier check on all changed files
- git diff --check

### DIFF
--- a/src/components/profiles/ProfileSwitcher.tsx
+++ b/src/components/profiles/ProfileSwitcher.tsx
@@ -130,7 +130,7 @@ export function ProfileSwitcher({ activeApp }: ProfileSwitcherProps) {
           sideOffset={6}
           className="z-[100] w-64 p-0"
         >
-          <Command>
+          <Command label={t("profiles.searchPlaceholder")}>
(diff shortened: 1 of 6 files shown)


farion1231:
Merged — thank you for the quick follow-through! Both follow-up items landed exactly as hoped, and the regression test was a great unprompted addition: asserting the combobox by its accessible name means the labelling fix can't silently regress, and the vendor-keyword filtering is covered too. Net −61 lines while finishing the consolidation — Hermes and OpenClaw model pickers get search for free. Nice pair of PRs 🎉

PR #6362, opened 2026-08-11
fix(provider): compact StepFun before context limit

### LINKED ISSUE #6363: bug(claude): StepFun can reach context limit before auto-compaction
### Self Checks / 自检

- [x] I have searched for existing issues, including closed ones.

### CC Switch Version / 版本号

3.19.2

### Operating System / 操作系统

Windows

### Related App / 涉及应用

Claude Code

### Steps to Reproduce / 重现步骤

1. Add or select the built-in `StepFun` or `StepFun en` provider for Claude Code.
2. Use `step-3.5-flash-2603` in a long-running Claude Code session with substantial tool output.
3. Continue until the conversation approaches StepFun's 262,144-token upstream context boundary.
4. Observe that the generated Claude Code `settings.json` does not declare StepFun's effective context window or an earlier auto-compact window.
5. Claude Code can therefore continue growing the request until StepFun rejects it at the upstream context limit instead of compacting with enough headroom.

This also affects StepFun providers saved before a preset-level fix: updating CC Switch alone does not rewrite their stored provider settings, so only changing the built-in preset leaves existing users unfixed unless they delete/recreate or reapply the provider.

### Expected Behavior / 期望行为

CC Switch should configure Claude Code with StepFun-specific context limits:

```json
{
  "env": {
    "CLAUDE_CODE_MAX_CONTEXT_TOKENS": "262144",
    "CLAUDE_CODE_AUTO_COMPACT_WINDOW": "220000"
  }
}
```

The defaults should apply to both the `.com` and `.ai` Step Plan endpoints. Existing saved providers should receive them in the effective live settings, while explicit user overrides remain authoritative.

### Actual Behavior / 实际行为

The StepFun Claude Code presets currently provide the endpoint and model routing variables but omit both context-related `CLAUDE_CODE_*` variables. Claude Code has no StepFun-specific early-compaction boundary and may reach the provider's hard limit.

### Additional Context / 补充信息

This is specifically a **Claude Code `settings.json` issue**. It is not the Codex `config.toml` problem described in #5611 or #5742, and it does not involve `model_context_window`, `model_auto_compact_token_limit`, or `tool_output_token_limit`.

A complete fix should include:

- preset defaults for newly created `StepFun` and `StepFun en` providers;
- effective-live-settings injection for already-saved providers identified by the normalized Step Plan endpoint;
- preservation of explicit user values;
- mirror cleanup during live-settings backfill so injected defaults do not become persisted user configuration;
- regression tests for both regional endpoints, existing providers, user overrides, and backfill behavior.

Dopetaiga:
## Summary

- declare StepFun Coding Plan's 262,144-token context window in both Claude Code regional presets
- trigger auto-compaction at 220,000 tokens, leaving about 42K tokens for generation, tool results, and token-estimation variance
- apply the same defaults to effective live settings for StepFun providers saved before this preset change
- preserve explicit user overrides and strip program-injected defaults during backfill
- add regression coverage for both presets, both regional endpoints, existing providers, user overrides, and backfill behavior

## Root cause

The StepFun Claude Code presets configured the correct `/step_plan` endpoint and model routing, but did not declare the provider's context window or an earlier auto-compact boundary. Long-running sessions could therefore approach StepFun's 262,144-token hard limit before Claude Code compacted them.

A preset-only change would fix newly created or reapplied providers but not providers already stored in CC Switch. The live-settings path now recognizes the normalized `.com` and `.ai` Step Plan endpoints and injects defaults only when the provider does not already contain explicit values.

## Reproduction

This was reproduced with StepFun Coding Plan and `step-3.7-flash`. Claude Code sent requests with input-related usage of 261,589 and 261,962 tokens, immediately below StepFun's 262,144-token hard limit, and the upstream returned HTTP 400.

The current preset defaults to `step-3.5-flash-2603`; this change intentionally does not alter that default. The context settings apply when users select `step-3.7-flash` as well because both routes use the Coding Plan 256K window.

Fixes #6363.

## Validation

- `git diff --check`
- Prettier check on the changed TypeScript files
- full TypeScript `tsc --noEmit`
- `tests/config/claudeProviderPresets.test.ts`: 19 tests passed
- Rust regression tests added for live injection, endpoint matching, user overrides, and backfill cleanup; not executed locally because the available environment does not include a Rust toolchain

### DIFF
--- a/src-tauri/src/services/provider/live.rs
+++ b/src-tauri/src/services/provider/live.rs
@@ -30,6 +30,8 @@ use super::normalize_claude_models_in_value;
 const CODEX_OAUTH_CLAUDE_MAX_CONTEXT_TOKENS: &str = "372000";
 const CODEX_OAUTH_CLAUDE_AUTO_COMPACT_WINDOW: &str = "372000";
 const KIMI_FOR_CODING_CONTEXT_TOKENS: &str = "262144";
+const STEPFUN_CONTEXT_TOKENS: &str = "262144";
+const STEPFUN_AUTO_COMPACT_WINDOW: &str = "220000";
 
 /// Model env keys Claude Code may route requests through. The defaults above
 /// are calibrated against gpt-5.6's Codex catalog, so every configured model
@@ -76,6 +78,21 @@ fn is_kimi_for_coding_provider(provider: &Provider) -> bool {
         .map(str::trim)
         .map(|url| url.trim_end_matches('/'))
         == Some("https://api.kimi.com/coding")
+}
+
+fn is_stepfun_provider(provider: &Provider) -> bool {
+    provider
+        .settings_config
+        .pointer("/env/ANTHROPIC_BASE_URL")
+        .and_then(Value::as_str)
+        .map(str::trim)
+        .map(|url| url.trim_end_matches('/'))
+        .is_some_and(|url| {
+            matches!(
+                url,
+                "https://api.stepfun.com/step_plan" | "https://api.stepfun.ai/step_plan"
+            )
+        })
 }
 
 /// Claude Code assigns unknown non-Claude model ids a 200K context window.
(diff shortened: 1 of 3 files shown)


farion1231:
@codex review

PR #6366, opened 2026-08-11
fix(codex): keep fresher auth.json when restoring takeover backup

luantu:
## Problem

After a CC Switch rebuild/upgrade (any restart that re-runs proxy-takeover restore), the user's Codex login silently breaks: Codex, CodexBar and any other tool that reads `~/.codex/auth.json` all demand re-authentication. The login was fine before the restart.

## Root cause

The Codex CLI refreshes `~/.codex/auth.json` on its own — token rotation plus a `last_refresh` bump. CC Switch, however, only snapshots `auth.json` into the live backup (`proxy_live_backup`) when takeover is (re)activated or a provider is switched; it never updates the snapshot as the CLI refreshes the file.

The restore path `write_codex_live_verbatim` then **unconditionally** overwrote `auth.json` with the backup snapshot:

- `restore_live_config_for_app_inner` → `write_codex_live` → `write_codex_live_verbatim` → `write_codex_live_atomic(auth, cfg)` rewrites `auth.json` from the backup.

So a stale backup OAuth token (e.g. from the last takeover activation weeks/months earlier, long expired) clobbers the still-valid login. Every client that reads `auth.json` then fails auth and prompts re-login. Reproduction is stable: rebuild → app exits (`cleanup_before_exit` restores from backup) → next start re-takes-over → stale token overwrites a fresh login.

The existing "empty auth does not destroy a login" guard (`write_codex_live_verbatim`) only covered `{}`; a non-empty but expired OAuth snapshot still clobbered the live file.

## Fix

When both the backup `auth` and the current `~/.codex/auth.json` carry OAuth login material (`tokens.*`), compare `last_refresh`:

- current `auth.json` is fresher → restore only `config.toml`, leave `auth.json` untouched;
- otherwise keep the previous behavior (backup wins).

Non-OAuth backups (API-key only) are unaffected. `(Some(auth), None)` (auth-only backup) gets the same protection for consistency.

## Tests

- New unit test `codex_backup_auth_is_staler_than_live_keeps_fresher_login` covering: stale backup + fresh live → keep live; fresh backup + stale live → restore backup; no live auth.json → restore backup; non-OAuth backup → unchanged; missing `last_refresh` handling.
- Full `services::proxy` test suite passes; `cargo clippy --lib` clean.

### DIFF
--- a/src-tauri/src/services/proxy.rs
+++ b/src-tauri/src/services/proxy.rs
@@ -2953,6 +2953,44 @@ impl ProxyService {
         self.write_codex_live_for_provider(config, provider)
     }
 
+    /// 备份里的 Codex OAuth 登录材料是否比当前 `~/.codex/auth.json` 更旧。
+    ///
+    /// Codex 客户端会自行刷新 `auth.json`（token 轮换 + `last_refresh` 更新），而
+    /// CC Switch 的接管备份只在「接管激活 / 切换供应商」时生成，不随刷新更新。恢复时
+    /// 若无条件用备份覆盖，会用过期 OAuth token 冲掉本地仍有效的登录——表现为每次
+    /// 重新编译/重启后，Codex、CodexBar 等直接读 `auth.json` 的工具集体失效并要求重登。
+    ///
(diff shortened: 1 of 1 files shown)


farion1231:
@codex review

PR #6372, opened 2026-08-11
fix(grok): deduplicate npm launcher and native canonical installation on Windows

Semidia:
## Problem

Grok's npm package installs its native binary into `~/.grok/bin/grok.exe` while the npm launcher script (`grok.cmd`) lives in a separate npm prefix directory. The `enumerate_tool_installations` function treated them as two independent installations of the same version, causing false conflict warnings in the upgrade UI.

## Solution

Added `dedupe_grok_npm_canonical_installations` which merges npm launcher + native binary into one entry (keeping the one with `is_path_default`), triggered only for `tool == "grok"` on Windows.

## Verification

- Targeted test: `grok_npm_launcher_and_canonical_binary_are_one_installation` — passes
- Same-version launcher + native → deduped to 1 entry
- Different-version launcher + native → kept as 2 entries (no false merge)

### DIFF
--- a/src-tauri/src/commands/misc.rs
+++ b/src-tauri/src/commands/misc.rs
@@ -2024,6 +2024,8 @@ fn enumerate_tool_installations(tool: &str) -> Vec<ToolInstallation> {
     #[cfg(not(target_os = "windows"))]
     use std::process::Command;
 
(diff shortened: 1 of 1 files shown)


farion1231:
@codex review

PR #6386, opened 2026-08-12
feat: discover Claude sessions across accounts

NeoSpecies:
## Summary

Claude Code session management currently scans only `~/.claude/projects`. Claude Desktop and Cowork store Claude-compatible transcripts under per-account local-agent session directories, so switching accounts or clients makes existing conversations appear to disappear.

This PR makes session discovery account-aware while preserving the existing provider model:

- Scan the standard Claude Code root plus macOS Claude Desktop and Claude-3p nested `.claude/projects` roots.
- Deduplicate by canonical transcript path, while keeping distinct sessions with the same session ID separate.
- Reuse the same complete root set for deletion safety checks.
- Surface the Claude account email/name in session metadata and the session list.
- Add a Claude account filter and include the account label in full-text search.
- Ignore Claude agent sidecars and queue-only JSONL files.

No transcript, OAuth token, Keychain credential, or account configuration is copied or modified by this change. Resuming an archived transcript still depends on the currently authenticated account and endpoint permissions.

## Verification

- `pnpm exec tsc --noEmit`
- `pnpm exec prettier --check src/components/sessions/SessionManagerPage.tsx src/components/sessions/SessionItem.tsx src/hooks/useSessionSearch.ts src/types.ts`
- `pnpm exec vitest run tests/components/SessionManagerPage.test.tsx tests/components/sessionUtils.test.ts`
- Full frontend suite: 104 files, 714 tests passed
- `git diff --check`

Rust checks could not be run in this environment because Cargo/Rust toolchain is not installed.

## Notes

The macOS paths are guarded with `cfg(target_os = "macos")`; other platforms retain the existing `~/.claude/projects` behavior and can add their native Desktop paths separately.


farion1231:
@codex review

PR #6417, opened 2026-08-13
fix(db): secure SQLite file permissions on Unix

### LINKED ISSUE #3265: Database files and .cc-switch directory lack explicit permission restrictions
## Problem

The `~/.cc-switch/` directory and database files (`cc-switch.db`) are created without explicit permissions, inheriting the system default umask (typically `0o755` for directories and `0o644` for files). This allows other users on the same machine to read these files.

## Current State

Some sensitive files already have strict permissions set:
- `settings.json` → `0o600` (`settings.rs:504`)
- `copilot_auth.json` → `0o600` (`copilot_auth.rs:1282,1288`)
- `codex_oauth_auth.json` → `0o600` (`codex_oauth_auth.rs:817,823`)
- Gemini `.env` file → `0o600`, parent dir → `0o700` (`gemini_config.rs:170,185`)

However, the following locations lack permission protection:
- **`~/.cc-switch/` directory itself** — created via multiple `create_dir_all()` calls with no explicit permissions (`config.rs:184,198,206`, `database/mod.rs:101`, `database/backup.rs:63,309`, etc.)
- **`cc-switch.db` database** — created via `rusqlite::Connection::open()` with no explicit permissions (`database/mod.rs:104`)
- **Backup database files** — same issue (`database/backup.rs:324`)
- **`atomic_write` first-time writes** — when the target file doesn't exist, `fs::metadata()` fails silently and the new file gets default umask permissions (`config.rs:230-237`)

## Suggested Fix

1. Set `0o700` permissions on the `~/.cc-switch/` directory when first created
2. Set `0o600` permissions on database files after creation
3. Have `atomic_write` set reasonable default permissions (e.g., `0o600`) when the target file doesn't exist yet

## Environment

- macOS / Linux (all permission-related code is inside `#[cfg(unix)]` blocks)
- Windows uses ACLs and is not affected

la2278647-arch:
## Summary

Fixes #3265 — SQLite database permissions on macOS/Linux.

cc-switch stores sensitive user data in its SQLite database
(credentials, tokens, config). Previously, newly created database
files and backups used the default umask, which on a typical system
produces mode 644 (world-readable). On a shared or multi-user machine
this exposes the database to other local users.

## Changes

All additions are gated behind `#[cfg(unix)]`, so Windows
behavior is completely unchanged.

### `src-tauri/src/config.rs` — `atomic_write`
- When the config file doesn't already exist, the temp file is
  explicitly set to `0o600` before rename.
- Existing file permissions are still preserved via the existing
  `fs::metadata` branch.
- Mirrors the pattern already used in `settings.rs::atomic_write`.

### `src-tauri/src/database/mod.rs` — `init`
- DB parent directory: `0o700` (owner-only rwx).
- New database file (only when not pre-existing): `0o600`.

### `src-tauri/src/database/backup.rs` — import & backup paths
- Temp database used for SQL import: `0o600`.
- `backups/` directory: `0o700`.
- Backup database files: `0o600`.

## Notes on design

- Uses `fs::set_permissions` + `Permissions::from_mode` with the
  `std::os::unix::fs::PermissionsExt` trait, the same idiom already
  present elsewhere in this crate (see `settings.rs`).
- Every `set_permissions` call is guarded by `#[cfg(unix)]`; on
  Windows they are compiled away.
- The backup path now also calls `authorizer` on the destination
  connection before the backup step, matching the intent of the
  existing authorizer usage (defence-in-depth).

## Risk / verification

- `rustfmt` passes on all three files.
- `cargo build` was not runnable locally (this host lacks the MSVC
  linker toolchain that cc-switch needs on Windows) — CI will cover
  that.
- The changes are purely additive permission-hardening; they do not
  alter any public API, schema, or control flow.
- Existing database files are left untouched (permissions only set on
  *new* files), so no migration risk for current users.

### DIFF
--- a/src-tauri/src/config.rs
+++ b/src-tauri/src/config.rs
@@ -377,9 +377,13 @@ pub fn atomic_write(path: &Path, data: &[u8]) -> Result<(), AppError> {
     #[cfg(unix)]
     {
         use std::os::unix::fs::PermissionsExt;
+        let exists = path.exists();
         if let Ok(meta) = fs::metadata(path) {
             let perm = meta.permissions().mode();
             let _ = fs::set_permissions(&tmp, fs::Permissions::from_mode(perm));
+        }
+        if !exists {
+            let _ = fs::set_permissions(&tmp, fs::Permissions::from_mode(0o600));
         }
     }
 
--- a/src-tauri/src/database/backup.rs
+++ b/src-tauri/src/database/backup.rs
@@ -169,8 +169,10 @@ impl Database {
             source: e,
         })?;
         let temp_path = temp_file.path().to_path_buf();
-        let temp_conn =
+        let mut temp_conn =
             Connection::open(&temp_path).map_err(|e| AppError::Database(e.to_string()))?;
+        #[cfg(unix)]
+        let _ = fs::set_permissions(&temp_path, fs::Permissions::from_mode(0o600));
 
         // authorizer 只覆盖外部 SQL，执行完立刻摘掉：紧随其后的
         // `create_tables_on_conn` / `apply_schema_migrations_on_conn` 是本程序自己的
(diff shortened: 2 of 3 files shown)


farion1231:
@codex review

PR #6421, opened 2026-08-13
fix(streaming): skip empty reasoning_content to prevent content block fragmentation

U1traTC:
## Problem

When using an OpenAI-compatible upstream that returns both `reasoning_content` and `content` in every SSE chunk (e.g. GLM-5.2 via internal LLM proxy), Claude Code renders each text fragment on a separate line instead of concatenating them into continuous paragraphs.

## Root Cause

The streaming converter in `streaming.rs` guards text `content` with `!content.is_empty()` but does **not** apply the same check to `reasoning_content`.

Some providers return `reasoning_content: ""` (empty string, not null/absent) in every chunk, even during the text-only phase. Without the emptiness guard, each such chunk triggers:

1. Close the current text `content_block`
2. Open a new `thinking` `content_block` (empty content)
3. Immediately close it
4. Open another text `content_block` for the next fragment

This fragments the output into many separate content blocks, which Claude Code renders line-by-line.

## Fix

Add `if !reasoning.is_empty()` around the reasoning block logic, matching the existing pattern already used for text content on line 315.

## Verification

Tested with GLM-5.2 via an OpenAI Chat Completions proxy through CC Switch's local proxy (Claude takeover):

**Before** — each text fragment in its own content block:

### DIFF
--- a/src-tauri/src/proxy/providers/streaming.rs
+++ b/src-tauri/src/proxy/providers/streaming.rs
@@ -268,45 +268,47 @@ pub fn create_anthropic_sse_stream<E: std::error::Error + Send + 'static>(
 
                                         // 处理 reasoning（thinking）
                                         if let Some(reasoning) = &choice.delta.reasoning {
-                                            if current_non_tool_block_type != Some("thinking") {
-                                                if let Some(index) = current_non_tool_block_index.take() {
+                                            if !reasoning.is_empty() {
+                                                if current_non_tool_block_type != Some("thinking") {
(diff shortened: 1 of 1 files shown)


farion1231:
@codex review

farion1231:
Superseded by #7227 (@gongzhenhu), merged to main as b78192e8f. The production change is identical to yours (treat an empty `reasoning_content` as absent); that PR also carries regression tests that pin block index, type and order in both directions, so it was the one to take.

Thank you for the fix. The release note will credit you as well. Closing this one.

PR #6456, opened 2026-08-14
feat(skills): group installed skills by source

Dramwig:
## Summary / 概述

- Group installed Skills by existing `repoOwner/repoName` source metadata, with source keys matched case-insensitively.
- Put historical or unmanaged Skills missing either source field into the existing localized Local group.
- Add per-source tri-state bulk controls for Claude, Codex, Gemini, Grok Build, OpenCode, Hermes, and Pi when Pi is the active context.
- Keep repository bulk scope independent from search filtering: hidden sibling Skills from the same source are still included.
- Add independent expand/collapse state per repository group, with localized accessible labels and hover/focus disclosure controls.
- Reserve a consistent action column so repository controls, app toggles, update actions, and row deletion remain aligned.
- Keep the change frontend-only: no database, Rust, data-model, or i18n-key changes.

Scope note: #4367 contains a broader Skills-management redesign. This PR is intentionally limited to the repository-source grouping requested in #4450, avoiding manual-group and schema complexity. Repository-level update execution is not changed by this PR.

## Related Issue / 关联 Issue

Related to #4450

## Screenshots / 截图

Repository-grouped Skills management page

Repository group disclosure control on hover

The screenshots demonstrate repository-level grouping, per-source app controls, the localized Local fallback, and the repository disclosure control aligned with row actions.

Component tests cover grouping, independent collapse state, hover/focus disclosure styling, mixed and pending states, case-insensitive sources, Pi visibility, action-column alignment, and repository-wide bulk operations when sibling Skills are hidden by search.

## Verification / 验证

- `pnpm test:unit -- tests/components/UnifiedSkillsPanel.test.tsx` — 36 passed
- `pnpm typecheck`
- `pnpm format:check`
- `pnpm build:renderer`
- `git diff --check`
- `pnpm exec vitest run --maxWorkers=1 --minWorkers=1` — 882 passed

## Checklist / 检查清单

- [x] `cargo clippy` passes (if Rust code changed) / 未修改 Rust，Not applicable
- [x] Updated i18n files if user-facing text changed / 复用现有翻译键，无需修改 i18n

### DIFF
--- a/src/components/skills/UnifiedSkillsPanel.tsx
+++ b/src/components/skills/UnifiedSkillsPanel.tsx
@@ -954,18 +954,20 @@ const SkillSourceGroupSection: React.FC<SkillSourceGroupSectionProps> = ({
           onToggle={onToggleGroupApp}
         />
 
-        <Button
-          type="button"
-          variant="ghost"
-          size="icon"
-          className="h-7 w-7 shrink-0 opacity-0 transition-opacity group-hover:opacity-100 focus-visible:opacity-100"
-          aria-label={disclosureLabel}
-          aria-expanded={isExpanded}
-          title={disclosureLabel}
-          onClick={() => setIsExpanded((expanded) => !expanded)}
-        >
-          {isExpanded ? <ChevronUp size={14} /> : <ChevronDown size={14} />}
-        </Button>
+        <div className="flex w-[3.625rem] shrink-0 justify-end">
+          <Button
+            type="button"
+            variant="ghost"
+            size="icon"
+            className="h-7 w-7 shrink-0 opacity-0 transition-opacity group-hover:opacity-100 focus-visible:opacity-100"
+            aria-label={disclosureLabel}
+            aria-expanded={isExpanded}
+            title={disclosureLabel}
+            onClick={() => setIsExpanded((expanded) => !expanded)}
+          >
+            {isExpanded ? <ChevronUp size={14} /> : <ChevronDown size={14} />}
+          </Button>
+        </div>
       </div>
 
       {isExpanded &&
@@ -1136,7 +1138,7 @@ const InstalledSkillListItem: React.FC<InstalledSkillListItemProps> = ({
       />
 
       <div
-        className="flex-shrink-0 flex items-center gap-0.5 opacity-0 group-hover:opacity-100 transition-opacity"
+        className="flex w-[3.625rem] shrink-0 items-center justify-end gap-0.5 opacity-0 transition-opacity group-hover:opacity-100"
         style={hasUpdate ? { opacity: 1 } : undefined}
       >
         {hasUpdate && onUpdate && (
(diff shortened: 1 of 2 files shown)


SaladDay:
Thanks for keeping this PR focused. I tested it after merging it with the current main: the 36 targeted tests, the full frontend suite (882 tests), type checking, formatting, and the renderer build all pass. The repository grouping, mixed-state bulk toggles, search behavior, and accessibility work look good. I think this is worth merging after two small changes. 1) Please change Fixes #4450 to Related to #4450. The Lark skills in that issue are marked Local, and this PR puts every skill without repository metadata into one Local group. A group toggle can therefore change unrelated local skills, so this does not fully solve manual or local grouping. 2) The group tooltip repeats the app name, for example: owner/repo: Claude Enable all for Claude. Please build it from groupLabel and actionLabel only. After these changes, this should be ready to merge. Thank you for the clean and well-tested implementation.

Dramwig:
> Thanks for keeping this PR focused. I tested it after merging it with the current main: the 36 targeted tests, the full frontend suite (882 tests), type checking, formatting, and the renderer build all pass. The repository grouping, mixed-state bulk toggles, search behavior, and accessibility work look good. I think this is worth merging after two small changes. 1) Please change Fixes #4450 to Related to #4450. The Lark skills in that issue are marked Local, and this PR puts every skill without repository metadata into one Local group. A group toggle can therefore change unrelated local skills, so this does not fully solve manual or local grouping. 2) The group tooltip repeats the app name, for example: owner/repo: Claude Enable all for Claude. Please build it from groupLabel and actionLabel only. After these changes, this should be ready to merge. Thank you for the clean and well-tested implementation.

Addressed both requested changes:

- Changed `Fixes #4450` to `Related to #4450`.
- Simplified the group tooltip so it is built from `groupLabel` and `actionLabel` only.
 
I also merged the current `origin/main`. The 36 targeted tests, the full 882-test frontend suite, type checking, formatting, and the renderer build pass locally.

One separate follow-up design observation—not something I propose adding to this PR:

The recently added per-Skill update action introduces a second action slot on each Skill row. Because the update button is hidden when no update is available, this leaves visible empty space immediately to the left of the delete button.

I think it may be worth reconsidering whether updates should operate at Skill granularity or repository granularity. For a repository containing multiple Skills, I believe a repository-scoped update is simpler and less ambiguous: update all installed Skills from that source together, rather than allowing individual Skills from the same repository to enter potentially inconsistent states.

My preferred follow-up UI would be:

- Keep only the repository-level update action.
- Make the non-interactive background area of the repository header clickable for expand/collapse, while preserving keyboard accessibility.
- Use the current disclosure-button position for the repository update button when an update is available.
- Keep only the delete action on individual Skill rows.

This would remove the dedicated disclosure column and the empty update slot, while keeping the row actions visually aligned. If Skill-level updates are still considered important, I think the action-area layout should at least be revisited so an unavailable update action does not leave a permanent blank column.

farion1231:
@codex review

PR #6460, opened 2026-08-14
fix(deeplink): preserve Grok Build model catalogs

### LINKED ISSUE #6457: bug(grokbuild): deeplink import collapses multi-model config to one profile
## Bug

CC Switch 3.19.2 collapses a Grok Build provider deeplink containing a valid multi-model `config.toml` to one generated model profile.

## Reproduction

1. Build a `ccswitch://v1/import` provider link with:
   - `app=grokbuild`
   - `model=grok-4.6`
   - explicit `endpoint` and `apiKey` URL parameters
   - `configFormat=json`
   - Base64 JSON config whose `config` TOML contains both `[model."grok-4.6"]` and `[model."grok-4.5"]`, each with distinct `name` and `description`.
2. Import the provider into CC Switch 3.19.2.
3. Inspect the generated Grok Build `config.toml` or run `/model` in Grok Build.

## Actual behavior

Only the selected `model` URL parameter survives. The generated profile uses the provider card name as the model display name and omits the configured description. `/model` therefore exposes one generic entry instead of both configured models.

## Expected behavior

The imported provider should preserve all safe model profiles from the inline TOML, including each model's ID, display name, description, and context window, while overriding every profile's endpoint and API key with the explicit deeplink URL parameters.

## Root cause

`build_grokbuild_settings` reconstructs a single profile from standard deeplink fields after `merge_grokbuild_config` has parsed only the selected/default profile. The full inline model catalog is never carried into `settings_config`.

## Security boundary

A fix should not preserve untrusted per-profile credentials, `env_key`, endpoints, or backend values. Those must continue to be replaced by the explicit deeplink endpoint/API key and the supported Responses backend.

## Environment

- CC Switch 3.19.2
- Grok Build 1.0.3
- macOS arm64

nobody396:
Fixes #6457.

## Summary

Grok Build provider deeplinks currently collapse an inline multi-model `config.toml` into one profile and use the provider card name as that profile's model name. This patch preserves the safe model catalog fields so Grok Build's `/model` picker receives the intended entries.

For each imported profile it preserves only:
- profile/model id
- display name and description
- context window

The explicit deeplink `endpoint` and `apiKey` override every profile, and `api_backend` is forced to `responses`. Inline endpoints, API keys, `env_key`, and backend selections are not trusted or carried over.

The legacy single-model path now uses the model id for its model name/description instead of the provider card name.

## Validation

- `pnpm typecheck`
- `pnpm format:check`
- `pnpm test:unit` — 122 files / 874 tests passed
- `cargo test --manifest-path src-tauri/Cargo.toml` — 2472 lib tests plus all integration suites passed
- clean in-memory provider import integration test verifies both profiles persist and are written to the live Grok Build config
- local debug app E2E with Grok Build 1.0.3: `/model` showed Grok 4.5 and Grok 4.6 with correct names/descriptions; both completed a minimal Responses request against a local no-secret stub

### DIFF
--- a/src-tauri/src/deeplink/provider.rs
+++ b/src-tauri/src/deeplink/provider.rs
@@ -148,7 +148,7 @@ pub(crate) fn build_provider_from_request(
         AppType::Claude | AppType::ClaudeDesktop => build_claude_settings(request),
         AppType::Codex => build_codex_settings(request),
         AppType::Gemini => build_gemini_settings(request),
-        AppType::GrokBuild => build_grokbuild_settings(request),
+        AppType::GrokBuild => build_grokbuild_settings(request)?,
         AppType::OpenCode => build_opencode_settings(request),
         AppType::OpenClaw => build_additive_app_settings(request),
         AppType::Hermes => build_hermes_settings(request),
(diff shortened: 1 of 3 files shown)


farion1231:
@codex review

PR #6463, opened 2026-08-14
feat(pi): add session usage statistics

SaladDay:
## Summary

- Import usage-bearing entries from native Pi JSONL sessions into the existing usage dashboard, preserving Pi token, cache, model, provider, cost, and failure semantics.
- Keep active sessions incremental and bounded, while deduplicating copied history across Pi forks and v1 migrations even after detail rows have been rolled up and pruned.
- Add Pi to the dashboard filter and localized display, with conservative partial cache-write coverage for mixed OpenAI/Anthropic Pi providers.

Follow-up to #6064.

## Database

- Bump schema v16 to v17.
- Add one isolated session_usage_dedup table; no existing table or column is changed.
- Keep the table local alongside usage details, rollups, and session cursors during configuration sync.

The durable identity ledger is needed because request details are pruned after daily rollup, while a later Pi fork can copy those older entries into a new session file. A cursor alone cannot prevent that history from being counted twice.

## Scope

- No Pi gateway, routing, failover, provider configuration, prompt, Skill, MCP, plugin, theme, or default-model behavior changes.
- Existing session importers and usage query behavior are unchanged apart from registering the Pi source and its display label.
- Pi session discovery continues to use the native root and layout rules introduced in #6064.

## Validation

- Real Pi 0.84.0 with a configured API: created a parent session and a native --fork session through two real API calls.
- Imported the resulting 3 physical assistant rows as 2 unique requests; verified 66 input and 214 output tokens.
- Rolled up and pruned both details, cleared the file cursor, then rescanned: 0 reimports, 3 duplicate skips, 2 durable identities, 2 rolled-up requests, 0 detail rows.
- cargo fmt --manifest-path src-tauri/Cargo.toml -- --check
- cargo clippy --manifest-path src-tauri/Cargo.toml -- -D warnings
- cargo test --manifest-path src-tauri/Cargo.toml --lib: 2488 passed, 5 ignored
- pnpm typecheck
- pnpm test:unit: 875 passed
- pnpm format:check
- pnpm build:renderer
- git diff --check
- Repeated independent blind reviews covering data integrity, transactions, migrations, incremental sync, malformed input, performance, cross-platform behavior, frontend semantics, regressions, and test coverage.

### DIFF
--- a/src-tauri/src/database/backup.rs
+++ b/src-tauri/src/database/backup.rs
@@ -90,6 +90,7 @@ const SYNC_SKIP_TABLES: &[&str] = &[
     "proxy_live_backup",
     "usage_daily_rollups",
     "session_log_sync",
+    "session_usage_dedup",
 ];
 
 /// Tables whose local data is preserved from the live database during WebDAV import.
@@ -100,6 +101,7 @@ const SYNC_PRESERVE_TABLES: &[&str] = &[
     "proxy_live_backup",
     "usage_daily_rollups",
     "session_log_sync",
+    "session_usage_dedup",
 ];
 
 /// A database backup entry for the UI
--- a/src-tauri/src/database/mod.rs
+++ b/src-tauri/src/database/mod.rs
@@ -53,7 +53,7 @@ use std::sync::Mutex;
 
 /// 当前 Schema 版本号
 /// 每次修改表结构时递增，并在 schema.rs 中添加相应的迁移逻辑
-pub(crate) const SCHEMA_VERSION: i32 = 16;
+pub(crate) const SCHEMA_VERSION: i32 = 17;
 
 /// 安全地序列化 JSON，避免 unwrap panic
 pub(crate) fn to_json_string<T: Serialize>(value: &T) -> Result<String, AppError> {
--- a/src-tauri/src/database/schema.rs
+++ b/src-tauri/src/database/schema.rs
@@ -308,6 +308,26 @@ impl Database {
         )
         .map_err(|e| AppError::Database(e.to_string()))?;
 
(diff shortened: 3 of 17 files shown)


farion1231:
Merged — thank you for another excellent contribution!

Highlights from review:

- The durable dedup ledger is the right call. Fork-copied history re-entering after detail rollup/prune is a real double-counting hazard unique to Pi, and a cursor alone genuinely cannot cover it. The dual-fingerprint design (entry-id-based request identity + id-free canonical semantic identity, gated by `has_entry_id`) cleanly handles both fork copies and the v1→v3 in-place migration without cross-contaminating the two cases.
- Independently verified the fresh-input semantics against real Pi 0.84.0 sessions on my side: `totalTokens = input + output + cacheRead + cacheWrite` holds even for OpenAI Responses-backed providers, so the `FRESH` designation is correct. Field names, entry-id shape, and stopReason values all match the parser's assumptions.
- The tagged-revision encoding reusing `last_synced_at` is safe as implemented — the marker bits can never collide with a real epoch timestamp, and any untagged value degrades to a safe full rescan backed by the ledger.

Two small follow-ups if you're ever interested (no urgency, neither blocks anything):

1. Pi usage currently has no reset/rebuild entry point (Codex has its rebuild flow). Since the ledger is durable by design, a user who wants to recount from scratch has no way to do it.
2. A leftover empty `.jsonl` (or a file whose first line isn't a session header) logs a warning on every sync cycle indefinitely. Treating empty files as deferred instead would keep the logs quiet.

Thanks again — the validation rigor on this one (real fork + rollup/prune/rescan round-trip) is exactly what usage-accounting changes need.

PR #6471, opened 2026-08-15
fix(codex): keep deeplink homepage optional

AiW520:
## Summary

- keep `homepage` unset for Codex provider deep links when the link only supplies an API endpoint
- retain homepage inference and validation for every other provider type
- add an integration regression for the `custom` provider imported from `https://api.tu-zi.com/coding`
- matching storefront change: tuziapi/faka#20

## Root cause

The provider importer inferred a homepage for every app. A Codex deep link without `homepage` therefore stored `https://tu-zi.com`, even though that field is optional metadata and is not used for Codex API requests.

## Validation

- `cargo test --manifest-path src-tauri/Cargo.toml --test deeplink_import` (3 passed)
- `cargo test --manifest-path src-tauri/Cargo.toml --lib deeplink` (42 passed)
- `git diff --check`

The new integration test exercises the actual parser and provider service with the storefront URL shape, then verifies the stored provider has no website and its TOML contains `[model_providers.custom]`, `name = "custom"`, and the expected base URL.

## Rollout

Publish this change in a CC Switch release before rolling out tuziapi/faka#20 for the complete experience. The storefront link itself remains backward compatible, but users will only see an empty website field after installing a release containing this change.

## Rollback

Revert this commit. That restores the previous behavior of deriving and requiring a homepage for Codex imports.

### DIFF
--- a/src-tauri/src/deeplink/provider.rs
+++ b/src-tauri/src/deeplink/provider.rs
@@ -41,6 +41,9 @@ pub fn import_provider_from_deeplink(
         .clone()
         .ok_or_else(|| AppError::InvalidInput("Missing 'app' field for provider".to_string()))?;
 
+    let app_type = AppType::from_str(&app_str)
+        .map_err(|_| AppError::InvalidInput(format!("Invalid app type: {app_str}")))?;
+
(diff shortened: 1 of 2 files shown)


farion1231:
@codex review

PR #6472, opened 2026-08-15
fix(ci): run WSL2 contract tests via prebuilt binaries

ISuuuu:
## Summary

Fixes the intermittent `Backend Checks (Windows + WSL2 home)` failure caused by MSVC `mt.exe` errors `c1010070` / `LNK1327`.

The workflow now:

- compiles the library test binary while `TEMP` and `TMP` point to the Windows runner's native temporary directory;
- reads the executable path from Cargo's JSON artifact output and waits for Cargo to finish;
- propagates Cargo compilation failures;
- runs the prebuilt test binary directly after switching `TEMP` and `TMP` to the WSL2 UNC directory.

This prevents Cargo from relinking under `\\wsl.localhost`, where `link.exe` / `mt.exe` cannot create manifest temporary files.

## Scope

This PR changes only `.github/workflows/ci.yml`. The scheduled full-suite WSL2 workflow is unchanged and can be handled independently.

## Related

Related to #6428.

## Checklist

- [x] Latest head passed the full CI workflow, including the Windows 2025 + WSL2 contract job
- [x] Workflow YAML parses successfully
- [x] Final diff received two independent reviews with no P0, P1, or P2 findings

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -205,15 +205,37 @@ jobs:
 
       - name: Compile backend tests with native temp
         # link.exe/mt.exe cannot create manifests when TEMP points at a WSL UNC path.
-        run: cargo test --lib --manifest-path src-tauri/Cargo.toml --no-run
+        # Record the built test binary path so the run step below can execute
+        # it directly instead of re-invoking cargo with TEMP on \\wsl.localhost.
+        env:
+          TEMP: ${{ runner.temp }}
+          TMP: ${{ runner.temp }}
+        shell: pwsh
+        run: |
+          $testBinary = cargo test --lib --manifest-path src-tauri/Cargo.toml --no-run --message-format=json |
+            ForEach-Object {
+              if ($_ -match '"reason":"compiler-artifact"') {
+                $artifact = $_ | ConvertFrom-Json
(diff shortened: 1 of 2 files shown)


farion1231:
@codex review

SaladDay:
I removed the wsl2-nightly.yml changes to keep this PR limited to the flaky per-PR Windows + WSL2 check. The nightly workflow is not exercised by this PR and currently has a separate 90-minute timeout issue, so those changes cannot be validated here.

We can move the nightly prebuilt-binary changes to a separate PR and test them with workflow_dispatch.

The current PR CI is green, including the Windows + WSL2 job. @farion1231, could you review this PR?

farion1231:
Thanks @ISuuuu for tracking down the root cause of the flaky Windows + WSL2 job and fixing it the right way — reading the test binary path from cargo's JSON artifact output keeps cargo entirely out of the UNC-TEMP environment, so the mt.exe failure mode is eliminated by construction rather than by luck. Thanks also @SaladDay for the hardening commit (`-Wait` + exit-code propagation) and for trimming the scope.

A follow-up PR applying the same prebuilt-binary approach to `wsl2-nightly.yml` would be very welcome — it can be validated via `workflow_dispatch` independently of the nightly timeout issue.

PR #6476, opened 2026-08-15
fix(openclaw): quote root keys so openclaw.json stays valid JSON

ntdatt812:
Refs #6467.

## What happens

`~/.openclaw/openclaw.json` is parsed back with JSON5, but it is named `.json`, it ships as strict JSON, and third-party tools read it with a strict parser. Saving an OpenClaw provider wrote a top-level key as a bare identifier, which only JSON5 accepts, so a previously valid file stopped parsing. The reporter's case is Feishu's aily-cli adapter refusing OpenClaw with `is not valid JSON`.

## Why

The writer was already inconsistent. Nested keys are produced by `serde_json::to_string_pretty` in `value_to_rt_value`, so they are always quoted. Only the root key of a section went through `make_json5_key`:

```rust
fn make_json5_key(key: &str) -> RtJSONValue {
    if is_identifier_key(key) {
        RtJSONValue::Identifier(key.to_string())   // bare -> JSON5 only
    } else {
        RtJSONValue::DoubleQuotedString(key.to_string())
    }
}
```

`models` is a valid identifier, so the output was `models: {` at the root with `"providers": {}` nested underneath — exactly the mixture in the report.

The repo's own test suite already pinned this: `default_model_write_preserves_top_level_comments` asserted `written.contains("agents: {")`. That assertion is updated here.

## Change

Always quote root keys. Quoted keys are valid in both JSON and JSON5, so OpenClaw keeps reading the file and strict parsers stop failing. `is_identifier_key` had no other caller and is removed.

Two adjacent paths needed the same treatment to actually fix the reported symptom:

- **Self-heal.** `set_root_section` previously replaced only `existing.value`, so a file already broken by an earlier version would keep its bare key forever. It now also normalizes the key of the section it is rewriting. Sections that are not touched keep whatever form the user wrote, so hand-maintained JSON5 is not reformatted wholesale.
- **Fresh installs.** `OPENCLAW_DEFAULT_SOURCE` was itself JSON5-only (bare keys, single quotes, trailing commas), so a brand-new config could never be read by a strict parser. It is now strict JSON and matches `default_openclaw_config_value()`.

## Tests

Four tests, all in `openclaw_config`:

- `root_section_write_keeps_config_strict_json_parseable` — strict-JSON fixture stays strict JSON after a section write. This is the regression test for the report.
- `config_created_from_scratch_is_strict_json` — covers the no-file path.
- `rewriting_a_section_requotes_a_bare_root_key` — an already-degraded file recovers on the next save.
- `new_root_keys_are_always_quoted` — unit-level guard on `make_json5_key`.

`default_model_write_preserves_top_level_comments` now expects `"agents": {`; the rest of its assertions (comment preservation, written value) are unchanged. `remove_last_provider_writes_empty_providers_without_panic` already asserted the quoted nested form and is untouched.

## Scope

`openclaw_config.rs` is the only writer of this file; every other reference in `src-tauri` reads it. Comment preservation, backup behaviour and the on-disk conflict check are all unchanged.

One thing worth your call: I kept the normalization limited to the section being written. If you would rather have a single save repair every bare key in the document, that is a small change to `set_root_section` — but it would rewrite keys the user may have deliberately written in JSON5 style, so I did not assume it.

### DIFF
--- a/src-tauri/src/openclaw_config.rs
+++ b/src-tauri/src/openclaw_config.rs
@@ -20,8 +20,12 @@ use std::fs;
 use std::path::{Path, PathBuf};
 use std::sync::{Mutex, OnceLock};
 
+/// 新建配置文件时的初始内容。
+///
+/// 用严格 JSON 书写：OpenClaw 自己按 JSON5 读取，但该文件同样会被第三方工具
+/// 用严格 JSON 解析器读取，所以不应一上来就写出只有 JSON5 能接受的内容。
 const OPENCLAW_DEFAULT_SOURCE: &str =
-    "{\n  models: {\n    mode: 'merge',\n    providers: {},\n  },\n}\n";
+    "{\n  \"models\": {\n    \"mode\": \"merge\",\n    \"providers\": {}\n  }\n}\n";
 const OPENCLAW_TOOLS_PROFILES: &[&str] = &["minimal", "coding", "messaging", "full"];
 
 // ============================================================================
@@ -293,6 +297,9 @@ impl OpenClawConfigDocument {
             .iter_mut()
             .find(|pair| json5_key_name(&pair.key) == Some(key))
         {
+            // 该 section 本来就要重写，顺带把键名归一为带引号形式，
+            // 让旧版本写出的裸键在下一次保存时自愈。
+            existing.key = make_json5_key(key);
             existing.value = new_value;
             return Ok(());
         }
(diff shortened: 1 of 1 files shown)


ntdatt812:
Verified locally against the pinned toolchain from `rust-toolchain.toml` (1.95).

**`cargo test --lib openclaw_config`** — 9 passed, 0 failed.

To confirm the four new tests actually pin the bug rather than just passing, I reverted the three source changes and re-ran with the tests left in place:

```
test openclaw_config::tests::root_section_write_keeps_config_strict_json_parseable ... FAILED
test openclaw_config::tests::config_created_from_scratch_is_strict_json ... FAILED
test openclaw_config::tests::rewriting_a_section_requotes_a_bare_root_key ... FAILED
test openclaw_config::tests::new_root_keys_are_always_quoted ... FAILED
test openclaw_config::tests::default_model_write_preserves_top_level_comments ... FAILED

test result: FAILED. 4 passed; 5 failed
```

All five go green with the change applied.

**Full crate** — `cargo test --lib`: 2474 passed, 2 failed. Both failures are `codex_config::tests::resolve_catalog_rejects_symlink_escaping_config_dir` and `services::session_usage_grokbuild::tests::symlink_cycle_does_not_cause_stack_overflow`, which panic with `Os { code: 1314, message: "A required privilege is not held by the client." }` — Windows refuses symlink creation without elevation. I re-ran both with this branch's change stashed and they fail identically on the base commit, so they are environmental and pre-existing. (`proxy::hyper_client::tests::bytes_with_limit_aborts_reqwest_response_before_full_body_arrives` failed once and passed on re-run; it looks timing-sensitive and is unrelated to this change.)

**`cargo fmt --check`** and **`cargo clippy --lib --all-features -- -D warnings`** both clean.

farion1231:
@codex review

PR #6479, opened 2026-08-15
fix(zhipu): read the quota response when it has no data envelope

ntdatt812:
Refs #6402.

## What happens

The GLM Coding Plan card renders no tiers at all for some accounts, while the request itself succeeds — pointing a custom extractor script at the same endpoint shows the quota fine. So the transport is healthy; the built-in parser is what drops the data.

## Why

`zhipu_quota_from_body` requires `body.data`:

```rust
let data = match body.get("data") {
    Some(d) => d,
    None => return make_error("Missing 'data' field in response".to_string()),
};
```

and `parse_zhipu_token_tiers` then requires `data.limits`.

`GET /api/monitor/usage/quota/limit` has also been observed returning the limit list as a bare top-level array, with no envelope at all. From the dump in the issue:

```json
[
  {"type":"CREDIT_LIMIT","unit":3,"number":5,"usage":28000,"currentValue":2585,"remaining":25414,"percentage":9,"nextResetTime":1786592963348},
  {"type":"CREDIT_LIMIT","unit":6,"number":1,"usage":140000,"currentValue":58386,"remaining":81613,"percentage":41,"nextResetTime":1786692650981}
]
```

Against that body `body.get("data")` is `None`, so the parser returns `Missing 'data' field in response` before it ever looks at the entries.

## Scope note

The issue titles the `CREDIT_LIMIT` type filter, but that half is **already fixed** on `main` — the filter accepts `CREDIT_LIMIT` alongside `TOKENS_LIMIT`, case-insensitively. The part that still reproduces is the envelope, which the reporter flagged separately at the end of the report. This PR covers only that, hence `Refs` rather than `Fixes`.

## Change

Find the limit array where it actually is, rather than pinning one envelope shape — the bare array, or a `limits` array on whatever object we were handed. `unit`-based window classification and every other behaviour is untouched.

A body with no limits anywhere still returns an error, so a genuinely broken response is not silently converted into an empty-but-successful quota.

## Tests

Three added:

- `zhipu_quota_reads_a_top_level_limit_array` — the reporter's real response verbatim, asserting both tiers and their percentages
- `zhipu_quota_reads_limits_without_a_data_envelope` — `{ "limits": [...] }` with no `data`
- `zhipu_quota_still_reports_a_body_with_no_limits_anywhere` — an unrecognised body must still surface an error

Reverting the two source hunks and leaving the tests in place fails both new envelope tests with `error: Some("Missing 'data' field in response")` — the exact symptom from the report.

All 20 pre-existing `zhipu*` tests still pass unchanged.

`cargo test --lib`: 2480 passed, 2 failed. Both failures are `codex_config::tests::resolve_catalog_rejects_symlink_escaping_config_dir` and `services::session_usage_grokbuild::tests::symlink_cycle_does_not_cause_stack_overflow`, which panic with `Os { code: 1314, "A required privilege is not held by the client." }` — Windows refuses symlink creation without elevation, and they fail identically on the base commit. `cargo fmt --check` and `cargo clippy --lib --all-features -- -D warnings` are clean.

### DIFF
--- a/src-tauri/src/services/coding_plan.rs
+++ b/src-tauri/src/services/coding_plan.rs
@@ -252,13 +252,25 @@ fn classify_zhipu_window(item: &serde_json::Value) -> Option<ZhipuWindow> {
 ///
 /// 老套餐（2026-02-12 前订阅）只回 1 条
 /// `TOKENS_LIMIT`，自然降级为仅展示 `five_hour`；新套餐回 2 条。
+/// 定位配额条目数组。
+///
+/// 该端点的外层结构不止一种：历史上是 `{ "data": { "limits": [...] } }`，
+/// 线上也出现过直接返回顶层数组、完全没有外层包裹的情况（issue #6402）。
+/// 与其钉死一种形状，不如按已知的几种位置依次查找。
+fn zhipu_limit_entries(value: &serde_json::Value) -> Option<&Vec<serde_json::Value>> {
+    if let Some(entries) = value.as_array() {
+        return Some(entries);
+    }
+    value.get("limits").and_then(|v| v.as_array())
+}
+
 fn parse_zhipu_token_tiers(data: &serde_json::Value) -> Vec<QuotaTier> {
     type Entry = (Option<i64>, f64, Option<String>);
     let mut five_hour: Option<Entry> = None;
     let mut weekly: Option<Entry> = None;
     let mut unclassified: Vec<Entry> = Vec::new();
 
-    if let Some(limits) = data.get("limits").and_then(|v| v.as_array()) {
+    if let Some(limits) = zhipu_limit_entries(data) {
         for limit_item in limits {
             let limit_type = limit_item
                 .get("type")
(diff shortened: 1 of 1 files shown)


farion1231:
@codex review

ntdatt812:
Rebased onto `main` — mergeable again. The conflict was only in the test-module `use` list: `main` had added `detect_provider` and `parse_opencode_go_tiers` while this branch added `zhipu_quota_from_body`. Resolved as the union of both; no logic changed.

I re-checked that the fix is still needed rather than assuming. `main` today still has:

```rust
let data = match body.get("data") {
    Some(d) => d,
    None => return make_error("Missing 'data' field in response".to_string()),
};
```

which is exactly the path #6402 hits — the live response is a top-level array with no `data` envelope, so the user sees "Missing 'data' field" instead of their quota.

`cargo check --lib --tests` clean. I could not run the test binary here: linking fails on the `x86_64-pc-windows-gnu` toolchain because MSVC-style flags (`/MANIFEST:EMBED`) reach the GNU `ld`. That is my environment, not this branch — saying so rather than implying I ran a suite I didn't.

Issue #5802, opened 2026-07-27
Feature Request: Add WorkBuddy (CodeBuddy) as a supported tool


yanglei9491-hue:
### Self Checks / 自检

- [x] I have read the [FAQ](https://github.com/farion1231/cc-switch#faq) section in README.
我已阅读 README 中的[常见问题](https://github.com/farion1231/cc-switch#常见问题)。

- [x] I have searched for [existing issues](https://github.com/farion1231/cc-switch/issues), including closed ones.
我已搜索过[已有的 Issue](https://github.com/farion1231/cc-switch/issues)，包括已关闭的。

### Related App / 涉及应用

General / 通用

### Description
Please add **WorkBuddy** (also known as CodeBuddy) as a supported tool in CC-Switch.

### About WorkBuddy
WorkBuddy is a powerful AI coding assistant / IDE developed by Tencent. It supports MCP servers, Skills, plugins, and various AI model providers. Many CC-Switch users also use WorkBuddy as their daily AI development environment.

### What I'd like to see
1. Ability to launch WorkBuddy from CC-Switch's app bar
2. Manage WorkBuddy's provider configurations through CC-Switch's unified provider interface
3. MCP server management for WorkBuddy through CC-Switch
4. System tray quick switching for WorkBuddy

### Why it makes sense
- CC-Switch already supports 8 AI tools (Claude Code, Codex, Gemini CLI, etc.)
- WorkBuddy has a large and growing user base in China
- Many developers use both CC-Switch (for CLI tools) and WorkBuddy (as IDE) simultaneously
- WorkBuddy has configurable provider/model settings that could benefit from CC-Switch's unified management

### Additional Info
- WorkBuddy website: https://www.codebuddy.cn
- It supports MCP server configuration similar to Claude Desktop
- It has Skills plugin system
- It supports multiple AI model providers

Thanks for considering this feature!

### Contribution / 参与贡献

- [ ] I am interested in contributing to this feature.
我有兴趣参与开发此功能。

zhengsilin22:
As a dual user of ChatGPT Work and WorkBuddy, I'm really in favor of this suggestion. Both tools are my main tools for daily development. If we could manage provider configurations and MCP servers through CC-Switch, it would significantly reduce the cost of manually switching configurations between the two tools. Looking forward to seeing this feature available soon.

farion1231:
[...]

Issue #1855, opened 2026-04-02
🗺️ New CLI Tool Support Tracker / 新 CLI 工具支持追踪


farion1231:
# New CLI Tool Support Tracker / 新 CLI 工具支持追踪

This issue tracks all requests for supporting new CLI tools in CC Switch.
此 issue 统一追踪所有新 CLI 工具的支持请求。

## Currently Supported / 已支持
- ✅ Claude Code
- ✅ Codex CLI
- ✅ Gemini CLI
- ✅ OpenCode
- ✅ OpenClaw

## Requested Tools / 请求中的工具

| Tool | Issue(s) | Votes | Status |
|------|----------|-------|--------|
| Qwen Code | #937, #1154, #1793 | 3 | 📋 Requested |
| Pi Coding Agent / Pi-Mono | #1795, #1575, #2185 | 3 | 📋 Requested |
| Copilot CLI | #2173 | 1 | 📋 Requested |
| iFlow | #1793 | 1 | 📋 Requested |
| Qoder CLI | #1793 | 1 | 📋 Requested |
| Factory Droid | #1655 | 1 | 📋 Requested |
| VSCode-based Apps | #1618 | 1 | 📋 Requested |
| Openfang | #1221 | 1 | 📋 Requested |
| AMP | #1215 | 1 | 📋 Requested |
| ZeroClaw | #1200 | 1 | 📋 Requested |
| Xcode Claude Code | #933 | 1 | 📋 Requested |

## How to Vote / 如何投票
👍 React to this issue or leave a comment with the tool name you'd like supported.
请在此 issue 下点赞或评论你希望支持的工具名称。

## Criteria for Adding Support / 添加支持的标准
- Tool must have a stable CLI interface / 工具需有稳定的 CLI 接口
- Tool must use file-based configuration / 工具需使用文件配置
- Sufficient community demand / 足够的社区需求

Ttungx:
Qwen Code

gabrielmoreira:
Please consider adding support for https://github.com/can1357/oh-my-pi
 (oh-my-pi), which is a fork of the pi coding agent, using the directory ~/.omp instead of ~/.pi.
The authentication and model registration parts should be the same, or at least very similar.

izhoubin:
Xcode

nicriverhoo:
pi-mono +1

openapphub:
pi-mono +1

lfyzjck:
Pi Coding Agent  +1

[...]

PR #6480, opened 2026-08-15
feat(workbuddy): add WorkBuddy support

chenp0401:
## Summary

WorkBuddy (the agent shipped with Tencent CodeBuddy) keeps its model list in a flat `~/.workbuddy/models.json` array and has no notion of a provider — every entry carries its own `url` and `apiKey`:

```json
[
  {
    "id": "glm-5.1",
    "model": "glm-5.1",
    "name": "GLM 5.1",
    "url": "https://api.example.com/v1",
    "apiKey": "sk-xxx",
    "supportsToolCall": true,
    "contextWindow": 200000
  }
]
```

So switching gateways today means hand-editing that array. This PR maps the format onto CC Switch's provider model and surfaces WorkBuddy as a first-class app. Requested in #1855 (`CodeBuddy / WorkBuddy`, also #2816 / #4802 / #5802).

## Native data and compatibility

- **Aggregation** — model entries are grouped by `(url, apiKey)`; each group becomes one provider with `settings_config = { baseUrl, apiKey, models: [...] }`. Writing flattens them back into one array, re-injecting each model's `url` / `apiKey`. Additive mode, like OpenCode / OpenClaw / Hermes / Pi.
- **Provider ids** are derived from the gateway host's second-level domain (`https://api.deepseek.com/v1` → `deepseek`), with `-2`, `-3`… on collision. A UUID would not work here: the live file has no provider concept, so the id must stay re-derivable from the gateway address or delete / `isInConfig` would mismatch after a restart. The frontend mirrors this rule (`workbuddyProviderId.ts`) and both sides are tested against the same cases; each carries a comment pointing at the other.
- **Unknown model fields are passed through verbatim**, so a WorkBuddy release that adds a capability flag cannot lose data on round-trip.
- **Symlinks are preserved** — users often symlink `models.json` into a cloud-synced folder, and a `tmp` + `rename` write would replace the link with a regular file and break their sync. Writes canonicalize the target instead.
- **Only `id` is required** on a model entry, matching what WorkBuddy itself requires, so imported minimal entries stay editable without being forced to fill in `name` / `contextWindow` / `maxTokens`.

## Scope

Deliberately narrow, mirroring Pi (#6064):

- excluded from MCP, prompts and skills sync, and from the legacy v3.6.x MCP / prompt migration structs
- no local proxy takeover
- no config-dir override (the path is fixed), so WorkBuddy is excluded from directory settings
- hidden by default in `visibleApps`, opt-in from settings
- no DB schema change

## Validation

- `pnpm typecheck`, `pnpm format:check` — clean
- `pnpm test:unit` — 123 files / 880 tests passed
- `cargo fmt --check`, `cargo clippy --all-targets -- -D warnings` — clean
- `cargo test` — 2619 passed; 0 failed; 5 ignored
- 6 new backend unit tests: aggregation by gateway, write round-trip with credential re-injection, add/remove provider, symlink preservation (`#[cfg(unix)]`), provider-id derivation, malformed-JSON health scan. They use `#[serial]` and restore `CC_SWITCH_TEST_HOME` on drop so they cannot interfere with other modules' tests.
- 4 new frontend unit tests for the provider-id derivation.

## Notes

- The bundled brand icon is the official logo, downscaled from 1620×560 to 405×140 (30 KB) to match the other bitmap icons in `src/icons/extracted/`.
- #1855 lists the entry as `CodeBuddy / WorkBuddy`. I used app id `workbuddy` with display name "WorkBuddy"; the `AppType` parser also accepts `codebuddy` / `work-buddy` / `work_buddy`. Happy to rename if you prefer `codebuddy`.
- i18n covers en, zh, zh-TW and ja.
- Happy to split this into separate backend / frontend / icon PRs if that is easier to review — I have the branches prepared.


farion1231:
@codex review

chenp0401:
Thanks for the review — all four are real, and I've pushed fixes in `ccb00cd1`. Details below, plus one related issue I found while fixing them and one open design question I'd like your call on.

### P1 — Preserve provider identity when writing live models

Fixed by carrying the DB provider id through the write: `set_typed_provider(&provider.id, &cfg)`. I'd got this wrong by inventing a signature that re-derives the key, where `opencode_config::set_typed_provider` and `openclaw_config::set_typed_provider` both already take the id as their first argument — WorkBuddy was the odd one out.

The function now also drops any entry pointing at the same gateway under a different key, so two DB rows can't leave residue for each other.

**Related issue this exposed, not in your list:** `import_workbuddy_providers_from_live` matched existing DB rows with `existing_ids.contains(&derived_id)`. The live file has nowhere to store a CC Switch provider id, so a provider whose `baseUrl` had been edited never matched and was imported as a *second* row — the same duplication you described, arriving from the import side. Import now matches on `baseUrl` first (keeping the existing row's id) and only falls back to the derived id for gateways it has genuinely not seen.

### P1 — Derive the WorkBuddy ID before deep-link import

Took the "or reject this import path" option, mirroring Pi: deep links for WorkBuddy now return `InvalidInput` pointing users at the WorkBuddy provider page, which reserves the derived id. Duplicating the derivation into the deep-link path would have been a third copy of the rule to keep in sync, and the surrounding caller would still assign the `name-timestamp` id at lines 101-108. Removed the now-unused `build_workbuddy_settings` with it.

### P1 — Keep WorkBuddy credential files private

Fixed. The temp file is created with `0600` via `OpenOptions::mode`, and when `models.json` already exists its current mode is carried over instead, so a user who tightened it to `0400` doesn't get it widened by the next sync. A missing `~/.workbuddy` is created `0700`. Both are behind `#[cfg(unix)]` with a plain fallback elsewhere.

### P2 — Reject non-model documents in the health scan

Fixed. The scan now requires the root to be an array or the supported `{ "models": [...] }` wrapper, and warns otherwise, so `{}` / `null` / a bare string are reported before the next sync replaces the file.

### Tests

Four regression tests, one per issue:

- `editing_base_url_replaces_the_previous_gateway` — the stale gateway must be gone from the live file
- `suffixed_id_does_not_overwrite_its_host_sibling` — writing `alpha-2` leaves `alpha` intact
- `models_json_is_not_world_readable` — `0600` on create, and an existing `0400` is preserved
- `scan_health_rejects_non_model_documents` — `{}` / `null` / `"oops"` / `42` / `{"models":{}}` warn; `[]` and `{"models":[]}` don't

Local run: WorkBuddy suite 10/10, full `cargo test` 2622 passed / 0 failed, `cargo fmt --check`, `cargo clippy --all-targets -- -D warnings`, `pnpm typecheck`, `pnpm format:check` and `pnpm test:unit` all clean. CI on the new commit is waiting for approval.

### One open question

`provider_exists_in_live_config` and `remove_workbuddy_provider_from_live` are keyed by provider id, and the live file has no place to record one. So if a user edits a provider's gateway to a different host, the id derived from the live file still won't match the DB id, and the "is this provider in live config?" check returns false — on delete, its models would be left behind as orphans.

Two ways to close that hole, and I didn't want to pick unilaterally:

1. **Tag each model entry with the owning provider id** (e.g. `ccSwitchProviderId`). Unknown fields are already passed through verbatim, so this round-trips cleanly and removes every remaining identity mismatch. The downside is writing a foreign key into a third-party config file, and I'd want to verify WorkBuddy itself tolerates the extra field first.
2. **Match on `baseUrl` in those two functions**, which means widening their signatures to take the provider (or an app-specific lookup). Cleaner conceptually, but it touches the shared path used by OpenCode / OpenClaw / Hermes / Pi, so I didn't want to change it without your sign-off.

Happy to implement either. Option 1 is self-contained; option 2 is more correct but broader.

Issue #6473, opened 2026-08-15
[Bug] Anthropic → Codex Responses bridge does not preserve optional tool-argument semantics


kirkchinese:
## Self Checks / 自检

- [x] I have read the FAQ section in the README.
- [x] I have searched existing open and closed issues.

Related reports checked and determined not to be duplicates:

- #5774 concerns a model rejecting `json_schema` during Codex auto-approval.
- #6394 concerns optional sampling parameters such as `temperature` being rejected by a generic Responses upstream.
- Neither covers Anthropic optional tool properties being materialized by the Codex OAuth Responses route.

## CC Switch Version / 版本号

3.19.2

## Operating System / 操作系统

Linux

## Related App / 涉及应用

- Claude Code protocol route
- Other: DeepSeek Harness using the `anthropic-messages` protocol against the CC Switch local proxy

### Route

Configure the Claude/Anthropic-facing CC Switch route with a Codex OAuth provider:

```text
Client
  → Anthropic Messages
  → CC Switch 127.0.0.1:15721
  → apiFormat=openai_responses
  → https://chatgpt.com/backend-api/codex/responses
  → gpt-5.6-sol
```

The issue was reproduced with the current provider metadata equivalent to:

```json
{
  "providerType": "codex_oauth",
  "apiFormat": "openai_responses"
}
```

### Minimal direct probe

This reproduction bypasses DeepSeek Harness and sends an Anthropic tool schema directly to CC Switch.

Request body:

```json
{
  "model": "gpt-5.6-sol",
  "max_tokens": 256,
  "stream": false,
  "messages": [
    {
      "role": "user",
      "content": "Call the probe tool once with required_value set to ok. Do not invent optional values."
    }
  ],
  "tools": [
    {
      "name": "probe",
      "description": "Schema optional-field probe",
      "input_schema": {
        "type": "object",
        "properties": {
          "required_value": { "type": "string" },
          "optional_text": { "type": "string" },
          "optional_mode": {
            "type": "string",
            "enum": ["alpha", "beta"]
          }
        },
        "required": ["required_value"]
      }
    }
  ],
  "tool_choice": { "type": "tool", "name": "probe" }
}
```

Only `required_value` is listed in `required`.

The returned Anthropic `tool_use` block nevertheless contains values for every declared property:

```json
{
  "type": "tool_use",
  "name": "probe",
  "input": {
    "required_value": "ok",
    "optional_text": "",
    "optional_mode": "alpha"
  }
}
```

[...]

PR #6481, opened 2026-08-15
fix(proxy): fall back to buffered deltas when Read tool done event carries empty arguments

ashel-wang:
## Summary / 概述

The streaming Responses bridge buffers `function_call_arguments.delta` chunks for tools named `Read` (for `pages` sanitization) and emits them at the `done` event. Some OpenAI-Responses-compatible gateways (e.g. recent claude-code-router builds) send `response.function_call_arguments.done` with an **empty-string** `arguments` field while the complete arguments only exist in the deltas and the later `output_item.done` item.

The `done` handler only fell back to the buffered chunks when the `arguments` field was **missing** — not when present-but-empty. Result: every `Read` tool_use reached the Anthropic client with empty `input`, permanently breaking the tool on that route (tools with other names stream through unaffected).

Fix: treat empty-string `arguments` on the done event as absent (one-line `.filter`), mirroring the existing `output_item.done` handler which already filters empty strings. Added a regression test reproducing the exact event sequence captured from such a gateway.

## Related Issue / 关联 Issue

Related (same bridge, different symptom): #6473

## Steps to verify / 验证方式

Captured from a live claude-code-router container (podman, `/v1/responses`, GLM-5.3):

| Probe | Result before | Result after |
|---|---|---|
| CC Switch 15721 → openai_responses → CCR, `Read` tool call | tool_use `input: {}` | parameters intact |
| Same route, other tool names | unaffected | unaffected |
| `apiFormat=anthropic` pass-through | working | working |

```bash
cd src-tauri && cargo test --lib streaming_responses
# 22 passed; 0 failed (includes new regression test)
```

## Screenshots / 截图

N/A (protocol-level fix, covered by unit test)

## Checklist / 检查清单

- [x] `cargo clippy --lib -- -D warnings` passes / 通过 Clippy 检查
- [x] Rust unit tests pass (`cargo test --lib streaming_responses` — 22/22) / 单测通过
- [ ] `pnpm typecheck` / `pnpm format:check` — no TS/frontend change / 未改动前端代码

### DIFF
--- a/src-tauri/src/proxy/providers/streaming_responses.rs
+++ b/src-tauri/src/proxy/providers/streaming_responses.rs
@@ -879,6 +879,7 @@ pub fn create_anthropic_sse_stream_from_responses<E: std::error::Error + Send + 
                                             .get("arguments")
                                             .or_else(|| data.pointer("/item/arguments"))
                                             .and_then(|v| v.as_str())
+                                            .filter(|value| !value.is_empty())
                                             .map(str::to_string)
                                             .unwrap_or_else(|| {
                                                 tool_args_by_index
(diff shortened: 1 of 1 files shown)


farion1231:
@codex review

PR #6482, opened 2026-08-15
fix(updater): preserve command error details in failed update checks

SailingLoong:
Hi, and first of all — thank you for maintaining cc-switch. I maintain a small derivative project that keeps syncing with upstream, and I'd like to start giving back some fixes we've found along the way. This is the smallest one, so I figured it's a good icebreaker. 🙂

## Summary

- Replace `err instanceof Error ? err.message : fallback` with the existing `extractErrorMessage` helper in `UpdateProvider`.
- The updater plugin rejects with plain strings, so `instanceof Error` is always false and the real failure reason (network error, rate limit, malformed `latest.json`, …) gets replaced by the generic fallback message shown to the user.
- `extractErrorMessage` already exists in `src/utils/errorUtils.ts` and is used elsewhere; this is just the one call site that missed it.

## Validation

- `npx tsc --noEmit`
- `npx prettier --check src/contexts/UpdateContext.tsx`

Happy to adjust the approach, wording, or split it differently if you prefer. And if there's a contribution guide or a preferred channel for this kind of small fix, just point me at it.

### DIFF
--- a/src/components/settings/AboutSection.tsx
+++ b/src/components/settings/AboutSection.tsx
@@ -503,7 +503,10 @@ export function AboutSection({ isPortable }: AboutSectionProps) {
       }
     } catch (error) {
       console.error("[AboutSection] Check update failed", error);
-      toast.error(t("settings.checkUpdateFailed"));
+      toast.error(t("settings.checkUpdateFailed"), {
+        description: extractErrorMessage(error) || undefined,
+        closeButton: true,
+      });
     }
   }, [checkUpdate, hasUpdate, isPortable, resetDismiss, t]);
 
--- a/src/contexts/UpdateContext.tsx
+++ b/src/contexts/UpdateContext.tsx
@@ -8,6 +8,7 @@ import React, {
 } from "react";
 import type { UpdateInfo } from "../lib/updater";
 import { checkForUpdate } from "../lib/updater";
+import { extractErrorMessage } from "../utils/errorUtils";
 
 interface UpdateContextValue {
   // 更新状态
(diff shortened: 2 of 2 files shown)


farion1231:
@codex review

farion1231 (approved):
Thanks for the icebreaker — and the diagnosis is exactly right. The updater plugin serializes its error via `impl Serialize for Error` as `serialize_str(self.to_string())`, and Tauri's `invoke` rejects with that value unchanged, so `err instanceof Error` can never match and the real reason was always discarded. That's a general property of Tauri command/plugin rejections rather than something specific to this one call site, so it's a genuinely useful thing to have found.

One thing the change couldn't reach on its own: the `error` state you were populating has no consumers — `useUpdate()` is destructured in only two places and neither reads it — and the rethrown error was landing in a `toast.error()` with no `description`. So the fix was correct but not yet visible to anyone. I pushed a follow-up commit to your branch (`10641ac67`) that passes the recovered message into that toast, mirroring what the install-failure branch a few lines above already does. With both commits, the network error / rate limit / malformed `latest.json` reason now actually shows up under the failure toast.

Verified: merge tree against current `main` is clean, `tsc --noEmit` clean, 1087/1087 unit tests pass, Prettier clean, CI green.

On your question about a contribution guide — there's a short Contributing section in the README: `pnpm typecheck`, `pnpm format:check`, `pnpm test:unit` before opening a PR, and for new features please open an issue first so we can agree on the shape. Small focused fixes like this one need no preamble at all; just send them. Two things worth knowing since you maintain a downstream fork: presets and UI strings must land in all four locales (zh/en/ja/zh-TW) or the untranslated key leaks into the UI, and anything on the proxy forwarding path is held to a higher bar — request format and headers must stay byte-identical, since even a semantically equivalent rewrite breaks upstream prompt-cache prefixes.

Looking forward to the rest of what you've run into downstream.

PR #6483, opened 2026-08-15
feat(codex): allow custom reasoning effort levels

dulong:
## Background

`main` already has the per-model Reasoning Effort editor from #6228. Its UI and catalog generator intentionally expose only the built-in canonical levels (`none` through `ultra`), and the backend drops any unknown value.

That behavior prevents typos, but it also prevents configuring a provider that legitimately supports an additional effort value before CC Switch's built-in list is updated (for example, `adaptive`). This PR adds **manual provider-specific effort values** without changing the existing built-in shortcuts.

### UI

- Keep the existing canonical values as the primary selectable list, in their current ascending-depth order.
- Add a compact custom-level input and **Add** action inside the existing Reasoning Levels popover.
- Normalize custom input by trimming whitespace, converting it to lowercase, and de-duplicating it.
- Display custom values as removable chips so users can see exactly which non-standard values will be written.
- Include both canonical and custom selected values in the existing default-level selector.
- Show an explicit warning that the value must be supported by both Codex and the upstream provider; CC Switch does not claim that an arbitrary value is valid.

No model-list API probing or automatic capability inheritance is added. A provider-specific value is only emitted after the user explicitly enters it.

### Persistence and generated catalog

- `normalizeCodexCatalogModelsForSave` applies the same trim/lowercase/de-duplication rules before saving the provider configuration.
- A default level is saved only if it is present in that model's declared level list.
- The Rust catalog generator preserves non-empty custom values instead of dropping them.
- Known levels keep their canonical ordering and official descriptions. Custom levels are appended after the canonical values, retain their declaration order, and receive the safe fallback description:

  ```json
  { "effort": "adaptive", "description": "Reasoning effort: adaptive" }
  ```

- Existing provider configurations that use only canonical values retain the current behavior and catalog output.

## Why keep unknown values?

The model catalog is an extension point for custom/upstream providers. Rejecting every value outside CC Switch's static list means users cannot use a capability that their provider and Codex version already support. The UI warning and generic description make this an explicit user choice rather than silently treating the value as universally supported.

## Reviewer test plan

1. Open a Codex provider's **Model Mapping** and add a model row.
2. Open **Reasoning Levels**.
3. Enter `Adaptive` in the custom input and click **Add**.
4. Verify that it becomes the `adaptive` chip, can be removed, and is available as a default level.
5. Save the provider and verify the generated `supported_reasoning_levels` catalog contains `adaptive` with the fallback description.

## Validation performed

- `./node_modules/.bin/vitest run tests/components/ProviderForm.codexCatalog.test.ts`
- `./node_modules/.bin/vite build`
- `RUSTUP_TOOLCHAIN=stable cargo test codex_config` (83 tests passed)
- `RUSTUP_TOOLCHAIN=stable cargo fmt --check`
- Manual test in the locally built `/Applications/CC Switch.app`: add `Adaptive` through the Codex Reasoning Levels popover and verify it is normalized to `adaptive` and rendered as a removable custom-level chip.


farion1231:
@codex review

farion1231 (commented):
Thanks for the PR — the implementation itself is clean: normalization is applied consistently on both the form side and the catalog generator, the default level is validated against the declared list, locale strings are complete, and tests were updated alongside. It builds naturally on #6228.

Before deciding on the direction, one question: **do you have a concrete provider/model today that needs an effort value outside the built-in eight?** So far we haven't seen a provider document a value like `adaptive` — the vendors we track all fit within `none…ultra`, and Codex's own enum matches that list. Knowing the real use case would help us judge whether to take this now or hold it until actual demand shows up, since every extension point we ship becomes a compatibility surface we maintain.

One related note that may affect your answer: custom values currently only reach the upstream on the native Responses path. When a provider is routed through the local proxy's Chat or Anthropic conversion, unknown efforts are dropped or rewritten by the converters (see [`map_reasoning_effort`](https://github.com/farion1231/cc-switch/blob/97ae84aea18f1a1f5df5e9115ff2960d2fc4d308/src-tauri/src/proxy/providers/transform_codex_chat.rs#L458-L499)) — the request still succeeds, but the custom level silently does nothing. So if your target scenario is a Chat-routed provider, the feature as it stands wouldn't take effect there yet. Curious which path your use case is on.

dulong:
Closing this proposal as requested; the design is being reconsidered.

PR #6486, opened 2026-08-15
fix(codex): preserve user config sections during live takeover rebuild

RemiliaNyaa:


RemiliaNyaa:
由提交者主动关闭：本人需先在本地完整测试后再重新提交。

PR #6490, opened 2026-08-16
feat(provider): add unipaxtools presets

timneal0301-ship-it:
## Summary / 概述

Adds **unipaxtools** (`https://us-proxy.unipaxtools.com`) as a built-in provider preset for Claude Code. It is an aggregator relay serving both the Claude and GPT families behind one key, so it is registered under `category: "aggregator"`.

新增 **unipaxtools** 中转服务商预设（Claude Code）。该站同时提供 Claude 与 GPT 系列，一个密钥通用，因此归类为 `aggregator`（聚合网站）。

**Claude Code only, on purpose / 只做了 Claude Code，是有意为之**

The OpenAI-format surfaces (Codex, Claude Desktop, Hermes, OpenCode, OpenClaw) each require explicit model IDs in their preset. I do not have a verified model catalog for this provider yet, so those are left out rather than guessed at — a wrong model ID in a preset fails at request time with a confusing error, which is worse than no preset. Happy to follow up with a second small PR once the list is confirmed.

Codex 等 OpenAI 格式的入口，预设里必须写死具体模型名。目前没有经过实测确认的模型清单，与其猜一个写上去（写错会在调用时报出难以定位的错误），不如先不放。等确认后再单独提一个小 PR。

**One provider-specific gotcha, pinned by a test / 一个坑，用测试锁住了**

This provider requires the **bare domain** for `ANTHROPIC_BASE_URL` — a `/v1` suffix or a trailing slash both get rejected upstream as an invalid path and surface as a `401`, which reads like a key problem and sends users down the wrong debugging path. `src/config/unipaxtoolsProviderPresets.test.ts` asserts the base URL and every `endpointCandidates` entry has neither, so a future edit can't silently reintroduce it.

这家的 base url 必须是裸域名：带 `/v1` 或者结尾带斜杠都会被上游判成非法路径，直接返回 `401` —— 看起来像密钥问题，实际不是，很容易查错方向。测试里对 base url 和 `endpointCandidates` 都断言了这两点。

**Icon / 图标**

`src/icons/extracted/unipaxtools.png` is the vendor's own logo (256×256, downscaled from their 512×512 original), registered in `iconUrls` + `metadata.ts` alongside the other URL-based icons. Brand color `#0FA37F` is taken from the site's own CSS token, not eyeballed.

## Related Issue / 关联 Issue

None — new provider preset. / 无，新增服务商预设。

## Screenshots / 截图

Not included: this machine has no Rust toolchain, so the full Tauri app was not launched. The change is data-only (a preset entry plus its icon) and renders through the existing provider-list UI with no layout changes.

未提供：本机没有 Rust 工具链，Tauri 整包没跑起来。本次改动是纯数据（一条预设 + 图标），走现有服务商列表 UI，没有任何布局改动。

## Checklist / 检查清单

- [x] `pnpm test:unit` passes (880 tests, incl. 4 new) / 单元测试通过（880 条，含新增 4 条）
- [x] `pnpm build:renderer` passes / 前端构建通过
- [ ] `cargo clippy` — N/A, no Rust code changed / 不适用，未改动 Rust 代码
- [ ] i18n — N/A, no user-facing strings added (the provider name comes from the preset data) / 不适用，未新增界面文案，服务商名称来自预设数据本身


farion1231:
@codex review

PR #6505, opened 2026-08-16
feat(codex): map fixed reasoning levels to provider values

dulong:
## Background

PR #6228 added per-model Reasoning Effort declarations to the generated Codex model catalog. A follow-up experiment in #6483 allowed arbitrary catalog effort strings, but end-to-end testing with Codex Desktop showed an important limitation: the Desktop picker filters catalog values through a built-in Codex effort vocabulary instead of rendering arbitrary provider-native strings.

This PR takes a different approach and is intentionally independent from #6483:

- expose a stable set of Codex-facing levels in the generated catalog;
- let users map each Codex level to the value required by the upstream provider;
- let users customize the catalog description for every level.

#6483 is left unchanged so the two designs can be reviewed independently.

### 1. Six Codex-facing levels enabled by default

Every new custom catalog model starts with these six levels selected:

- `low` — shown as **Light** by Codex Desktop
- `medium` — **Medium**
- `high` — **High**
- `xhigh` — **Extra High**
- `max` — **Max**
- `ultra` — **Ultra**

These canonical values are written into `supported_reasoning_levels` in ascending order. Existing configurations may still select a subset.

Codex Desktop contains labels for all six values, but the final visibility of `max` and `ultra` can still depend on the installed Codex version, account feature flags, and Desktop settings. CC Switch's responsibility is to declare them in the catalog by default rather than silently discarding them.

`none` and `minimal` are intentionally not enabled by this UI because this design focuses on the six user-requested reasoning-depth levels.

### 2. Per-level provider mapping

Each selected Codex-facing level stores an optional provider-native value:

```json
{
  "reasoningEffortMappings": [
    {
      "level": "low",
      "upstreamValue": "light",
      "description": "Fast provider reasoning"
    },
    {
      "level": "xhigh",
      "upstreamValue": "max",
      "description": "Maximum provider reasoning"
    },
    {
      "level": "ultra",
      "upstreamValue": "unlimited"
    }
  ],
  "defaultReasoningLevel": "xhigh"
}
```

The generated Codex catalog exposes the canonical values (`low`, `xhigh`, `ultra` in this example). It never exposes `light` or `unlimited` as picker values.

When a request passes through CC Switch local routing, the selected Codex level is replaced with its model-specific `upstreamValue`. Leaving `upstreamValue` blank preserves the original Codex value.

This separation is important because provider vocabularies are not uniform and may evolve independently from Codex's picker vocabulary.

### 3. User-authored descriptions

Every selected level has an optional description input. When blank, CC Switch writes its maintained default description. When provided, the custom description is written into the model catalog beside the canonical Codex effort.

Provider-native values are deliberately never written into the generated catalog description or effort field unless the user explicitly enters them as description text.

### 4. Protocol coverage

Mappings are model-specific and cover every Codex routing path:

- native Responses passthrough: rewrites `reasoning.effort`;
- Responses -> Chat Completions: writes the mapped value using the configured parameter shape (`reasoning_effort` or `reasoning.effort`);
- Responses -> Anthropic Messages: applies the mapped value after conversion, writing `output_config.effort` without changing `thinking.type`.

Applying the mapping at the protocol boundary prevents provider-native values from being passed through converters that only understand canonical Codex levels.

## UI

The existing model row stays compact. Its Reasoning Levels field opens a popover containing:

1. the six Codex-facing levels, selected by default for a new model;
2. one compact row per selected level with:
   - Codex level;
   - editable upstream value;
   - editable description;
3. the default-level selector.

No separate large settings section is added to the provider form.

An explicit empty selection remains distinguishable from a model that has never configured mappings, so reopening the form does not unexpectedly re-enable levels the user removed.

## Catalog and routing boundaries

The two value domains remain separate throughout the implementation:

```text
Codex picker level
        ↓
cc-switch-model-catalog.json: supported_reasoning_levels[].effort
        ↓ request enters CC Switch routing
model-specific reasoningEffortMappings lookup
        ↓
provider-native request value
```

For example:

```text
Codex xhigh → catalog xhigh → route mapping → Anthropic output_config.effort=max
```

Without routing takeover, CC Switch cannot rewrite the outgoing request, so `upstreamValue` mappings are not applied. Catalog levels and custom descriptions still work without routing.

## Compatibility and migration

- Existing `reasoningLevels` arrays from #6228 remain readable.
- On load/save they migrate to `reasoningEffortMappings`.
- Legacy `max` values are now retained instead of being discarded.
- Unknown values remain filtered out of the Codex-facing catalog.
- Providers without an explicit saved mapping receive the six default rows when edited as a custom catalog model.
- Explicit mapping subsets and explicit empty mappings are preserved by the frontend state/normalization path.
- Upstream mapping only takes effect when requests pass through CC Switch routing.

## Validation

Completed locally on August 16, 2026:

```bash
pnpm typecheck
pnpm test:unit
pnpm format:check
pnpm build:renderer
cargo test --manifest-path src-tauri/Cargo.toml codex_config::tests:: --lib
cargo test --manifest-path src-tauri/Cargo.toml proxy::providers::codex::tests:: --lib
rustfmt --edition 2021 src-tauri/src/codex_config.rs
git diff --check
```

Results:

- frontend tests: **122 files / 876 tests passed**;
- Codex catalog/config tests: **82 passed**;
- Codex provider/routing tests: **45 passed**;
- TypeScript, Prettier, renderer production build, Rust formatting, and diff checks passed.

## Reviewer test plan

1. Add or edit a custom Codex provider with a model catalog.
2. Add a model and open its Reasoning Levels popover.
3. Confirm `low`, `medium`, `high`, `xhigh`, `max`, and `ultra` are selected by default.
4. Remove one level, save, reopen the provider, and confirm the subset is preserved.
5. Enter an upstream value and a custom description for one or more levels.
6. Select a default and save.
7. Confirm the provider stores `reasoningEffortMappings`.
8. Confirm the generated catalog contains the canonical Codex levels and supplied descriptions, not provider-native values.
9. With local routing enabled, send a request using a mapped level and confirm the upstream request carries its configured provider-native value.
10. If testing `max` or `ultra` in Codex Desktop, account for the Desktop build/settings/feature-gate visibility rules described above.

### DIFF
--- a/src-tauri/src/codex_config.rs
+++ b/src-tauri/src/codex_config.rs
@@ -548,9 +548,9 @@ fn codex_catalog_input_modalities(
     modalities.iter().map(|item| (*item).to_string()).collect()
 }
 
-/// Stable reasoning levels currently rendered by Codex Desktop for custom
-/// catalog entries. Provider-native values are kept in CC Switch settings and
-/// are never written into `supported_reasoning_levels`.
+/// Codex-facing reasoning levels exposed by default for custom catalog entries.
+/// Provider-native values are kept in CC Switch settings and are never written
+/// into `supported_reasoning_levels`.
 const CODEX_REASONING_LEVEL_DESCRIPTIONS: &[(&str, &str)] = &[
     ("low", "Fast responses with lighter reasoning"),
     (
@@ -559,6 +559,8 @@ const CODEX_REASONING_LEVEL_DESCRIPTIONS: &[(&str, &str)] = &[
     ),
     ("high", "Greater reasoning depth for complex problems"),
     ("xhigh", "Extra high reasoning depth for complex problems"),
+    ("max", "Maximum reasoning depth for the hardest problems"),
+    ("ultra", "Ultra reasoning depth"),
 ];
 
 fn codex_reasoning_level_description(effort: &str) -> Option<&'static str> {
@@ -3539,6 +3541,8 @@ base_url = "https://production.api/v1"
                             { "level": "xhigh", "upstreamValue": "max", "description": "Maximum provider reasoning" },
                             { "level": "low", "upstreamValue": "light", "description": "Fast provider reasoning" },
                             { "level": "medium" },
+                            { "level": "max", "upstreamValue": "maximum" },
+                            { "level": "ultra", "description": "Unlimited provider reasoning" },
                             { "level": "bogus", "upstreamValue": "turbo" }
                         ],
                         "defaultReasoningLevel": "xhigh"
@@ -3568,7 +3572,7 @@ base_url = "https://production.api/v1"
                 .filter_map(|level| level.get("effort").and_then(Value::as_str))
                 .collect()
         };
-        assert_eq!(efforts(0), vec!["low", "medium", "xhigh"]);
+        assert_eq!(efforts(0), vec!["low", "medium", "xhigh", "max", "ultra"]);
         assert_eq!(
             models[0]["supported_reasoning_levels"][0]["description"],
             "Fast provider reasoning"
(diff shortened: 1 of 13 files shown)


dulong:
Closing this proposal as requested; the reasoning-effort design is being reconsidered.

PR #6506, opened 2026-08-16
fix(codex): cancel stale device login after clear

SaladDay:
## Summary

Fix a race between starting Codex Device Code login and clearing all Codex auth.

A response started before the clear can no longer add a pending login afterward.

Follow-up to #3879.

## Tests

- `cargo test --manifest-path src-tauri/Cargo.toml --lib`
- `cargo clippy --manifest-path src-tauri/Cargo.toml --lib --tests -- -D warnings`

### DIFF
--- a/src-tauri/src/proxy/providers/codex_oauth_auth.rs
+++ b/src-tauri/src/proxy/providers/codex_oauth_auth.rs
@@ -22,7 +22,10 @@ use std::collections::HashMap;
 use std::fs;
 use std::io::Write;
 use std::path::PathBuf;
-use std::sync::Arc;
+use std::sync::{
+    atomic::{AtomicU64, Ordering},
+    Arc,
+};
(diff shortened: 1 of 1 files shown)


farion1231:
@codex review

farion1231:
Thanks @SaladDay for the quick turnaround on this — it closes out the login_epoch follow-up from the #3879 review exactly as scoped: the epoch bump sits inside the pending-map write lock so the check-and-insert has no TOCTOU window, invalidation is deliberately limited to clear-all (removing account A can't kill a parallel login of account B), and extracting the registration step made the race window unit-testable without mocking the network. Clean work.

The other two follow-up items from #3879 (WSL/exFAT hard-link probe fallback, and the native-login-exists loading flash) are still open if you're ever interested — no pressure either way.

PR #6507, opened 2026-08-16
fix(provider): harden IME-safe provider fields

yovinchen:
## Summary

- adopt ImeSafeInput for the shared provider name, notes, and website URL fields
- force-commit interrupted IME compositions on blur and reconcile idle drafts with the external value
- commit OpenCode model and option key edits from the blur event value instead of stale local React state
- add component and form-level regression coverage for the follow-up paths

## Review follow-up

This is a follow-up to PR #6333 based on the maintainer review comment:
https://github.com/farion1231/cc-switch/pull/6333#issuecomment-5289108906

It addresses both non-blocking follow-ups from that review. The underlying bug is the macOS IME issue tracked in #6308, where parent re-renders during marked-text composition could produce duplicated or reordered characters. The blur hardening also covers the window-switch path where compositionend may not arrive, and the OpenCode JSON option path where parse/stringify can leave the prop textually unchanged while the local draft diverges.

## Validation

- focused IME/provider tests: 5 files, 34 tests passed
- pnpm test:unit: 130 test files, 980 tests passed
- pnpm typecheck
- Prettier check on all changed files
- git diff --check

Related to #6308 and follow-up to #6333.

### DIFF
--- a/src/components/providers/forms/BasicFormFields.tsx
+++ b/src/components/providers/forms/BasicFormFields.tsx
@@ -8,7 +8,7 @@ import {
   FormLabel,
   FormMessage,
 } from "@/components/ui/form";
-import { Input } from "@/components/ui/input";
+import { ImeSafeInput } from "@/components/ui/ime-safe-input";
 import { Button } from "@/components/ui/button";
 import { ArrowLeft } from "lucide-react";
 import {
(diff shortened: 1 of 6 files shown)


farion1231:
@codex review

farion1231:
Merged — thanks for coming back to close out both follow-ups so thoroughly! 🎉

A couple of things I appreciated in review:

- The blur hardening covers exactly the two paths that needed it: force-committing interrupted compositions on window-switch blurs (with the duplicate-event guard still armed for a late `compositionend`), and using idle blur as the reconciliation point for parse→stringify round-trips that leave the prop textually unchanged.
- Reading the committed value from the blur event itself in the OpenCode fields was the right call — without it, the force-committed composition would have needed a second blur to reach the form. Nice catch on the closure staleness there.
- The test coverage is genuinely valuable here: these event-order edge cases only reproduce on real macOS IME setups, so having them locked into regression tests is worth every line.

Great follow-through from #6333 — the IME story is in solid shape now.

PR #6509, opened 2026-08-16
fix(proxy): preserve terminal web search output order

allenxu09:
## What changed

- Reuse the established Responses JSON conversion when an SSE stream contains only `response.completed`.
- Preserve terminal output order for `message("Before") -> web_search_call -> message("After")`.
- Add a regression test that asserts `text -> server_tool_use -> web_search_tool_result -> text`.

## Why

Follow-up to #5681. The terminal-only SSE path collected messages for later but emitted WebSearch blocks immediately, moving text that preceded the search behind its result.

## Impact

Incremental SSE streams keep using the existing reconciliation state machine. Only the no-increment `response.completed` fallback is routed through the already-correct JSON conversion path.

## Validation

N/A

### DIFF
--- a/src-tauri/src/proxy/providers/streaming_responses.rs
+++ b/src-tauri/src/proxy/providers/streaming_responses.rs
@@ -3489,7 +3489,9 @@ fn create_anthropic_sse_stream_from_responses_raw<E: std::error::Error + Send + 
                                     continue;
                                 }
                                 if event_name == "response.completed"
-                                    && !has_sent_message_start
+                                    && next_content_index == 0
+                                    && !has_substantive_output
+                                    && open_indices.is_empty()
                                     && response_obj.get("output").and_then(Value::as_array).is_some()
                                 {
                                     for event in responses_json_to_anthropic_sse(
@@ -3497,6 +3499,11 @@ fn create_anthropic_sse_stream_from_responses_raw<E: std::error::Error + Send + 
                                         Some(hosted_web_search_name.as_str()),
                                         max_web_search_uses,
                                     ) {
+                                        if has_sent_message_start
+                                            && event.starts_with(b"event: message_start\n")
+                                        {
+                                            continue;
+                                        }
                                         yield Ok(event);
                                     }
                                     terminated = true;
@@ -6447,6 +6454,8 @@ mod tests {
     #[tokio::test]
     async fn test_terminal_only_web_search_preserves_output_order() {
         let input = concat!(
+            "event: response.created\n",
+            "data: {\"type\":\"response.created\",\"response\":{\"id\":\"resp_order\",\"model\":\"gpt-5.6\"}}\n\n",
             "event: response.completed\n",
             "data: {\"type\":\"response.completed\",\"response\":{\"id\":\"resp_order\",\"model\":\"gpt-5.6\",\"status\":\"completed\",\"output\":[{\"type\":\"message\",\"role\":\"assistant\",\"content\":[{\"type\":\"output_text\",\"text\":\"Before\",\"annotations\":[]}]},{\"id\":\"ws_order\",\"type\":\"web_search_call\",\"status\":\"completed\",\"action\":{\"type\":\"search\",\"query\":\"Rust\"}},{\"type\":\"message\",\"role\":\"assistant\",\"content\":[{\"type\":\"output_text\",\"text\":\"After\",\"annotations\":[]}]}],\"usage\":{\"input_tokens\":8,\"output_tokens\":4}}}\n\n"
         );
@@ -6473,6 +6482,22 @@ mod tests {
             vec!["text", "server_tool_use", "web_search_tool_result", "text"]
         );
         assert_eq!(text_deltas, vec!["Before", "After"]);
+        assert_eq!(merged.matches("event: message_start").count(), 1);
+        assert!(merged.contains("event: message_stop"));
+    }
+
+    #[tokio::test]
+    async fn test_terminal_output_does_not_duplicate_done_reasoning() {
+        let input = concat!(
+            "event: response.output_item.done\n",
+            "data: {\"type\":\"response.output_item.done\",\"output_index\":0,\"item\":{\"id\":\"rs_done\",\"type\":\"reasoning\",\"summary\":[{\"type\":\"summary_text\",\"text\":\"Need a tool.\"}]}}\n\n",
+            "event: response.completed\n",
+            "data: {\"type\":\"response.completed\",\"response\":{\"id\":\"resp_reasoning\",\"model\":\"gpt-5.6\",\"status\":\"completed\",\"output\":[{\"id\":\"rs_done\",\"type\":\"reasoning\",\"summary\":[{\"type\":\"summary_text\",\"text\":\"Need a tool.\"}]}]}}\n\n"
+        );
+
+        let merged = convert_stream_text(input).await;
+
+        assert_eq!(merged.matches("\"type\":\"thinking_delta\"").count(), 1);
         assert!(merged.contains("event: message_stop"));
     }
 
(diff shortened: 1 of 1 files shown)


farion1231:
@codex review

farion1231 (commented):
Thanks for picking up the terminal-only ordering follow-up — routing the no-increment fallback through the already-verified JSON conversion path is exactly the right approach, and the patch is admirably tight (+46 lines with a regression test). I verified the new test passes and the full backend suite stays green on a merge tree against current main.

That said, I found two P2s while replaying gateway shapes against this branch. They share a single root cause: the new guard uses `has_sent_message_start` ([L3491-L3493](https://github.com/farion1231/cc-switch/blob/18c0a427bbe2800b8f46facd03129499c3376949/src-tauri/src/proxy/providers/streaming_responses.rs#L3491-L3493)) as a proxy for "nothing was emitted yet", but that flag diverges from actual content emission in both directions.

**[P2] `response.created` + `response.completed` streams still emit out of order.** `response.created` unconditionally sets `has_sent_message_start` ([L2533](https://github.com/farion1231/cc-switch/blob/18c0a427bbe2800b8f46facd03129499c3376949/src-tauri/src/proxy/providers/streaming_responses.rs#L2533)) without emitting any content, so a gateway that forwards the standard lifecycle events (`created` → `completed`) but no deltas is excluded from the new branch and falls back to the old reordering path. Repro: take the input from `test_terminal_only_web_search_preserves_output_order` and prepend a `response.created` event — the block order becomes `server_tool_use → web_search_tool_result → text("Before") → text("After")` instead of `text("Before") → server_tool_use → web_search_tool_result → text("After")`. Since `response.created` is the standard opening event of a Responses SSE stream, this shape is at least as common in the wild as the completed-only one the PR fixes.

**[P2] A stream of only `output_item.done`(reasoning) + `response.completed` now replays the thinking block twice.** The `output_item.done` reasoning handler yields a thinking block without setting `has_sent_message_start` ([L4314-L4337](https://github.com/farion1231/cc-switch/blob/18c0a427bbe2800b8f46facd03129499c3376949/src-tauri/src/proxy/providers/streaming_responses.rs#L4314-L4337)), so the terminal event enters the new branch and re-converts the full output. Measured on this branch: 2 `thinking_delta` events for the same summary text (and the first event of the stream is a `content_block_start` before any `message_start`). The same input on the merge base emits the thinking content exactly once — the duplication is introduced by this PR. This matches what the inline bot comment flagged.

**Suggested fix (one change closes both, ~10-15 lines):** gate the new branch on "no content block has been emitted yet" instead of `has_sent_message_start` — e.g. a dedicated flag, or a combination like `next_content_index == 0 && !has_substantive_output && open_indices.is_empty()`. When `message_start` was already sent (the `created` one carries zero usage, so nothing is lost), skip the `message_start` at the head of the `responses_json_to_anthropic_sse` sequence — but note the error path returns an error event as its first element, so don't blindly `skip(1)`; matching on the event content or adding an `include_message_start` parameter to the function is safer. Regression tests: (1) the prepended-`created` variant above asserting block order, (2) the done(reasoning)-only stream asserting exactly one `thinking_delta`.

Optional while you're in there: `response.incomplete` shares the same match arm but is excluded from the new branch, so truncated terminal-only responses keep the old ordering. The widened gate could cover it too.

With the guard reworked and the two regression tests added, this is good to merge — happy to re-review quickly.

allenxu09:
Addressed both terminal-replay review findings in de851456: lifecycle-only response.created no longer suppresses terminal replay, and output_item.done reasoning is not replayed twice. Re-verified today with cargo test --manifest-path src-tauri/Cargo.toml test_terminal_ (7 passed), cargo fmt --check, and git diff --check. Current CI is green. Ready for re-review.

farion1231 (commented):
Thanks for the quick turnaround on the gate rework — this landed exactly as discussed. I re-verified both findings from the last round against `de851456`:

- `response.created` (with or without `response.in_progress`) followed by a terminal-only `response.completed` now yields `text → server_tool_use → web_search_tool_result → text` with a single `message_start`.
- `output_item.done`(reasoning) + `response.completed` emits the thinking block exactly once — the `done` handler consumes a content index, so the new branch correctly stays out of the way.

Full backend suite is green on a merge tree against current main, and CI is green on all four platforms. A nice side effect worth calling out: a terminal-only stream carrying a `function_call` item (no `output_item.added` beforehand) used to produce `stop_reason: "tool_use"` with zero `tool_use` blocks — the `done` handler skips items that were never opened ([L4040](https://github.com/farion1231/cc-switch/blob/de851456e5e19af3d241c2271179c71cad53dca6/src-tauri/src/proxy/providers/streaming_responses.rs#L4040)) and the old terminal path only collects `message`/`web_search_call` items ([L3571-L3577](https://github.com/farion1231/cc-switch/blob/de851456e5e19af3d241c2271179c71cad53dca6/src-tauri/src/proxy/providers/streaming_responses.rs#L3571-L3577)). Routing through the JSON converter now emits the proper `tool_use` block.

One remaining P2, then two follow-up-grade notes:

**[P2] A terminal response without `status` now loses its `stop_reason`.** The old terminal path fell back to the event name when `response.status` was missing ([L3908-L3914](https://github.com/farion1231/cc-switch/blob/de851456e5e19af3d241c2271179c71cad53dca6/src-tauri/src/proxy/providers/streaming_responses.rs#L3908-L3914)), so `response.completed` still mapped to `end_turn`/`tool_use`. The new branch ([L3491-L3511](https://github.com/farion1231/cc-switch/blob/de851456e5e19af3d241c2271179c71cad53dca6/src-tauri/src/proxy/providers/streaming_responses.rs#L3491-L3511)) hands `response_obj` straight to the converter, which passes `body.status` through unchanged ([transform_responses.rs L2760-L2765](https://github.com/farion1231/cc-switch/blob/de851456e5e19af3d241c2271179c71cad53dca6/src-tauri/src/proxy/providers/transform_responses.rs#L2760-L2765) → [`map_responses_stop_reason`](https://github.com/farion1231/cc-switch/blob/de851456e5e19af3d241c2271179c71cad53dca6/src-tauri/src/proxy/providers/transform_responses.rs#L2061) returns `None` for `None`), so `message_delta` carries `stop_reason: null`. Repro: a `response.completed` event whose response object has `output` but no `status` → `stop_reason: null`; the same payload preceded by a single `output_text.delta` (old path) → `end_turn`. With a `function_call` item it becomes a `tool_use` block paired with `stop_reason: null`, which is the combination most likely to confuse a client that keys off the stop reason. Omitting `status` is off-spec so this is narrow, but it's a regression relative to the path being replaced, and the fix is a few lines at the call site (keeps the non-streaming JSON path untouched):

```rust
let mut terminal_body = response_obj.clone();
if terminal_body.get("status").and_then(Value::as_str).is_none() {
    terminal_body["status"] = json!("completed");
}
for event in responses_json_to_anthropic_sse(
    terminal_body,
    Some(hosted_web_search_name.as_str()),
    max_web_search_uses,
) {
```

Two regression tests would lock it: status-less `completed` with a text item asserting `end_turn`, and with a `function_call` item asserting `tool_use`. (One side effect to be aware of: with `status` injected, the converter treats malformed `function_call` arguments strictly and emits an error event instead of substituting `{}` — consistent with `completed` semantics, so fine.)

**[P3] `response.incomplete` is still excluded.** The gate is `event_name == "response.completed"` only, so a truncated terminal-only response with a search keeps the old ordering (`server_tool_use → web_search_tool_result → text → text`, `stop_reason: max_tokens`). Fine as a follow-up; if you widen it, the `status` injection above should pick `"incomplete"` for that event.

**[P3] The original `completed`-only shape no longer has a dedicated test.** [`test_terminal_only_web_search_preserves_output_order`](https://github.com/farion1231/cc-switch/blob/de851456e5e19af3d241c2271179c71cad53dca6/src-tauri/src/proxy/providers/streaming_responses.rs#L6455) now prepends `response.created`, so the exact shape from #5681 (no lifecycle events at all) is only covered indirectly. I verified it still behaves correctly (same block order, and `message_start` carries the input usage), but keeping both variants as separate tests would lock both halves of the gate.

nit: `open_indices.is_empty()` is implied by `next_content_index == 0` (every content block index is allocated from that counter). Harmless as a defensive check; drop or keep as you prefer.

With the P2 fixed and its two tests added, this is good to merge — happy to re-review quickly.

allenxu09:
@farion1231 ptal!

PR #6427, opened 2026-08-13
refactor(grokbuild): align provider form with Codex

YUZHEthefool:
## Summary / 概述

- Align the Grok Build custom-provider form with the existing Codex advanced-options layout without modifying shared Codex form components.
- Remove the redundant client profile and API Backend inputs while retaining the context-window setting.
- Keep the Grok client config on the Responses backend, while preserving the selected upstream API format in provider metadata for routing conversion.
- Add regression coverage for the advanced section and non-Responses upstream formats.

## Related Issue / 关联 Issue

N/A

## Screenshots / 截图

N/A

## Checklist / 检查清单

- [x] `pnpm typecheck` passes / 通过 TypeScript 类型检查
- [x] `pnpm format:check` passes / 通过代码格式检查
- [x] `cargo clippy` passes (N/A: no Rust code changed) / 通过 Clippy 检查（不适用：未修改 Rust 代码）
- [x] Updated i18n files (N/A: no new user-facing strings) / 已更新国际化文件（不适用：未新增用户可见文本）

## Tests / 测试

- `pnpm exec vitest run tests/components/GrokBuildProviderForm.test.tsx` (6 passed)
- `pnpm typecheck`
- `pnpm format:check`

Issue #5758, opened 2026-07-25
[grokbuild] extra_headers is dropped, so api_backend="messages" configs lack anthropic-version


szsnzz:
### Self Checks / 自检

- [x] I have read the FAQ section in README.
- [x] I have searched for existing issues, including closed ones.

### CC Switch Version / 版本号

3.18.0

### Operating System / 操作系统

Windows 11

### Related App / 涉及应用

Grok Build

### Summary / 概述

When CC Switch writes a Grok Build provider whose `api_backend = "messages"` (Anthropic Messages protocol), the generated `~/.grok/config.toml` has no way to carry `extra_headers`. Per the official Grok Build docs, the Anthropic backend needs `anthropic-version` (and, strictly, authenticates via the `x-api-key` header rather than `Authorization: Bearer`). Both must be passed through `extra_headers`, which `grok_config.rs` neither models nor preserves.

Related but distinct: #5678 is about proxy-mode route registration. This issue is about the config document CC Switch generates in direct (non-takeover) mode.

### Steps to Reproduce / 重现步骤

1. Add a Grok Build provider in CC Switch pointing at an Anthropic-compatible relay, e.g. `base_url = "https://<relay>/v1"`, `api_backend = "messages"`.
2. Switch to that provider so CC Switch writes `~/.grok/config.toml`.
3. Inspect the generated file.

Generated (abridged):

```toml
[models]
default = "opus-5"

[model]
[model.opus-5]
model = "claude-opus-5"
base_url = "https://<relay>/v1"
name = "right"
api_backend = "messages"
context_window = 1000000
api_key = "sk-..."
```

No `extra_headers` entry is emitted, and adding one by hand is lost on the next switch.

### Root Cause / 根因

In `src-tauri/src/grok_config.rs`:

- `GrokModelConfig` has no `extra_headers` field, so `extract_model_config` drops it on read.
- `validate_config_toml` never considers it, so a `messages` backend without `anthropic-version` validates as correct.
- `update_selected_model_string` only writes scalar strings via `toml_edit::value(value)`, so an inline table like `extra_headers = { ... }` cannot be produced by that path at all.

Since `write_grok_provider_live` round-trips the snapshot through these helpers, any hand-added `extra_headers` does not survive a provider switch.

### Expected Behavior / 期望行为

1. `GrokModelConfig` models `extra_headers` and `extract_model_config` preserves it, so hand-written headers survive a switch.
[...]

Issue #5678, opened 2026-07-23
[grokbuild] Missing chat_completions route causes fallback to Codex channel


DuanXiaoWen:
### Self Checks / 自检

- [x] I have read the [FAQ](https://github.com/farion1231/cc-switch#faq) section in README.
我已阅读 README 中的[常见问题](https://github.com/farion1231/cc-switch#常见问题)。

- [x] I have searched for [existing issues](https://github.com/farion1231/cc-switch/issues), including closed ones.
我已搜索过[已有的 Issue](https://github.com/farion1231/cc-switch/issues)，包括已关闭的。

### CC Switch Version / 版本号

3.18.0

### Operating System / 操作系统

macOS

### Related App / 涉及应用

Other / 其他

### Steps to Reproduce / 重现步骤

1. Set Grok Build config `api_backend = "chat_completions"`, `base_url` pointing at CC Switch (`http://127.0.0.1:15721`).
2. Enable grokbuild takeover in CC Switch with an OpenAI-compatible provider (e.g. idealab internal `/v1` endpoint).
3. Send any prompt via `grok -m grok-4.5 "hello"`.
4. Observe the request hits the **Codex** channel provider (OpenAI Official) instead of the grokbuild provider.

Error in Grok Build output:
```
upstream_status: HTTP 404 ... endpoint=/chat/completions ... provider=OpenAI Official
```

Root cause: CC Switch only registers `/grokbuild/v1/responses` route. There is no `/grokbuild/v1/chat/completions` handler, so the request falls through to the Codex channel's catch-all.

### Expected Behavior / 期望行为

`/grokbuild/v1/chat/completions` should route to the currently active grokbuild provider (with any needed format translation), not fall through to the Codex channel.

All three `api_backend` values supported by Grok Build (`chat_completions`, `responses`, `messages`) should be handled by the grokbuild channel when takeover is active.

### Actual Behavior / 实际行为

Requests to `/grokbuild/v1/chat/completions` silently fall through to the Codex channel's default provider (OpenAI Official), which returns HTTP 404 because the configured Codex endpoint doesn't match.

This is extremely confusing — no error message indicates the request was misrouted. The user sees an OpenAI 404 and has no way to know it's a CC Switch routing issue.

### Additional Context / 补充信息

**Environment:**
- CC Switch v3.18.0 (Tauri/Rust)
- Grok Build v0.2.111
- macOS arm64 (Apple Silicon)

**Suggested fix:**
[...]

PR #6510, opened 2026-08-16
fix(grokbuild): preserve api_backend in form and detect it for proxy protocol conversion

zayokami:
## Problem

Grok Build providers support three upstream protocols via `api_backend` in `[model."<profile>"]` (`responses` / `chat_completions` / `messages`). Two bugs currently break non-Responses upstreams:

**1. The provider form silently reverts `api_backend` to `"responses"` on every save (regression from #6427).**

#6427 removed the form's `apiBackend` state; every write path (`syncStructuredConfig`, `handleSubmit`) now hardcodes `GROK_BUILD_DEFAULT_API_BACKEND`. `parseGrokBuildConfig` still parses the field correctly — the parsed value is just never used. So a hand-written `api_backend = "messages"` in the raw TOML editor, or a provider saved before #6427 with `chat_completions`, is silently rewritten to `"responses"` by any edit-save, with no UI indication. In direct mode the Grok client then speaks Responses to a Chat/Anthropic-only upstream and fails.

**2. The proxy's upstream-protocol detection never reads Grok Build's `api_backend` key.**

Grok Build providers reuse `CodexAdapter` in the proxy (credentials / model / base_url already resolve via `grok_config`). But `codex_provider_uses_chat_completions` / `codex_provider_uses_anthropic` only probe `meta.api_format`, settings `api_format`, and the Codex-shaped `wire_api` TOML key — never `api_backend`. A provider whose stored config declares `api_backend = "chat_completions"` (savable before #6427, or hand-imported) with no `meta.api_format` gets no Responses→Chat conversion under takeover: the client speaks Responses to the proxy (takeover forces `api_backend = "responses"` in the live file), and the proxy forwards the Responses body unchanged to a Chat-only upstream — a silently broken pipe. Same for `messages` (Anthropic).

## Fix

- **Form**: thread the parsed `api_backend` through component state again — raw-TOML edits parse-back into it, submit writes the state value, choosing a preset resets it to `responses` (all current presets are Responses). The selector removed in #6427 is **not** restored: upstream protocol selection stays on `meta.apiFormat` + proxy conversion per that refactor's direction; this fix only stops the silent data loss. `meta.apiFormat` and the TOML `api_backend` stay independent (locked by test).
- **Proxy**: after the `wire_api` probe and before the base-url heuristic, both detection functions now probe `api_backend` via `grok_config::extract_model_config`. Codex configs have no `[models]` / `[model.*]` tables, so the probe returns `None` and stays inert for Codex; `is_chat_wire_api` / `is_anthropic_wire_api` already accept `chat_completions` / `messages`, so the value domains align with no matcher changes.

## Tests

- Form (component level): editing a `messages` provider keeps `messages` after save; an `api_backend` typed into the raw TOML editor survives submit; one pre-existing #6427 test asserted the buggy behavior (forced `"responses"`) and is flipped to lock the correct one (keeping its `meta.apiFormat` independence assertion). All three go red when the fix is reverted (verified).
- Proxy: `chat_completions` → Chat conversion and `messages` → Anthropic conversion (both red before the fix), plus three guards that stay green throughout: `responses` stays native, plain Codex TOML without `wire_api` is not misread as Grok, and explicit `meta.api_format` overrides the TOML probe.

## Round 2 (review feedback)

- **[P2]** The form wrote the default `meta.apiFormat` (`"openai_responses"`) on every save, and both detection functions early-return when meta is present — so a hand-imported `api_backend = "messages"/"chat_completions"` provider lost its takeover conversion after any form save. `meta.apiFormat` is now only written when it existed on load or the user touched the format selector / applied a preset (`apiFormatTouched` state), keeping meta absent so the TOML probe stays authoritative.
- **[P3]** `handleRawConfigChange` skipped the parse-back while the config was semantically incomplete (e.g. `api_key` not yet filled), so "edit `api_backend` in the raw editor first, then fill in the API key" clobbered the typed value with the stale `"responses"` via `syncStructuredConfig`. The parse-back is now gated on syntax only (new `isSyntacticallyValidGrokToml` helper); the mid-typing guard against broken TOML is unchanged.
- **nit** — deduplicated the comment above the second probe in `codex.rs`.
- Three new regression tests (meta absent on save for an untouched meta-less provider; meta written after touching the selector; the P3 raw-first ordering), each revert-verified.

## Round 3 (CI)

`Backend Checks (Windows + WSL2 home)` failed on cold builds — fork PRs cannot restore the upstream actions cache, so after the `--no-run` compile step the next run-mode cargo invocation re-links the lib test binary, and `link.exe`/`mt.exe` cannot create manifests when TEMP points at a WSL UNC path (LNK1327). The `--list` invocation now runs before TEMP is redirected into WSL2 and absorbs any re-link with a native temp; the test run then executes the fresh binary without touching the linker.

## Notes / limits

- #5758's `extra_headers` modeling is still open — this PR makes `api_backend = "messages"` survivable (a precondition), but doesn't model `extra_headers` in `GrokModelConfig`. Happy to follow up separately.
- Under takeover, the canonical setup for Chat/Anthropic upstreams remains `meta.apiFormat` (the form's upstream-format selector); the TOML probe covers providers whose meta is absent.

Related to #5758, #5678.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -210,8 +210,6 @@ jobs:
       - name: Run Windows-to-WSL2 filesystem contract
         shell: pwsh
         run: |
-          $env:TEMP = $env:CC_SWITCH_WSL_TEST_TEMP
-          $env:TMP = $env:CC_SWITCH_WSL_TEST_TEMP
           $testName = "config::tests::atomic_write_replaces_existing_wsl_unc_file"
           $testList = cargo test --lib --manifest-path src-tauri/Cargo.toml -- --ignored --list
           if ($LASTEXITCODE -ne 0) {
@@ -221,6 +219,15 @@ jobs:
           if ($testList -notcontains $expected) {
             throw "Expected ignored test was not discovered: $testName"
           }
+          # TEMP only points into WSL2 once the binary is fully linked. On cold
+          # builds (fork PRs cannot restore the upstream cache) the next run-mode
+          # cargo invocation re-links the lib test binary even though --no-run
+          # built it above, and link.exe/mt.exe cannot create manifests when TEMP
+          # is a WSL UNC path (LNK1327). The --list call above absorbs that
+          # rebuild with a native TEMP; the run below executes the fresh binary
+          # without re-linking.
+          $env:TEMP = $env:CC_SWITCH_WSL_TEST_TEMP
+          $env:TMP = $env:CC_SWITCH_WSL_TEST_TEMP
           cargo test --lib --manifest-path src-tauri/Cargo.toml $testName -- --ignored --exact --nocapture
           if ($LASTEXITCODE -ne 0) {
             exit $LASTEXITCODE
--- a/src-tauri/src/proxy/providers/codex.rs
+++ b/src-tauri/src/proxy/providers/codex.rs
@@ -210,10 +210,8 @@ pub fn codex_provider_uses_anthropic(provider: &Provider) -> bool {
         return is_anthropic_wire_api(&wire_api);
     }
 
-    // Grok Build TOML declares the upstream protocol via `api_backend` inside
-    // [model."<profile>"], which the Codex-shaped wire_api probe above can never
-    // match. Codex configs have no [models]/[model.*] tables, so
-    // extract_model_config returns None for them and this branch stays inert.
+    // Same Grok Build `api_backend` probe as codex_provider_uses_chat_completions
+    // above — see there for why it stays inert for Codex configs.
     if let Some(api_backend) = provider
         .settings_config
         .get("config")
--- a/src-tauri/src/proxy/providers/codex.rs
+++ b/src-tauri/src/proxy/providers/codex.rs
@@ -55,6 +55,20 @@ pub fn codex_provider_uses_chat_completions(provider: &Provider) -> bool {
         return is_chat_wire_api(&wire_api);
     }
 
+    // Grok Build TOML declares the upstream protocol via `api_backend` inside
+    // [model."<profile>"], which the Codex-shaped wire_api probe above can never
+    // match. Codex configs have no [models]/[model.*] tables, so
+    // extract_model_config returns None for them and this branch stays inert.
+    if let Some(api_backend) = provider
+        .settings_config
+        .get("config")
+        .and_then(|v| v.as_str())
+        .and_then(crate::grok_config::extract_model_config)
+        .map(|config| config.api_backend)
+    {
+        return is_chat_wire_api(&api_backend);
+    }
+
     if let Some(base_url) = provider
         .settings_config
         .get("base_url")
@@ -163,8 +177,9 @@ pub fn inject_codex_chat_prompt_cache_key(
 /// Messages protocol (`/v1/messages`). The local Codex client always talks to CC
 /// Switch through the Responses API, so CC Switch bridges Responses ⇄ Anthropic.
 ///
-/// Determined solely from explicit config (apiFormat / wire_api); no base_url
-/// guessing — Anthropic gateway addresses vary widely and guessing easily misfires.
+/// Determined solely from explicit config (apiFormat / wire_api / Grok Build
+/// api_backend); no base_url guessing — Anthropic gateway addresses vary widely
+/// and guessing easily misfires.
 pub fn codex_provider_uses_anthropic(provider: &Provider) -> bool {
     if let Some(api_format) = provider
         .meta
@@ -186,13 +201,30 @@ pub fn codex_provider_uses_anthropic(provider: &Provider) -> bool {
         return is_anthropic_wire_api(api_format);
     }
 
-    provider
+    if let Some(wire_api) = provider
         .settings_config
         .get("config")
         .and_then(|v| v.as_str())
         .and_then(extract_codex_wire_api_from_toml)
-        .map(|wire_api| is_anthropic_wire_api(&wire_api))
-        .unwrap_or(false)
+    {
+        return is_anthropic_wire_api(&wire_api);
+    }
+
+    // Grok Build TOML declares the upstream protocol via `api_backend` inside
+    // [model."<profile>"], which the Codex-shaped wire_api probe above can never
+    // match. Codex configs have no [models]/[model.*] tables, so
+    // extract_model_config returns None for them and this branch stays inert.
+    if let Some(api_backend) = provider
+        .settings_config
+        .get("config")
+        .and_then(|v| v.as_str())
+        .and_then(crate::grok_config::extract_model_config)
+        .map(|config| config.api_backend)
+    {
+        return is_anthropic_wire_api(&api_backend);
+    }
+
+    false
 }
 
 pub fn should_convert_codex_responses_to_anthropic(provider: &Provider, endpoint: &str) -> bool {
@@ -1048,6 +1080,7 @@ context_window = 500000
     }
 
     #[test]
+    #[test]
     fn explicit_codex_official_cards_use_chatgpt_backend() {
         let mut provider = create_provider(json!({
             "auth": {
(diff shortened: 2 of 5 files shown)


farion1231 (commented):
Reviewed at 4af2c573. Both bugs are real and verified on my side: the form regression from #6427 (which never shipped in a release — good timing to catch it), and the proxy detection gap, which does exist in released builds (#5678 is an adjacent report of the same family). The fix is the minimal shape for both halves — form state threading plus a probe that stays inert for Codex configs (`extract_model_config`'s all-fields-required contract guarantees `None` there, and your regression guard test locks it). I also re-verified the merge against current `main` (which now includes #6511): clean merge, typecheck, full frontend suite and the codex module tests all green.

One gap worth closing, one narrower issue that can be a follow-up:

**[P2] Saving the form writes the default `meta.apiFormat`, which masks the new TOML probe for meta-less providers.**
`apiFormat` state initializes to `"openai_responses"` when the provider has no stored `meta.apiFormat` ([form L114-L117](https://github.com/farion1231/cc-switch/blob/4af2c573d877457b1a0554a97d6f262e02d5d5cc/src/components/providers/forms/GrokBuildProviderForm.tsx#L114-L117)), and submit writes it into meta unconditionally ([L390](https://github.com/farion1231/cc-switch/blob/4af2c573d877457b1a0554a97d6f262e02d5d5cc/src/components/providers/forms/GrokBuildProviderForm.tsx#L390)). Both detection functions early-return as soon as meta is present ([codex.rs L28-L46](https://github.com/farion1231/cc-switch/blob/4af2c573d877457b1a0554a97d6f262e02d5d5cc/src-tauri/src/proxy/providers/codex.rs#L28-L46), [L183-L201](https://github.com/farion1231/cc-switch/blob/4af2c573d877457b1a0554a97d6f262e02d5d5cc/src-tauri/src/proxy/providers/codex.rs#L183-L201)), so a hand-imported `api_backend = "messages"` / `"chat_completions"` provider loses the takeover conversion after any form save (even just editing the API key) — the exact population the TOML probe is meant to cover. Smallest fix I see: only include `apiFormat` in meta when it existed on load or the user actually touched the selector / applied a preset, e.g.

```tsx
const [apiFormatTouched, setApiFormatTouched] = useState(false);
// set it true in the upstream-format selector's onChange and in handleApplyPreset
...
apiFormat:
  initialData?.meta?.apiFormat !== undefined || apiFormatTouched
    ? apiFormat
    : undefined,
```

That keeps meta absent → the TOML probe keeps working under takeover, and direct mode is unaffected either way.

**[P3] Raw-TOML `api_backend` edits are lost in one ordering during provider creation.**
`handleRawConfigChange` bails before parse-back while the config is still invalid ([L295](https://github.com/farion1231/cc-switch/blob/4af2c573d877457b1a0554a97d6f262e02d5d5cc/src/components/providers/forms/GrokBuildProviderForm.tsx#L295)), so "edit `api_backend` in the raw editor first, then fill in the API key" ends with `syncStructuredConfig` writing the stale `responses` back. I'd rate this P3 rather than a blocker: on `main` the value is clobbered unconditionally in every flow, so this PR is strictly an improvement — only this one ordering window remains, and fixing it properly means teaching that early return to distinguish syntactically broken TOML from semantically incomplete TOML (the guard exists to protect state while the user is mid-typing). Fine as a follow-up.

nit: the explanatory comment above the two new probes is duplicated verbatim; could point one at the other.

With the P2 addressed — in this PR or as an immediate follow-up, your call — this is mergeable. Happy to re-review quickly.

zayokami:
Thanks for the detailed review — all three points are addressed in the updated branch (rebased on current main, 3 commits):

**P2 — meta.apiFormat masking the TOML probe.** Fixed as you sketched: `apiFormatTouched` state, set in the upstream-format selector's onChange and in the preset branch, and meta only gets `apiFormat` when it existed on load or was touched. A meta-less hand-imported provider now keeps `meta.apiFormat` absent through any form save, so both detection functions fall through to the `api_backend` probe under takeover. Direct mode is unaffected.

**P3 — raw-TOML api_backend clobbered by the ordering window.** Fixed rather than deferred, along the line you suggested: the parse-back is now gated on syntax only via a new `isSyntacticallyValidGrokToml` helper (try-parse, no semantic checks), instead of the full `validateGrokBuildConfig`. Semantically-incomplete TOML (e.g. `api_key` not yet filled) still gets read back, so "edit api_backend in the raw editor first, then fill in the API key" keeps the typed value; the mid-typing guard against syntactically broken TOML is unchanged.

**nit — duplicated probe comment.** The second occurrence now points at the first.

Tests: three new regression tests in `GrokBuildProviderForm.test.tsx` — meta absent on save for an untouched meta-less provider, meta written after touching the selector, and the P3 raw-first ordering. Each was revert-verified (exactly its test goes red when the fix is reverted; the other ten stay green). Full frontend suite, typecheck and prettier pass locally, and the codex module tests are green under cargo.

zayokami:
Added a fix for #6679 on this branch (commit 69665457): the Codex auth file transaction no longer uses hard links, which WSL UNC paths reject with os error 50. Install/rollback now go through a create_new + copy protocol that keeps the same no-clobber atomicity guarantee, plus regression tests and an extended Windows + WSL2 CI contract job.

zayokami:
Also added a fix for #6789 on this branch (commit a15bb138): Claude Code's per-turn \<total_tokens>\ system message was accumulating inside the merged system string on the Claude→OpenAI and Claude→Gemini paths, breaking byte-prefix prompt caches (DeepSeek/GLM/Kimi hit rates collapsed from ~99% to ~20%). The tag is now stripped in both merge paths with regression tests.

farion1231 (commented):
Re-reviewed at head a15bb13. The Round 2 follow-ups close out both points from my earlier review cleanly — `apiFormatTouched` keeps `meta.apiFormat` absent for untouched meta-less providers, the syntax-only gate fixes the raw-first ordering, and both land with regression tests. I also ran the merge tree against current main: backend 2695 tests, frontend typecheck + 992 tests, all green. The Grok Build portion of this branch is mergeable as far as I'm concerned.

The blocker is the two riders. The branch now carries two commits unrelated to the PR title — 69665457 (#6679, Codex auth transaction) and a15bb138 (#6789, tag stripping). I appreciate that you flagged both in comments, but they need to come out of this branch and into their own PRs: the first has a serious defect, and the second is a proxy-conversion policy decision that shouldn't ride along with a Grok Build form fix.

**[P1] In 69665457: a mid-copy failure deletes the original credentials.**
[`install_file_if_vacant`](https://github.com/farion1231/cc-switch/blob/a15bb1389747bb2a3398f8194d49f24a8326d921/src-tauri/src/services/proxy.rs#L320-L337) creates the destination with `create_new` and then copies into it. If the copy or flush fails partway — disk full, WSL UNC disconnect; no crash required — the partially-written destination stays in place, and the [error path in `install`](https://github.com/farion1231/cc-switch/blob/a15bb1389747bb2a3398f8194d49f24a8326d921/src-tauri/src/services/proxy.rs#L229-L237) sees `self.path.exists() == true`, concludes Codex must have written a newer file, and calls [`discard_quarantined`](https://github.com/farion1231/cc-switch/blob/a15bb1389747bb2a3398f8194d49f24a8326d921/src-tauri/src/services/proxy.rs#L372-L381) — permanently deleting the quarantined original `auth.json` — and sets `finished = true`, so the `Drop` rollback never runs. End state: a corrupt `auth.json` at the live path and the user's original credentials gone.

The `exists()` heuristic was sound under the hard-link protocol — a failed `hard_link` never creates the destination, so an existing path really could only mean a concurrent Codex write. The new primitive breaks that invariant. When you resubmit this as its own PR: keep `hard_link` as the primary install and fall back to `create_new` + copy only when it fails with `ErrorKind::Unsupported` (the os error 50 case from #6679), and in the fallback delete the partially-created destination before returning, so the heuristic keeps its meaning. A test that injects a mid-copy failure would lock this in — the current tests only cover the vacant/occupied cases. The WSL2 lifecycle CI test is genuinely valuable and transfers as-is. Note `ci.yml` is entangled: the `--list`-before-TEMP reorder belongs to this PR, the second test's wiring goes with the rider.

**[P2] In a15bb138: the tag stripping is broader than the problem, and lossy.**
Three concrete issues: (1) [`strip_claude_context_indicator_tags`](https://github.com/farion1231/cc-switch/blob/a15bb1389747bb2a3398f8194d49f24a8326d921/src-tauri/src/proxy/providers/transform.rs#L66-L92) removes any `<total_tokens>…</total_tokens>` span from any system text — it isn't anchored to Claude Code's standalone injected message or its `N tokens left` format, so legitimate system content that mentions the tag (a CLAUDE.md documenting Claude Code, say) gets silently rewritten. (2) The tag is not an inert placeholder — its value changes every turn and is the model's signal of remaining context budget; stripping it means every model driven through the proxy flies blind on budget. (3) The [single-system-message path](https://github.com/farion1231/cc-switch/blob/a15bb1389747bb2a3398f8194d49f24a8326d921/src-tauri/src/proxy/providers/transform.rs#L374-L385) now rewrites array-form content into a joined string even when no tag is present, changing request shape for unrelated traffic. Structurally, the underlying #6789 problem is that mid-conversation system messages get hoisted and merged at all — stripping one known tag stabilizes a single source of drift. Please pull this commit and take the approach discussion to #6789 first.

**[P3] Duplicated `#[test]` attribute** ([codex.rs#L1080-L1081](https://github.com/farion1231/cc-switch/blob/a15bb1389747bb2a3398f8194d49f24a8326d921/src-tauri/src/proxy/providers/codex.rs#L1080-L1081)), in the Grok Build probe commit, so it survives the split. It compiles but warns `duplicate_macro_attributes` and registers the test twice. One-line fix.

So: drop the two riders, delete the duplicated attribute, and the remaining four commits are good to merge from my side. Happy to re-review quickly, and to look at the #6679 PR when you open it.

zayokami:
Thanks for the thorough review — all three points addressed:

**Riders removed.** Both commits are out of this branch (force-pushed); the five remaining commits are Grok Build–scoped plus the one-line CI temp-path fix.

**[P3]** The duplicated `#[test]` is gone (now head commit).

**[P1]** Resubmitted as its own PR: #6884, reworked per your prescription — `hard_link` stays the primary atomic install, the `create_new` + copy fallback triggers only on `ErrorKind::Unsupported`, and a failed copy/flush deletes the partially created destination before returning so the `exists()` heuristic keeps its meaning. A test-only injection arms the fallback and fails the copy after a configurable prefix: one test locks the helper contract (destination vacant, source intact), another locks the transaction-level guarantee (quarantined generation survives the failed install instead of being discarded by the heuristic, and rollback restores it). The ignored WSL2 lifecycle test transfers as-is, rewired onto main's current `CC_SWITCH_TEST_EXE` step shape (no `--list`/TEMP reorder needed there anymore).

**[P2]** The tag-stripping commit is pulled and I've moved the approach discussion to #6789, covering the anchoring/signal-loss/array-shape concerns and a structural option (keep mid-conversation system blocks unmerged) alongside narrower alternatives.

Happy to re-review whatever remains here.

farion1231 (commented):
Re-checked `b2a1b9f2f`: the cleanup is complete — both unrelated commits are gone, and the duplicate `#[test]` attribute fix is exactly right. I verified the four topic commits locally by cherry-picking them onto current main: full backend suite green (2869 tests), clippy clean, frontend typecheck and unit tests green (1017 tests). The topic changes are ready.

One new thing since your last push: main has since merged #6472, which reworked the WSL2 contract step more thoroughly — the run step now executes the compiled test binary directly, so the cold-build re-link problem your `e30dde7ca` patched no longer exists. That commit is now superseded and conflicts with `.github/workflows/ci.yml` on main.

So one last ask: rebase onto latest main and drop `e30dde7ca`. The remaining four commits apply cleanly (verified locally), and after that this is good to merge — I'll turn it around quickly.

zayokami:
Done — rebased onto latest main (054673e0) and dropped e30dde7ca.

The remaining four topic commits apply cleanly, no conflicts. Verified locally on the rebased branch:

- `cargo fmt --check` clean, `cargo clippy -- -D warnings` clean
- Backend tests: codex provider 115 passed, api_backend 4 passed, grokbuild 45 passed
- Frontend typecheck green, GrokBuildProviderForm unit tests green

(Note: on this machine two unrelated test files fail on plain upstream main too — codexReasoningLevelPresets grok-4.5 cases and App integration — pre-existing locally, not touched by this branch.)

Ready to merge.

zayokami:
Added one follow-up commit: `test(codex): sync grok-4.5 reasoning levels with probed presets` — the failing CI check.

Root cause is a pre-existing main regression, not this branch: your own c08040e9 (2026-08-30, probed native /v1/responses) added `xhigh` to the grok-4.5 xAI presets but didn't update `codexReasoningLevelPresets.test.ts`, which still expected `['low','medium','high']`. Full-suite CI picked it up. The commit syncs the test expectations to the probed four-tier list; nothing else changed.

zayokami:
All seven CI checks are now green (the 21e5acd0 test-sync commit resolved the pre-existing main regression from c08040e9). Nothing left pending on this side — ready for merge whenever you are.

zayokami:
Small housekeeping: rebased onto the current main (11317c62). While doing so I found that upstream's own 21fda0ea already aligned the grok-4.5 reasoning-tier expectations (same values as our earlier test-sync commit), so I dropped our now-redundant 21e5acd0 — the branch is back to the four topic commits you verified, with no content changes. CI is re-running on the new base.

PR #6511, opened 2026-08-16
fix(grokbuild): stop showing Codex-specific copy in the Grok Build form

zayokami:
## Problem

#6427 aligned the Grok Build provider form with Codex by reusing `CodexFormFields`, but five Codex-facing strings render unchanged in the Grok Build form:

1. **Default-model hint** — "The model Codex requests by default … If left empty while model mapping is configured, the first mapping row is used." Grok Build has no model-mapping catalog, so both the app name and the fallback clause are wrong here.
2. **Default-model placeholder** — suggests a GPT model (`e.g. gpt-5.6-sol`).
3. **Advanced-section hint** — lists model mapping as contained, which the Grok form doesn't offer.
4. **Reasoning-effort hint** — references Codex's `reasoning.effort`.
5. **Max-output-tokens hint** — describes Codex's request body ("Codex does not forward model_max_output_tokens …").

## Fix

- Branch those five strings on `appId === "grokbuild"` inside `CodexFormFields` and add matching `grokBuild.*` entries to all four locales (en/zh/ja/zh-TW), worded for Grok Build semantics (e.g. no mapping clause, "the request's reasoning effort" instead of "Codex's reasoning.effort").
- Codex rendering is unchanged — same keys, same text.

The catalog-gated strings (`defaultModelNotInCatalog`, `addToModelMapping`, …) were checked and are safe: they require `catalogRows.length > 0`, which never happens for Grok Build since the catalog section isn't rendered.

## Tests

- Two new regression tests in `GrokBuildProviderForm.test.tsx` assert the Grok copy renders and the Codex text never appears (model field + collapsed advanced hint + reasoning hint on the Chat format; max-tokens hint on the Anthropic format).
- Revert-verified: forcing the appId branch off makes exactly the two new tests fail; the six pre-existing tests stay green either way.
- `pnpm typecheck`, `pnpm format:check` and the full `pnpm test:unit` pass.

## Note

`grokBuild.apiBackend` / `grokBuild.profile` became orphaned locale keys when #6427 removed those controls. Left in place intentionally since pruning is unrelated to this fix — happy to remove them here if you'd rather.

### DIFF
--- a/src/components/providers/forms/CodexFormFields.tsx
+++ b/src/components/providers/forms/CodexFormFields.tsx
@@ -450,6 +450,10 @@ export function CodexFormFields({
   //（填了才生成 catalog）。两者都已与「路由接管」概念解耦。
   const isChatFormat = apiFormat === "openai_chat";
   const isAnthropicFormat = apiFormat === "anthropic";
+  // Grok Build 复用本表单，但语义与 Codex 有差异（无模型映射、协议由 TOML 的
+  // api_backend 声明、请求体也不是 Codex 发出的）——提示文案按 appId 分流，
+  // 对应词条在 grokBuild.* 下。
+  const isGrokBuild = appId === "grokbuild";
   const canEditCatalog = Boolean(onCatalogModelsChange);
   const canEditReasoning = Boolean(onCodexChatReasoningChange);
   const supportsThinking =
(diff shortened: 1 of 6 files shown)


farion1231:
Thanks @zayokami — verified end to end and merging. The coverage is exactly right: scanning every codexConfig key the shared form uses across all four locales, the five strings you branched are precisely the Codex-specific copy reachable in the Grok Build form, and the catalog-gated ones (`defaultModelNotInCatalog` / `modelMappingHint`) indeed can't render there since the Grok form never passes catalog props. The revert check reproduced cleanly on our side too — exactly the two new tests fail with the branch forced off. Well-scoped fix with honest notes.

Two optional follow-ups if you're interested, no pressure:
- Pruning the orphaned `grokBuild.apiBackend` / `grokBuild.profile` keys you spotted — a small PR is welcome.
- Interestingly, your Grok wording of the advanced-section hint is the more complete one: the Codex-side `advancedSectionHint` only mentions Chat Completions needing routing takeover, but Anthropic Messages needs it as well. Aligning the Codex copy would be a nice tiny follow-up.

PR #6524, opened 2026-08-17
fix(codex-oauth): support multiple accounts under same ChatGPT workspace (#5885)

### LINKED ISSUE #5885: Codex OAuth: Team workspace members overwrite each other in OAuth Auth Center
## Bug description

When adding multiple ChatGPT accounts from the same Team/Business workspace into **ChatGPT (Codex OAuth) Auth Center**, the second account replaces the first one instead of keeping both accounts.

This prevents managing multiple Team members' Codex OAuth accounts in CC Switch.

## Environment

- OS: Windows (tested)
- CC Switch version: latest version
- Provider: ChatGPT (Codex OAuth)

## Steps to reproduce

1. Have two different ChatGPT user accounts (A and B) under the same Team/Business workspace.
2. Open CC Switch → OAuth Auth Center.
3. Login with account A.
4. Account A appears normally in the account list.
5. Login with account B using "Add Another Account".
6. Account B appears, but account A disappears automatically.

## Expected behavior

Both accounts should remain available:

```
Team Workspace
├── Member A → OAuth credential A
└── Member B → OAuth credential B
```

Users should be able to select which credential/provider to use.

## Actual behavior

The second login overwrites the first one:

```
Login A
→ Account A stored

Login B
→ Account B stored
→ Account A removed
```

## Possible cause

It looks like the account storage key may be based on `chatgpt_account_id` / workspace-level identifier rather than a unique user credential identifier.

For users in the same Team workspace, multiple members may share the same account/workspace identifier, causing a collision.

A possible solution would be separating:

- user credential identity (unique OAuth user/member)
- workspace/account identity (used for ChatGPT-Account-ID header)

so multiple users under the same workspace can coexist.

## Additional notes

Independent ChatGPT accounts from different workspaces appear to work correctly; the issue seems specific to multiple members inside the same Team/Business workspace.

Thanks for maintaining this project!

yulin0629:
## Summary / 概述

Fixes #5885. Resolves an issue where multiple ChatGPT accounts overwrite each other in CC Switch by adopting a composite natural key (`user_identity::workspace_id`).

Before this change the stored account key was the workspace ID itself (`chatgpt_account_id`, falling back to the first `organizations[].id`), so two collision scenarios were indistinguishable:

1. **Multiple users in the same Team/Enterprise workspace** (sharing the same `chatgpt_account_id`).
2. **The same user account subscribed across multiple workspaces** (e.g. personal space + company team).

### Key Changes

- **Composite Primary Key (`user_identity::workspace_id`)**: `account_id` resolves to `format!("{user}::{workspace}")` (or `user` for personal workspaces without a workspace UUID). User identity is ranked `openai_auth.user_id` > `sub` > `email` and is taken **from the `id_token` only**; the `access_token` is a fallback used only when the `id_token` carries no personal claim at all. Both choices exist for stability: `user_id`/`sub` are opaque and immutable while an email can be changed, and the Codex CLI rotates the access token during self-refresh — letting it decide the key would compute a *different* key after a rotation, at which point `~/.codex/auth.json` no longer maps back to the account and the CLI-rotated `refresh_token` is orphaned.
- **Separate Workspace UUID**: The upstream `chatgpt_account_id` is preserved independently for HTTP headers (`chatgpt-account-id` on `/wham/usage` & `/codex/models`) and for writing into the native `~/.codex/auth.json` (`tokens.account_id`).
- **Case-insensitive email**: When email *is* the key (no `user_id`/`sub` available), it is lower-cased so `Alice@example.com` and `alice@example.com` cannot split one account into two entries. `user_id` / `sub` are opaque identifiers and are kept verbatim.
- **Pre-upgrade keys are never reused**: A login always writes the composite key. It deliberately does *not* adopt the pre-upgrade workspace-ID entry, because a re-login by member B of a shared workspace is byte-for-byte indistinguishable from a re-login by member A — adopting that key could silently overwrite someone else's credentials. The old entry is left untouched (the UI already shows it as needing re-login) and the user re-binds it by picking the account once. As a result the storage key is always identical to the identity in the claims, which is what makes every ownership check below decidable.
- **Shared identity resolution**: The login path and the managed live-auth marker in `codex_config.rs` now go through the same resolver, so the same account can no longer be assigned two different keys (which would make a CLI-rotated `refresh_token` unrecognizable).
- **Workspace resync**: A refreshed or adopted token re-derives the workspace ID from the new `id_token` claims rather than short-circuiting on the stored non-empty field, so a changed workspace actually replaces the stale one instead of being re-emitted in `chatgpt-account-id` and in the native `auth.json`.
- **Managed Official session validation**: The expected workspace ID is now resolved from `CodexOAuthManager`. A managed provider only persists placeholder auth, so reading it from `settings_config` would reject every request from the Team/Enterprise accounts this change is meant to support.
- **Display Name with Workspace Tag**: Formats `account.login` as `email (short_workspace_uuid)` (e.g. `user@example.com (e7a7bd67)`) so users can distinguish multiple workspace cards.
- **Claims-first live-auth ownership**: `codex_live_auth_is_managed_chatgpt_login` now matches `~/.codex/auth.json` against the identity derived from its own token claims. `tokens.account_id` only holds the workspace UUID, so matching on it alone would let one member adopt or delete another member's live credentials. Workspace-level matching is kept only for keys without `::` (pre-upgrade entries) and for logins whose claims cannot be parsed at all. Ownership is also **decoupled from the managed-bundle shape whitelist** — an unknown field added by a future Codex CLI would otherwise make the whitelist fall through and every composite-key account fail to recognize its own `auth.json` — and it compares against **every** identity claim present in the tokens rather than only the highest-ranked one, so a rotation that changes which claims are present still resolves to the same account.
- **Rollback no longer conflates two members of one workspace**: `restore_preserving_newer_same_account_auth` keeps the on-disk auth only when the *same* account demonstrably refreshed it. That check used `tokens.account_id`, i.e. the workspace, so a failed `alice -> bob` transaction would treat bob's newer `auth.json` as "the same account, newer generation" and leave bob's credentials on disk while cc-switch considers alice active. It now uses the claims identity.
- **`tokens.account_id` never receives a composite key**: writing the native `auth.json` previously fell back to `account_id` when no workspace was known, which would put an email into the field the Codex CLI sends as `ChatGPT-Account-Id`. The field is now omitted instead; the fallback is limited to pre-upgrade bare-workspace keys.
- **One default-account resolution per request**: the forwarder resolved the default account twice (once inside `get_valid_token()`, once for the header), so a concurrent `set_default_account` between them could pair account A's bearer token with account B's workspace header.
- **Unit Tests**: Same-workspace multi-user coexistence, same-user multi-workspace coexistence, a re-login leaving the pre-upgrade entry untouched, email case normalization, the key ignoring access-token-only claims, ownership surviving both an unknown extra field and a changed claim set, rollback keeping two workspace members apart, a workspace-less key never leaking into the native `auth.json`, and workspace resync on refresh.

### Known limitations / 已知限制

- **The official passthrough still cannot tell two members of one workspace apart.** A Codex session only ever sends the workspace UUID in `chatgpt-account-id`, so `validate_codex_official_authorization` can confirm the workspace but not which member's session it is. This is pre-existing behavior (the previous code compared against the workspace too) and is not made worse here; distinguishing members would require validating bearer-token ownership, which is out of scope for this PR.
- **Accounts stored before the upgrade need one re-login.** They keep their workspace-ID key until the user picks the account again; see above for why auto-migration is unsafe.
- **Until that re-login, a pre-upgrade account's live-auth ownership check is workspace-level.** If a second member of the same workspace logs in on the same machine before the old entry is re-bound, that check can still misjudge. Re-logging in resolves it permanently.
- **The on-disk store is still `version: 1` while the key semantics changed.** Downgrading to an older build is not handled: it would read a composite `account_id` as if it were a workspace UUID. Happy to bump the version if you prefer that.

## Related Issue / 关联 Issue

Fixes #5885

## Checklist / 检查清单

### DIFF
--- a/src-tauri/src/codex_config.rs
+++ b/src-tauri/src/codex_config.rs
@@ -283,12 +283,9 @@ impl CodexLiveStateSnapshot {
         if auth.get("auth_mode").and_then(Value::as_str) != Some("chatgpt") {
             return None;
         }
-        let account_id = auth
-            .pointer("/tokens/account_id")
-            .and_then(Value::as_str)
-            .map(str::trim)
-            .filter(|account_id| !account_id.is_empty())?
-            .to_string();
+        // 必须用 claims 推导的身份：`tokens.account_id` 只是工作区 ID，会把同工作区的
+        // alice / bob 判成同一个账号。
+        let account_id = chatgpt_auth_identity(&auth)?;
         let last_refresh_ms = auth
             .get("last_refresh")
             .and_then(Value::as_str)
@@ -373,12 +370,31 @@ pub(crate) fn codex_managed_oauth_live_auth_marker_exists() -> bool {
     get_codex_managed_oauth_live_auth_marker_path().exists()
 }
 
+/// 从 ChatGPT `auth` 推导账号主键：claims 里的身份优先，完全推不出来才退回
+/// `tokens.account_id`。
+///
+/// 这条 fallback ladder 只能有一份：marker 记录与回滚身份判定若各写一套，改动其中
+/// 一处就会让两边算出不同的 key。
+fn chatgpt_auth_identity(auth: &Value) -> Option<String> {
+    parse_codex_auth_identity(auth)
+        .and_then(|(identity, workspace)| {
+            crate::proxy::providers::codex_oauth_auth::resolve_account_identity_from(
+                &identity, workspace,
+            )
+            .0
+        })
+        .or_else(|| {
+            auth.as_object()
+                .and_then(tokens_account_id)
+                .map(str::to_string)
+        })
+}
+
 /// 从 live/备份的 Codex `auth` 中提取 `account_id`，用于 marker 记录/比对。
 ///
 /// 仅接受 ChatGPT 登录形状（`auth_mode == "chatgpt"`、`OPENAI_API_KEY` 可清空）。
 /// 托管账号写入的完整 bundle 会额外带 `tokens.refresh_token` 与顶层 `last_refresh`，
-/// 这里一并容忍。所有权按 account-scoped 内容判断；Codex CLI 自刷新会轮换
-/// access_token，因此短期 token 指纹不能作为稳定的删除谓词。
+/// 这里一并容忍。
 fn extract_codex_managed_oauth_account_id(auth: &Value) -> Option<String> {
     let auth_obj = auth.as_object()?;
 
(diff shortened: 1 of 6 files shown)


farion1231:
Thanks a lot for this PR — and apologies it sat while the bigger rework was in flight. The same problem was ultimately solved by #6780, which re-keys accounts locally and keeps the OIDC subject as proof of user identity (plus identity validation on the takeover path), and it shipped in v3.20.1. Closing this one in its favor — the work on the composite-key approach is appreciated.

https://github.com/farion1231/cc-switch/releases/tag/v3.20.1

PR #6527, opened 2026-08-17
fix(codex): handle native login availability checks

SaladDay:
## Summary

- Keep the Follow Codex login option visible and disabled while availability is checked.
- Show a retry action when the check fails.
- Block saving until a native-login choice is confirmed.

## Related Issue

Follow-up to #3879.

## Checklist

- [x] `pnpm test:unit`
- [x] `pnpm typecheck`
- [x] `pnpm build:renderer`
- [x] `pnpm format:check`
- [x] Updated i18n files for the new messages.


SaladDay:
Closing this follow-up because the next change will restore the previous behavior where multiple Official providers can follow the current Codex login. The native-login availability check will no longer be needed.

PR #6530, opened 2026-08-17
fix(codex): coalesce commentary with tool calls

### LINKED ISSUE #6529: Codex Responses→Chat splits commentary and tool calls into consecutive assistant messages
## Summary

The Codex Responses → Chat Completions translator emits two consecutive `assistant` messages when one Responses model turn contains an assistant commentary `message` immediately followed by one or more `function_call` items.

This is not only a cosmetic shape difference. Chat models can learn or imitate the first text-only assistant message as a complete turn and return `finish_reason=stop` before emitting the tool call that Codex expects next. In long agent tasks this appears as a turn ending after a progress update even though the task is unfinished.

中文概述：同一 Responses turn 中的 commentary `message` 与紧随其后的 `function_call` 被投影成两条连续 assistant。Chat 模型可能在第一条进度播报后直接 `stop`，导致长任务提前结束。

## Reproduction

On current `main` (`3d126f45`), pass this history to `responses_to_chat_completions`:

```json
[
  {"type":"reasoning","summary":[{"type":"summary_text","text":"need to update the file"}]},
  {"type":"message","role":"assistant","content":"Part 1 written. Appending sections 4-5."},
  {"type":"function_call","call_id":"call_1","name":"apply_patch","arguments":"{\"patch\":\"next section\"}","reasoning_content":"need to update the file"},
  {"type":"function_call_output","call_id":"call_1","output":"Success"}
]
```

### Actual Chat history

```json
[
  {"role":"assistant","content":"Part 1 written. Appending sections 4-5.","reasoning_content":"need to update the file"},
  {"role":"assistant","content":null,"tool_calls":[...]},
  {"role":"tool","tool_call_id":"call_1","content":"Success"}
]
```

### Expected Chat history

```json
[
  {"role":"assistant","content":"Part 1 written. Appending sections 4-5.","reasoning_content":"need to update the file","tool_calls":[...]},
  {"role":"tool","tool_call_id":"call_1","content":"Success"}
]
```

The regression test fails on current `main` with `left: 3, right: 2`.

## Root cause

`flush_pending_tool_calls()` always creates a new assistant message. It does not recognize that the directly adjacent assistant commentary item and the pending calls belong to the same Responses model turn.

The same reasoning can also appear on both the commentary item and the function-call item, so coalescing must avoid duplicating `reasoning_content`.

## Impact and validation

This was reproduced with Qwen through the Codex Responses → Chat path during long multi-tool tasks. After coalescing the adjacent items in an installed build, transparent upstream logs showed the commentary text and `tool_calls` on the same assistant message, and a real long task continued through repeated tool iterations instead of ending at a progress-only message.

## Proposed fix

When flushing pending calls, merge them into the directly adjacent assistant message only if that message has no existing `tool_calls`; preserve its content and attach only reasoning not already present. Keep the existing new-message path for all other role and boundary shapes.

A PR with a focused RED/GREEN regression and the minimal implementation will follow.

BigStrongSun:
## What changed

- merge a directly adjacent assistant commentary message with pending Chat `tool_calls`
- preserve the commentary content instead of emitting a second `content: null` assistant message
- avoid duplicating `reasoning_content` when the same reasoning is attached to both Responses items
- add a Qwen-shaped regression covering commentary, repeated reasoning, a function call, and its output

## Root cause

One Codex Responses model turn can contain a commentary `message` followed by separate `function_call` output items. The translator serialized those items as consecutive Chat assistant messages. Chat models could then treat the text-only progress message as a completed turn and return `finish_reason=stop` before producing the tool call expected by the agent loop.

The fix is intentionally limited to the directly adjacent assistant message with no existing `tool_calls`. User/tool boundaries and the existing fallback that creates a new assistant tool-call message are unchanged.

Fixes #6529.

## Validation

- RED: focused regression failed on `upstream/main@3d126f45` with 3 messages instead of 2
- GREEN: focused regression passed
- `cargo test proxy::providers::transform_codex_chat::tests --lib` — 90 passed
- `git diff --check`
- strict UTF-8 decode, no BOM or U+FFFD

The same transformation has also been exercised in an installed CCSwitchMulti build: transparent upstream logs showed commentary and `tool_calls` on one assistant message, and a real long Qwen task continued through repeated tool iterations without the previous progress-only premature stop.

### DIFF
--- a/src-tauri/src/proxy/providers/transform_codex_chat.rs
+++ b/src-tauri/src/proxy/providers/transform_codex_chat.rs
@@ -892,6 +892,28 @@ fn flush_pending_tool_calls(
     // new assistant tool-call turn. Consecutive outputs do not enter here
     // because `pending_tool_calls` is empty after the first output.
     flush_pending_chat_tool_media(messages, pending_media);
+
+    // A Responses turn may emit a commentary `message` item followed by one
+    // or more tool-call items. Chat Completions represents that as one
+    // assistant message containing both `content` and `tool_calls`. Keeping
+    // the items as two consecutive assistant messages teaches chat models
+    // that a text-only progress update is a complete turn, so they can stop
+    // before emitting the tool call on the next sample.
+    if let Some(previous) = messages.last_mut().filter(|message| {
+        message.get("role").and_then(Value::as_str) == Some("assistant")
+            && message.get("tool_calls").is_none()
+    }) {
+        if let Some(previous_obj) = previous.as_object_mut() {
+            previous_obj.insert(
+                "tool_calls".to_string(),
+                Value::Array(std::mem::take(pending_tool_calls)),
+            );
+            attach_unique_pending_reasoning_to_assistant(previous, pending_reasoning);
+            *last_assistant_index = Some(messages.len() - 1);
+            return;
+        }
+    }
+
     let mut message = json!({
         "role": "assistant",
         "content": null,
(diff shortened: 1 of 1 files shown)


farion1231:
@codex review

farion1231:
Superseded by #7280 (@fszcd), merged to main as 5e0f3442e. The production change is the same coalescing rule as yours (merge pending tool calls into the directly adjacent assistant message when it has no `tool_calls` yet); that PR also deduplicates pending reasoning per segment for parallel tool calls, which addresses the review comment left here, and carries seven more boundary regressions, so it was the one to take.

Thank you for the fix and for the Qwen replay that confirmed the premature `finish_reason=stop`. The release note will credit you as well. Closing this one.

PR #6532, opened 2026-08-17
fix(docs): repair broken star history chart

Dessalines39394:
The star history chart in the README has stopped rendering because it relies on the GitHub stargazer API, which now restricts access. This swaps the chart to an alternative source that needs no API token, so it works reliably again. The badge remains untouched. Updated in README, README_ZH, README_JA, and README_DE.


Dessalines39394:
?

PR #6534, opened 2026-08-17
fix(codex): reconcile edit form key with live bearer token

### LINKED ISSUE #6414: 同一个 API 请求地址 ,按照不同key 配置不同的供应商。经常会自动修改apikey 改成和其他一样的
### Self Checks / 自检

### CC Switch Version / 版本号

3.19.2

### Operating System / 操作系统

macOS

### Related App / 涉及应用

Codex

### Steps to Reproduce / 重现步骤

Image上面公司代理的地址，按照不同模型分配不同key，经常发现使用某一个时，突然就提示“model xxx not found”，一排查发现，我使用的，apikey被修改了

### Expected Behavior / 期望行为

不要自动更换apikey

### Actual Behavior / 实际行为

_No response_

### Additional Context / 补充信息

_No response_

YUZHEthefool:
## Problem

Fixes #6414.

When multiple Codex providers share the same base URL but use **different** API keys (e.g. a company proxy that maps different keys to different models), cc-switch silently overwrites one provider's stored API key with another's. Symptom reported in the issue: a provider suddenly starts failing with `model xxx not found`, and investigation shows its `apiKey` has been changed to match a different provider's.

> **Note on commit history:** the first two commits (`fix ... backfill` + its `Revert`) cancel out and are left in only because force-pushing the branch was not authorized; the **net change** of this PR is the third commit, `fix(codex): reconcile edit form key with live bearer token`. Reviewers may squash-merge.

## Root cause

My first hypothesis (the switch-away backfill adopting the shared `auth.json`) was **wrong** — `tests/provider_commands.rs::switch_provider_updates_codex_live_and_state` encodes that adoption as **intentional** (it captures a user's manual live `auth.json` edit). The CI failure on that first attempt confirmed it. The real root cause is in the **edit form**:

In **bearer-token mode** (`preserveCodexOfficialAuthOnSwitch` enabled — the setting that keeps a ChatGPT login while routing third-party providers), switching to a provider writes its key to `config.toml`'s `experimental_bearer_token` and **preserves** the shared `~/.codex/auth.json`. `auth.json` is a single shared slot with no provider identity, so it may hold **another provider's stale key** (left over from a previous default-mode switch or a different provider).

When the user then opens the edit dialog for the current provider:
1. `EditProviderDialog` loads `liveSettings` as the form base (only for the current provider).
2. `useCodexConfigState` initializes `codexAuth` from the stale `auth.json`.
3. `pickCodexApiKey` prefers `auth.OPENAI_API_KEY` over the bearer token, so the form **displays the stale key**.
4. On save, `ProviderForm` writes `codexAuth` straight into `settingsConfig.auth` (`JSON.parse(codexAuth)`), persisting the stale key back into the provider's DB record.

Repeated edits made keys silently converge across providers sharing a base URL, surfacing as `model xxx not found` once the wrong key reached a model it was not entitled to.

## Fix

When loading Codex config into the form (`useCodexConfigState` init + `resetCodexConfig`), lift the config's `experimental_bearer_token` into `auth.OPENAI_API_KEY` when present and differing, so both the displayed key and the saved auth carry the correct per-provider key. This mirrors the backend `restore_codex_provider_token_for_backfill` lift (which already treats the bearer as authoritative on switch-away backfill).

### Why this is safe

The reconciliation is gated on the config actually carrying a bearer token:
- **Bearer present** (bearer-token mode): the bearer is the per-provider authoritative key written on the last switch; the shared `auth.json` is stale → reconcile. Fixes the contamination.
- **No bearer** (default mode, or a manual live edit with no bearer): no reconciliation → `auth.json` is the active key slot (in sync with the current provider) and manual live edits are preserved exactly. This is the intentional behavior covered by existing tests (`switch_provider_updates_codex_live_and_state`, `EditProviderDialog.test.tsx` "保留 Codex 数据库中的 modelCatalog").
- Official providers (native ChatGPT auth, no bearer) are unaffected.
- The reconciliation preserves all other `auth` fields (e.g. a preserved `tokens` login), only setting `OPENAI_API_KEY`.

## Tests

Added `tests/hooks/useCodexConfigState.bearer.test.ts` (4 cases):
- lifts the config bearer over a stale shared `auth.json` key;
- surfaces the bearer when `auth.json` has no key (only preserved login material), keeping other auth fields;
- does **not** reconcile when the config has no bearer (manual live edit preserved);
- no-op when bearer already matches `auth.OPENAI_API_KEY`.

Verified locally:
- `tests/hooks/useCodexConfigState.*` — 6 passed (2 existing + 4 new)
- `tests/components/EditProviderDialog.test.tsx` — 7 passed (incl. the live-adopt + proxy-takeover cases)
- `tests/components/ProviderForm.codexCatalog` / `CodexOAuthSection` / `tests/utils/providerConfigUtils.codex` — 42 passed

No backend change, so the previously failing `switch_provider_updates_codex_live_and_state` passes again.

Fixes #6414.

### DIFF
--- a/src-tauri/src/codex_config.rs
+++ b/src-tauri/src/codex_config.rs
@@ -2846,9 +2846,20 @@ pub fn prepare_codex_provider_live_config(
 /// `auth.OPENAI_API_KEY` so the stored provider keeps its canonical shape
 /// and generated live tokens don't leak into stored provider TOML.
 ///
-/// Only intervenes when the live config actually carries a bearer token —
-/// otherwise the function is a no-op so the caller's normal backfill path
-/// (which keeps live `auth` as the authoritative source) is unaffected.
+/// When the live config carries a per-provider `experimental_bearer_token`,
+/// that token is provider-scoped and safe to lift back into the stored auth.
+///
+/// When it does NOT (the default `preserve_codex_official_auth_on_switch =
+/// false` mode keeps the active key in the shared `auth.json` instead), the
+/// live `auth` is a single-slot file with no provider identity. It may hold a
+/// *different* provider's key after a backup/restore cycle, an in-app ChatGPT
+/// login, or any current/live divergence. Adopting it here would overwrite
+/// this provider's stored key with another's, so keys silently converge across
+/// providers that share a base URL (#6414). In that case keep the provider's
+/// own stored `auth` instead — the live `config.toml` changes below the auth
+/// slot are still captured by the caller. This mirrors the #6277 restore-side
+/// rule of never letting the shared live credential clobber per-provider
+/// storage.
 pub fn restore_codex_provider_token_for_backfill(
     settings: &mut Value,
     template_settings: &Value,
@@ -2862,6 +2873,18 @@ pub fn restore_codex_provider_token_for_backfill(
     };
 
     let Some(token) = extract_codex_experimental_bearer_token(&config_text) else {
+        // No per-provider bearer token: do NOT adopt the shared live
+        // `auth.json` — it may belong to another provider. Preserve this
+        // provider's own stored auth so its key is not silently overwritten.
+        if let Some(obj) = settings.as_object_mut() {
+            if let Some(stored_auth) = template_settings
+                .get("auth")
+                .filter(|value| value.is_object())
+                .cloned()
+            {
+                obj.insert("auth".to_string(), stored_auth);
+            }
+        }
         return Ok(());
     };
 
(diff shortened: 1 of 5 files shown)


YUZHEthefool:
Addressed the review finding while preserving issue #6414's intended current-provider convergence.

Changes:
- moved bearer-to-auth reconciliation out of the generic useCodexConfigState hook;
- apply it only when editing the current Codex provider from a real live snapshot;
- rebuild auth from the stored provider auth and replace only OPENAI_API_KEY, avoiding persistence of shared auth.json login material;
- keep non-current DB providers and presets auth-first;
- preserve official and OAuth-only provider semantics.

Added regression coverage for current-live reconciliation, non-current DB precedence, preset precedence, retained provider auth fields, and OAuth-only auth.

Rebase update:
- rebased the PR onto the latest main;
- resolved the EditProviderDialog import conflict by retaining main's current structure and only the bearer-token parser required by this fix;
- GitHub now reports the PR as mergeable.

Post-rebase verification:
- pnpm format:check
- pnpm typecheck
- targeted Vitest regression tests (14/14)
- pnpm build:renderer
- git diff --check

Commit: 41613272

farion1231:
@codex review

YUZHEthefool:
rebase it to main

farion1231 (approved):

farion1231:
Merged — thank you for this fix, and especially for the honest root-cause writeup.

What stood out in review:

- The self-correction process was exemplary: the first hypothesis (switch-away backfill) got caught by the existing `switch_provider_updates_codex_live_and_state` test, and instead of forcing it through, you revisited the premise and found the real leak in the edit form's live-adoption path.
- The final scope is exactly right: reconciling only at the EditProviderDialog boundary (current provider + real live snapshot) keeps non-current providers, presets, and default-mode manual live edits untouched — the two earlier, broader placements would each have broken one of those.
- Rebuilding auth from the stored provider template and replacing only `OPENAI_API_KEY` mirrors the backend backfill lift and keeps shared login material out of provider rows. The regression tests lock all of these boundaries down.

Verified on merge trees against latest main: typecheck + full frontend suite green. This also composes cleanly with the upcoming Codex switch refactor, where it becomes even more important. #6414 is closed by this merge.

PR #6535, opened 2026-08-17
fix(codex): allow multiple follow-login providers

SaladDay:
Follow-up to #3879.

This keeps the previous Codex Official behavior and makes OAuth account binding optional.

- allow multiple Codex Official providers to use the current Codex login
- keep provider IDs unchanged when binding or unbinding an account
- keep existing provider data unchanged on startup
- preserve direct API-key and custom-upstream behavior

Tests:
- pnpm typecheck
- pnpm vitest run
- cargo test --manifest-path src-tauri/Cargo.toml --lib
- cargo test --manifest-path src-tauri/Cargo.toml --test provider_service
- cargo clippy --manifest-path src-tauri/Cargo.toml --lib -- -D warnings
- real Codex CLI request

### DIFF
--- a/src-tauri/src/lib.rs
+++ b/src-tauri/src/lib.rs
@@ -764,16 +764,6 @@ pub fn run() {
                 Err(e) => log::warn!("✗ Failed to seed official providers: {e}"),
             }
 
-            match crate::services::provider::ProviderService::migrate_legacy_codex_official_managed_binding(
-                &app_state,
-            ) {
-                Ok(Some(provider_id)) => log::info!(
-                    "✓ Migrated legacy Codex Official account binding to {provider_id}"
-                ),
-                Ok(None) => {}
-                Err(e) => log::warn!("✗ Failed to migrate legacy Codex Official binding: {e}"),
-            }
-
             {
                 let db_for_codex_history_migration = app_state.db.clone();
                 tauri::async_runtime::spawn_blocking(move || {
--- a/src-tauri/src/proxy/providers/codex.rs
+++ b/src-tauri/src/proxy/providers/codex.rs
@@ -219,20 +219,91 @@ pub fn provider_needs_responses_namespace_flatten(provider: &Provider) -> bool {
     provider.is_xai_oauth()
 }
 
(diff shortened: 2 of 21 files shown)


farion1231:
@codex review

farion1231:
Thanks @SaladDay — this is the right call, and the timing is what makes it cheap: unwinding the mandatory-binding / single-card / ID-swap trio from #3879 before it ever ships means zero migration debt, and the net −361 lines say the old design was fighting the grain. What stood out in review: the content-based official detection fails safe in both directions (a stale `official` label with a third-party base_url or a stored key can no longer be sent to the ChatGPT backend, while an unbound official card simply follows the current login again), the frontend/backend mirrors are pinned by matching fixtures on both sides, and deleting the whole ID-replacement/startup-migration machinery is the kind of negative-code fix we like best. Bonus: dropping the `native-login-exists` query closes the loading-flicker follow-up from #3879 for free.

About the conflict with #6537: we pushed a reconciliation merge to your branch so this could land without another round-trip. The semantic merge point — on codex, `isCodexOauth` now means "has a bound managed account", which is exactly the intersection of your two PRs (#6537 widened it to bound cards missing `providerType`; this one narrows it to exclude unbound follow-login cards). Both test probes were kept (quota-footer props recorder + duplicate/configure-usage buttons). Full suite re-verified on the merged tree: 2677 backend / 990 frontend, all green.

A few non-blocking leftovers if you ever feel like a small follow-up — no pressure:
- The per-request official check parses TOML up to four times on the forwarding hot path; harmless next to network latency, but the verdict is cacheable per provider.
- The unified-session projection is now recognized in three places (Rust inject/strip + the TS four-key whitelist); if the injected shape ever grows a key, all three need the same change.
- The strip-on-update guard still keys on `category == "official"` — aligning it with the new content predicate would cover category-less managed rows too.
- Dead-code sweep: `replace_provider_id` (DAO), the `ensure_codex_official_provider` command + frontend wrapper, the redundant branch in `supportsOfficialProxyTakeover`, and the two orphaned locale keys (`codex.accountNotSelected` / `codex.chooseAccount`).

PR #6537, opened 2026-08-17
fix(codex): make OAuth usage queries configurable

SaladDay:
Summary:
- Allow bound Codex Official providers to configure OAuth usage queries.
- Keep the existing enabled default and allow disabling or changing the polling interval.
- Keep account-scoped quota data out of the app-wide tray cache.

Tests:
- pnpm test:unit
- pnpm typecheck
- cargo check --manifest-path src-tauri/Cargo.toml
- cargo test --manifest-path src-tauri/Cargo.toml managed_codex_quota_stays_out_of_the_app_wide_tray_cache


farion1231:
@codex review

farion1231:
Thanks @SaladDay — another solid follow-up to #3879. What stood out in review: the tray guard is exactly the right instinct (the tray's subscription cache is app-wide, so account-scoped quota from a bound card has no safe representation there — refusing to write beats mislabeling another account's numbers), the modal's test-query now hits the bound account and writes the per-account cache key instead of the live-login one, and the hook change mirrors the existing `useSubscriptionQuota` pattern instead of inventing a new knob. The three ProviderCard tests pin each behavior (default-on, saved-off, custom interval) nicely.

A few non-blocking leftovers if you ever feel like a small follow-up — no pressure:
- In the UsageScriptModal reset branch (official provider carrying a stale non-official `templateType`), `enabled` is preserved but the saved `autoQueryInterval` falls back to 5; carrying it over is a one-liner.
- The native-login card (`providerType: codex_oauth`, unbound) still polls at a fixed 5 minutes with the configure button hidden — the same treatment you gave bound cards would extend to it naturally.
- The tray now intentionally shows no quota for bound cards; a per-account tray query would be the longer-term way to fill that gap.

Heads-up: now that this is merged, your #6535 will need a rebase — both touch `ProviderCard.tsx` and `ProviderCard.codexAccount.test.tsx`.

PR #6597, opened 2026-08-19
fix(i18n): align Grok Build and Codex routing copy

xwil1:
## Summary

- remove the orphaned `grokBuild.apiBackend` and `grokBuild.profile` locale keys from all four supported locales
- update the Codex advanced-section hint and fallback copy to state that both Chat Completions and Anthropic Messages upstreams require routing takeover
- add locale coverage so every supported translation keeps both routed protocols in the hint

## Background

This follows up on the two optional cleanup suggestions from #6511.

The Grok Build API-backend and profile controls were removed previously, but their translation keys remained unused. The Codex hint also mentioned only Chat Completions even though Anthropic Messages is bridged through the local routing layer as well.

This PR changes only bundled copy and dead locale entries; it does not change routing or protocol-conversion behavior.

## Validation

- `pnpm test:unit -- tests/config/localeCoverage.test.ts tests/components/GrokBuildProviderForm.test.tsx` (21 tests passed)
- `pnpm exec prettier --check ...`
- `pnpm typecheck`
- Codex Security diff review: complete coverage, no findings

### DIFF
--- a/src/components/providers/forms/CodexFormFields.tsx
+++ b/src/components/providers/forms/CodexFormFields.tsx
@@ -902,7 +902,7 @@ export function CodexFormFields({
                   })
                 : t("codexConfig.advancedSectionHint", {
                     defaultValue:
-                      "包含上游格式、模型映射、思考能力与自定义 User-Agent。使用 Chat Completions 协议的供应商需开启路由接管才能使用。",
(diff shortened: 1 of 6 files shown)


farion1231:
@codex review

PR #6621, opened 2026-08-19
fix(tools): consult GitHub Releases when local Hermes version leads PyPI

### LINKED ISSUE #6618: Hermes 最新版本显示 0.19.0 但本地已安装 0.20.4（PyPI 滞后于 git 主线）
### 版本号 / CC Switch Version
v3.20.0 (Windows)

### 描述 / Description
Hermes 的「最新版本」显示 0.19.0，但本地已安装 0.20.4（git 安装，GitHub tag v2026.8.18），出现「最新版本 < 当前版本」的矛盾。

### 复现步骤 / Steps to reproduce
1. 打开 设置 → 关于（工具管理面板）
2. 查看 Hermes 一行的 当前版本 / 最新版本

### 预期行为 / Expected Behavior
最新版本应 >= 当前版本。Hermes 最新主线为 v0.20.4（2026-08-18 发布，tag v2026.8.18）。

### 实际行为 / Actual Behavior
当前版本 0.20.4，最新版本 0.19.0。

### 根因推测 / Root cause hypothesis
Hermes 的「最新版本」检测似乎从 PyPI 查询（pypi.org/pypi/hermes-agent/json -> 0.19.0，发布于 2026-07-20），但 Hermes 官方主线已迁移到 git/GitHub releases 分发，PyPI 发布滞后于主线（0.20.x 尚未发布到 PyPI），导致最新 < 当前。

### 环境 / Environment
- CC Switch v3.20.0 (Windows)
- Hermes 0.20.4（git 安装，HERMES_HOME=%LOCALAPPDATA%\hermes）

### 建议 / Suggestion
Hermes 的 latest 版本检测改用 GitHub Releases API（api.github.com/repos/NousResearch/hermes-agent/releases/latest），与 CC Switch 自身的 updater 一致；或对 git 安装方式的 Hermes 跳过 PyPI 检查。

SailingLoong:
Hi 👋 Another small one from the same sync-with-upstream effort as #6482.

Fixes #6618 (and the earlier report #6475).

## Problem

The tools panel queries PyPI for Hermes' "latest version", but Hermes' mainline now ships through GitHub Releases and PyPI lags behind (currently `0.19.0` vs `0.20.4`). Git-installed users get a contradictory "latest < current" display.

One extra wrinkle: the repo tags releases with calendar versions (`v2026.8.18`) while the CLI reports semver (`0.20.4`), and the semver only survives in the release *name* ("Hermes Agent v0.20.4 (2026.8.18)") — so a plain tag-based lookup wouldn't be comparable either.

## Fix

- Keep PyPI as the primary source — it remains the channel cc-switch installs/upgrades Hermes through, so pip users keep seeing exactly what they can upgrade to, with no extra request.
- Only when the local version strictly leads PyPI's `latest` (i.e. a git/mainline build — the situation in the issue), or when PyPI is unreachable, consult the GitHub latest release. Same shape as `fetch_npm_latest_for_tool`'s "local ahead → check the pre-release channel" logic.
- `fetch_github_latest_version` now prefers a semver parsed from the release `name`, falling back to the `tag_name` as before. For opencode this is behavior-neutral — its current releases use `name == tag_name` (`v1.18.18`).

## Validation

- `cargo test --lib commands::misc::tests` — 96 passed, including 3 new tests: name-over-calendar-tag extraction, tag fallback when the name has no semver, and the strictly-greater gating.
- Spot-checked the live endpoints: PyPI `hermes-agent` → `0.19.0`; GitHub `releases/latest` → name "Hermes Agent v0.20.4 (2026.8.18)", tag `v2026.8.18`. A local `0.20.4` build now resolves latest `0.20.4` instead of `0.19.0`.
- `cargo fmt` clean.

Happy to adjust — e.g. if you'd rather always show the GitHub release for Hermes regardless of install channel, that's a two-line change here.

### DIFF
--- a/src-tauri/src/commands/misc.rs
+++ b/src-tauri/src/commands/misc.rs
@@ -821,7 +821,7 @@ async fn get_single_tool_version_impl(
             }
         }
         "openclaw" => fetch_npm_latest_for_tool(&client, "openclaw", tool, local).await,
-        "hermes" => fetch_pypi_latest_version(&client, "hermes-agent").await,
+        "hermes" => fetch_hermes_latest_version(&client, local).await,
         "pi" => {
             fetch_npm_latest_for_tool(&client, "@earendil-works/pi-coding-agent", tool, local).await
         }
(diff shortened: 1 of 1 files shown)


farion1231:
@codex review

farion1231:
Thanks for the PR and the careful write-up. The release-name observation (semver lives in the name, the tag is calendar-versioned) was exactly right, and the fix that landed on main (b1250dc7a) borrows that idea.

Closing this in favour of that commit, mainly because the premise of keeping PyPI as the primary source doesn't hold for cc-switch: it installs Hermes through the official install.sh (a git clone of main) and upgrades through `hermes update` (git pull), so PyPI is never the channel users actually get Hermes from. With PyPI first, anyone still on 0.19.0 would keep seeing "up to date" and never get the upgrade button.

The landed version therefore queries GitHub Releases first with PyPI only as a fallback (the "always GitHub" variant you offered), rejects calendar-looking numbers on both the name and tag paths so 2026.x.y can never be shown as a version, hides a fallback value that is already behind the local version, and adds a 15s request timeout since api.github.com is now hit on every refresh.

Appreciate the work here.

PR #6622, opened 2026-08-19
fix(deeplink): preserve the full model catalog when importing Grok Build links

### LINKED ISSUE #6457: bug(grokbuild): deeplink import collapses multi-model config to one profile

SailingLoong:
Fixes #6457.

## Problem

Importing a `ccswitch://v1/import` grokbuild link whose inline config carries multiple `[model.*]` profiles collapses the catalog: `merge_grokbuild_config` lifts only the selected profile's fields into the flat request, and `build_grokbuild_settings` then rebuilds a single profile from those flat fields — keyed by the URL `model` param, named after the provider card, description dropped. `/model` in Grok Build ends up with one generic entry instead of the configured models.

## Fix

- The merged request now carries the sanitized model catalog (`grokbuild_models`, `#[serde(skip)]` — deeplink-internal, never parsed from or serialized to the URL): per `[model.*]` table the profile key, `model`, `name`, `description` and `context_window`, plus the original `[models] default`.
- `build_grokbuild_settings` emits every profile when the catalog is present. Each profile's `base_url` / `api_key` are overridden with the deeplink's explicit endpoint/key (or the existing merged fallback when URL params are absent), and `api_backend` is forced to the supported `responses` backend.
- Security boundary unchanged from the issue's ask: per-profile `api_key` / `env_key` / `base_url` / `api_backend` from the untrusted link are dropped at extraction time, so the existing env_key rejection and "never inline a resolved env secret" behavior is untouched.
- Default selection: URL `model` matched against profile keys, then profile `model` fields, then the config's original default — `[models] default` always points at an emitted table.
- Links without an inline config keep the previous single-profile output verbatim.

## Tests

- `grokbuild_deeplink_preserves_multi_model_catalog_from_inline_config` — three-profile catalog (one sparse): id/name/description/context_window preserved, URL endpoint+key override every profile, per-profile credentials/URLs/backends absent, output passes `validate_config_toml`.
- `grokbuild_deeplink_default_follows_config_when_url_model_absent` — default follows the config's `[models] default`; the merged fallback key overrides all profiles.
- Full `cargo test --lib`: 2681 passed, 0 failed (pre-existing env_key safety tests included). `cargo fmt` clean.

Happy to reshape the carrier (a side-channel instead of a request field) or move the override boundaries if you'd prefer different trade-offs.

### DIFF
--- a/src-tauri/src/deeplink/mod.rs
+++ b/src-tauri/src/deeplink/mod.rs
@@ -136,4 +136,39 @@ pub struct DeepLinkImportRequest {
     /// Auto query interval in minutes (0 to disable)
     #[serde(skip_serializing_if = "Option::is_none")]
     pub usage_auto_interval: Option<u64>,
+
+    // ============ Grok Build inline config fields ============
+    /// Sanitized model catalog extracted from an inline Grok Build config
+    /// (`grokbuild_models`). Deeplink-internal only: populated by config
+    /// merging, never parsed from or serialized to the URL itself.
+    #[serde(skip)]
+    pub grokbuild_models: Option<GrokDeeplinkModels>,
+}
+
+/// 从 deeplink 内联 config.toml 提取的 Grok Build 模型目录（已消毒）。
+///
+/// 只保留可安全透传的展示/容量字段。`api_key` / `env_key` / `base_url` /
+/// `api_backend` 有意不进入这里：导入结果必须统一使用 deeplink 显式端点与
+/// 密钥（或既有的合并回退），不能让不可信链接按档案夹带凭据或私有 backend。
+#[derive(Debug, Clone, PartialEq)]
+pub struct GrokDeeplinkModels {
+    /// 内联 config 原有的 `[models] default`（指向档案键）
+    pub default_profile: String,
+    /// 全部 `[model.*]` 档案
+    pub profiles: Vec<GrokDeeplinkModel>,
+}
+
+/// 单个模型档案的安全字段（`[model.<profile>]` 表）。
+#[derive(Debug, Clone, PartialEq)]
+pub struct GrokDeeplinkModel {
+    /// `[model.<profile>]` 表键，`[models] default` 指向的就是它
+    pub profile: String,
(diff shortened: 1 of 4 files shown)


farion1231:
@codex review

SailingLoong:
Apologies for the week-long red CI — the Clippy failure (needless `to_string` on `Display` types in the format args) is fixed in 9b6a5468, and the push retriggers the full matrix including the Windows run that hit the flaky mt.exe LNK1327 last time.

PR #6623, opened 2026-08-19
fix: attribute usage to selected upstream endpoint

glyahh:
## Summary

- Add immutable usage endpoint identities and schema v18 migration.
- Attribute proxy usage to the endpoint actually selected after failover, retry, and streaming response processing.
- Keep Codex JSONL imports as Unknown Endpoint when the protocol cannot prove a URL.
- Split Provider statistics and request logs by endpoint, with an endpoint filter query that is independent of pagination.
- Label the dashboard metric as model-processed tokens and local pricing estimate.

## Verification

- `pnpm typecheck`
- `pnpm format:check`
- `pnpm build:renderer`
- Rust lib tests: 2668 passed, 2 pre-existing Windows symlink-privilege failures, 6 ignored.
- Frontend unit tests: endpoint-related tests pass; the full suite retains unrelated existing timeout/duplicate-DOM failures.

The full Windows EXE release build completed. Installer bundling was blocked only by the environment failing to download WiX/NSIS dependencies (`http response missing version`).


farion1231:
@codex review

PR #6643, opened 2026-08-20
fix(hermes): pin discover_models: false when models declared

AlkaidSTART:
## Summary
When writing custom provider configs to `~/.hermes/config.yaml`, CC Switch now automatically writes `discover_models: false` whenever explicit `models` are declared for that provider.

## Background & Root Cause
Hermes defaults to live-probing `/v1/models` from upstream endpoints and treats dict-shaped `models` in its config as per-model metadata rather than an allowlist. As a result, Hermes' model picker would populate extra models discovered live from the endpoint and overwrite the explicit model list configured in CC Switch.

Hermes provides the `discover_models: false` flag to opt out of endpoint discovery and use the declared model list as the authoritative source of truth.

## Changes
- In `src-tauri/src/hermes_config.rs`:
  - Added helper `pin_discover_models_for_write(&mut serde_json::Value)` to inject `discover_models: false` if `models` map is non-empty.
  - Called `pin_discover_models_for_write` in `set_provider` before YAML serialization.
  - Added assertions to unit tests verifying `discover_models: false` is written when models are present, and omitted when no models are declared.

### 1. Unit Tests
All 57 hermes config unit tests pass:
```text
cargo test --manifest-path src-tauri/Cargo.toml -p cc-switch --lib hermes_config::tests::

running 57 tests
test hermes_config::tests::provider_with_models_array_writes_dict_to_yaml ... ok
test hermes_config::tests::provider_without_models_is_unaffected ... ok
...
test result: ok. 57 passed; 0 failed; 0 ignored; 0 measured; 2627 filtered out; finished in 0.03s

### DIFF
--- a/src-tauri/src/hermes_config.rs
+++ b/src-tauri/src/hermes_config.rs
@@ -588,9 +588,9 @@ fn normalize_provider_models_for_write(config: &mut serde_json::Value) {
 /// expect their configured model list to be authoritative. This must run
 /// after forward-compatible fields are restored from disk: an unrelated UI
 /// update can omit `models` while the final provider still declares them.
-fn pin_discover_models_for_write(config: &mut serde_yaml::Value) {
+fn pin_discover_models_for_write(config: &mut serde_yaml::Value) -> bool {
     let Some(map) = config.as_mapping_mut() else {
-        return;
+        return false;
     };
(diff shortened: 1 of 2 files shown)


farion1231:
@codex review

PR #5444, opened 2026-07-16
feat(skills): filter skills by app

zichen0116:
## Summary
- Skills 管理页默认展示全部 Skills，点击 Claude、Codex 等应用标签后，只显示已配置到对应应用的 Skills；点击全部恢复完整列表。
- Add app-based filtering to the Skills management page.
- Keep the default view as all skills and rename the Skills total trigger to the localized All label.
- Add filtered empty-state copy and interaction coverage.
- close #5446

## Verification
- Targeted Skills tests passed.
- pnpm typecheck passed.
- pnpm format:check passed.
- pnpm build:renderer passed.
- Full unit suite still has two existing App.test.tsx integration failures; the file passes in isolation with --testTimeout=15000.

## show
- 点击Claude的效果
9873dc521ca0530d2daf466b607c7ef3

- 点击codex的效果
aa701889ddbefd0d3f17efd62bef14e4

- 点击全部的效果（默认）
85431eab3f0727eec3d0b23d395ab2f0

PR #6644, opened 2026-08-20
feat(skills): undo app toggles and confirm unsaved changes on leave

chronicle12345:
## Summary / 概述

Clicking an app badge at the top of the Skill panel (`Claude: 0` and friends) bulk-enables **every** installed Skill for that app, with no way back: the previous per-Skill assignment is simply gone, and clicking again does not disable them either. That is what [#6536](https://github.com/farion1231/cc-switch/issues/6536) reports.

This PR implements the approach the issue author suggested in the comments:

**1. Undo stack + Ctrl/⌘+Z**

Every app toggle — a row switch or a badge bulk toggle — now pushes its **inverse** onto an undo stack. The entry has the exact shape the bulk-toggle mutation already takes (`{ ids, app, enabled }`), so both entry points share one replay path. `Ctrl/⌘+Z`, or the new "Undo" button (fades in only when there is something to undo, next to "Update All"), replays the most recent step.

**2. Confirm before leaving with unsaved changes**

Going back to the main page (back button or `Esc`) while the stack is non-empty opens a confirmation dialog with three options:

| Option | Behaviour |
|--------|-----------|
| Cancel | Stay on the page, changes untouched |
| Don't save | Replay the whole stack in reverse — older values land last, restoring the state from when the panel was opened — then leave |
| Save | Changes are already on disk; clear the stack and leave |

Toggles still **write immediately**. The stack only remembers how to write them back, so the panel does not need a draft/apply rewrite and the diff stays small.

### Implementation notes

- `UnifiedSkillsPanel.tsx`: undo stack, `handleUndoLast`, keyboard listener, save/discard handlers, undo button, `UnsavedSkillChangesDialog`
- `App.tsx`: the back button and `Esc` now share a single `navigateBack()`, which asks the panel's new `confirmLeave` handle whether to intercept
- Existing guards are reused rather than duplicated: undo and discard both go through `beginWrite`/`endWrite` (so concurrent writes and a double `Ctrl+Z` cannot race), and the dialog joins `dialogOpen`, which already disables the back button, `Esc`, and the toggles behind it
- `Ctrl+Z` inside a text field does not trigger undo (guarded by `isTextEditableTarget`)
- The stack is panel-scoped: it clears on unmount, so nothing leaks across visits
- If a write fails while restoring, the stack is left intact and the error is surfaced as a toast — the user can go back again to retry, or choose "Save" to leave anyway

### Scope

- The MCP and Prompts panels have the same bulk-toggle badges but are untouched here, to keep this PR to one change. Happy to follow up with a separate PR if maintainers want the same treatment.
- The other idea in the issue — turning the badges into filters (#5446, and PR #5444) — is out of scope for this PR.

## Related Issue / 关联 Issue

Refs #6536

### Before

Clicking an app badge enables everything, with no way to undo.

Before

### After

An "Undo" button appears once something is changed; `Ctrl/⌘+Z` does the same.

Undo button

Leaving with unsaved changes asks whether to keep them — cancel / don't save / save.

Unsaved changes dialog

## Checklist / 检查清单

- [x] `cargo clippy` passes (if Rust code changed) / no Rust code changed in this PR
- [x] Updated i18n files if user-facing text changed / en, zh, zh-TW and ja all updated

### Testing

Five new cases in `tests/components/UnifiedSkillsPanel.test.tsx`: `Ctrl+Z` undoes a bulk toggle, the undo button reverts a single row toggle, leaving with no changes is not intercepted, "don't save" replays the stack in reverse and then proceeds, and "save" keeps the changes while clearing the stack. `managementListLocales.test.ts` gained coverage for the new strings across all four locales.

`pnpm test:unit` reports 985 passed / 7 failed. Those 7 (`PiProviderForm.test.tsx` ×5, `App.test.tsx` ×2) fail identically on a clean `main` with this branch stashed, so they are pre-existing and unrelated.


chronicle12345:
> 

Fixed, please re-review

farion1231:
@codex review

chronicle12345:
@codex review

chronicle12345:
@codex review

chronicle12345:
@farion1231 please review

PR #6655, opened 2026-08-20
fix(usage): take cumulative input_tokens from message_delta per Anthr…

dchaofei:
---

Summary

Fixes input_tokens being severely understated for Anthropic-compatible relays that report the final (cumulative) input count in the message_delta usage event.

Per the Anthropic streaming spec (https://platform.claude.com/docs/en/build-with-claude/streaming), the usage object in message_delta is cumulative — its input_tokens is the request's final input count. The previous code only applied the delta value when it was smaller than the message_start estimate, on the assumption that some providers report a corrected "fresh" input in delta. For relays that follow the official cumulative semantics (delta carries the full total, which is larger than the start snapshot), the real value was dropped and input_tokens was pinned to the start estimate, badly understating recorded usage and cost.

This changes the delta handling to:
- adopt input_tokens from message_delta whenever present and > 0 (overriding the message_start estimate)
- keep the message_start value when delta omits input_tokens (official Claude only reports output_tokens there)

It also flips the existing test that encoded the wrong assumption and adds a case covering delta-without-input keeping the start value.

Related Issue

N/A — no related issue.

Screenshots

N/A — backend-only change.

Checklist

- [x] pnpm typecheck passes (no frontend changes in this PR)
- [x] pnpm format:check passes (no frontend changes in this PR)
- [x] cargo clippy passes (cargo clippy --lib -- -D warnings clean)
- [x] cargo test --lib proxy::usage::parser passes (28 tests, including the flipped and new cases)
- [x] Updated i18n files if user-facing text changed (no user-facing text changed)

---

### DIFF
--- a/src-tauri/src/proxy/usage/parser.rs
+++ b/src-tauri/src/proxy/usage/parser.rs
@@ -729,7 +729,10 @@ mod tests {
         ];
 
         let usage = TokenUsage::from_claude_stream_events(&events).unwrap();
-        assert_eq!(usage.input_tokens, 100, "delta 无 input_tokens 时保留 start 值");
+        assert_eq!(
+            usage.input_tokens, 100,
+            "delta 无 input_tokens 时保留 start 值"
+        );
         assert_eq!(usage.output_tokens, 50);
         assert_eq!(usage.cache_read_tokens, 20);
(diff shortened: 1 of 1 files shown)


farion1231:
@codex review

Issue #5745, opened 2026-07-24
database is locked storm on Claude sessions (SESSION-SYNC long transaction) — Claude-side of #3189


yaluba88:
### Summary
On v3.18.0, the `session_usage` SESSION-SYNC routinely scans a large number of Claude `.jsonl` session files inside a single long transaction. Combined with concurrent realtime writes from `proxy_request_logs` (one row per proxied request), this produces a sustained `database is locked` storm that eventually forces the user to kill the app, and on restart `Failed to init database: database is locked` because WAL has un-checkpointed writes from the killed process.

This is the **Claude-side** equivalent of the problem PR #3189 fixes for Codex — but #3189 only aggregates Codex `token_count` events. The Claude path still produces the lock storm.

### Environment
- CC Switch v3.18.0 (latest, no newer release available)
- Windows 11
- DB: `journal_mode=wal` (persisted), `synchronous=FULL`, `busy_timeout=10000` (set by app at startup)
- ~1320 Claude session `.jsonl` files under `~/.claude/projects` (~1.4 GB)

### Reproduction
1. Heavy Claude usage (frequent provider hot-switching + many session files accumulated)
2. SESSION-SYNC runs periodically, scanning all session files in one transaction
3. `cc-switch.log` fills with `database is locked` for `SESSION-SYNC` inserts + `proxy_request_logs` writes
4. App becomes unresponsive → user exits → restart shows `Failed to init database: database is locked`

### Evidence (from `cc-switch.log`)
```
[SESSION-SYNC] 插入失败 (msg_...): 数据库错误: 插入会话日志失败: database is locked
[SESSION-SYNC] 文件解析失败: ...jsonl: 数据库错误: 更新同步状态失败: database is locked
[USG-001] 记录使用量失败: 数据库错误: 记录请求日志失败: database is locked
=== CC Switch v3.18.0 started ===
Failed to init database: 数据库错误: database is locked   (on restart)
```
- 200+ `database is locked` lines within ~10 minutes during a single storm
- `proxy_request_logs` grew to 35k+ rows (Claude side, glm/kimi models)
- After clearing `proxy_request_logs` detail (rollups already aggregated, zero loss) + WAL checkpoint, the storm stops and does not recur until the detail table re-inflates

### Root cause (analysis)
- `session_usage::sync` wraps the scan of all session files in one transaction → holds the write lock for the full scan duration (1320 files → well beyond `busy_timeout=10s`)
[...]

Issue #2879, opened 2026-05-18
Performance: session usage sync scans thousands of local logs every minute


majiayu000:
### Self Checks / 自检

- [x] I have searched existing issues, including closed ones.
- [x] I checked that this is not the same as #2858; that issue is about unknown model usage records, while this one is about background CPU/file scanning.

### CC Switch Version / 版本号

3.14.1 observed locally. The local checkout is behind upstream, but the installed app bundle reports `CFBundleShortVersionString = 3.14.1`.

### Operating System / 操作系统

macOS

### Related App / 涉及应用

CC Switch background app, Claude Code session usage sync, Codex session usage sync.

### Problem / 问题

CC Switch can periodically consume noticeable CPU because the background session usage sync scans thousands of local Claude/Codex session JSONL files every minute.

On my machine, Activity Monitor / `ps` showed `cc-switch` temporarily around ~25% CPU after running for several days. It later returned to idle, so this looks like a periodic spike rather than constant CPU usage.

### Evidence / 证据

The installed process:

```text
/Applications/CC Switch.app/Contents/MacOS/cc-switch
Bundle ID: com.ccswitch.desktop
Version: 3.14.1
Notarized Developer ID app
```

Runtime files opened by the process:

```text
/Users/<user>/.cc-switch/logs/cc-switch.log
/Users/<user>/.cc-switch/cc-switch.db
/tmp/com_ccswitch_desktop_si.sock
```

The local database is already fairly large:

```text
cc-switch.db: 81 MB
proxy_request_logs: 174391 rows
session_log_sync: 9869 rows
```

Recent log lines show minute-level syncs scanning thousands of files:

```text
[SESSION-SYNC] 同步完成: 导入 ... 条, 跳过 0 条, 扫描 3217 个文件
[CODEX-SYNC] 同步完成: 导入 ... 条, 跳过 0 条, 扫描 5474 个文件
```

The current source appears to hard-code a 60 second periodic sync in `src-tauri/src/lib.rs`:

```rust
const SESSION_SYNC_INTERVAL_SECS: u64 = 60;
```

The sync implementations enumerate all relevant JSONL files each run before skipping unchanged files:

- `sync_codex_usage()` collects `~/.codex/sessions/**/**/*.jsonl` and archived sessions.
- `sync_claude_session_logs()` collects `~/.claude/projects/*/*.jsonl`.
[...]

PR #6667, opened 2026-08-20
fix(usage): index Pi session dedup lookups

### LINKED ISSUE #6624: [Feature Request] Add a setting to disable session usage sync / 请求添加"禁用会话用量同步"开关
## Self Checks / 自检
- [x] I have read the FAQ section in README. / 我已阅读 README 中的常见问题。
- [x] I have searched for existing issues, including closed ones. / 我已搜索过已有的 Issue，包括已关闭的。

## Related App / 涉及应用
General / 通用

I never use the usage statistics / session management features, but CC Switch unconditionally imports **all** local session logs (Claude / Codex / Gemini / OpenCode / GrokBuild / pi) into its internal database — once at startup and then every 60 seconds — with no way to opt out.

This caused a severe freeze for me on v3.20.0 (macOS, Apple Silicon):

- On the first launch after upgrading, the newly added pi session import scanned 143 session files and inserted **~39k rows** into `proxy_request_logs`. The import ran for **~35 minutes at 100% CPU** (one tiny transaction per row, judging by journal churn).
- During the entire import, the main thread was blocked waiting on the global database `Mutex<Connection>` (confirmed via `sample`: `startURLSchemeTask → ... → __psynch_mutexwait`), so the window stayed frozen / beachballing, and afterwards the UI stuttered every few seconds whenever the sync task grabbed the lock.
- My `cc-switch.db` grew to **~230 MB** just from these imported usage rows.

Users who don't use the usage dashboard still pay the full cost of scanning, parsing and storing every session on their machine, forever.

## Proposed Solution / 建议方案

Add a setting (e.g. "Enable session usage statistics", default on to preserve current behavior) that, when turned off:

1. Skips the startup sync and the 60-second periodic sync entirely (`sync_all_unlocked` and the cost backfill);
2. Optionally hides or shows a "disabled" placeholder on the usage dashboard.

A minimal implementation only needs to gate the `spawn` block in `lib.rs` on a new `AppSettings` flag.

顺带一提：本次首启导入耗时 35 分钟，可能是因为逐行独立事务写入。批量事务或 WAL 模式应该能把同样的导入压缩到秒级。Related: #2879 #2883 #5745

## Additional Context / 补充信息

- Version: v3.20.0 (latest release)
- OS: macOS 15 (Apple Silicon)
- Sanitized log excerpt / 脱敏日志节选:

```
[22:50:33] [OPENCODE-SYNC] 同步完成: 导入 6474 条, 跳过 2170 条, 扫描 1065 个会话
[23:26:14] [PI-SYNC] 同步完成: 导入 39446 条, 跳过 771 条, 扫描 143 个文件
```

（两条日志间隔 35 分钟，期间 CPU 100%、界面全程假死。）

- Database table sizes after the freeze / 假死后的表体积（`dbstat`）:
  - `proxy_request_logs`: ~95 MB data + ~90 MB indexes, ~278k rows total

SaladDay:
## Summary

- Replace the per-record OR-based Pi dedup probe with short-circuit, identity-specific lookups.
- Ensure request IDs use the primary-key index and semantic fallbacks use the complete semantic index, including `has_entry_id` for legacy matching.
- Preserve the existing file-level transaction, cursor, retry, pricing, schema, scheduling, and UI behavior.

## Root cause

Without SQLite statistics, the previous `request_id OR semantic_id` predicate selected only the `data_source` prefix and scanned the Pi portion of `session_usage_dedup` for every imported row. Because the file transaction held the shared database connection, a large ledger could make startup appear frozen.

## Scope

This change is limited to the Pi session usage importer and one co-located query-plan regression test. It does not add a setting or change the schema, scheduling, retention, other importers, or UI.

## Validation

- `cargo test --manifest-path src-tauri/Cargo.toml --lib services::session_usage_pi::tests` — 16 passed
- `cargo test --manifest-path src-tauri/Cargo.toml --all-features` — all unit and integration tests passed (2,680 unit tests passed, 5 ignored)
- `cargo clippy --manifest-path src-tauri/Cargo.toml --lib --tests -- -D warnings`
- Real CC Switch v3.20.0 startup test with an isolated 129 MB database, 270,000 Pi dedup rows, no `sqlite_stat1`, 143 Pi JSONL files, and 39,446 pending records: all 39,446 imported in 16–19 seconds, with 0 skipped. A runtime sample kept the macOS main thread in the normal AppKit event loop while Pi imported in the worker.
- Stress subset: 20,000 existing and 5,000 incoming rows shared one semantic identity; the final count was 25,000, confirming distinct entry IDs remained distinct while the complete index path was used.

Fixes #6624.

Related context: #2879, #5745.

### DIFF
--- a/src-tauri/src/services/session_usage_pi.rs
+++ b/src-tauri/src/services/session_usage_pi.rs
@@ -34,6 +34,18 @@ const REVISION_COMPLETE_SHIFT: u32 = 60;
 const REVISION_SIZE_SHIFT: u32 = 32;
 const REVISION_MARKER: u64 = 0b101;
 const REVISION_SIZE_MASK: u64 = (1 << 28) - 1;
+const PI_REQUEST_DEDUP_SQL: &str = "SELECT EXISTS(
+         SELECT 1 FROM session_usage_dedup
+         WHERE data_source = ?1 AND request_id = ?2
+     )";
+const PI_SEMANTIC_DEDUP_SQL: &str = "SELECT EXISTS(
+         SELECT 1 FROM session_usage_dedup
+         WHERE data_source = ?1 AND semantic_id = ?2
+     )";
+const PI_LEGACY_SEMANTIC_DEDUP_SQL: &str = "SELECT EXISTS(
+         SELECT 1 FROM session_usage_dedup
+         WHERE data_source = ?1 AND semantic_id = ?2 AND has_entry_id = 0
+     )";
 
 #[derive(Debug, Clone, Copy, Default)]
 struct PiCosts {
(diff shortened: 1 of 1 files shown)


farion1231:
@codex review

SaladDay:
CI note: the `Backend Checks (Windows + WSL2 home)` failure happens before the contract test runs. `link.exe` invokes `mt.exe` with `TEMP` on the WSL UNC path and fails with `c1010070` / `LNK1327` while creating the temporary manifest.

The same job fails with the same linker error on the unmodified base commit `0b5da510` in [the main-branch CI run](https://github.com/farion1231/cc-switch/actions/runs/32118251143), and in [the latest WSL2 nightly run](https://github.com/farion1231/cc-switch/actions/runs/32406824011). The regular Windows, Ubuntu, and macOS backend jobs for this PR all passed, including formatting, Clippy, and tests.

This PR does not change the WSL2 contract, CI workflow, filesystem handling, or build configuration, so I have kept that pre-existing CI issue outside this PR's scope.

farion1231:
Thanks for the excellent work here — this fixes the most impactful regression in v3.20.0, and the quality of the submission made it a pleasure to review:

- The root-cause analysis is exactly right. I independently reproduced the query plans: the old OR predicate degrades to a `data_source`-prefix scan, while all three new probes are covering-index point lookups.
- The `EXPLAIN QUERY PLAN` regression test is the real gem — silent performance regressions like this are precisely what functional tests can't catch, and now the query shape is locked.
- The verifiable CI note (WSL2 failure pre-existing on main, with run links) saved real review time. Much appreciated.

Two optional follow-ups if you're interested — neither blocks anything:

1. The per-row probes re-prepare their SQL on every iteration (`query_row` doesn't use the statement cache). Switching to `prepare_cached` or preparing once outside the loop should shave the import time further.
2. The two dedup probes share an identical error message, so a failure can't tell you which probe tripped.

Merging now. Thanks again! 🎉

PR #6673, opened 2026-08-21
feat(settings): add provider-only management scope

hehuaiyu:
## Summary / 概述

- add a device-local `managementScope` setting for MCP, skills, and sessions, with all fields defaulting to `true` for backward compatibility
- add a "Providers only" master toggle plus independent controls in General Settings
- guard backend commands, startup migrations/imports, profile sync, provider projection, cloud skill sync, and periodic session usage work when the corresponding scope is disabled
- preserve unmanaged MCP configuration while switching Codex, Grok Build, and Gemini providers
- hide disabled MCP, Skills, and Sessions surfaces instead of leaving controls that can no longer succeed
- add localized copy for English, Simplified Chinese, Traditional Chinese, and Japanese, plus component and Rust regression coverage

This allows CC Switch to remain responsible for provider/account switching without also owning shared MCP, Skills, or session state maintained by other tools.

## Behavior / 行为

`managementScope` is stored as device-local settings:

```json
{
  "managementScope": {
    "mcp": true,
    "skills": true,
    "sessions": true
  }
}
```

When a scope is disabled:

- MCP import/write/projection is skipped, and existing live MCP sections are preserved during provider switches.
- Skills initialization, migration, mutation, provider/profile sync, and cloud sync are rejected or skipped.
- Session scanning, reading, deletion, migration, usage import/rebuild, and unified Codex session history injection are disabled.

Older settings files deserialize all three fields as enabled, so upgrading does not silently change existing behavior.

## Related Issue / 关联 Issue

N/A. This is a self-contained management-scope feature and does not claim to close an existing upstream issue.

## Screenshots / 截图

Not included. The UI change is limited to localized toggle rows in General Settings and is covered by component tests.

## Validation / 验证

- `pnpm typecheck`
- `pnpm format:check`
- full Vitest suite: 131 files, 987 tests passed
- `cargo check`
- `cargo clippy`
- settings command tests: 9 passed
- unmanaged MCP TOML preservation tests: 2 passed
- production renderer build
- `pnpm tauri build --no-bundle`
- `git diff --check`

## Checklist / 检查清单

- [x] `cargo clippy` passes (Rust code changed) / 通过 Clippy 检查
- [x] Updated i18n files for all currently shipped locales / 已更新所有当前语言文件

### DIFF
--- a/src-tauri/src/codex_config.rs
+++ b/src-tauri/src/codex_config.rs
@@ -768,6 +768,15 @@ pub fn delete_codex_provider_config(
     Ok(())
 }
 
+fn preserve_live_mcp_when_unmanaged(incoming: &str) -> Result<String, AppError> {
+    if crate::settings::mcp_management_enabled() {
+        return Ok(incoming.to_string());
+    }
+
+    let current = read_codex_config_text()?;
+    crate::mcp::replace_toml_mcp_sections(&current, incoming)
+}
+
 /// 原子写 Codex 的 `auth.json` 与 `config.toml`，在第二步失败时回滚第一步
 pub fn write_codex_live_atomic(
     auth: &Value,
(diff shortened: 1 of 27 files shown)


farion1231:
@codex review

PR #6714, opened 2026-08-22
fix(theme): dynamic system theme synchronization on Linux/KDE via Portal

eacheat53:
## Summary
Fixes an issue on Linux (e.g. KDE Plasma / GNOME with XDG Desktop Portal) where CC Switch in "System" theme mode does not dynamically follow desktop theme switches without restarting the application.

### Root Cause
On Wayland, GTK3 has no live settings channel for desktop theme changes, and WebKitGTK reads `gtk-application-prefer-dark-theme` from `GtkSettings`, so WebKit's `prefers-color-scheme` never flips dynamically. Reading the XDG Desktop Portal (`org.freedesktop.appearance/color-scheme`) is the proper live data source.
*(Note: As a pre-existing upstream behavior, `tao` maps `set_theme(None)` on Linux to `gtk-application-prefer-dark-theme = false`. With this PR the page root classes `.dark`/`.light` will synchronize dynamically, though `prefers-color-scheme` inside the webview may still disagree for native controls like scrollbars until a separate follow-up).*

### Changes
1. **Rust Backend (`src-tauri`)**:
   - Added `dark-light = "3.0.0"` dependency.
   - Implemented async `get_system_theme()` command reading the active XDG Desktop Portal theme via `tauri::async_runtime::spawn_blocking` bounded by a 1500ms `tokio::time::timeout` to prevent any D-Bus / portal hangs from blocking the GTK main loop or worker threads.
   - Added `map_system_theme_mode` helper and unit tests for theme mapping and async command execution.
2. **Frontend React (`src/components/theme-provider.tsx`)**:
   - Updated `ThemeProvider` to query `get_system_theme` on initial mount / mode change.
   - Added an `inFlight` guard to prevent stacking concurrent IPC requests.
   - Gated the 2s fallback polling interval on `isLinux()` so macOS/Windows users incur no polling overhead.
   - Wrapped `window.matchMedia` fallback in `Boolean(...)` for safe `classList.toggle` under jsdom / fallback environments.
   - Added unit test suite in `tests/components/ThemeProvider.test.tsx` covering theme detection, Linux 2s polling, non-Linux polling omission, and in-flight request deduplication.
3. **Dependency & Environment Notes**:
   - `.gitignore` tracks `.pnpm-store/`.
   - `Cargo.lock` updates `zbus` family (5.14 → 5.19) and related dependencies (such as `syn 3.0.3`, `winnow 1.0.4`) required by `dark-light 3.0.0`.

### Verification
- Frontend Vitest tests pass (136 test files, 1093 tests passed).
- Backend Rust unit tests pass (`cargo test --lib system_theme`).
- TypeScript typecheck (`pnpm run typecheck`) passes with 0 errors.
- Verified on Fedora KDE Plasma 6: dynamically toggles UI theme between dark and light modes on desktop theme changes.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -29,3 +29,4 @@ copilot-api
 .history
 CODEBUDDY.md
 mainWindow.js
+.pnpm-store/
--- a/src-tauri/Cargo.lock
+++ b/src-tauri/Cargo.lock
@@ -769,6 +769,7 @@ dependencies = [
  "brotli 7.0.0",
  "bytes",
  "chrono",
+ "dark-light",
  "dirs 5.0.1",
  "flate2",
  "futures",
@@ -1163,6 +1164,23 @@ checksum = "32a2785755761f3ddc1492979ce1e48d2c00d09311c39e4466429188f3dd6501"
 dependencies = [
  "quote",
  "syn 2.0.117",
+]
+
+[[package]]
+name = "dark-light"
+version = "3.0.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "0f66b71ac6b73812b0bf51cda8dd646a268fbb38623f45b2b7907593c323f745"
+dependencies = [
+ "futures-channel",
+ "futures-core",
+ "objc2 0.6.4",
+ "objc2-foundation 0.3.2",
+ "wasm-bindgen",
+ "web-sys",
+ "windows-sys 0.61.2",
+ "winreg 0.56.0",
+ "zbus",
 ]
 
 [[package]]
@@ -4687,7 +4705,7 @@ dependencies = [
  "errno",
  "libc",
  "linux-raw-sys 0.4.15",
- "windows-sys 0.52.0",
+ "windows-sys 0.59.0",
 ]
 
 [[package]]
@@ -5414,6 +5432,17 @@ name = "syn"
 version = "2.0.117"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "e665b8803e7b1d2a727f4023456bbbbe74da67099c585258af0ad9c5013b9b99"
+dependencies = [
+ "proc-macro2",
+ "quote",
+ "unicode-ident",
+]
+
+[[package]]
+name = "syn"
+version = "3.0.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "53e9bae58849f64dfa4f5d5ae372c8341f7305f82a3868709269343628b659a3"
 dependencies = [
  "proc-macro2",
  "quote",
@@ -7513,6 +7542,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "winnow"
+version = "1.0.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "23b97319f7b8343df12cc98938e5c3eb436064524c8d2b4e30a1d3a36eecdf81"
+dependencies = [
+ "memchr",
+]
+
+[[package]]
 name = "winreg"
 version = "0.10.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -7539,6 +7577,16 @@ checksum = "cb5a765337c50e9ec252c2069be9bf91c7df47afb103b642ba3a53bf8101be97"
 dependencies = [
  "cfg-if",
  "windows-sys 0.59.0",
+]
+
+[[package]]
+name = "winreg"
+version = "0.56.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "7d6f32a0ff4a9f6f01231eb2059cc85479330739333e0e58cadf03b6af2cca10"
+dependencies = [
+ "cfg-if",
+ "windows-sys 0.61.2",
 ]
 
 [[package]]
@@ -7770,9 +7818,9 @@ dependencies = [
 
 [[package]]
 name = "zbus"
-version = "5.14.0"
+version = "5.19.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "ca82f95dbd3943a40a53cfded6c2d0a2ca26192011846a1810c4256ef92c60bc"
+checksum = "5db4be7c075cb421e4b7ee645541604239bd243ba7c357511f4ff3a74b555907"
 dependencies = [
  "async-broadcast",
  "async-executor",
@@ -7797,7 +7845,7 @@ dependencies = [
  "uds_windows",
  "uuid",
  "windows-sys 0.61.2",
- "winnow 0.7.15",
+ "winnow 1.0.4",
  "zbus_macros",
  "zbus_names",
  "zvariant",
@@ -7805,14 +7853,14 @@ dependencies = [
 
 [[package]]
 name = "zbus_macros"
-version = "5.14.0"
+version = "5.19.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "897e79616e84aac4b2c46e9132a4f63b93105d54fe8c0e8f6bffc21fa8d49222"
+checksum = "2990635d09ade6df1868f72f8cac69a876a90981e8bd3c40b1be413f8dc88f40"
 dependencies = [
  "proc-macro-crate 3.5.0",
  "proc-macro2",
  "quote",
- "syn 2.0.117",
+ "syn 3.0.3",
  "zbus_names",
  "zvariant",
  "zvariant_utils",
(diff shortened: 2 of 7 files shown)


farion1231:
@codex review

eacheat53:
Hi @farion1231 , this PR has been waiting for review since August 22.
All checks are passing and GitHub reports the branch as mergeable.

Could you let me know whether:
1. the implementation needs changes;
2. the PR is waiting for a later release cycle; or
3. this behavior is not planned to be merged?

A short status update would help avoid repeatedly rebuilding this patch locally.

farion1231 (commented):
Thanks for this, and apologies for the long wait. The problem is real and the diagnosis is in the right neighbourhood: on Wayland, GTK3 has no live channel for desktop theme changes, so WebKitGTK's `prefers-color-scheme` (which it derives from `GtkSettings`) never flips, and reading the XDG Portal `color-scheme` key is the right data source. The new ThemeProvider tests are also the first time that component has been testable under jsdom, which is a nice side effect.

I built the merge tree against current `main` (2d54e261c): no conflicts, `cargo test` / clippy / `pnpm typecheck` / `pnpm test:unit` all green. I could not verify on real Linux hardware, so the P1 below is reasoned from source: the mechanism is confirmed, the "how bad can it get" part is not measured.

**[P1] `get_system_theme` blocks the GTK main thread on a D-Bus round-trip, every 2 s**

- The command is a plain `pub fn` ([misc.rs#L4687-L4699](https://github.com/farion1231/cc-switch/blob/e4c21dc1c70b90d67551bb7fc3df88c14057bc1a/src-tauri/src/commands/misc.rs#L4687-L4699)). Tauri runs non-async commands on the main thread ([docs](https://v2.tauri.app/develop/calling-rust/#async-commands)); on Linux that is the GTK loop, since wry dispatches IPC from the `script-message-received` signal.
- `dark_light::detect()` on Linux opens a fresh `zbus::blocking::Connection::session()` on every call and then does a blocking `ReadOne` with no call timeout ([dark-light 3.0.0 freedesktop.rs](https://docs.rs/crate/dark-light/3.0.0/source/src/platforms/freedesktop.rs)). zbus also spawns an executor thread per connection.
- The frontend calls it on mount, on every `matchMedia` change, and from a 2 s `setInterval` ([theme-provider.tsx#L76-L109](https://github.com/farion1231/cc-switch/blob/e4c21dc1c70b90d67551bb7fc3df88c14057bc1a/src/components/theme-provider.tsx#L76-L109)) with no in-flight guard, so if one call stalls the next ticks queue up behind it and each one blocks the main thread again once the queue drains.

When the portal is healthy this is a short blocking round-trip every 2 s and nobody notices. When the portal is installed but its backend is missing or stuck (minimal WMs, WSLg, containers), D-Bus service activation can block until dbus-daemon gives up on starting the service, and the whole window freezes for that long, repeatedly. Since `defaultTheme` is `"system"`, every Linux user is on this path out of the box, which turns "theme doesn't follow" into "window hangs". Note that `#[tauri::command(async)]` alone is not enough: for a non-`async` fn the macro just wraps the synchronous call in an async task, so the block moves onto a tokio worker instead of the main thread.

Suggested fix (same pattern `s3_sync.rs` and `config.rs` already use):

```rust
#[tauri::command]
pub async fn get_system_theme() -> Result<Option<&'static str>, String> {
    const DETECT_TIMEOUT: std::time::Duration = std::time::Duration::from_millis(1500);
    let detect = tauri::async_runtime::spawn_blocking(detect_system_theme_blocking);
    match tokio::time::timeout(DETECT_TIMEOUT, detect).await {
        Ok(Ok(mode)) => Ok(mode),
        Ok(Err(e)) => {
            log::debug!("System theme detect task failed: {e}");
            Ok(None)
        }
        Err(_) => {
            log::debug!("System theme detect timed out; falling back to matchMedia");
            Ok(None)
        }
    }
}

fn detect_system_theme_blocking() -> Option<&'static str> {
    match dark_light::detect() {
        Ok(dark_light::Mode::Dark) => Some("dark"),
        Ok(dark_light::Mode::Light) => Some("light"),
        Ok(dark_light::Mode::Unspecified) => None,
        Err(e) => {
            log::debug!("Failed to detect system theme via dark-light: {e}");
            None
        }
    }
}
```

and on the frontend an in-flight guard so a slow call cannot pile up requests:

```ts
let inFlight = false;
const syncSystemTheme = async () => {
  if (!isMounted || inFlight) return;
  inFlight = true;
  try {
    // existing body
  } finally {
    inFlight = false;
  }
};
```

**[P2] The 2 s poll runs on every platform, but only Linux needs it**

On macOS and Windows the `matchMedia` `change` event already fires live, so the interval there is pure overhead (cheap, but permanent, and it changes behaviour on platforms this PR isn't fixing). `src/lib/platform.ts` already exports `isLinux()`; gating the interval on it is a two-line change:

```ts
const intervalId = isLinux() ? setInterval(syncSystemTheme, 2000) : null;
// cleanup: if (intervalId !== null) clearInterval(intervalId);
```

Longer term, dark-light 3.0 ships `subscribe()`, which on Linux is driven by the portal's `SettingChanged` signal (no polling at all). A `#[cfg(target_os = "linux")]` subscription in the backend that emits an event to the frontend would remove the interval entirely. Happy to take that as a follow-up rather than in this PR.

**[P3] / nits**

- Root-cause wording: it's less "WebKit doesn't subscribe to the portal" and more "GTK3 on Wayland has no live settings channel, and WebKitGTK reads `gtk-application-prefer-dark-theme` from `GtkSettings`". Related and pre-existing (not this PR's fault): tao maps `set_theme(None)` on Linux to `gtk-application-prefer-dark-theme = false` ([tao 0.34.6 event_loop.rs](https://docs.rs/crate/tao/0.34.6/source/src/platform_impl/linux/event_loop.rs), `WindowRequest::SetTheme` arm), so after this PR the page classes will be right but `prefers-color-scheme` inside the webview may still disagree and native controls (scrollbars, `<select>`) can mismatch. Worth a sentence in the description; a proper fix is a follow-up.
- `get_system_theme_returns_valid_variant` ([misc.rs#L7066-L7073](https://github.com/farion1231/cc-switch/blob/e4c21dc1c70b90d67551bb7fc3df88c14057bc1a/src-tauri/src/commands/misc.rs#L7066-L7073)) can never fail: the function never returns `Err`, and `Ok(None)` passes too. Either assert the mapping on the blocking helper or drop it.
- `window.matchMedia && window.matchMedia(...).matches` evaluates to `undefined` when `matchMedia` is absent, and `classList.toggle(cls, undefined)` then *toggles* instead of setting. Unreachable in a real webview, but `Boolean(...)` around it makes the fallback safe under jsdom.
- Unrelated changes worth a line in the description: `.gitignore` gains `.pnpm-store/`, and `Cargo.lock` bumps the zbus family 5.14 → 5.19 (dark-light pins `zbus = "5.19.0"`, so this is forced, but it also moves `tauri-plugin-opener` / `tauri-plugin-single-instance` on Linux) plus new `syn 3.0.3`, `winnow 1.0.4`, `winreg 0.56`.

With the P1 and P2 fixed this is good to merge. I have the full patch (including three extra ThemeProvider tests for the Linux-only poll and the in-flight guard) verified green on the merge tree and can push it to your branch or paste it here, whichever you prefer. Will re-review quickly once it's updated.

eacheat53:
Thanks a lot for the thorough and constructive review!

I have addressed all the feedback and updated the PR branch:

1. **[P1] Async D-Bus with bounded timeout & In-flight guard**:
   - Converted `get_system_theme()` to an async Tauri command, offloading `detect_system_theme_blocking` via `tauri::async_runtime::spawn_blocking` with a 1500ms `tokio::time::timeout`. This guarantees that stuck or missing portal backends will not block the GTK main loop or worker threads.
   - Added an `inFlight` guard in `ThemeProvider`'s `syncSystemTheme` to prevent stacking concurrent requests.
2. **[P2] Platform-gated polling**:
   - Gated the 2s fallback polling on `isLinux()`, so Windows and macOS rely strictly on native `matchMedia` change events with zero polling overhead.
3. **[P3] / Nits**:
   - Refined the root-cause description and added notes on `tao`'s Linux `set_theme(None)` behavior and the lockfile dependencies bump.
   - Added `map_system_theme_mode` helper with explicit unit test coverage alongside the async command test.
   - Wrapped `window.matchMedia` fallback in `Boolean(...)` for safe `classList.toggle` execution.
   - Added test cases in `tests/components/ThemeProvider.test.tsx` covering 2s Linux polling, non-Linux polling omission, and in-flight request deduplication.
   - Rebased cleanly onto current main (`2d54e26`).

All unit tests (136 Vitest files, 1093 tests, cargo unit tests) and TypeScript typecheck are green.

PR #6743, opened 2026-08-23
fix(usage): reprice history rows when models.dev sync changes prices

V-dev-388:
### Background

Stale-price pollution timeline (observed in the wild on a real machine, 2.6k+ rows):

1. On startup, `run_session_sync(backfill=true)` calls `backfill_missing_usage_costs()` before the frontend has run anything network-side. Every zero-cost usage row gets priced with whatever the pricing table currently holds — built-in seeds that may lag models.dev.
2. Seconds later, the models.dev auto-sync (when enabled) fetches fresh prices and calls `updateModelPricingBatch`.
3. That batch updated the pricing table and re-ran the zero-cost backfill — but rows already priced in step 1 were never revisited, because the backfill only selects `total_cost_usd <= 0`.
4. Result: usage rows permanently priced with stale costs. The only recovery was manually zeroing the cost columns to make the rows backfill-eligible again.

### Fix

- `update_model_pricing_batch_inner` now records which entries actually changed a value (the pricing upsert already reports per-row change counts).
- After the transaction commits, each changed model goes through the new `reprice_usage_costs_for_model`:
  - zero the five cost columns of detail rows in that model's scope — reusing exactly the `model_pricing_candidates` + `log_pricing_scope_matches` scope of the per-model backfill, so rows stored under raw aliases are covered by the same normalization instead of a second one;
  - chunked `UPDATE ... WHERE request_id IN (...)` (500 per chunk) inside one transaction;
  - then rerun `backfill_missing_usage_costs_for_model` so the rows are recomputed at the new price with all existing semantics preserved (cost multiplier, cache-inclusive input deduction, `pricing_model` resolution).
- The existing full zero-cost backfill after the batch is kept.
- Rows whose pricing cannot be resolved stay at 0 and get picked up by a later backfill, same as any other zero-cost row.

### Scope / non-goals

- Only the batch path changes (its only caller is the models.dev sync). Manual single-price edits keep the current semantics — someone fixing one price by hand should not silently rewrite historical costs. Deletions/tombstones still never touch existing costs.
- Daily rollups freeze summed costs and the rollup table has no `cost_multiplier` column, so they cannot be safely recomputed here; they converge as new days accumulate. Left as a follow-up.

### Also included

- The `stepfun` provider joins the common-model families (`step-` prefix), so the default models.dev selection covers StepFun text models; audio/TTS stay excluded by the existing non-text filters.

### Testing

- 4 new Rust tests in `model_pricing.rs`: reprice recalculates history at the new price (exact amounts asserted on all five cost columns), models whose price did not change in the same batch stay untouched to the cent, manual single edit keeps existing costs, pricing deletion keeps existing costs. Reverse-verified: disabling the reprice call turns the new tests red.
- Extended the picker family test: `stepfun/step-3` enters `getCommonModelKeys`; `step-tts-mini` / `step-audio-2` are excluded at both the flatten level and the family-bucket level.
- `cargo test`: 2805 pass / 0 fail / 5 pre-existing ignored (baseline 2801 + 4 new); `cargo clippy --all-targets` 0 warnings; `cargo fmt --check` clean.
- `pnpm typecheck` and `format:check` green; full `test:unit` passes aside from the pre-existing load-related flaky failures in the PiProviderForm / integration suites, which have no import relationship with the modules touched here (verified by A/B reruns on a clean checkout).

### DIFF
--- a/src-tauri/src/services/model_pricing.rs
+++ b/src-tauri/src/services/model_pricing.rs
@@ -881,6 +881,87 @@ mod tests {
 
     #[test]
     #[serial]
+    fn batch_reprice_preserves_provider_reported_costs() {
+        // Pi and GrokBuild session imports may carry provider-reported costs
+        // (record.costs.reported()). Repricing must not zero and overwrite
+        // those authoritative values with locally-calculated pricing.
+        with_test_home(|db, _path| {
+            {
+                let conn = db.conn.lock().expect("lock test database");
+                // Standard proxy row: locally-calculated costs → repriced.
+                insert_usage_row(&conn, "proxy-target", "custom-model");
+                // Pi session row with provider-reported costs → must be left alone.
+                conn.execute(
+                    "INSERT INTO proxy_request_logs (
+                        request_id, provider_id, app_type, model, request_model,
+                        input_tokens, output_tokens, cache_read_tokens, cache_creation_tokens,
+                        input_cost_usd, output_cost_usd, cache_read_cost_usd,
+                        cache_creation_cost_usd, total_cost_usd, latency_ms,
+                        status_code, created_at, data_source
+                    ) VALUES (
+                        'pi-reported', 'pi-provider', 'pi', 'custom-model', 'custom-model',
+                        1000000, 1000000, 500000, 250000,
+                        '0.50', '2.00', '0.10', '0.75', '3.35', 100, 200, 1, 'pi_session'
+                    )",
+                    [],
+                )
+                .expect("insert pi row");
+                // GrokBuild session row with provider-reported total → must be left alone.
+                conn.execute(
+                    "INSERT INTO proxy_request_logs (
+                        request_id, provider_id, app_type, model, request_model,
+                        input_tokens, output_tokens, cache_read_tokens, cache_creation_tokens,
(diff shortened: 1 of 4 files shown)


farion1231:
@codex review

V-dev-388:
Thanks for the review — addressed in commit c7813eb.

**P1: Preserve provider-reported costs during repricing**
- The repricing SELECT now excludes rows whose `data_source` is `pi_session` or `grok_session` — the two data sources that may carry provider-reported costs (`insert_pi_record` prefers `record.costs.reported()`; GrokBuild may use `turn.reported_cost_usd()` for `total_cost`).
- `COALESCE(data_source, 'proxy')` handles legacy NULL `data_source` rows from pre-schema-v9, treating them as proxy (always locally priced, so still eligible for repricing).
- All other data sources (proxy, session_log, codex_session, gemini_session, opencode_session) use locally-calculated pricing and remain eligible for repricing when models.dev sync changes prices.

**New test**: `batch_reprice_preserves_provider_reported_costs` — inserts proxy, pi_session, and grok_session rows; verifies batch repricing updates the proxy row to v2 costs while pi_session and grok_session rows keep their original provider-reported costs. All 15 model_pricing tests and 32 usage_stats tests pass; clippy and fmt clean.

PR #6755, opened 2026-08-24
fix(codex): infer http type for url-only MCP servers on import

szupzj18:
## Summary

Fixes the url-only MCP corruption symptom from #6719.

Codex configs may omit `type` on URL-based MCP servers (e.g. `[mcp_servers.esp-cn-docs] url = "https://..."`). The import path defaulted to `stdio`, so the live projection wrote `command = ""` and the server broke.

## Changes

- **Import path** (`mcp/codex.rs`): when `type` is absent, infer `http` for entries that have a `url` but no `command`; otherwise keep `stdio`.
- **Write path** (`json_server_to_toml_table`): no longer emits `command = ""` for stdio specs that lack a command, or `url = ""` for http/sse specs that lack a URL. This prevents broken projections even for specs that were misclassified before this fix.

## Test plan

- `infer_type_for_url_only_server_is_http` — url-only entry infers `http`
- `infer_type_defaults_to_stdio_without_url` — command-only entry stays `stdio`
- `stdio_spec_without_command_does_not_emit_empty_command` — RED on old code (writes `command = ""`), GREEN after fix
- `http_spec_without_url_does_not_emit_empty_url` — RED on old code (writes `url = ""`), GREEN after fix
- `url_only_server_round_trip_preserves_http_type` — full import → project pipeline for a url-only server: type is `http`, URL preserved, no `command` emitted

All 45 `mcp::` tests pass; `cargo clippy --lib -D warnings` clean.

## Note

This addresses the most critical symptom of #6719 (url-only MCP written as `type="stdio" command=""`). The broader save-pipeline issues described in that issue (smol-toml comment loss, key reordering, common-config stripping) are separate concerns not covered here.

### DIFF
--- a/src-tauri/src/mcp/codex.rs
+++ b/src-tauri/src/mcp/codex.rs
@@ -625,6 +625,23 @@ pub(super) fn json_server_to_toml_table(spec: &Value) -> Result<toml_edit::Table
 
     let mut t = Table::new();
     let typ = spec.get("type").and_then(|v| v.as_str()).unwrap_or("stdio");
+    // Repair legacy malformed specs: url-only servers misclassified as stdio
+    // by the old import default should be projected as http.
+    let has_command = spec
+        .get("command")
+        .and_then(|v| v.as_str())
+        .map(|s| !s.is_empty())
+        .unwrap_or(false);
+    let has_url = spec
+        .get("url")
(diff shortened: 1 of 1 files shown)


farion1231:
@codex review

PR #6766, opened 2026-08-24
fix(proxy): avoid replaying terminal response text

### LINKED ISSUE #6756: OpenAI Responses 流式文本在终态事件中被重复输出
## 问题

Claude Code 通过 CC Switch 使用 `openai_responses` 格式时，部分响应会把同一段 assistant 文本输出两遍。切换同一 provider 到 `openai_chat` 后问题消失。

典型事件流程：

```text
response.output_text.delta
response.output_item.done
response.completed
```

第一份文本来自正常的流式 delta，第二份文本来自终态补偿。

## 影响

1. 输出两遍相同内容
2. sub2Api缓存不命中

## 原因

`streaming_responses.rs` 会记录已经转发给 Claude Code 的流式文本，并在终态事件中通过 `output_index` / `item_id` / `content_index` 判断是否只需补发缺失后缀。

部分兼容网关在 delta 与终态事件中使用的关联键不完整或不一致。此时 `StreamedTextState::missing_suffix()` 找不到已经发送的文本，并把“键关联失败”误判成“此前没有收到 delta”，返回完整的 `full_text`。随后 `missing_message_text_parts()` / `text_block_events()` 将全文再次发送。

该逻辑由 #5681 引入。#6509 的 review 也报告了相邻终态处理路径会重放已输出 reasoning，但 #6509 本身不修复普通文本的键关联问题。

## 临时规避

将 provider API Format 改为 OpenAI Chat Completions（`openai_chat`）。

Image

OiAnthony:
## Summary

- reconcile terminal response text with uniquely compatible streamed parts when identity keys drift
- emit only the missing suffix instead of replaying the full terminal snapshot
- apply the same reconciliation to buffered citation text and preserve terminal-only recovery

## Problem

Some Responses-compatible gateways omit or change `output_index` / `item_id` between `response.output_text.delta` and `response.completed`. The existing terminal compensation path then treats an association miss as if no delta was emitted and forwards the full text again.

This is the failure described in #6756, including the case where reasoning occupies output index 0 during streaming but is omitted from the terminal output array.

## Verification

- `cargo test --manifest-path src-tauri/Cargo.toml streamed_text`
- `cargo test --manifest-path src-tauri/Cargo.toml terminal_output`
- `cargo test --manifest-path src-tauri/Cargo.toml output_item_done_emits_text_when_deltas_are_missing`

Closes #6756

### DIFF
--- a/src-tauri/src/proxy/providers/streaming_responses.rs
+++ b/src-tauri/src/proxy/providers/streaming_responses.rs
@@ -286,6 +286,18 @@ struct StreamedTextPart {
     discarded: bool,
 }
 
+fn compatible_terminal_suffix<'a>(streamed: &str, terminal: &'a str) -> Option<&'a str> {
+    terminal
+        .strip_prefix(streamed)
+        .or_else(|| streamed.starts_with(terminal).then_some(""))
+}
+
+enum CompatiblePartMatch {
+    None,
+    Unique(usize),
+    Ambiguous,
+}
+
 #[derive(Default)]
 struct StreamedTextState {
     parts: Vec<StreamedTextPart>,
@@ -456,6 +468,26 @@ impl StreamedTextState {
                 && (self.unkeyed.starts_with(full_text) || full_text.starts_with(&self.unkeyed)))
     }
 
+    fn compatible_part_match(&self, full_text: &str) -> CompatiblePartMatch {
+        let mut candidates = self.parts.iter().enumerate().filter(|(_, part)| {
+            !part.discarded
+                && !part.text.is_empty()
+                && compatible_terminal_suffix(&part.text, full_text).is_some()
+        });
+        let Some((index, part)) = candidates.next() else {
+            return CompatiblePartMatch::None;
+        };
+        let suffix = compatible_terminal_suffix(&part.text, full_text)
+            .expect("compatible candidates must have a terminal suffix");
+        if candidates.any(|(_, candidate)| {
+            compatible_terminal_suffix(&candidate.text, full_text) != Some(suffix)
+        }) {
+            CompatiblePartMatch::Ambiguous
+        } else {
+            CompatiblePartMatch::Unique(index)
+        }
+    }
+
     fn record_delta(&mut self, data: &Value, delta: &str) {
         if delta.is_empty() {
             return;
@@ -517,7 +549,19 @@ impl StreamedTextState {
         if self.key_pair_conflicts(output_key, item_key.as_ref()) {
             return String::new();
         }
-        let keyed_index = self.existing_keyed_part_index(output_key, item_key.as_ref());
+        let mut keyed_index = self.existing_keyed_part_index(output_key, item_key.as_ref());
+        if keyed_index.is_none() && self.unkeyed.is_empty() {
+            match self.compatible_part_match(full_text) {
+                CompatiblePartMatch::Unique(index) => keyed_index = Some(index),
+                CompatiblePartMatch::Ambiguous => {
+                    log::warn!(
+                        "[Claude/Responses] Multiple streamed text parts match terminal text after identity drift; avoiding duplicate replay"
+                    );
+                    return String::new();
+                }
+                CompatiblePartMatch::None => {}
+            }
+        }
         let emitted = keyed_index.map(|index| self.parts[index].text.clone());
 
         let missing = if !self.unkeyed.is_empty() {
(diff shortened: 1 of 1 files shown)


farion1231 (commented):
Thanks for taking this on — the core fix is real and verified: I replayed the minimal repro from #6756 (terminal `message` with **no id** at a shifted output index) against `main` and got the text emitted twice; on this branch it is emitted exactly once. That form isn't covered by the PR's own integration test (which uses a *different* id), so the fix is actually broader than its test suite shows. The suffix-invariant design (uniqueness defined on the *suffix*, not the part) and the key re-binding that dedupes the `output_item.done` → `response.completed` double pass are both solid.

That said, I found one regression this change introduces, plus one gap it leaves open.

**[P1] A streamed part can be adopted by multiple terminal items within the same snapshot, so a genuinely new terminal-only message loses its prefix — or is swallowed entirely.**

[`compatible_part_match`](https://github.com/farion1231/cc-switch/blob/c3891ce7012538fcfa2c359cff8735f7141c7337/src-tauri/src/proxy/providers/streaming_responses.rs#L471-L489) doesn't track whether a candidate was already accounted for by an earlier item of the same terminal snapshot. A terminal-only item's keys *always* miss (it never streamed), so the reconciliation at [L552-L564](https://github.com/farion1231/cc-switch/blob/c3891ce7012538fcfa2c359cff8735f7141c7337/src-tauri/src/proxy/providers/streaming_responses.rs#L552-L564) participates unconditionally — no identity drift required. If its text happens to be prefix-compatible with an already-streamed part, it gets misread as that part's drift. Both of these fail on this branch and pass on `main`:

```rust
#[test]
fn terminal_only_item_keeps_its_prefix() {
    let mut state = StreamedTextState::default();
    state.record_delta(
        &json!({"item_id": "msg_first", "output_index": 0, "content_index": 0}),
        "Hello",
    );
    assert_eq!(state.missing_suffix("Hello", Some(0), Some("msg_first"), 0), "");
    // A brand-new terminal-only item; expected: full text. Actual: " again".
    assert_eq!(
        state.missing_suffix("Hello again", Some(1), Some("msg_second"), 0),
        "Hello again"
    );
}

#[test]
fn terminal_only_item_shorter_than_streamed_text_is_not_swallowed() {
    let mut state = StreamedTextState::default();
    state.record_delta(
        &json!({"item_id": "msg_first", "output_index": 0, "content_index": 0}),
        "Hello again",
    );
    assert_eq!(state.missing_suffix("Hello again", Some(0), Some("msg_first"), 0), "");
    // Terminal text is a *prefix* of the streamed text → empty suffix → the whole
    // message is silently dropped. Expected: full text. Actual: "".
    assert_eq!(state.missing_suffix("Hello", Some(1), Some("msg_second"), 0), "Hello");
}
```

Note `msg_first` resolves via an **exact key match** in both repros — the bug needs no drift at all, only the prefix coincidence. [`compatible_streamed_part_match`](https://github.com/farion1231/cc-switch/blob/c3891ce7012538fcfa2c359cff8735f7141c7337/src-tauri/src/proxy/providers/streaming_responses.rs#L970-L990) has the same issue on the buffered/citation side (the equivalent `render_message_part` sequence returns `Some(" again")`). Silent content loss is worse than the duplication this PR fixes, so I'd treat this as blocking.

Suggested fix — a per-terminal-event consumption set (e.g. a local `HashSet<usize>` of part indices) threaded through `missing_message_text_parts` into `missing_suffix` / `render_message_part`, with these constraints:

1. **Scope it to one upstream event**: the whole `output` loop of a single `response.completed` shares one set. A single `output_item.done` needs its own set too (one message can carry several content parts). Don't scope per item / per call — the second item of the same snapshot must see what the first consumed.
2. **Exact-key matches must also insert into the set** — `msg_first` above never touches `compatible_part_match`, so marking only content-reconciled parts won't fix these repros.
3. **Do not persist the marker across events.** `output_item.done` (exact keys) followed by `response.completed` (drifted keys) must still be able to re-match the *same* part — that cross-event re-match is exactly what makes this PR's main fix work, and a global "consumed" flag would break it.
4. Apply the same constraint to both `StreamedTextState` and `BufferedCitationTextState`.

One implementation caveat: `merge_part_indices` can invalidate stored part indices mid-loop, so check/update the set after any merge.

**[P2] Text streamed *before* web-search buffering turns on is still replayed in full on terminal identity drift.**

The buffered branch is guarded by [`has_part_matching_terminal`](https://github.com/farion1231/cc-switch/blob/c3891ce7012538fcfa2c359cff8735f7141c7337/src-tauri/src/proxy/providers/streaming_responses.rs#L455-L469) ([call site](https://github.com/farion1231/cc-switch/blob/c3891ce7012538fcfa2c359cff8735f7141c7337/src-tauri/src/proxy/providers/streaming_responses.rs#L1853-L1858)), which only checks exact keys / unkeyed text. `preserve_web_search_citations` only flips on at `output_item.added(web_search_call)`, so keyed text streamed before that point exists only in `StreamedTextState` — on terminal drift the guard misses, and the empty citation buffer's fallthrough replays the full text. I verified this behaves identically on `main`, so it's a pre-existing gap rather than something this PR introduces — but it *is* the same root cause #6756 describes, just in the web-search combination. Two acceptable resolutions: extend the guard to reuse the content-compatibility match, or narrow `Closes #6756` to `Refs` and note the remaining case.

⚠️ Ordering matters if you fix it here: don't just reuse the current `compatible_part_match` in the guard *before* the P1 consumption set exists. Today that path is accidentally safe (the buffer has no candidates for pre-buffer text, so terminal-only items fall through to the full-text default); wiring the guard to `StreamedTextState`'s candidates first would extend the P1 mis-adoption into it. Build the consumption-aware reconciliation first, then share it between both paths.

**nit (non-blocking, fine to leave):** (a) the `Ambiguous` arm can suppress a genuinely missing suffix when multiple prefix-overlapping parts coexist under drift — conservative and logged, reasonable trade-off; (b) buffered parts with only a `content_part.added` snapshot and no deltas (`received_delta == false`) are excluded from reconciliation and keep the old full-replay behavior — pre-existing, and the conservative filter choice seems deliberate.

With the P1 fixed (the two tests above would make good regression tests, plus a buffered-side twin) and the `Closes`/guard question settled either way, this is good to merge. Happy to re-review quickly.

farion1231:
@codex review

PR #6768, opened 2026-08-24
fix(i18n): add the missing pi.form.providerKeyDuplicate and common.collapse keys

ntdatt812:
## The defect

`translatePiProviderMutationError` returns a translation key for the duplicate-key error the Pi backend raises (`src/utils/errorUtils.ts:55-57`):

```ts
if (message.includes("Pi provider") && message.includes("already exists")) {
  return t("pi.form.providerKeyDuplicate");
}
```

That key exists in **none** of the four locale files. `fallbackLng: "en"` cannot rescue it, because `en` lacks it too, so `mutations.ts:144` renders the raw key into the toast:

```
Failed to add provider: pi.form.providerKeyDuplicate
```

Resolved against the real JSON, before this change:

```
en     pi.form.providerKeyDuplicate = MISSING    pi.provider.writeConflict = OK
zh     pi.form.providerKeyDuplicate = MISSING    pi.provider.writeConflict = OK
zh-TW  pi.form.providerKeyDuplicate = MISSING    pi.provider.writeConflict = OK
ja     pi.form.providerKeyDuplicate = MISSING    pi.provider.writeConflict = OK
```

`common.collapse` — the aria-label of the thinking-map toggle at `PiProviderForm.tsx:1747` — was missing in all four the same way, so it rides along.

The backend really does emit the matching string: `src-tauri/src/services/provider/pi.rs:51` and `:58`.

## Why it is a one-key omission, not a missing namespace

The branch one line above resolves everywhere, and the same duplicate-key message already exists for three other providers:

```
opencode.providerKeyDuplicate      OK in all four
openclaw.providerKeyDuplicate      OK in all four
hermes.form.providerKeyDuplicate   OK in all four
```

The wording follows Hermes', which is the closest match — *"This provider key already exists"* / 该供应商标识已存在 / 該供應商標識已存在 / このプロバイダーキーは既に存在します.

## Why the existing test did not catch it

`tests/utils/errorUtils.test.ts:22-32` mocks `t` as the identity function and asserts the return **is** the string `"pi.form.providerKeyDuplicate"`. It passes whether or not the key resolves, which is exactly how this survived.

So the guard belongs where keys are resolved against the real files: both paths are added to `requiredPaths` in `tests/config/managementListLocales.test.ts`, which already walks that list across all four locales.

## Verification

```
tests/config/managementListLocales.test.ts + tests/utils/errorUtils.test.ts   10 passed, 0 failed
tests/config/managementListLocales.test.ts (after adding the two paths)        7 passed
```

Reverting only the locale files, with the new guard in place, fails all four locale cases with `expected [ Array(2) ] to deeply equal []` — so the guard pins the keys rather than passing incidentally.

The diff is additive: 4 insertions / 2 deletions per locale file, the two deletions being the trailing-comma lines the insertions follow. No reformatting.

### DIFF
--- a/src/i18n/locales/en.json
+++ b/src/i18n/locales/en.json
@@ -42,7 +42,8 @@
     "deleting": "Deleting...",
     "auto": "Auto",
     "enabled": "Enabled",
-    "notSet": "Not Set"
+    "notSet": "Not Set",
+    "collapse": "Collapse"
   },
   "firstRunNotice": {
     "title": "Welcome to CC Switch",
@@ -978,7 +979,8 @@
       "effectiveApiRequired": "Model {{id}} needs an API, either directly or from the provider",
       "effectiveBaseUrlRequired": "Model {{id}} needs a base URL, either directly or from the provider",
       "duplicateModel": "Model ID {{id}} is duplicated",
-      "modelRequired": "Add at least one model"
+      "modelRequired": "Add at least one model",
+      "providerKeyDuplicate": "This provider key already exists"
     },
     "current": {
       "readFailed": "Could not read Pi's provider configuration",
--- a/src/i18n/locales/ja.json
+++ b/src/i18n/locales/ja.json
@@ -42,7 +42,8 @@
     "deleting": "削除中...",
     "auto": "自動",
     "enabled": "有効",
-    "notSet": "未設定"
+    "notSet": "未設定",
+    "collapse": "折りたたむ"
   },
   "firstRunNotice": {
     "title": "CC Switch へようこそ",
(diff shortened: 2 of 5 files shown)


farion1231:
@codex review

farion1231 (approved):
Thanks for the fix. Confirmed end to end: loading the four locale files from main into a real i18next instance with the app's config, the duplicate-key toast rendered as `添加供应商失败：pi.form.providerKeyDuplicate` (i18next returns the missing key verbatim as a non-empty string, so the `|| detail` fallback never kicks in); with this branch it becomes `…该供应商标识已存在` / `This provider key already exists`, and the toggle's aria-label resolves to `收起` / `Collapse`. I also scanned every literal `t("…")` key under `src/` against the four JSON files — these two were the only missing ones. The `requiredPaths` guard is real: reverting the locale files while keeping the guard turns 7 cases red.

One optional nit: in zh-TW the Pi form labels the field 供應商識別碼, while the new message follows Hermes' 供應商標識. Either is fine — just flagging in case you want them to match.

Merging. Thanks!

ntdatt812:
Thanks for the thorough verification — and you were right about the nit. I checked the other three locales and zh-TW was the only one where the form's label and its messages used different terms:

| locale | label | required | duplicate |
|---|---|---|---|
| en | Provider key | Provider key is required | This provider key already exists |
| ja | プロバイダーキー | プロバイダーキーを入力してください | このプロバイダーキーは既に存在します |
| zh | 供应商标识 | 供应商标识不能为空 | 该供应商标识已存在 |
| zh-TW | 供應商識別碼 | 供應商識別碼為必填 | 該供應商**標識**已存在 |

So it was zh-TW out of step rather than a choice worth keeping. #7187 changes that one string to 該供應商識別碼已存在. `hermes.form.providerKeyDuplicate` has the same sentence and is left alone — its own label is 供應商標識, so it is already consistent.

PR #6777, opened 2026-08-24
fix(provider-form): prevent config editors from overwriting each other

YUZHEthefool:
## Summary

Split the frontend portion of #6187 into its own PR. This fixes the two `ProviderForm` JSON editors overwriting each other when an external `value` synchronization was echoed back as user input.

## Changes

- Mark CodeMirror value-to-document updates with an `ExternalSync` annotation and ignore only updates made entirely of external-sync transactions.
- Read `onChange` through a ref so the editor keeps the latest form closure without rebuilding the editor on every render.
- Read `settingsConfig` at event time throughout the structured provider-form hooks, preserving edits made in the JSON editor and common-config editor.
- Use the current config for submit-time endpoint/API-key validation and provider identity detection.
- Add regression tests for the CodeMirror no-echo invariant and the settings snapshot invariant.

## Review follow-up

This is the frontend half of closed PR #6187 by BingZi-233. The original author is retained as a co-author in the commit (`Co-authored-by: BingZi-233 <lhby233@outlook.com>`).

## Verification

- `pnpm exec vitest run tests/components/JsonEditorExternalSync.test.tsx tests/components/ProviderFormSettingsConfigSnapshot.test.tsx`
- `pnpm run typecheck`

### DIFF
--- a/src/components/JsonEditor.tsx
+++ b/src/components/JsonEditor.tsx
@@ -3,13 +3,16 @@ import { EditorView, basicSetup } from "codemirror";
 import { json } from "@codemirror/lang-json";
 import { javascript } from "@codemirror/lang-javascript";
 import { oneDark } from "@codemirror/theme-one-dark";
-import { EditorSelection, EditorState } from "@codemirror/state";
+import { Annotation, EditorSelection, EditorState } from "@codemirror/state";
 import { placeholder } from "@codemirror/view";
 import { linter, Diagnostic } from "@codemirror/lint";
 import { useTranslation } from "react-i18next";
 import { Wand2 } from "lucide-react";
 import { toast } from "sonner";
 import { formatJSON } from "@/utils/formatters";
+
+/** 标记把 value 属性推进编辑器的事务，避免回吐为 onChange。 */
+const ExternalSync = Annotation.define<boolean>();
 
 interface JsonEditorProps {
   id?: string;
@@ -210,8 +213,13 @@ const JsonEditor: React.FC<JsonEditorProps> = ({
       ),
       EditorView.updateListener.of((update) => {
         if (!readOnly && update.docChanged) {
-          const newValue = update.state.doc.toString();
-          onChangeRef.current(newValue);
+          // CodeMirror 会把外部同步 dispatch 也标记为 docChanged。只有当
+          // 本次 ViewUpdate 中的事务全部来自外部同步时才跳过；使用 every
+          // 保证与用户输入合并的更新不会被误吞。
+          if (update.transactions.every((tr) => tr.annotation(ExternalSync))) {
+            return;
+          }
+          onChangeRef.current(update.state.doc.toString());
         }
       }),
     ];
@@ -316,6 +324,7 @@ const JsonEditor: React.FC<JsonEditorProps> = ({
           contextualRanges,
           viewRef.current.state.selection.mainIndex,
         ),
+        annotations: ExternalSync.of(true),
       });
       viewRef.current.dispatch(transaction);
     }
--- a/src/components/providers/forms/ProviderForm.tsx
+++ b/src/components/providers/forms/ProviderForm.tsx
@@ -61,6 +61,7 @@ import { HermesFormFields } from "./HermesFormFields";
 import type { UniversalProviderPreset } from "@/config/universalProviderPresets";
 import {
   applyTemplateValues,
+  getApiKeyFromConfig,
   hasApiKeyField,
 } from "@/utils/providerConfigUtils";
 import { mergeProviderMeta } from "@/utils/providerMetaUtils";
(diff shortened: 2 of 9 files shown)


YUZHEthefool:
@codex review

farion1231 (commented):
Thanks for splitting this out of #6187 — the split makes both halves much easier to review. The core fix is the right approach: the `ExternalSync` annotation is exactly what CodeMirror designed for this, the `every` check correctly avoids swallowing user keystrokes batched with a sync dispatch, and I verified the main regression test is real (reverting the listener skip turns it red). Merge tree against main is clean: typecheck + 998 frontend tests all green.

That said, the "read at event time" principle this PR establishes isn't applied consistently yet. Two P2s below are one-line fixes in the same family; I'd like those addressed before merging.

**[P2] Empty config falls back to the stale snapshot in `useApiKeyState`**

[`useApiKeyState.ts#L66`](https://github.com/farion1231/cc-switch/blob/365baeb1287252a9d256153d9adb0b72bb74163c/src/components/providers/forms/hooks/useApiKeyState.ts#L66) uses `getConfig?.() || initialConfig || "{}"`. When the user clears the JSON editor and then types an API key, `getConfig()` returns `""`, which is falsy — so the code rebuilds from the render-time `initialConfig` snapshot and resurrects every field the user just deleted. I reproduced this with a harness mirroring the ProviderForm wiring: a deleted `env` field comes back verbatim. Note the other hooks already get this right (`useBaseUrlState` does `JSON.parse(read() || "{}")`, `useModelState` falls back to `{ env: {} }`); this is the only one that treats `""` as "use the old snapshot". Fix:

```ts
getConfig ? getConfig() || "{}" : initialConfig || "{}",
```

**[P2] "Extract common config" still reads the render-time snapshot**

[`useCommonConfigSnippet.ts#L338-L345`](https://github.com/farion1231/cc-switch/blob/365baeb1287252a9d256153d9adb0b72bb74163c/src/components/providers/forms/hooks/useCommonConfigSnippet.ts#L338-L345) — `handleExtract` passes the closure `settingsConfig` to the backend. This PR threads `getSettingsConfig` through the other three read sites in this same file but misses this fourth one. The natural flow ("finish editing the JSON, then click Extract") hits it: the extracted snippet is computed from the pre-edit config and then persisted via `setCommonConfigSnippet`, so a stale snippet propagates to every provider sharing it. It doesn't write back into `settingsConfig`, so it won't clobber the user's edits — but the persisted snippet is wrong. Fix: use `getSettingsConfig?.() ?? settingsConfig` here too and add it to the deps.

**[P3] The onChange-swap test is a silent no-op, and needs two more fixes to actually run**

[`JsonEditorExternalSync.test.tsx#L86-L94`](https://github.com/farion1231/cc-switch/blob/365baeb1287252a9d256153d9adb0b72bb74163c/tests/components/JsonEditorExternalSync.test.tsx#L86-L94) reads `cmView` off `.cm-editor`, but CodeMirror attaches it to `.cm-content` — I probed this in the test environment: `view` is always `undefined`, so the test always hits the early `return` and passes without asserting anything. Two caveats for the fix, both verified locally:

1. Just fixing the selector makes the test fail for a different reason: `screen.getByText("swap").click()` is a native click outside `act()`, so React hasn't re-rendered when `view.dispatch` runs and the listener still sees the old closure. Use `fireEvent.click(...)` instead.
2. With all three changes — selector `.cm-content`, `fireEvent.click`, and asserting the view exists instead of silently returning — the test goes green, and injecting a frozen-closure bug into the listener turns it red. That's the version worth keeping.

**[nit] `readSubmitTimeCredentials` has a dead gemini branch**

The `GOOGLE_GEMINI_BASE_URL` branch in [`ProviderForm.tsx#L558-L577`](https://github.com/farion1231/cc-switch/blob/365baeb1287252a9d256153d9adb0b72bb74163c/src/components/providers/forms/ProviderForm.tsx#L558-L577) is never consumed — `submitBaseUrl` is only read inside the `appId === "claude"` paths, and gemini validation uses its own `geminiBaseUrl` state (which doesn't go through `settingsConfig`, so it isn't affected by the snapshot bug). Suggest dropping the branch or adding a comment, so it doesn't read as if gemini validation were covered here.

With the two P2s fixed this is good to merge — the P3/nit can ride along or follow up. Happy to re-review quickly.

YUZHEthefool:
已按 review 逐项修复并推送 `cdb3c3b2`：

- `useApiKeyState` 现在在传入 `getConfig` 时始终使用当前配置；即使当前值是空字符串，也不会回退到旧的 `initialConfig` 快照。
- `useCommonConfigSnippet.handleExtract` 改为调用 `getSettingsConfig()`，因此 Extract 使用 JSON 编辑器事件发生时的最新内容。
- `JsonEditorExternalSync` 的 onChange-swap 测试改为从 `.cm-content` 读取 CodeMirror view，使用 `fireEvent.click` 等待 React 更新，并在 view 缺失时明确失败，不再静默跳过。
- 移除了 `readSubmitTimeCredentials` 中不会被使用的 Gemini 分支。

验证通过：

- `pnpm exec vitest run tests/components/JsonEditorExternalSync.test.tsx tests/components/ProviderFormSettingsConfigSnapshot.test.tsx`
- `pnpm run typecheck`
- `pnpm exec prettier --check src/components/providers/forms/hooks/useApiKeyState.ts src/components/providers/forms/hooks/useCommonConfigSnippet.ts src/components/providers/forms/ProviderForm.tsx tests/components/JsonEditorExternalSync.test.tsx`

PR #6779, opened 2026-08-24
fix(provider): always project edits to live configuration

YUZHEthefool:
## Summary

Split the backend portion of closed PR #6187 into a separate PR. This fixes provider updates that reported success while leaving the live configuration file unchanged, including the universal-provider path.

## Changes

- Centralize provider-to-live synchronization and proxy-takeover ownership decisions for saves and explicit resyncs.
- Treat stale backup rows as recoverable: refresh the backup and write the live file when there is no corroborating takeover evidence.
- Do not let a persisted `proxy_config.enabled` flag alone suppress a live write after an interrupted teardown; use live placeholders, active per-app switch locks, and running-proxy evidence.
- Refresh Claude, Codex, and Grok Build proxy-safe projections during active takeover, including the Grok Build path.
- Re-project a generated universal-provider child when it is the effective current provider, continue syncing other apps if one projection fails, and report failed app names instead of returning a false success.
- Add regression tests for normal edits, stale backups, interrupted teardown flags, and universal-provider live projection.

## Review follow-up

This is the backend half of closed PR #6187 by BingZi-233. The Codex P1 review was addressed by requiring corroborating takeover evidence instead of trusting `proxy_config.enabled` unconditionally. The original author is retained as a co-author in the commit (`Co-authored-by: BingZi-233 <lhby233@outlook.com>`).

## Verification

- `cargo check --all-targets`
- Targeted provider regression tests pass.
- `cargo test --lib` passes except two pre-existing Windows symlink tests that require the SeCreateSymbolicLink privilege.

### DIFF
--- a/src-tauri/src/proxy/switch_lock.rs
+++ b/src-tauri/src/proxy/switch_lock.rs
@@ -39,4 +39,21 @@ impl SwitchLockManager {
         };
         lock.lock_owned().await
     }
+
+    /// 该应用当前是否正有切换 / 接管操作在进行中。
+    ///
+    /// 用途是判定"接管激活窗口"：`set_takeover_for_app` 全程持有本锁，而它在
+    /// 提交 `proxy_config.enabled`、给 live 打上占位符之前就已经写好了恢复备份。
+    /// 那段窗口里只看标志和占位符会误判成"未接管"，从而覆盖正在接管的 live。
+    ///
+    /// 关键在于这个信号是**按应用**的。全局的"代理进程在跑"标志做不到：A 应用
+    /// 开着接管时，B 应用一行残留的备份会被误算成 B 也在接管，于是 B 的保存又
+    /// 只写备份不写文件。
+    pub async fn is_locked_for_app(&self, app_type: &str) -> bool {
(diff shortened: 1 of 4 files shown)


YUZHEthefool:
@codex review

farion1231:
Merged — thank you for taking over the backend half of #6187 and keeping the original author credited, and for addressing the earlier review feedback properly instead of just trusting `proxy_config.enabled`.

The ownership predicate consolidation is the right shape: the three drifted copies of the `has_backup || placeholder` gate were exactly how the stale-row bug survived, and collapsing them into one function with corroborating evidence (placeholder / enabled+backup+running proxy / per-app activation lock) closes it at the root. I verified the stale-backup and universal-provider regression tests genuinely pin the fix — reverting the predicate or the projection turns them red.

One known residue we deliberately left for a follow-up: the provider *switch* path still treats a bare backup row as takeover ownership, so an orphaned row (enabled off, proxy stopped) can still divert a switch even though saves now heal. If you feel like a sequel, gating the switch path on the same predicate (evaluated before it acquires its own per-app lock) — or deleting the backup row when it's unambiguously stale — would finish the job.

PR #6780, opened 2026-08-24
fix(codex-oauth): isolate shared workspace accounts

### LINKED ISSUE #2245: OAuth 认证中心登录同一个team下多个不同chatgpt business账号会相互覆盖
### Self Checks / 自检

### CC Switch Version / 版本号

v3.14.0

### Operating System / 操作系统

macOS

### Related App / 涉及应用

Codex

### Steps to Reproduce / 重现步骤

OAuth 认证中心登录同一个team下多个不同chatgpt business账号会相互覆盖，原因可能是account_id是一致的，所以导致覆盖。但其实id_token，access_token，refresh_token都不一致。

### Expected Behavior / 期望行为

同一个team的chatgpt business账号可以同时登录，不会相互替换。

### Actual Behavior / 实际行为

_No response_

### Additional Context / 补充信息

_No response_

SaladDay:
## Summary

- assign each managed Codex OAuth login a stable local account ID so different users in one ChatGPT workspace can coexist
- keep local identity separate from the upstream workspace ID, and validate official passthrough against the resolved workspace plus the exact live bearer
- reauthenticate an existing account in place after verifying workspace and stable OIDC subject, preserving provider bindings and quarantining ambiguous legacy records
- linearize refresh, persistence, cancellation, expiry, and concurrent reauthentication so stale flows cannot publish credentials
- migrate legacy live-auth ownership only when stored identity proves the account before removal

## Scope

This is limited to Codex OAuth identity, persistence, request validation and selection, targeted reauthentication, native Codex auth compatibility paths, and the corresponding UI and regression tests. It does not change other auth providers, auth.json structure, OAuth revocation, imports, sessions, provider-binding removal semantics, or the pre-existing cross-file remove/logout transaction design.

## Tests

- cargo test --manifest-path src-tauri/Cargo.toml --lib
- cargo clippy --manifest-path src-tauri/Cargo.toml --lib --all-features -- -D warnings
- cargo fmt --manifest-path src-tauri/Cargo.toml -- --check
- pnpm test:unit
- pnpm typecheck
- pnpm format:check
- git diff --check

Fixes #2245

### DIFF
--- a/src-tauri/src/codex_config.rs
+++ b/src-tauri/src/codex_config.rs
@@ -488,38 +488,30 @@ fn extract_codex_auth_user_identity(auth: &Value) -> Option<String> {
     extract_codex_id_token_user_identity(id_token)
 }
 
-fn extract_codex_id_token_user_identity(id_token: &str) -> Option<String> {
-    let claims: Value = match id_token
-        .split('.')
-        .nth(1)
-        .and_then(|payload| URL_SAFE_NO_PAD.decode(payload).ok())
-        .and_then(|decoded| serde_json::from_slice(&decoded).ok())
-    {
-        Some(claims) => claims,
+pub(crate) fn extract_codex_id_token_user_identity(id_token: &str) -> Option<String> {
+    match extract_codex_id_token_subject(id_token) {
+        Some(subject) => Some(format!("sub:{subject}")),
         None => {
             #[cfg(test)]
             return Some("test-user".to_string());
             #[cfg(not(test))]
             return None;
         }
-    };
-    let chatgpt_user_id = claims
-        .get("https://api.openai.com/auth")
-        .and_then(Value::as_object)
-        .and_then(|claim| claim.get("chatgpt_user_id"))
-        .and_then(Value::as_str)
-        .or_else(|| claims.get("chatgpt_user_id").and_then(Value::as_str))
-        .map(str::trim)
-        .filter(|value| !value.is_empty());
-    if let Some(user_id) = chatgpt_user_id {
-        return Some(format!("chatgpt_user_id:{user_id}"));
     }
+}
+
+pub(crate) fn extract_codex_id_token_subject(id_token: &str) -> Option<String> {
+    let claims: Value = id_token
+        .split('.')
+        .nth(1)
+        .and_then(|payload| URL_SAFE_NO_PAD.decode(payload).ok())
+        .and_then(|decoded| serde_json::from_slice(&decoded).ok())?;
     claims
         .get("sub")
         .and_then(Value::as_str)
         .map(str::trim)
         .filter(|value| !value.is_empty())
-        .map(|subject| format!("sub:{subject}"))
+        .map(str::to_string)
 }
 
 #[cfg(test)]
@@ -617,6 +609,21 @@ fn migrate_legacy_codex_managed_oauth_live_auth_marker(
     record_codex_managed_oauth_live_auth(auth, managed_account_id)
 }
 
+/// Before removing a manager record, make any legacy live-auth ownership
+/// provable with the manager's persisted user identity. Failure is surfaced so
+/// callers keep the manager record and marker instead of orphaning auth.json.
+pub(crate) fn prepare_codex_live_auth_for_managed_account_removal(
+    managed_account_id: &str,
+    managed_id_token: Option<&str>,
+) -> Result<(), AppError> {
+    let auth_path = get_codex_auth_path();
+    if !auth_path.exists() {
+        return Ok(());
+    }
+    let auth: Value = read_json_file(&auth_path)?;
+    migrate_legacy_codex_managed_oauth_live_auth_marker(&auth, managed_account_id, managed_id_token)
+}
+
 pub fn codex_auth_matches_recorded_managed_oauth(
     auth: &Value,
     account_id: &str,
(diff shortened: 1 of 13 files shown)


farion1231:
@codex review

farion1231 (commented):
Thanks for taking on the shared-workspace isolation — decoupling a stable local account ID from the upstream workspace ID is the right architecture for the collision reports, and the forwarder injection path resolving local → workspace before stamping the outbound header ([forwarder.rs#L1741-L1755](https://github.com/farion1231/cc-switch/blob/0629cd7aa41d13f8900d5284254f5e93a338a6d0/src-tauri/src/proxy/forwarder.rs#L1741-L1755), [#L2216-L2220](https://github.com/farion1231/cc-switch/blob/0629cd7aa41d13f8900d5284254f5e93a338a6d0/src-tauri/src/proxy/forwarder.rs#L2216-L2220)) shows exactly the intended two-ID discipline. The problem is that three pre-existing consumers still assume the old invariant `local id == workspace id`, and they break once the two diverge:

**[P1] Proxied requests for a newly bound account are always rejected by the official-auth validation**

Every login now mints a fresh local UUID ([codex_oauth_auth.rs#L1496](https://github.com/farion1231/cc-switch/blob/0629cd7aa41d13f8900d5284254f5e93a338a6d0/src-tauri/src/proxy/providers/codex_oauth_auth.rs#L1496)), and the managed live `auth.json` is written with the upstream workspace ID ([live.rs#L814-L844](https://github.com/farion1231/cc-switch/blob/0629cd7aa41d13f8900d5284254f5e93a338a6d0/src-tauri/src/services/provider/live.rs#L814-L844)), so Codex CLI sends `chatgpt-account-id: <workspace-id>`. But `validate_codex_official_authorization` still compares that header against the provider's bound **local** ID ([forwarder.rs#L58-L77](https://github.com/farion1231/cc-switch/blob/0629cd7aa41d13f8900d5284254f5e93a338a6d0/src-tauri/src/proxy/forwarder.rs#L58-L77)), and it gates every proxied Codex-official request before the injection path runs ([forwarder.rs#L1184-L1186](https://github.com/farion1231/cc-switch/blob/0629cd7aa41d13f8900d5284254f5e93a338a6d0/src-tauri/src/proxy/forwarder.rs#L1184-L1186)). For any post-PR account the two values differ by design, so takeover users get a permanent "请重启 Codex 或新建会话后重试" that no restart can fix. Legacy migrated accounts (local id == workspace id) mask this, which is why the existing test's happy path still passes — [`managed_codex_official_rejects_a_different_session_account`](https://github.com/farion1231/cc-switch/blob/0629cd7aa41d13f8900d5284254f5e93a338a6d0/src-tauri/src/proxy/forwarder.rs#L4567-L4591) binds `account-b` and sends `account-b`, an equality that can no longer occur for new accounts.

Minimal fix: resolve the expected header value at the call site (async context is available there) and compare against that:

```rust
let expected = match provider.meta.as_ref()
    .and_then(|m| m.managed_account_id_for("codex_oauth"))
{
    Some(local_id) => Some(codex_auth.chatgpt_account_id_for_account(&local_id).await
        .map_err(|e| ProxyError::AuthError(format!("Codex OAuth 账号解析失败: {e}")))?),
    None => None,
};
validate_codex_official_authorization(headers, provider, expected.as_deref())?;
```

plus a regression test where the bound local ID and the workspace ID differ.

**[P1] "重新登录" cannot repair an account — it strands the old one and creates an unbound new one**

`add_account_internal` unconditionally generates a new UUID with no dedup by stable user identity + workspace ([codex_oauth_auth.rs#L1472-L1533](https://github.com/farion1231/cc-switch/blob/0629cd7aa41d13f8900d5284254f5e93a338a6d0/src-tauri/src/proxy/providers/codex_oauth_auth.rs#L1472-L1533)), and the re-auth button on a `reauth_required` account is a plain `addAccount` with no target ([CodexOAuthSection.tsx#L501-L512](https://github.com/farion1231/cc-switch/blob/0629cd7aa41d13f8900d5284254f5e93a338a6d0/src/components/providers/forms/CodexOAuthSection.tsx#L501-L512)). Previously the manager was keyed by workspace ID, so a same-workspace re-login overwrote the entry in place: reauth cleared, provider bindings kept working. Now the old account stays `reauth_required` forever, every provider bound to it keeps failing, and the fresh account starts unbound — then hits the P1 above once bound. The one flow the UI offers for recovery funnels users into a broken state.

Minimal fix: in the login-completion path, match an existing account by (stable user identity from the id_token, workspace ID) and apply the new tokens in place, preserving its `account_id` — or thread the target account through the re-auth flow. Either restores binding continuity.

**[P2] `remove_account` / `clear_auth` orphan live credentials for pre-upgrade (v1/v2 marker) accounts**

The deletion predicate hard-returns `false` for any marker version other than 3 ([codex_config.rs#L645-L661](https://github.com/farion1231/cc-switch/blob/0629cd7aa41d13f8900d5284254f5e93a338a6d0/src-tauri/src/codex_config.rs#L645-L661)), so `clear_codex_live_auth_for_managed_account_if_unchanged` leaves `auth.json` on disk while still deleting the marker ([codex_config.rs#L727-L760](https://github.com/farion1231/cc-switch/blob/0629cd7aa41d13f8900d5284254f5e93a338a6d0/src-tauri/src/codex_config.rs#L727-L760)), and the manager record — including the `id_token` that could prove ownership — is removed in the same operation. Net effect: real, refreshable ChatGPT credentials stay on the machine while every piece of bookkeeping that could ever claim them is destroyed, which contradicts the removal contract stated right above the call ([codex_oauth_auth.rs#L1258-L1268](https://github.com/farion1231/cc-switch/blob/0629cd7aa41d13f8900d5284254f5e93a338a6d0/src-tauri/src/proxy/providers/codex_oauth_auth.rs#L1258-L1268)) and OpenAI's own guidance that logout should clear stored credentials (https://developers.openai.com/codex/auth). The identity-based migration you already wrote for activation ([`migrate_legacy_codex_managed_oauth_live_auth_marker`](https://github.com/farion1231/cc-switch/blob/0629cd7aa41d13f8900d5284254f5e93a338a6d0/src-tauri/src/codex_config.rs#L591-L618)) is exactly the right tool — it's just never invoked on the removal paths.

Minimal fix: before the deletion predicate, attempt the same id_token-based v1/v2→v3 migration using the manager's stored `id_token`; only when ownership genuinely cannot be proven, keep the conservative skip but surface it to the user (log + error/warning) instead of silently deleting the marker and manager record.

---

The two P1s are what block merging — with those fixed (and ideally the P2, since the upgrade window is exactly when users will hit remove/re-login) this is in good shape, and the migration/ownership discipline elsewhere in the PR is genuinely careful work. Happy to re-review quickly once it's updated.

farion1231:
感谢两轮高质量的工作！🎉

第一轮把本地账号 ID 与上游 workspace ID 彻底解耦、迁移走 fail-closed 路线，方案在几个竞争实现里是最稳的；第二轮又把官方直通的会话校验、定向重认证的原位更新（保留绑定）、登录/切换/取消的全生命周期竞态一口气闭环，测试覆盖也非常扎实。同 workspace 多账号互相覆盖是近期重复报告最多的问题之一，这个修复价值很大。

升级后存量托管账号需要在认证中心重新登录一次（绑定会保留），发版说明会写清这一点。

两个小的 follow-up 不着急，有兴趣欢迎继续：
- 「添加账号」重复登录同一账号目前会新建重复条目，可以按 (sub, workspace) 合并到既有账号
- `extract_codex_id_token_user_identity` 里 `#[cfg(test)]` 的 test-user 分支，等测试都换成真 JWT 构造后可以移除

再次感谢！

PR #6793, opened 2026-08-25
fix(proxy): prevent duplicate Responses text in Anthropic SSE

kpkhxlgy0:
## Summary

- Normalize OpenAI Responses `resp_*` identifiers to Anthropic-compatible `msg_*` message IDs.
- Track the transcript actually emitted downstream and reconcile `response.completed` aggregates against it, so already streamed text is not replayed while genuine missing suffixes and citations are still delivered.
- Flush pending citation-buffered output before terminal reconciliation and add regression coverage for normal, indexed, annotated, refusal, interleaved, reversed, and buffered streams.
- Keep ambiguous cases conservative and add no request/response-content logging or diagnostic instrumentation.

## Root cause

A Responses stream may emit several message items through `response.output_item.done`, then collapse their combined text into a differently indexed aggregate message in `response.completed`. The previous item-keyed reconciliation treated that aggregate as new output and emitted text the user had already received.

This branch is intentionally based directly on v3.20.0. Upstream `main` is currently three commits ahead of that tag; a fresh three-way merge check reports no conflicts, and the PR changes one file only.

## Related Issue

N/A

## Screenshots

N/A — protocol conversion only; no UI changes.

## Verification

- `cargo clippy --all-targets --manifest-path src-tauri/Cargo.toml -- -D warnings` with three pre-existing v3.20.0 lint classes allowed locally (`for_kv_map`, `result_large_err`, and `useless_borrows_in_formatting`); the five affected locations are all in untouched files
- `cargo test --manifest-path src-tauri/Cargo.toml` (all suites passed)
- `tsc --noEmit`
- `prettier --check 'src/**/*.{js,jsx,ts,tsx,css,json}'`
- `vitest run` (131 files, 987 tests passed)
- Manual Claude Desktop soak testing, including multi-agent/retry workflows

## Checklist

- [x] Frontend formatting check passes
- [x] Rust formatting and all-target Clippy pass for changed code (baseline exceptions documented above)
- [x] Full Rust and frontend test suites pass
- [x] No user-visible text or i18n changes

### DIFF
--- a/src-tauri/src/proxy/providers/streaming_responses.rs
+++ b/src-tauri/src/proxy/providers/streaming_responses.rs
@@ -27,6 +27,12 @@ fn response_object_from_event(data: &Value) -> &Value {
     data.get("response").unwrap_or(data)
 }
 
+fn anthropic_message_id(response_id: &str) -> String {
+    response_id
+        .strip_prefix("resp_")
+        .map_or_else(|| response_id.to_string(), |suffix| format!("msg_{suffix}"))
+}
+
 fn anthropic_sse(event_name: &str, payload: &Value) -> Bytes {
     Bytes::from(format!(
         "event: {event_name}\ndata: {}\n\n",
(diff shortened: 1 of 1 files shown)


farion1231:
@codex review

PR #6808, opened 2026-08-25
fix(codex): proxy xAI deferred tool_search over native Responses

### LINKED ISSUE #6795: [Bug][Codex][xAI OAuth] Native Responses disables tool_search and sends 744 flattened tools, exceeding xAI's 350-tool limit
## Summary

Using Codex through CC Switch's managed **xAI (Grok) OAuth** provider fails when a large MCP/plugin tool catalog is enabled:

```text
Provider: xAI (Grok) OAuth
model: grok-4.6
upstream_status: HTTP 400
cause: Maximum tools limit reached. 744 tools have been provided but the maximum is 350.
url: http://127.0.0.1:15721/v1/responses
```

This is not an OAuth, network, or quota failure. The request reaches xAI, but the final request sent to `api.x.ai/v1/responses` contains 744 top-level tool schemas, exceeding xAI's current per-request limit of 350.

## Environment

- OS: macOS
- Reproduced with CC Switch runtime: `3.19.2`
- Client: Codex
- Endpoint: `/v1/responses`
- Provider: `xAI (Grok) OAuth`
- Model: `grok-4.6`
- Large MCP/plugin/dynamic-tool catalog enabled

I also inspected tag `v3.20.0`. The catalog, namespace-flattening, and xAI-sanitizer path described below still exists there. I have not claimed a runtime reproduction on v3.20.0.

## Expected behavior

Codex should be able to keep a large tool catalog registered without sending every full JSON schema to xAI on every turn.

For deferred tools, the intended flow should be:

1. Expose only built-in/hot tools and `tool_search` initially.
2. Let the model request tool discovery.
3. Translate that request back into a Codex `tool_search_call`.
4. Let the Codex client search its local deferred tool registry and return `tool_search_output`.
5. Expose only the matching tools to xAI on the next turn.
6. Let Codex execute the selected MCP tool locally.

## Actual behavior

The Native Responses catalog used by xAI OAuth disables search-tool support:

```json
"supports_search_tool": false
```

Source:

```text
src-tauri/src/resources/codex_native_responses_template.json
```

The native xAI Responses path then:

1. Flattens every Codex `namespace` child into a top-level `function` tool.
2. Does not distinguish deferred namespace children during flattening.
3. Removes native `type: "tool_search"`, because xAI's strict Responses parser does not accept that OpenAI-private type.
4. Sends the entire flattened function list to xAI.
5. Receives HTTP 400 because 744 exceeds the upstream limit of 350.

## Relevant request path

```text
Codex POST /v1/responses
  -> CC Switch local proxy
  -> provider_needs_responses_namespace_flatten(provider)
  -> flatten_request_namespaces()
  -> sanitize_xai_responses_request()
  -> POST https://api.x.ai/v1/responses
  -> HTTP 400: 744 tools > 350
```

### xAI OAuth enables namespace flattening

```text
src-tauri/src/proxy/providers/codex.rs
provider_needs_responses_namespace_flatten()
```

```rust
pub fn provider_needs_responses_namespace_flatten(provider: &Provider) -> bool {
    provider.is_xai_oauth()
}
```

### Namespace children are eagerly flattened

```text
src-tauri/src/proxy/providers/transform_codex_responses_namespace.rs
flatten_request_namespaces()
```

A Codex private tool such as:

```json
{
  "type": "namespace",
  "name": "mcp__example__",
  "tools": ["..."]
}
```

is expanded into top-level `function` tools. The current transform does not preserve deferred children as cold/searchable tools.

### Native tool_search is removed

```text
src-tauri/src/proxy/providers/transform_codex_responses_xai_sanitize.rs
sanitize_xai_responses_request()
filter_unsupported_tools()
```

`XAI_SUPPORTED_TOOL_TYPES` intentionally does not contain `tool_search`. Sending the original OpenAI-private shape directly to xAI would be incorrect, but deleting it without a compatible replacement removes Codex's deferred-discovery path.

### Transform order

```text
src-tauri/src/proxy/forwarder.rs
```

The native xAI path runs namespace flattening and then the xAI sanitizer before forwarding the body upstream.

## Root cause

This appears to be the combination of:

```text
xAI OAuth catalog: supports_search_tool=false
  +
all namespace children eagerly flattened
  +
native tool_search removed by the xAI sanitizer
  =
all 744 schemas sent in one xAI request
```

The upstream tool-count limit cannot be solved by increasing a local limit. The large catalog needs deferred discovery.

### 1. Enable search-tool support only for managed xAI OAuth

Do not globally change `codex_native_responses_template.json`, because other strict Native Responses providers may not support this compatibility path.

When generating the model catalog for `provider.is_xai_oauth()`, override:

```json
"supports_search_tool": true
```

This allows Codex to produce its deferred tool-discovery flow for this provider.

### 2. Request translation: tool_search -> function

Before the xAI sanitizer, translate:

```json
{"type":"tool_search"}
```

into an xAI-compatible function tool, for example:

```json
{
  "type": "function",
  "name": "tool_search",
  "description": "Search and load Codex tools, plugins, connectors, and MCP namespaces for the current task.",
  "parameters": {
    "type": "object",
    "properties": {
      "query": {"type": "string"},
      "limit": {"type": "integer"}
    },
    "required": ["query"]
  }
}
```

The existing Chat compatibility path already has reusable behavior:

```text
src-tauri/src/proxy/providers/transform_codex_chat.rs
add_tool_search_tool()
```

The original `type: "tool_search"` should not be added to xAI's native type allowlist. It must first become a normal function.

### 3. Do not eagerly flatten deferred namespace children

The first outbound request should include only:

- Built-in/hot tools
- Non-deferred namespace children
- Tools loaded by a prior `tool_search_output`
- Tools already used in the current conversation when required for replay
- The function-form `tool_search`

Cold/deferred namespace children should remain registered in Codex but should not enter xAI's `tools[]` until discovered.

### 4. Response translation: function_call -> tool_search_call

If xAI returns a normal function call named `tool_search`, translate it into Codex's client-executed item:

```json
{
  "type": "tool_search_call",
  "call_id": "...",
  "status": "completed",
  "execution": "client",
  "arguments": {"query": "..."}
}
```

Then Codex can search its own local registry and return `tool_search_output`.

The Chat compatibility path already contains similar response construction:

```text
transform_codex_chat.rs
response_tool_search_call_item()
```

Both streaming and non-streaming native Responses must be covered.

### 5. Restore tool_search even when the namespace map is empty

The current native response restore path may be skipped when the namespace restore map is empty.

On the first turn, there may be no loaded namespace tool yet, but `function_call(name="tool_search")` still needs to become `tool_search_call`. Tool-search restoration therefore cannot depend only on a non-empty namespace restore map.

### 6. Handle legacy eager overflow explicitly

If CC Switch receives an already-eager request with more than 350 function tools and no usable deferred/search context:

- Do not silently truncate tools while claiming all capabilities remain available.
- Prefer a clear local error asking the user to regenerate/reselect the xAI catalog and start a new session.
- A best-effort fallback should be explicitly logged, e.g. `xai_tool_overflow`.

## Why the local proxy should not execute MCP tools

The CC Switch proxy does not own Codex's MCP runtime, dynamic registry, permission state, or plugin lifecycle.

The proxy should perform protocol translation only:

- Namespace flatten/restore
- `tool_search` <-> `function` conversion
- Strict xAI request sanitization

Codex should continue to perform discovery, permission checks, and MCP execution locally.

## Suggested tests

1. xAI OAuth catalog sets `supports_search_tool: true` without changing other Native Responses providers.
2. A request containing 700+ deferred namespace children produces fewer than 350 upstream tools.
3. The xAI body contains no `type: "namespace"`.
4. The xAI body contains no native `type: "tool_search"`.
5. The xAI body contains the function-form `tool_search`.
6. Cold deferred tools do not appear before discovery.
7. Tools returned by `tool_search_output` appear on the following upstream request.
8. Non-streaming `function_call(name="tool_search")` becomes `tool_search_call` with `execution: "client"`.
9. Streaming tool-search calls receive the same restoration.
10. Tool-search restoration works when the namespace restore map is empty.
11. Existing flattened MCP function calls still restore to `{name, namespace}`.
12. Requests with <=350 non-deferred tools preserve current behavior.
13. Non-xAI providers are unaffected.
14. Repeated transforms are deterministic to preserve prompt-cache stability.
15. `/responses/compact` follows the same compatibility contract where applicable.

## Related discussions

- OpenCode deferred tools for provider tool-count limits: https://github.com/anomalyco/opencode/issues/20500
- Grok Build maximum tools limit: https://github.com/anomalyco/opencode/issues/28911
- CC Switch deferred tools when tool_search is omitted: https://github.com/farion1231/cc-switch/pull/5799
- Codex deferred namespace replay issue: https://github.com/openai/codex/issues/23839
- Codex tool_search on non-native endpoints: https://github.com/openai/codex/issues/20574
- CLIProxyAPI Codex tool_search payload compatibility: https://github.com/router-for-me/CLIProxyAPI/pull/3158

## Current workarounds

- Disable enough MCP/plugin tools to keep the final upstream tool count below 350.
- Split MCP servers across multiple Codex profiles.
- Put a lazy-loading Meta-MCP gateway in front of the existing MCP servers.

These workarounds are less desirable than preserving Codex's native deferred tool-discovery flow.

jarvisluk:
## Summary

Fix Codex requests routed through xAI OAuth exceeding xAI's 350-tool limit by translating Codex's private deferred-tool protocol at the proxy boundary.

The proxy now keeps deferred MCP and namespace tools hidden until Codex selects them, while preserving standard Responses behavior for clients such as Grok Build.

## Changes

- Advertise deferred tool search only for the proxied xAI native Responses profile; direct native Responses behavior is unchanged.
- Add a shared Codex tool bridge for Chat Completions and Anthropic adapters, covering function, custom, namespace, and deferred `tool_search` identities.
- Translate Codex's private `tool_search` to the internal upstream name `ccswitch_tool_search`, allowing an ordinary user function named `tool_search` to coexist safely.
- Reserve `ccswitch_tool_search` only while the private search shim is actually published; ordinary no-search requests may use that function name.
- Promote tools returned by `tool_search_output` and `additional_tools`; omit deferred catalog entries of every tool kind until they are loaded.
- Replace an outer-deferred namespace with the authoritative loaded subset so untouched catalog children never become visible during namespace flattening.
- Enforce xAI's 350-tool limit against the final deduplicated visible catalog after deferred results are promoted, including the search shim and loaded namespace children.
- Detect ambiguous function/custom/namespace name collisions and return an explicit transform error instead of silently choosing one declaration.
- Restore proxied function calls to their original Codex `tool_search_call` or namespaced function-call shape in both streaming and non-streaming responses.
- Normalize loaded function schemas, including root `object | null` unions, before forwarding them to xAI.
- Restrict Codex-private namespace, deferred-loading, and tool-search handling to the Codex client. Grok Build keeps standard Responses tool and reasoning semantics.
- Limit namespace restoration to actual Responses output containers so nested business JSON is not rewritten accidentally.

Search and tool execution remain local to Codex; this PR only translates the wire protocol.

## Testing

- `cargo test --manifest-path src-tauri/Cargo.toml --lib` — 2748 passed, 5 ignored.
- `cargo check --manifest-path src-tauri/Cargo.toml`.
- `cargo fmt --manifest-path src-tauri/Cargo.toml --all -- --check`.
- `git diff --check`.

## Notes

- Fixes #6795.
- No user-facing UI or i18n changes.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -210,10 +210,14 @@ jobs:
       - name: Run Windows-to-WSL2 filesystem contract
         shell: pwsh
         run: |
-          $env:TEMP = $env:CC_SWITCH_WSL_TEST_TEMP
-          $env:TMP = $env:CC_SWITCH_WSL_TEST_TEMP
           $testName = "config::tests::atomic_write_replaces_existing_wsl_unc_file"
-          $testList = cargo test --lib --manifest-path src-tauri/Cargo.toml -- --ignored --list
+          $testExe = Get-ChildItem -Path "src-tauri\target\debug\deps" -Filter "cc_switch_lib-*.exe" |
+            Sort-Object LastWriteTime -Descending |
+            Select-Object -First 1 -ExpandProperty FullName
+          if (-not $testExe) {
+            throw "Expected compiled test executable was not found"
+          }
+          $testList = & $testExe --ignored --list
           if ($LASTEXITCODE -ne 0) {
             exit $LASTEXITCODE
           }
@@ -221,7 +225,9 @@ jobs:
           if ($testList -notcontains $expected) {
             throw "Expected ignored test was not discovered: $testName"
           }
-          cargo test --lib --manifest-path src-tauri/Cargo.toml $testName -- --ignored --exact --nocapture
+          $env:TEMP = $env:CC_SWITCH_WSL_TEST_TEMP
+          $env:TMP = $env:CC_SWITCH_WSL_TEST_TEMP
+          & $testExe $testName --ignored --exact --nocapture
           if ($LASTEXITCODE -ne 0) {
             exit $LASTEXITCODE
           }
--- a/src-tauri/src/codex_config.rs
+++ b/src-tauri/src/codex_config.rs
@@ -311,6 +311,11 @@ impl CodexLiveStateSnapshot {
 pub enum CodexCatalogToolProfile {
     ProxyChat,
     NativeResponses,
+    /// Codex talks through cc-switch's takeover proxy to a strict native
+    /// Responses gateway whose private `tool_search` protocol is translated by
+    /// the proxy. This profile keeps the clean native tool surface while
+    /// advertising deferred search support to Codex.
+    ProxiedNativeResponses,
     /// Codex talks (through cc-switch's proxy) to a native Anthropic Messages
     /// gateway. Like `NativeResponses` it must suppress Codex's freeform custom
     /// tools — the Responses→Anthropic transform keeps only `function` tools.
@@ -1252,6 +1257,12 @@ fn codex_catalog_model_entry(
         }
     }
 
+    if profile == CodexCatalogToolProfile::ProxiedNativeResponses {
+        // This capability is valid only while cc-switch takeover owns the live
+        // endpoint and translates Codex's private tool-search protocol.
+        entry_obj.insert("supports_search_tool".to_string(), json!(true));
+    }
+
     // Per-model reasoning levels override the template's conservative
     // none/high default (e.g. a LiteLLM gateway serving a model that accepts
     // low/medium/high/xhigh/max). Applies to every profile.
@@ -1912,9 +1923,9 @@ fn codex_model_catalog_from_settings(
     // no cache dependency); proxy-chat providers keep cloning Codex's gpt-5.5
     // entry so the proxy can rewrite custom<->function tools as before.
     let template = match profile {
-        CodexCatalogToolProfile::NativeResponses | CodexCatalogToolProfile::Anthropic => {
-            load_codex_native_responses_template()
-        }
+        CodexCatalogToolProfile::NativeResponses
+        | CodexCatalogToolProfile::ProxiedNativeResponses
+        | CodexCatalogToolProfile::Anthropic => load_codex_native_responses_template(),
         CodexCatalogToolProfile::ProxyChat => load_codex_model_catalog_template()?,
     };
     Ok(Some(codex_model_catalog_from_specs(
@@ -2020,7 +2031,8 @@ pub fn prepare_codex_config_text_with_model_catalog(
             // The Responses→Anthropic transform silently drops the Codex web_search
             // hosted tool, so always disable it here rather than present a dead tool.
             CodexCatalogToolProfile::Anthropic => true,
-            CodexCatalogToolProfile::NativeResponses => {
+            CodexCatalogToolProfile::NativeResponses
+            | CodexCatalogToolProfile::ProxiedNativeResponses => {
                 codex_native_gateway_rejects_web_search(&config_text)
             }
             CodexCatalogToolProfile::ProxyChat => false,
@@ -4045,6 +4057,32 @@ base_url = "https://production.api/v1"
         assert!(template.get("supports_search_tool").is_none());
         assert!(template.get("supports_image_detail_original").is_none());
         assert!(template.get("web_search_tool_type").is_none());
+    }
+
+    #[test]
+    fn proxied_native_catalog_alone_enables_deferred_search() {
+        let settings = json!({
+            "modelCatalog": {"models": [{"model": "grok-4.6"}]}
+        });
+        let direct = codex_model_catalog_from_settings(
+            &settings,
+            "",
+            CodexCatalogToolProfile::NativeResponses,
+        )
+        .unwrap()
+        .unwrap();
+        let proxied = codex_model_catalog_from_settings(
+            &settings,
+            "",
+            CodexCatalogToolProfile::ProxiedNativeResponses,
+        )
+        .unwrap()
+        .unwrap();
+
+        assert_eq!(direct["models"][0]["supports_search_tool"], json!(false));
+        assert_eq!(proxied["models"][0]["supports_search_tool"], json!(true));
+        assert!(proxied["models"][0].get("tools").is_none());
+        assert_eq!(proxied["models"][0]["shell_type"], "shell_command");
     }
 
     #[test]
--- a/src-tauri/src/proxy/forwarder.rs
+++ b/src-tauri/src/proxy/forwarder.rs
@@ -1564,23 +1564,51 @@ impl RequestForwarder {
             mapped_body
         };
 
-        // Native Responses passthrough to a strict third-party gateway (xAI):
-        // flatten Codex's private `namespace`/plugin tool declarations into
-        // top-level function tools so the upstream's strict serde parser does
-        // not 422 on `unknown variant "namespace"`. The Chat/Anthropic paths
-        // above already unwrap namespaces, so this only fires on the native
-        // passthrough. The response handler restores the flat names using a map
-        // re-derived from the same request tools.
-        if matches!(app_type, AppType::Codex | AppType::GrokBuild)
+        let is_xai_native_responses = matches!(app_type, AppType::Codex | AppType::GrokBuild)
             && !codex_responses_to_chat
             && !codex_responses_to_anthropic
-            && super::providers::provider_needs_responses_namespace_flatten(provider)
+            && super::providers::provider_needs_responses_namespace_flatten(provider);
+
+        // Codex's private deferred-search items are not part of xAI's strict
+        // Responses schema. Proxy them as an ordinary function before namespace
+        // flattening; discovered namespaces are promoted here so only selected
+        // tools become model-visible.
+        if is_xai_native_responses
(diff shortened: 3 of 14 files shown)


farion1231:
@codex review

jarvisluk:
@codex review

Updated in d032bf28:
- Fixed the active P2 by ensuring namespaced functions named `tool_search` are not restored as the xAI shim.
- Avoid xAI's reserved `tool_search` function name by using an internal `ccswitch_tool_search` proxy name for native Responses.
- Addressed the low-risk suppressed findings around no-search oversized/deferred catalogs, schema alias selection, synthetic root unions, and MCP `server_label` deduping.
- Fixed the Windows + WSL2 CI step to run the already-built test executable under the UNC temp contract instead of invoking Cargo/linker from that temp path.

The remaining suppressed streaming lifecycle note is intentionally left for a separate repro/fix because it requires SSE buffering semantics changes.

jarvisluk:
@codex review

Updated in 0eb7d0d3:
- Fixed the P2 by counting the same promotion dedup keys when enforcing the no-search 350-tool catalog limit, so a top-level tool repeated in an `additional_tools` carrier is not double-counted (350 unique tools no longer look like 700).

jarvisluk:
Updated in `9ee6a375`:

- Centralized the flat-tool compatibility layer for Chat Completions and Anthropic in a shared Codex tool bridge, so function, custom, namespace, and deferred-tool identities use the same request/response mapping.
- Kept an ordinary user function named `tool_search` distinct from Codex's private search shim by reserving the internal upstream name `ccswitch_tool_search` for the shim.
- Restricted Codex-private `tool_search`, namespace, deferred-loading, and `additional_tools` semantics to the Codex client; Grok Build now retains standard Responses behavior.
- Omit deferred catalog entries until they are loaded through `tool_search_output` or `additional_tools`, and reject ambiguous cross-kind or flattened namespace name collisions instead of silently selecting one tool.
- Tightened namespace restoration to actual Responses output containers so nested business JSON that resembles a `function_call` is not rewritten.
- Applied the same identity restoration to streaming and non-streaming Chat/Anthropic responses.

Validated with the full Rust library suite (`2741 passed, 5 ignored`), `cargo check`, `cargo fmt --check`, and `git diff --check`.

@codex review

jarvisluk:
@codex review

jarvisluk:
@codex review

Updated in `25acc4be`:

- Fixed the P2 where deferred non-function tools such as top-level `mcp` declarations remained visible: deferred filtering now applies to every top-level tool kind, and retained tools have stale `defer_loading` / `deferLoading` markers removed.
- Fixed the P2 where promoted search results could exceed xAI's 350-tool limit: the final deduplicated visible catalog is now validated after `tool_search_output` / `additional_tools` promotion and deferred omission.
- The final count includes the internal search shim and loaded namespace children, while excluding catalog entries that remain deferred.
- Added regression coverage for deferred MCP omission, post-promotion overflow through a loaded namespace, and the exact 350-tool boundary.

Validated with the focused xAI tool-search suite (`27 passed`), the full Rust library suite (`2744 passed, 5 ignored`), `cargo check`, `cargo fmt --check`, and `git diff --check`.

jarvisluk:
@codex review

Updated in `d86b7024`:

- Fixed the P2 where loading a subset of an outer-deferred namespace retained untouched children from the stale catalog.
- An outer-deferred namespace is now replaced by the normalized `tool_search_output` / `additional_tools` copy, treating that loaded copy as the authoritative selected subset.
- Non-deferred namespaces still merge loaded children normally, preserving their already-visible catalog while refreshing selected schemas.
- Added a regression test proving that an initial two-tool deferred namespace exposes only the single selected child after promotion and flattening.

Validated with the focused xAI tool-search suite (`28 passed`), the full Rust library suite (`2745 passed, 5 ignored`), `cargo check`, `cargo fmt --check`, and `git diff --check`.

jarvisluk:
@codex review

Updated in `9e14f242`:

- Fixed the P2 by reserving `ccswitch_tool_search` only when the request actually publishes Codex's private `tool_search` shim.
- Applied the same conditional reservation to both xAI native Responses translation and the shared Chat Completions / Anthropic tool bridge.
- No-search requests can now use an ordinary top-level or dynamically loaded function named `ccswitch_tool_search`; responses retain the normal `function_call` identity because shim restoration is disabled.
- Requests that do publish the private shim still reject top-level and dynamically loaded collisions with its reserved upstream name.
- Added regression coverage for allowed no-shim functions and loaded tools, plus rejected collisions while the shim is present.

Validated with the shared bridge suite (`14 passed`), the focused xAI tool-search suite (`30 passed`), the full Rust library suite (`2748 passed, 5 ignored`), `cargo check`, `cargo fmt --check`, and `git diff --check`.

PR #6810, opened 2026-08-25
fix(prompts): keep unmanaged prompt files intact when a restore enables none

### LINKED ISSUE #6778: WebDAV 下载会清空本地 AGENTS.md（当云端 DB 无启用的 prompt 时）
## 问题描述

使用 WebDAV 云同步的「从云端下载」时，本地的 `~/.config/opencode/AGENTS.md` 会被整体清空成空文件。

## 复现步骤

1. 在某台机器 A 的 WebDAV 云端开启云同步（会有 `db.sql` + `skills.zip`），但云端 DB 中 OpenCode 没有「已启用」的 prompt 记录
2. 在另一台机器 B 使用 WebDAV 云同步「从云端下载」
3. 下载完成后，机器 B 的 `~/.config/opencode/AGENTS.md` 被清空为 `""`

## 根因

下载只还原两个快照：`db.sql` 和 `skills.zip`。下载完成后执行 `run_post_import_sync()` → `PromptService::sync_all_to_live()`，会把数据库中的 prompts 反向投影回各应用提示词文件（OpenCode 对应 `~/.config/opencode/AGENTS.md`）。

投影逻辑 `project_prompt_set_to_path`（`src-tauri/src/services/prompt.rs`）采用「以 DB 为权威、整体覆盖」的语义，当 DB 中没有任何启用的 prompt 时，直接用空串覆盖本地文件：

```rust
let enabled = prompts.iter().filter(|(_, p)| p.enabled).collect();
if let Some((_, prompt)) = enabled.first() {
    write_text_file(target_path, &prompt.content)?;
} else if target_path.exists() {
    write_text_file(target_path, "")?;   // ← 本地已有 AGENTS.md 被直接清空
}
```

`AGENTS.md` 并不在 `db.sql` / `skills.zip` 同步清单内，本身不会被上传，却会被下载后的「投影」步骤覆盖。现有测试也固化了此行为：`restored_prompt_projection_clears_a_stale_file_when_none_are_enabled` 断言文件会被清空。

## 期望行为

当本地目标文件已存在、而 DB 中又没有启用项时，不应将其清空（保守行为：保留本地文件，或至少通过配置可禁用该 prompt 投影/将 AGENTS.md 纳入同步保护清单）。

## 补充

同样的下载投影机制也会让 `~/.agents/skills` 目录被云端 `skills.zip` 整体覆盖，机制类似，建议一并评估。

SailingLoong:
Hi 👋 Another one from the open-issue pile — this time the WebDAV AGENTS.md wipe reported in #6778.

Fixes #6778

## Problem

WebDAV/S3 "download" restores `db.sql` and then runs `run_post_import_sync` → `PromptService::sync_all_to_live`, which projects the imported prompt rows back to the live prompt files. When the restored database has **no enabled prompt** for an app, `project_prompt_set_to_path` took the `else` branch and overwrote the local file (e.g. `~/.config/opencode/AGENTS.md`) with an empty string. Those files are never part of the upload payload, so a machine whose cloud DB happens to have no enabled prompt gets its local AGENTS.md wiped by a plain download — exactly the data destruction #6778 describes.

## Fix

Drop the clear-on-nothing-enabled branch from `project_prompt_set_to_path` and leave the target file untouched instead.

- This is safe because that branch was only ever reachable from restore paths: `project_prompt_set_to_path` ← `sync_to_live` ← `sync_all_to_live` ← `run_post_import_sync`, and `sync_to_live` has no other production caller. On a restore, "no enabled prompt in the snapshot" says nothing about the local file, so the conservative choice is to leave it alone.
- The interactive behavior is unchanged: disabling the last prompt from the UI still clears the file via `PromptService::upsert_prompt`.
- The pre-existing test pinned the wiping behavior (`restored_prompt_projection_clears_a_stale_file_when_none_are_enabled`); it is renamed and now asserts the local content survives.

## Tests

- `restored_prompt_projection_preserves_the_live_file_when_none_are_enabled` (replaces the old clearing test): seeded local file stays intact when the restored set has no enabled prompt. Revert-verified: it fails on `main`.
- Full `cargo test` green (14 test binaries, 0 failed), `cargo fmt --check` and `cargo clippy --all-targets -- -D warnings` clean. No frontend files touched; `pnpm typecheck` / `pnpm format:check` pass on the identical frontend tree.

## Checklist / 检查清单

Happy to adjust — e.g. if you'd rather surface a note that an unmanaged local file was left untouched, I can add that on top.

### DIFF
--- a/src-tauri/src/services/prompt.rs
+++ b/src-tauri/src/services/prompt.rs
@@ -30,11 +30,12 @@ fn project_prompt_set_to_path(
 
     if let Some((_, prompt)) = enabled.first() {
         write_text_file(target_path, &prompt.content)?;
-    } else if target_path.exists() {
-        // Match the existing "disable the last prompt" behavior without
-        // creating an otherwise unused application config directory.
-        write_text_file(target_path, "")?;
     }
+    // With nothing enabled, leave the target file untouched. This projection
+    // only runs after a database restore, and the live file is not part of
+    // the sync payload — clearing it here would wipe local content the
+    // restored snapshot never contained. Disabling the last prompt from the
+    // UI still clears the file via `PromptService::upsert_prompt`.
 
     if enabled.len() <= 1 {
         return Ok(None);
@@ -523,15 +524,21 @@ mod tests {
     }
 
     #[test]
-    fn restored_prompt_projection_clears_a_stale_file_when_none_are_enabled() {
+    fn restored_prompt_projection_preserves_the_live_file_when_none_are_enabled() {
         let temp = tempdir().expect("tempdir");
         let path = temp.path().join("AGENTS.md");
-        std::fs::write(&path, "stale").expect("seed stale prompt");
-        let prompts = IndexMap::new();
+        std::fs::write(&path, "local content").expect("seed live prompt file");
+        let mut prompts = IndexMap::new();
+        prompts.insert("off".to_string(), prompt("off", "managed", false));
 
-        let warning = project_prompt_set_to_path(&prompts, &path).expect("clear prompt");
+        let warning = project_prompt_set_to_path(&prompts, &path).expect("project prompt");
         assert!(warning.is_none());
-        assert_eq!(std::fs::read_to_string(path).expect("read prompt"), "");
+        // The live file is not part of the sync payload, so a restore with no
+        // enabled prompt must not wipe local content it never contained.
+        assert_eq!(
+            std::fs::read_to_string(path).expect("read prompt"),
+            "local content"
+        );
     }
 
     #[test]


farion1231:
@codex review

farion1231:
Merged — thanks @SailingLoong! This was a model fix for a data-loss bug: the call-chain reachability argument (the cleared branch was only reachable from restore paths) checked out exactly as described, and flipping the test that had been pinning the wiping behavior — with revert-verification — is precisely the right way to lock this in. The explanatory comment on why the file must be left untouched will keep this from being "fixed back" later. Much appreciated.

PR #6814, opened 2026-08-26
fix(proxy): preserve Responses reasoning summaries for Claude Code

### LINKED ISSUE #5362: GPT 5.6 Codex route emits redacted_thinking blocks in Claude Code VS Code
### Self Checks / 自检

### CC Switch Version / 版本号

3.17.0

### Operating System / 操作系统

macOS

### Related App / 涉及应用

Claude Code

## Environment

- CC Switch: 3.17.0
- OS: macOS
- Claude Code CLI: 2.1.209
- Claude Code VS Code extension: 2.1.209
- Provider: Codex / ChatGPT Codex
- Model mapping: `claude-opus-4-8` → `gpt-5.6-terra`
- Local proxy: `http://127.0.0.1:15721`

## Steps to reproduce

1. Start CC Switch and select the Codex provider.
2. Use Claude Code in VS Code with the Opus route mapped to `gpt-5.6-terra`.
3. Create a new Claude Code session.
4. Send a prompt that triggers a tool, Skill, Agent, or Bash call.
5. Before tool calls, VS Code repeatedly displays:

   `Unsupported content type: redacted_thinking`

## Actual behavior

The Claude Code session transcript contains assistant content blocks with:

```json
{ "type": "redacted_thinking", "..." : "..." }

<img width="360" height="50" alt="Image" src="https://github.com/user-attachments/assets/7ace7cd5-9944-43f0-897b-9280be7fc309" />

### Expected Behavior / 期望行为

When CC Switch routes Claude Code requests to Codex/GPT, internal reasoning data should not be exposed as an unsupported visible content block in VS Code.

The VS Code Claude Code timeline should display normal assistant and tool messages without:

`Unsupported content type: redacted_thinking`

If `redacted_thinking` is required for upstream reasoning or multi-turn tool context, CC Switch should retain it internally, or translate/filter it before returning the client-facing Claude Code-compatible stream.


### Actual Behavior / 实际行为

When CC Switch routes Claude Code Opus requests to `gpt-5.6-terra`, the VS Code Claude Code timeline repeatedly displays:

`Unsupported content type: redacted_thinking`

The issue appears before assistant tool calls such as Skill, Agent, and Bash.

The Claude Code session transcript contains assistant content blocks with:

```json
{ "type": "redacted_thinking", "..." : "..." }


### Additional Context / 补充信息

- CC Switch: 3.17.0
- OS: macOS
- Claude Code CLI: 2.1.209
- Claude Code VS Code extension: 2.1.209
- Provider: Codex / ChatGPT Codex
- Mapping: `claude-opus-4-8` → `gpt-5.6-terra`
- Local proxy: `http://127.0.0.1:15721`

CC Switch forwards requests to:

`https://chatgpt.com/backend-api/codex/responses`
`(model=gpt-5.6-terra)`

I updated CC Switch, Claude Code, and the VS Code extension; restarted VS Code; and reproduced the issue in a new session.

Screenshot attached. Full logs and transcripts are intentionally omitted because they may contain project content or credentials.

<img width="360" height="50" alt="Image" src="https://github.com/user-attachments/assets/2e702ff4-3bb6-419d-8814-1fd4d6d964e1" />

### PR DESCRIPTION
**Pull request by gadfly3173**
## Summary

- Request visible Responses reasoning summaries when Claude Code thinking is enabled, while honoring explicit omitted display.
- Translate Responses reasoning summary events into replayable Anthropic thinking blocks instead of exposing unsupported `redacted_thinking` content.
- Preserve encrypted reasoning signatures and reconcile summary events across streaming, aggregation, history replay, and non-streaming fallbacks.
- Keep client-aware Responses conversion on the Messages handler path.

Fixes #5362
Related to #5494
Related to #5643

## Testing

- `cargo test --manifest-path src-tauri/Cargo.toml proxy::providers::reasoning_bridge::tests --lib` (6 passed)
- `cargo test --manifest-path src-tauri/Cargo.toml proxy::providers::streaming_responses::tests --lib` (94 passed)
- `cargo test --manifest-path src-tauri/Cargo.toml proxy::providers::transform_responses::tests --lib` (121 passed)
- `cargo test --manifest-path src-tauri/Cargo.toml proxy::handlers::tests --lib` (53 passed)
- `git diff --check main...HEAD`

### DIFF
--- a/src-tauri/src/proxy/handlers.rs
+++ b/src-tauri/src/proxy/handlers.rs
@@ -20,6 +20,7 @@ use super::{
         codex_chat_common::extract_reasoning_field_text,
         codex_chat_history::record_responses_sse_stream,
         get_adapter, get_claude_api_format,
+        reasoning_bridge::reasoning_summary_text,
         streaming::create_anthropic_sse_stream,
         streaming_codex_anthropic::{
             create_responses_sse_stream_from_anthropic_with_context,
@@ -27,10 +28,7 @@ use super::{
         },
         streaming_codex_chat::create_responses_sse_stream_from_chat_with_context,
         streaming_gemini::create_anthropic_sse_stream_from_gemini,
-        streaming_responses::{
-            create_anthropic_sse_stream_from_responses,
-            create_anthropic_sse_stream_from_responses_with_web_search_options,
-        },
+        streaming_responses::create_anthropic_sse_stream_from_responses_with_web_search_options_for_client,
         transform, transform_codex_anthropic, transform_codex_chat,
         transform_codex_responses_namespace, transform_codex_responses_xai_sanitize,
         transform_gemini, transform_responses,
@@ -53,6 +51,7 @@ use bytes::Bytes;
 use futures::StreamExt;
 use http_body_util::BodyExt;
 use serde_json::{json, Value};
+use std::collections::{BTreeMap, HashMap};
 
 // ============================================================================
 // 健康检查和状态查询（简单端点）
@@ -175,6 +174,7 @@ async fn handle_messages_for_app(
     let method = parts.method.clone();
     let uri = parts.uri;
     let headers = parts.headers;
+    let preserve_redacted_thinking = !is_claude_vscode_client(&headers);
     let extensions = parts.extensions;
     let body_bytes = body
         .collect()
@@ -252,6 +252,7 @@ async fn handle_messages_for_app(
             &body,
             is_stream,
             &api_format,
+            preserve_redacted_thinking,
             connection_guard,
         )
         .await;
@@ -389,6 +390,7 @@ async fn handle_claude_transform(
     original_body: &Value,
     is_stream: bool,
     api_format: &str,
+    preserve_redacted_thinking: bool,
     connection_guard: Option<ActiveConnectionGuard>,
 ) -> Result<axum::response::Response, ProxyError> {
     let status = response.status();
@@ -429,13 +431,21 @@ async fn handle_claude_transform(
             dyn futures::Stream<Item = Result<Bytes, std::io::Error>> + Send + Unpin,
         > = if api_format == "openai_responses" {
             if hosted_web_search_name.is_none() && hosted_web_search_max_uses.is_none() {
-                Box::new(Box::pin(create_anthropic_sse_stream_from_responses(stream)))
+                Box::new(Box::pin(
+                    create_anthropic_sse_stream_from_responses_with_web_search_options_for_client(
+                        stream,
+                        None,
+                        None,
+                        preserve_redacted_thinking,
+                    ),
+                ))
             } else {
                 Box::new(Box::pin(
-                    create_anthropic_sse_stream_from_responses_with_web_search_options(
+                    create_anthropic_sse_stream_from_responses_with_web_search_options_for_client(
                         stream,
                         hosted_web_search_name.clone(),
                         hosted_web_search_max_uses,
+                        preserve_redacted_thinking,
                     ),
                 ))
             }
@@ -557,11 +567,12 @@ async fn handle_claude_transform(
                 )));
             }
             let response_headers = response.headers().clone();
-            let message = responses_sse_stream_to_anthropic_message(
+            let message = responses_sse_stream_to_anthropic_message_for_client(
                 response.bytes_stream(),
                 hosted_web_search_name.clone(),
                 hosted_web_search_max_uses,
                 body_timeout,
+                preserve_redacted_thinking,
             )
             .await?;
             (response_headers, Some(message), None)
@@ -634,10 +645,11 @@ async fn handle_claude_transform(
     let transform_result = match (direct_anthropic_response, upstream_response) {
         (Some(response), _) => Ok(response),
         (None, Some(response)) if api_format == "openai_responses" => {
-            transform_responses::responses_to_anthropic_with_web_search_options(
+            transform_responses::responses_to_anthropic_with_web_search_options_for_client(
                 response,
                 hosted_web_search_name.as_deref(),
                 hosted_web_search_max_uses,
+                preserve_redacted_thinking,
             )
         }
         (None, Some(response)) if api_format == "gemini_native" => {
@@ -2152,6 +2164,13 @@ pub async fn handle_gemini(
     .await
 }
 
+fn is_claude_vscode_client(headers: &axum::http::HeaderMap) -> bool {
+    headers
+        .get(axum::http::header::USER_AGENT)
+        .and_then(|value| value.to_str().ok())
+        .is_some_and(|value| value.to_ascii_lowercase().contains("claude-vscode"))
+}
+
 fn should_use_claude_transform_streaming(
     requested_streaming: bool,
     upstream_is_sse: bool,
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

**Comment by farion1231** [comment]
@codex review

**Comment by gadfly3173** [comment]
Resolved all issues.

**Comment by gadfly3173** [comment]
Fixed the Rust formatting and Clippy errors that caused CI to fail; all local backend checks now pass, with 2,735 tests passing and zero failures.

**Comment by gadfly3173** [comment]
@farion1231 请审查一下此PR

**Review by farion1231 — COMMENTED** [review]
Thanks for taking this on — #5362 has been hurting VS Code users for a while. The core move is right: an empty `thinking` block carrying the replay envelope in `signature` is exactly the shape Anthropic itself returns for Opus 4.7+ when `thinking.display` is omitted, so every Anthropic-protocol client already has to render it. I also verified the User-Agent detection against the real extension bundle (it spawns the CLI with `CLAUDE_CODE_ENTRYPOINT="claude-vscode"`, so the substring match holds). Merged onto current `main` (f3b18df12) the tree is clean: 2842 backend tests pass, clippy `-D warnings` and fmt are clean, no conflicts.

Reviewed at c08db3f2. Three things need to change before merge; the rest are optional.

**[P1] Requesting `reasoning.summary` by default breaks API-key routes on unverified OpenAI orgs**

https://github.com/farion1231/cc-switch/blob/c08db3f2412d818c03ea0db2e3c81d296b9ec1cf/src-tauri/src/proxy/providers/transform_responses.rs#L1835-L1848

When thinking is enabled and the client sends no `display`, this now adds `reasoning.summary: "auto"` to every Responses provider, not just Codex OAuth. OpenAI's platform API rejects that for organizations that have not completed verification:

```
400 invalid_request_error — param: "reasoning.summary", code: "unsupported_value"
"Your organization must be verified to generate reasoning summaries."
```

That is a hard failure on requests that work today (`effort` alone is accepted). Same failure has been reported across litellm (BerriAI/litellm#16032), OpenHands (#11552) and Azure OpenAI users. We also keep a project rule of not changing the request shape on paths that already work, because gateway-specific quirks belong in config generation, not on every request.

Codex OAuth is safe here — Codex CLI's own default is `model_reasoning_summary = "auto"`, so the ChatGPT backend already sees this field from the native client — and that is where all the #5362 reporters are. Gate the default on it and keep the explicit `display: "summarized"` opt-in for everyone:

```rust
let summary_requested = body
    .pointer("/thinking/display")
    .and_then(Value::as_str)
    .map(|display| display.eq_ignore_ascii_case("summarized"))
    .unwrap_or(effort.is_some() && is_codex_oauth);
```

`is_codex_oauth` is already a parameter of `anthropic_to_responses`. The three tests that assert `summary == "auto"` with `is_codex_oauth = false` (`..._adaptive_thinking_requests_visible_summary_without_display` and the two updated `resolve_reasoning_effort` tests) need to flip that argument or assert absence.

**[P2] The `preserve_redacted_thinking` split is not honored on the JSON fallback path, and I don't think the split buys anything**

https://github.com/farion1231/cc-switch/blob/c08db3f2412d818c03ea0db2e3c81d296b9ec1cf/src-tauri/src/proxy/providers/streaming_responses.rs#L235-L241

When the client asked for streaming but the upstream returned a JSON body, this branch rewrites `redacted_thinking` into `thinking` + `signature_delta` regardless of the flag — I fed the same JSON through with `preserve_redacted_thinking = true` and `false` and got byte-identical output. So the earlier review comment about honoring the flag is still open for this path.

Stepping back: no client depends on `redacted_thinking` here, while the `(external, jetbrains)` entrypoint (https://github.com/farion1231/cc-switch/blob/c08db3f2412d818c03ea0db2e3c81d296b9ec1cf/src-tauri/src/proxy/handlers.rs#L2152-L2157 only matches `claude-vscode`) stays on the shape that VS Code cannot render. My preference is to drop the split entirely: always emit the empty `thinking` + `signature` form, delete the `_for_client` wrappers and the `ClaudeTransformRequest` plumbing, and keep the `redacted_thinking.data` decode in `openai_reasoning_item_from_anthropic_block` for old session history. If you'd rather keep the split, at minimum this branch has to respect the flag. Either way, please do one or the other, not both.

**[P2] Terminal-snapshot replay duplicates the thinking block for gateways that omit reasoning ids**

https://github.com/farion1231/cc-switch/blob/c08db3f2412d818c03ea0db2e3c81d296b9ec1cf/src-tauri/src/proxy/providers/streaming_responses.rs#L4099-L4116
https://github.com/farion1231/cc-switch/blob/c08db3f2412d818c03ea0db2e3c81d296b9ec1cf/src-tauri/src/proxy/providers/streaming_responses.rs#L2201-L2231

Minimal repro through `create_anthropic_sse_stream_from_responses_with_web_search_options_for_client(…, false)`:

```
event: response.reasoning_summary_text.delta
data: {"type":"response.reasoning_summary_text.delta","delta":"Need a tool."}

event: response.completed
data: {"type":"response.completed","response":{"status":"completed","output":[
  {"type":"reasoning","summary":[{"type":"summary_text","text":"Need a tool."}],"encrypted_content":"opaque"},
  {"type":"message","role":"assistant","content":[{"type":"output_text","text":"Done."}]}]}}
```

The keyless delta lands on `legacy_reasoning_index` (block 0). The terminal item has no `id`, so `resolve_reasoning_index` looks up `reasoning:out:0`, which the legacy path never registered, allocates block 1, and replays the full text plus signature again. Result: two thinking blocks with the same text, and block 0 (no signature) stays open until the end of the message. `main` emits one block here (it drops the envelope, which this PR correctly fixes). When `item_id` is absent and neither key lookup hits, fall back to `legacy_reasoning_index` before allocating a new index, and only allocate when that block has already completed. The stream above makes a good regression test; the id-carrying variant (done + terminal both with `rs_…`) already behaves correctly.

**[P3] Optional, small**

- https://github.com/farion1231/cc-switch/blob/c08db3f2412d818c03ea0db2e3c81d296b9ec1cf/src-tauri/src/proxy/providers/streaming_responses.rs#L4708 — marking a reasoning item as substantive output changes the clean-EOF case: an encrypted-only reasoning item followed by EOF with no terminal event used to surface a stream error (so the client retries); it now ends as a silent `max_tokens` turn with no text. `main` already treats streamed reasoning deltas as substantive, so this only extends it to delta-less items; I'd not set it in the `output_item.done` reasoning arm.
- https://github.com/farion1231/cc-switch/blob/c08db3f2412d818c03ea0db2e3c81d296b9ec1cf/src-tauri/src/proxy/providers/streaming_responses.rs#L2172-L2200 — if a `…text.done` snapshot drifts from the streamed deltas (deltas `"Hello"` + `" world"`, done `"Hello  world"` with a double space), the whole snapshot is appended again, so the visible thinking reads `Hello worldHello  world`. When `previous_part` is non-empty and no prefix relation holds, returning an empty string is the safer choice. `merge_responses_sse_reasoning_summary_part` in handlers.rs (L2274) has the same append fallback.
- https://github.com/farion1231/cc-switch/blob/c08db3f2412d818c03ea0db2e3c81d296b9ec1cf/src-tauri/src/proxy/handlers.rs#L2339-L2347 — the non-streaming aggregation returns early once the final item carries any summary text, so if the SSE events delivered two summary parts and `output_item.done` carries only the first, the second is lost from both the visible text and the envelope; the streaming path keeps both in the same scenario. Mirror `preserve_streamed_reasoning_summary`: overwrite when the collected text starts with the item's text and is longer.

nit: `reasoning_summary_text` now joins parts with `\n`, falls back to `content[]`, and accepts string parts — all reasonable, but worth a line in the PR description since they change output for every Responses provider.

Fix the P1 and the two P2s and this is good to merge; the P3s can ride along or go to a follow-up. Ping me when it's pushed and I'll re-check quickly.

gadfly3173:
Pushed `de62b39c` addressing the review: `reasoning.summary` now defaults only on Codex OAuth (P1), the JSON fallback honors `preserve_redacted_thinking` (P2), keyless terminal snapshots reuse the streamed anonymous block instead of replaying a duplicate thinking block (P2), and the P3s ride along — sparse `output_item.done` no longer turns clean EOF into `stream_truncated`, drifting summary snapshots are no longer appended twice, the second summary part survives enrichment — with regression tests added for each repro; all backend tests, clippy and fmt pass.

PR #6820, opened 2026-08-26
fix(proxy): normalize Grok Build response annotations

### LINKED ISSUE #6819: fix(proxy): normalize missing annotations in Grok Build Responses streams
## Environment

- CC Switch: 3.20.0
- OS: Windows
- Related app: Grok Build

## Steps to reproduce

1. Configure a Grok Build provider whose upstream uses the native OpenAI Responses API.
2. Enable CC Switch proxy takeover. Grok Build is routed through `127.0.0.1:57890/grokbuild/v1`.
3. Start a streaming inference request.
4. Observe a `response.content_part.added` SSE event from the local proxy whose output-text part has no `annotations` field:

```json
{"type":"response.content_part.added","part":{"text":"","type":"output_text"}}
```

The same provider works when Grok Build bypasses the local proxy and uses its Chat Completions endpoint directly. The local Grok Build route only exposes `/responses`; `/chat/completions` returns 404, so changing the backend cannot work around this while proxy takeover is enabled.

## Expected behavior

The native Responses passthrough should remain consumable by strict Responses clients. An `output_text` content part without citations should contain `"annotations": []`.

## Actual behavior

Grok Build treats `annotations` as required and terminates response deserialization:

```text
Internal error: serialization error: missing field annotations
```

This also causes follow-up operations such as title generation to fail.

## Proposed fix

For Grok Build native Responses passthroughs, recursively add an empty `annotations` array only to `type: "output_text"` objects that omit the field. Preserve existing annotation values and leave delta events and unrelated payloads unchanged. Apply the normalization to streaming and non-streaming Responses payloads, including `/responses/compact`.

moyuan10086:
## Summary / 概述

Fix Grok Build crashes when a native Responses upstream omits the `annotations` field from an `output_text` content part.

- normalize Grok Build native Responses payloads by adding `annotations: []` only when an `output_text` object omits the field
- cover both SSE and JSON responses for `/responses` and `/responses/compact`
- preserve existing annotations, delta events, unrelated payload fields, error responses, usage accounting, and Codex passthrough behavior
- reuse the existing native Responses SSE wrapper so namespace restoration and compatibility normalization happen in one parsing pass

Observed failing proxy event:

```json
{"type":"response.content_part.added","part":{"text":"","type":"output_text"}}
```

Grok Build treats `annotations` as required and exits with:

```text
Internal error: serialization error: missing field annotations
```

The upstream works when accessed directly via Chat Completions, but CC Switch proxy takeover exposes only the Grok Build `/responses` route (`/chat/completions` returns 404), so the proxy path needs to normalize the Responses payload.

## Related Issue / 关联 Issue

Fixes #6819

## Screenshots / 截图

Not applicable; this is a proxy protocol fix.

## Verification / 验证

- `cargo test --manifest-path src-tauri/Cargo.toml transform_native_responses --lib` (2 passed)
- `cargo test --manifest-path src-tauri/Cargo.toml transform_codex_responses_namespace --lib` (10 passed)
- Full `cargo test --manifest-path src-tauri/Cargo.toml --lib`: 2663 passed, 6 ignored, 3 unrelated failures. Two require Windows symlink privileges; one pre-existing Hermes provider test fails on shared-state count (`left: 2`, `right: 1`).

Manual reproduction confirmed the malformed SSE event through `127.0.0.1:57890/grokbuild/v1`. Bypassing proxy takeover and using the same provider directly produced two successful Grok inference runs, confirming the upstream credentials and model are healthy.

## Checklist / 检查清单

- [ ] `pnpm typecheck` passes / 通过 TypeScript 类型检查 (not run; no TypeScript changes)
- [ ] `pnpm format:check` passes / 通过代码格式检查 (not run; no frontend changes)
- [x] Updated i18n files if user-facing text changed / 如修改了用户可见文本，已更新国际化文件 (no user-facing text changed)

### DIFF
--- a/src-tauri/src/proxy/handlers.rs
+++ b/src-tauri/src/proxy/handlers.rs
@@ -936,13 +936,17 @@ async fn handle_responses_for_app(
     // Native Responses passthrough to a strict gateway (xAI): restore flattened
     // function-call names *and* rewrite whole-float tool arguments. The integer
     // rewrite must run even when the request had no namespace tools.
-    if super::providers::provider_needs_responses_namespace_flatten(&ctx.provider) {
+    let xai_native = super::providers::provider_needs_responses_namespace_flatten(&ctx.provider);
+    let add_annotations = matches!(app_type, AppType::GrokBuild);
+    if should_use_native_responses_compat(xai_native, add_annotations) {
         return handle_codex_xai_native_responses_rewrite(
             response,
             &ctx,
             &state,
             connection_guard,
             namespace_restore_map,
+            xai_native,
+            add_annotations,
         )
         .await;
     }
@@ -1128,13 +1132,17 @@ async fn handle_responses_compact_for_app(
         .await;
     }
 
-    if super::providers::provider_needs_responses_namespace_flatten(&ctx.provider) {
+    let xai_native = super::providers::provider_needs_responses_namespace_flatten(&ctx.provider);
+    let add_annotations = matches!(app_type, AppType::GrokBuild);
+    if should_use_native_responses_compat(xai_native, add_annotations) {
         return handle_codex_xai_native_responses_rewrite(
             response,
             &ctx,
             &state,
             connection_guard,
             namespace_restore_map,
+            xai_native,
+            add_annotations,
         )
         .await;
     }
@@ -1162,6 +1170,8 @@ async fn handle_codex_xai_native_responses_rewrite(
         String,
         transform_codex_responses_namespace::NamespacedName,
     >,
+    xai_native: bool,
+    add_output_text_annotations: bool,
 ) -> Result<axum::response::Response, ProxyError> {
     let status = response.status();
 
@@ -1186,6 +1196,8 @@ async fn handle_codex_xai_native_responses_rewrite(
             transform_codex_responses_xai_sanitize::create_xai_native_responses_sse_stream(
                 response.bytes_stream(),
                 restore_map,
+                xai_native,
+                add_output_text_annotations,
             );
         let usage_collector =
             create_usage_collector(ctx, state, status.as_u16(), &CODEX_PARSER_CONFIG);
@@ -1199,7 +1211,7 @@ async fn handle_codex_xai_native_responses_rewrite(
 
         let body = axum::body::Body::from_stream(logged_stream);
         return builder.body(body).map_err(|e| {
-            log::error!("[{}] 构建 namespace 还原流式响应失败: {e}", ctx.tag);
+            log::error!("[{}] 构建 Responses 兼容流式响应失败: {e}", ctx.tag);
             ProxyError::Internal(format!("Failed to build streaming response: {e}"))
         });
     }
@@ -1220,15 +1232,22 @@ async fn handle_codex_xai_native_responses_rewrite(
     // Restore names when the body parses as JSON; otherwise pass the bytes
     // through untouched (a native Responses non-stream body is always JSON, so
     // this only guards against a malformed upstream).
-    let restored_bytes = match serde_json::from_slice::<Value>(&body_bytes) {
+    let compatible_bytes = match serde_json::from_slice::<Value>(&body_bytes) {
         Ok(mut value) => {
             transform_codex_responses_namespace::restore_response_namespaces(
                 &mut value,
                 &restore_map,
             );
-            transform_codex_responses_xai_sanitize::normalize_xai_function_call_integer_arguments(
-                &mut value,
-            );
+            if xai_native {
+                transform_codex_responses_xai_sanitize::normalize_xai_function_call_integer_arguments(
+                    &mut value,
+                );
+            }
+            if add_output_text_annotations {
+                super::providers::transform_native_responses::ensure_output_text_annotations(
+                    &mut value,
+                );
+            }
             if let Some(usage) =
                 TokenUsage::from_codex_response_auto(&value).filter(TokenUsage::has_billable_tokens)
             {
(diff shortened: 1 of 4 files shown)


farion1231:
@codex review

moyuan10086:
Rebased onto current `main` (`054673e0`) and resolved the proxy conflicts in `2b791f99`.

The updated implementation now composes with the upstream xAI native Responses sanitizer added since this PR opened: whole-float tool-argument normalization remains intact, while Grok Build additionally fills missing `output_text.annotations`. The duplicate namespace SSE helper from the old base was removed.

Updated local verification:
- `transform_native_responses`: 2 passed
- `transform_codex_responses_xai_sanitize`: 38 passed

The PR is mergeable again; CI has restarted for the rebased commit.

farion1231 (commented):
Thanks for the thorough report and fix — the layer choice is right (the client↔proxy protocol is pinned to Responses, so the proxy is the only place this can be fixed), the normalization itself is well-behaved (additive-only, exact `output_text` match, existing annotations preserved, untouched SSE blocks passed through byte-identical), and composing it into the existing xAI sanitize pass instead of adding a second parse was the right call. Full local suite passes on your branch (2765 passed / 0 failed).

There is one blocking issue, introduced by the rebase:

**[P1] The rebase dropped the outer gate, so the fix never runs for the scenario #6819 reports.**

- Pre-rebase (312623726b, the commit the bot reviewed), the dispatch was `restore_namespaces || normalize_grokbuild` — every Grok Build native-Responses passthrough entered the compat handler.
- On the current head, the outer gate is unchanged from main: the handler is only reached behind [`provider_needs_responses_namespace_flatten`](https://github.com/farion1231/cc-switch/blob/2b791f99ad39ba629fda8ecc70d094b642ff773d/src-tauri/src/proxy/handlers.rs#L939) (same at [L1132](https://github.com/farion1231/cc-switch/blob/2b791f99ad39ba629fda8ecc70d094b642ff773d/src-tauri/src/proxy/handlers.rs#L1132) for `/responses/compact`). That predicate ([codex.rs#L221-L244](https://github.com/farion1231/cc-switch/blob/2b791f99ad39ba629fda8ecc70d094b642ff773d/src-tauri/src/proxy/providers/codex.rs#L221-L244)) only accepts managed xAI OAuth cards or Codex-style TOML with `wire_api = "responses"` + an `api.x.ai` base URL.
- Grok Build cards store Grok-style TOML (`[model.X]` tables with `api_backend`), which has no `wire_api` key, and [`extract_codex_wire_api_from_toml`](https://github.com/farion1231/cc-switch/blob/2b791f99ad39ba629fda8ecc70d094b642ff773d/src-tauri/src/proxy/providers/codex.rs#L756-L774) only reads Codex conventions. I verified with a unit probe: an API-key Grok Build card shaped like the #6819 scenario (third-party relay, `api_backend = "responses"`, `meta.apiFormat = "openai_responses"`) fails all three dispatch predicates and falls through to the plain passthrough — `ensure_output_text_annotations` is never reached. The same holds even for a first-party `api.x.ai` API-key card. The only cards that currently get the fix are xAI OAuth ones.

Minimal fix (~15 lines): restore the widened outer gate, and gate the xAI-specific rewrites separately so third-party Grok Build upstreams don't newly receive the whole-float integer rewrite that was deliberately scoped to xAI:

```rust
let xai_native = super::providers::provider_needs_responses_namespace_flatten(&ctx.provider);
let add_annotations = matches!(app_type, AppType::GrokBuild);
if xai_native || add_annotations {
    return handle_codex_xai_native_responses_rewrite(
        response,
        &ctx,
        &state,
        connection_guard,
        namespace_restore_map,
        xai_native,      // gates namespace restore + whole-float rewrite
        add_annotations, // gates ensure_output_text_annotations
    )
    .await;
}
```

(with the `xai_native` flag threaded through the SSE wrapper the same way `add_output_text_annotations` already is, so `restore_sse_event_namespaces` / `normalize_xai_function_call_integer_arguments` only run when it's set — both in the streaming and the non-streaming branch.)

**[P2] Add a dispatch-level regression test.** The new tests all exercise the leaf function with the flag hardwired to `true`, which is exactly how this gate regression slipped through the rebase unnoticed. One test that builds a Grok-TOML API-key provider and asserts the annotations fill actually applies on the dispatch path would pin this down for good.

**[nit]** Might be worth noting in #6819 that switching the card's API format to `openai_chat` in CC Switch routes the request through the Responses→Chat conversion path and avoids the crash — a usable interim workaround for affected users until this ships (your report confirms the upstream's Chat endpoint works).

Fix the P1 (ideally with the P2 test) and this is mergeable — happy to re-review quickly once it's pushed.

moyuan10086:
Addressed the P1 gate regression and added the requested dispatch-level regression test.

- Grok Build now enters the native Responses compat handler even when the provider is a third-party/API-key card that does not satisfy the xAI `wire_api` predicate.
- The `xai_native` flag independently gates namespace restoration and whole-float argument normalization, so those xAI-only rewrites are not applied to third-party Grok Build providers.
- `add_annotations` independently gates the additive `output_text.annotations` normalization.
- Added a dispatch predicate regression test covering Grok Build-only, xAI-only, and plain passthrough cases.

Verification: `cargo fmt --check`, `cargo clippy --lib -- -D warnings`, dispatch test, and 38 xAI sanitizer tests all pass. The branch is pushed at the updated head.

moyuan10086:
@farion1231 Friendly follow-up: the requested P1 gate fix and P2 dispatch-level regression test are included in the current head (dde8335), and all backend CI checks are green across Ubuntu, Windows, macOS, and Windows + WSL2. When you have a chance, could you please take another look? Thanks!

PR #6828, opened 2026-08-26
fix(claude): stop settings.local.json env from shadowing provider switch

googxi1310328414-afk:
## Summary

Claude Code merges `settings.local.json` over `settings.json`. If you launch Claude from the home directory, `~/.claude` is both the user config dir and the project config dir. Leftover `ANTHROPIC_*` keys in `settings.local.json` (a dead `http://127.0.0.1:3800` local proxy in the report that triggered this) keep winning after CC Switch writes live `settings.json`.

The user-visible result is: CC Switch switch/auto-config looks successful, Claude then warns that both `ANTHROPIC_AUTH_TOKEN` and `ANTHROPIC_API_KEY` are set, and requests fail with `Connection refused`.

## Changes

- When writing Claude live settings, strip `ANTHROPIC_*` from `settings.local.json`
- Keep unrelated local fields such as `permissions`
- Report those keys in env-conflict detection
- Unit tests for scrub, empty-env drop, live write, and conflict listing

## Test plan

- [x] `cargo test --lib settings_local`
- [x] `cargo test --lib write_claude_settings_file_scrubs`
- [x] Reproduced locally: home-dir `claude -p` failed against `:3800`, then succeeded against the live provider after clearing local env

### DIFF
--- a/src-tauri/src/config.rs
+++ b/src-tauri/src/config.rs
@@ -199,6 +199,77 @@ pub fn get_claude_settings_path() -> PathBuf {
     settings
 }
 
+/// Claude Code project-local settings that overlay `settings.json`.
+///
+/// Launching Claude from the home directory treats `~/.claude` as the project
+/// config dir, so this file and user `settings.json` are the same folder.
+pub fn get_claude_settings_local_path() -> PathBuf {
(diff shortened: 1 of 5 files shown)


farion1231:
@codex review

PR #6831, opened 2026-08-26
fix(codex-oauth): align identity tests with JWT parsing

SaladDay:
## Summary

- remove the test-only fallback from Codex id-token identity extraction
- require a valid compact JWT shape and algorithm header before trusting the subject claim
- use structurally valid JWT fixtures throughout managed-account identity flows

## Scope

This is the identity-fixture follow-up suggested after #6780. It does not add signature, issuer, or audience verification; change OAuth routing or account state; or implement the separate (sub, workspace) account-deduplication follow-up.

## Testing

- cargo test --lib
- cargo clippy --lib --tests -- -D warnings
- cargo fmt --all -- --check
- two independent blind review passes

### DIFF
--- a/src-tauri/src/codex_config.rs
+++ b/src-tauri/src/codex_config.rs
@@ -489,22 +489,31 @@ fn extract_codex_auth_user_identity(auth: &Value) -> Option<String> {
 }
 
 pub(crate) fn extract_codex_id_token_user_identity(id_token: &str) -> Option<String> {
-    match extract_codex_id_token_subject(id_token) {
-        Some(subject) => Some(format!("sub:{subject}")),
-        None => {
-            #[cfg(test)]
-            return Some("test-user".to_string());
-            #[cfg(not(test))]
-            return None;
-        }
-    }
+    extract_codex_id_token_subject(id_token).map(|subject| format!("sub:{subject}"))
 }
 
 pub(crate) fn extract_codex_id_token_subject(id_token: &str) -> Option<String> {
-    let claims: Value = id_token
-        .split('.')
-        .nth(1)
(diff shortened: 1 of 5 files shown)


farion1231:
@codex review

farion1231:
Merged — thank you @SaladDay!

This closes out exactly the fixture follow-up flagged after #6780, and it does it the right way: removing the `cfg(test)` "test-user" fallback means the identity gates (`ensure_account_ready_for_use`, `reauth_required`) are now exercised by tests under the same physics as production, instead of a regime where every account's identity compares equal. The tightened compact-JWT shape check is a nice bonus — real tokens are unaffected, and malformed tokens now converge on the fail-closed reauth path introduced in #6780. The scope discipline (no signature/issuer/audience verification, no dedup) and the two blind review passes are appreciated.

Two small follow-ups if you're interested — neither blocked the merge:

1. In `services/proxy.rs` (~line 5414), the backup assertion still checks `!contains("cli-id-a1")`, but nothing writes that literal anymore after the fixture swap, so that arm is vacuously true. The access/refresh arms still lock the invariant, but switching the third arm to `!backup.original_config.contains(&id_token_a)` restores it.
2. `test_codex_id_token` is deterministic, so a same-subject "rotated" id_token is now byte-identical to the manager's copy — the integration tests' CLI-refresh scenarios no longer drive the id-token half of `material_changed` (the adoption unit tests still cover it with opaque strings, which is why leaving those as-is was the right call). An optional extra-claims parameter (e.g. a nonce) would let integration fixtures model "same subject, new token" the way real refreshes behave.

Thanks again for the careful work here.

Issue #6836, opened 2026-08-26
Zhipu GLM5.3不支持视觉，但是CC生成的配置里标记为了支持视觉


SwakinX:
### Self Checks / 自检

- [x] I have read the [FAQ](https://github.com/farion1231/cc-switch#faq) section in README.
我已阅读 README 中的[常见问题](https://github.com/farion1231/cc-switch#常见问题)。

- [x] I have searched for [existing issues](https://github.com/farion1231/cc-switch/issues), including closed ones.
我已搜索过[已有的 Issue](https://github.com/farion1231/cc-switch/issues)，包括已关闭的。

### CC Switch Version / 版本号

v3.20.0

### Operating System / 操作系统

Windows

### Related App / 涉及应用

Codex

### Steps to Reproduce / 重现步骤

接入Zhipu GLM，codex自己在测试软件时截图发送给模型

### Expected Behavior / 期望行为

不识别为图像支持

### Actual Behavior / 实际行为

报错
{"error":{"message":"CC Switch local proxy failed while handling Codex endpoint /responses. Provider: Zhipu GLM; model: glm-5.3; upstream_status: HTTP 400; cause: messages.content.type 参数非法，取值范围 ['text']","type":"upstream_error","code":"1210","param":null,"provider":"Zhipu GLM","model":"glm-5.3","endpoint":"/responses","upstream_status":400}}

CC Switch local proxy failed while handling Codex endpoint /responses. Provider: Zhipu GLM; model: glm-5.3; upstream_status: HTTP 400; cause: messages.content.type 参数非法，取值范围 ['text']

### Additional Context / 补充信息

之前有同样的问题已被关闭https://github.com/farion1231/cc-switch/issues/4792
但是 glm-5.3 在 cc-switch-model-catalog.json 里仍然标记为支持图像

临时解决，自己修改文件把 input_modalities 里的 image 项去掉，保存为cc-switch-model-catalog-text-only.json，在CC配置里添加一项
model_catalog_json = "cc-switch-model-catalog-text-only.json"

teddyli18000:
I can take this. The issue appears to be that `glm-5.3` is missing from the exact text-only capability registry; I’ll prepare a focused fix with a regression test.

PR #6851, opened 2026-08-26
fix(codex): mark glm-5.3 as text-only

teddyli18000:
## Summary

- add `glm-5.3` to the exact text-only model capability registry
- keep `glm-5.3v` fail-open as image-capable
- extend the existing registry regression coverage

Fixes #6836

## Validation

Locally verified the registry behavior.

PR #6854, opened 2026-08-26
fix(model-catalog): add glm-5.3 to confirmed text-only model list (#6836)

aniruddhaadak80:
## What Problem This Solves

Zhipu GLM-5.3 does not support image input, but CC Switch's model catalog marks it as image-capable. When Codex sends a screenshot to GLM-5.3, the upstream rejects it with HTTP 400: `messages.content.type 参数非法，取值范围 ['text']`.

## Why This Change Was Made

The `CONFIRMED_TAILS` list in `model_capabilities.rs` contains `glm-5.1` and `glm-5.2` as confirmed text-only models, but `glm-5.3` was missing. The fail-open default (`Unknown → Supported`) treated it as image-capable.

## User Impact

GLM-5.3 is now correctly identified as text-only. Codex will not send image content to this model, preventing the HTTP 400 error.

## Evidence

- Issue: https://github.com/farion1231/cc-switch/issues/6836
- Root cause: `glm-5.3` missing from `CONFIRMED_TAILS` in `model_capabilities.rs:72-102`
- Fix: added `"glm-5.3"` to the list, matching the pattern of `glm-5.1` and `glm-5.2`
- 1 file changed, 1 insertion

### DIFF
--- a/src-tauri/src/model_capabilities.rs
+++ b/src-tauri/src/model_capabilities.rs
@@ -79,6 +79,7 @@ pub(crate) fn is_confirmed_text_only_model(model: &str) -> bool {
         // Exact rather than prefix matching: GLM visual models use a `v`
         // suffix (for example glm-5.2v), which must remain image-capable.
         "glm-5.2",
+        "glm-5.3",
         "kat-coder",
         "kat-coder-pro",
         "kat-coder-pro v1",


farion1231:
@codex review

farion1231:
Thanks for the patch! This was fixed by #6851, which was submitted a few hours earlier by the contributor who had claimed #6836 in the issue thread, and also ships regression tests for the `[1M]` normalization and the `glm-5.3v` fail-open case. Closing this one in favor of the earlier PR — appreciate you taking a look at the issue.

PR #6863, opened 2026-08-27
fix(codex): move $ref siblings into allOf for Moonshot/Kimi chat upstreams

### LINKED ISSUE #6867: [Bug] Codex 经本地代理接入 Moonshot(Kimi) 必现 400：工具 schema 中 $ref 带兄弟关键字未被清洗
## 环境

- CC Switch v3.20.0（Windows 11，本地代理模式）
- Codex 桌面端 26.820.60940
- 供应商：Moonshot 官方 `https://api.moonshot.cn/v1`，模型 `kimi-k3` / `kimi-k2.7-code`
- 链路：Codex → CC Switch 本地代理（`/responses` 转 `/v1/chat/completions`）→ Moonshot

## 现象

Codex 的每个请求都被上游拒绝，客户端报错：

```
CC Switch local proxy failed while handling Codex endpoint /responses. Provider: Kiminew; model: kimi-k3; upstream_status: HTTP 400; cause: Invalid request: tools.function.parameters is not a valid moonshot flavored json schema, details: <At path '$defs.__schema20': when using $ref, type should be defined in the referenced schema instead of the parent schema>
```

100% 必现，与 API key、额度、模型选择无关（换 kimi-k2.7-code 同样失败）。

代理日志确认转换与转发路径正常：

```
[2026-08-27][11:23:24][INFO][cc_switch_lib::proxy::forwarder] [Codex] >>> 请求目标: https://api.moonshot.cn/v1/chat/completions (model=kimi-k3)
```

## 根因分析

Codex 内置工具的参数 schema 由 Rust schemars 自动生成，遵循 JSON Schema 2020-12 规范——`$ref` 允许与 `type`、`description` 等兄弟关键字并列（报错中的 `$defs.__schema20` 即 schemars 生成的内联定义名）。

Moonshot 服务端使用的是旧式校验（draft-07 语义）：`$ref` 节点不允许任何兄弟关键字。CC Switch 在 responses → chat completions 转换时原样转发 `tools`，因此 Moonshot 对每个带工具的请求都返回 400。

## 最小复现（已直连 api.moonshot.cn 验证，可排除 CC Switch 因素）

工具参数中出现如下节点（`$ref` 带兄弟关键字）：

```json
{
  "type": "object",
  "properties": {
    "mode": { "$ref": "#/$defs/Mode", "type": "string", "description": "which mode" }
  },
  "$defs": { "Mode": { "type": "string", "enum": ["a", "b"] } }
}
```

→ HTTP 400：`tools.function.parameters is not a valid moonshot flavored json schema, details: <At path 'properties.mode': when using $ref, type should be defined in the referenced schema instead of the parent schema>`

对照实验（同一账号、同一模型）：

| 请求 | 结果 |
|---|---|
| 不带 tools 的普通对话 | 200 |
| `$ref` 带兄弟关键字（上述写法） | 400，错误一致 |
| 兄弟关键字合并进 `$defs` 目标，`$ref` 独占节点 | 200 |
| 完全内联、不使用 `$defs` | 200 |

可见 Moonshot 接受 `$defs` + `$ref` 本身，仅拒绝 `$ref` 带兄弟关键字这一种形态。

## 修复建议

在 Codex → chat completions 的转换链路中对 `tools[].function.parameters` 做一遍 schema 清洗：

1. 递归遍历 schema；
2. 遇到含 `$ref` 且有兄弟关键字的节点时，克隆被引用的 `$defs` 条目（生成唯一新名，如 `Mode__m1`），将兄弟关键字深合并进克隆体（兄弟优先）；
3. 将原节点替换为指向克隆体的独立 `$ref`。

该变换语义等价（2020-12 中 `$ref` 兄弟与引用目标取交集），且已验证能 100% 通过 Moonshot 校验。无法解析的 `$ref` 形态可退化为丢弃兄弟关键字，保证至少能通过校验。

我已按此方案实现了一个本地清洗代理（Python 约 100 行）挂在 CC Switch 上游验证：原先必现 400 的请求全部 200 通过，SSE 流式正常。如维护者指出转换器所在模块（Rust 侧 proxy/forwarder？），我可以按此思路提交 PR。

czhmartinez:
## Maintainer update (2026-09-04)

The fix has been redesigned on top of the original commit (second commit on this branch), following the review rounds and real-device verification with Codex Desktop → local proxy → Kimi For Coding.

**What changed**

- The global strip in `transform_codex_chat.rs` is reverted. Dropping the siblings was lossy (`description`, `minLength`, `format` disappeared for every upstream, including the Codex → Anthropic path, which shares the same tool context).
- New `transform_codex_chat_moonshot_schema.rs`: a vendor-gated, lossless rewrite applied after the Responses → Chat conversion. When the resolved upstream host is `moonshot.cn` / `moonshot.ai` / `kimi.com` (or a subdomain), each `$ref` that carries sibling keywords is moved into `allOf` and the siblings stay in place, i.e. the draft-07 idiom for the same constraint. Traversal only enters schema-valued keywords, so data values (`default`, `enum`, …) and unknown extensions are never touched. The rewrite is idempotent.
- Every other provider keeps byte-identical tool schemas, so prompt-cache prefixes are not disturbed. The Codex → Anthropic path is left alone because Moonshot's Anthropic-compatible endpoint accepts the original shape.

**Verification**

- Probes against Kimi For Coding (model `k3`): the chat endpoint rejects `$ref` + siblings (400) and accepts the `allOf` form at property, `$defs` entry, root and existing-`allOf` positions (200); the Anthropic-compatible endpoint returns 200 for both shapes.
- Codex Desktop → local proxy → Kimi For Coding on a real device: the 400 from #6867 becomes 200.
- `cargo fmt --check`, `cargo clippy --all-targets`, full backend suite 2785 passed / 0 failed on the merge with `main`.

**Known limit**: relays that forward to Moonshot under their own host are not matched; extend `MOONSHOT_HOST_SUFFIXES` if such a report shows up.

Fixes #6867

---

### DIFF
--- a/src-tauri/src/proxy/forwarder.rs
+++ b/src-tauri/src/proxy/forwarder.rs
@@ -1663,6 +1663,27 @@ impl RequestForwarder {
             );
         }
 
+        // Moonshot / Kimi Chat Completions reject `$ref` nodes that carry sibling
+        // keywords, which Codex Desktop's built-in tool schemas do (#6867). Move
+        // each such `$ref` into `allOf` for that upstream only; every other
+        // provider keeps byte-identical tool schemas (prompt-cache prefix intact).
+        if codex_responses_to_chat
+            && super::providers::transform_codex_chat_moonshot_schema::upstream_requires_ref_sibling_all_of(
+                &base_url,
+            )
+        {
+            let rewritten =
+                super::providers::transform_codex_chat_moonshot_schema::wrap_ref_siblings_in_chat_tools(
+                    &mut request_body,
+                );
+            if rewritten > 0 {
+                log::debug!(
+                    "[Codex] Moved `$ref` siblings into allOf for {rewritten} tool schema(s) (Moonshot upstream, provider={})",
+                    provider.id
+                );
+            }
+        }
+
         if matches!(app_type, AppType::Codex | AppType::GrokBuild) {
             self.apply_media_prevention(&mut request_body, provider);
         }
--- a/src-tauri/src/proxy/providers/mod.rs
+++ b/src-tauri/src/proxy/providers/mod.rs
@@ -34,6 +34,7 @@ pub mod streaming_responses;
 pub mod transform;
 pub mod transform_codex_anthropic;
 pub mod transform_codex_chat;
+pub mod transform_codex_chat_moonshot_schema;
 pub mod transform_codex_responses_namespace;
 pub mod transform_codex_responses_xai_sanitize;
 pub mod transform_gemini;
--- a/src-tauri/src/proxy/providers/transform_codex_chat.rs
+++ b/src-tauri/src/proxy/providers/transform_codex_chat.rs
@@ -1282,40 +1282,7 @@ fn normalize_function_parameters(params: Option<&Value>) -> Value {
             }
         }
     }
-    strip_ref_siblings(&mut params);
     params
-}
-
-/// Strip sibling keys from any schema node that contains `$ref`.
-///
-/// Pydantic-generated `$defs` (e.g. Codex's built-in `automation_update` tool)
-/// emit nodes like `{"type": "string", "format": "uuid", "$ref": "#/$defs/x"}`.
-/// Draft 2020-12 tolerates `$ref` siblings and OpenAI/Anthropic accept them,
-/// but strict validators (Moonshot's "moonshot flavored json schema" check,
-/// and other OpenAI-compatible gateways) reject the whole request with
-/// HTTP 400 ("type should be defined in the referenced schema instead of the
-/// parent schema"). Dropping the siblings preserves semantics: the referenced
-/// schema already carries the authoritative constraints.
-fn strip_ref_siblings(value: &mut Value) {
-    match value {
-        Value::Object(map) => {
-            if map.contains_key("$ref") && map.len() > 1 {
-                let ref_value = map.get("$ref").cloned().unwrap_or(Value::Null);
-                map.clear();
-                map.insert("$ref".to_string(), ref_value);
-                return;
-            }
-            for child in map.values_mut() {
-                strip_ref_siblings(child);
-            }
-        }
-        Value::Array(items) => {
-            for child in items.iter_mut() {
-                strip_ref_siblings(child);
-            }
-        }
-        _ => {}
-    }
 }
 
 fn responses_function_tool_to_chat_tool(tool: &Value, chat_name: &str) -> Option<Value> {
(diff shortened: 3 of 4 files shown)


farion1231:
@codex review

farion1231 (commented):
Thanks — this is a solid diagnosis and a well-verified fix: pinpointing the Pydantic-style `$ref`-with-siblings in Codex's built-in `automation_update` schema, ruling out user MCP servers, and confirming live 400→200 against the real Moonshot upstream. I've reproduced the verification locally: the branch merges clean against main, and with the fmt issue fixed the full backend suite and clippy are green.

**[P1] CI is red on all three platforms, but only because of `cargo fmt`.** The multi-line `assert_eq!` in `normalize_parameters_also_strips_ref_siblings` ([L2101-L2112](https://github.com/farion1231/cc-switch/blob/b5e5007ced200b35582eaaa335a2d54987175678/src-tauri/src/proxy/providers/transform_codex_chat.rs#L2101-L2112)) needs to collapse to one line ([CI log](https://github.com/farion1231/cc-switch/actions/runs/33036276117/job/98732854714)). Since fmt is the first CI step, clippy/tests never ran. Just run `cargo fmt` and push.

**[P2] "Dropping the siblings preserves semantics" ([L1297-L1298](https://github.com/farion1231/cc-switch/blob/b5e5007ced200b35582eaaa335a2d54987175678/src-tauri/src/proxy/providers/transform_codex_chat.rs#L1297-L1298)) overstates it — under draft 2020-12, `$ref` siblings are live (conjunctive), not ignored.** Your own `__schema20` example shows it: `minLength`, `format: "uuid"`, and `description` live in the parent node, and the referenced `__schema2` does *not* carry them ([spec §8.2.3.1](https://json-schema.org/draft/2020-12/json-schema-core.html#section-8.2.3.1)). So for upstreams that already accepted these schemas, stripping is a real (if small) regression: the model loses field descriptions and constraints for the affected tools. I'm fine taking the strip as a stopgap — for Moonshot the alternative is a 100% failure, and gating by provider isn't practical at this layer. Two asks:
- In this PR: soften the doc comment to something like "lossy for draft 2020-12 siblings, but safe: strict upstreams reject the request outright otherwise".
- Follow-up (no need in this PR): replace strip with inline-merge — dereference the target and merge it into the node (sibling keys winning), dropping `$ref`. That satisfies strict validators *and* keeps constraints/descriptions for everyone. I'll open an issue for it.

**[P3] A root-level `$ref` schema becomes a dangling reference.** Verified locally: `normalize_function_parameters` on `{"$ref": "#/$defs/Args", "$defs": {...}}` returns just `{"$ref": "#/$defs/Args"}` — `map.clear()` wipes `$defs` along with the siblings. Not blocking: MCP requires a top-level object schema and no real generator emits this shape today, and inline-merge would eliminate it anyway. Flagging so it's on record.

**[P3] The walk can't distinguish schema nodes from maps-of-schemas.** A tool parameter literally named `$ref` under `properties` (or a `$defs` entry named `$ref`, or user data containing a `$ref` key in `default`/`examples`) makes the containing map hit the `contains_key("$ref") && len > 1` branch and get wiped to a single bogus entry ([L1302-L1306](https://github.com/farion1231/cc-switch/blob/b5e5007ced200b35582eaaa335a2d54987175678/src-tauri/src/proxy/providers/transform_codex_chat.rs#L1302-L1306)). Extremely unlikely in practice; a ~10-line fix is to descend directly into the values of `properties`/`$defs`/`patternProperties` without testing those maps themselves. Fine as part of the follow-up.

nit: the Notes claim idempotency but there's no test asserting a second pass is a no-op.

Fix the P1 (one `cargo fmt`) and the doc wording, and this is mergeable as a stopgap — happy to re-review quickly. Also consider adding `Fixes #6867` to the description so the issue auto-closes on merge.

farion1231 (approved):
Redesigned on top of the original commit: vendor-gated allOf rewrite for Moonshot/Kimi chat upstreams, lossless and byte-identical for every other provider. Verified on a real device (Codex Desktop → Kimi For Coding, 400 → 200) and with the full backend suite on the merge with main.

farion1231:
Merged, thank you @czhmartinez! Your report, the `$defs.__schema20` root-cause analysis and the real-upstream 400 → 200 verification are what pinned this down; the final shape only narrows the rewrite to Moonshot/Kimi hosts and swaps the strip for the lossless `allOf` form, so descriptions and constraints survive for every upstream.

One note on the earlier review: the inline-merge follow-up mentioned there is no longer needed. Moving `$ref` into `allOf` is exact by construction, so there is nothing left to merge.

This will ship in the next release.

PR #6866, opened 2026-08-27
fix(codex): preserve non-managed [model_providers.*] sections on config write

Monking-21:
## Problem

CC Switch's config writer only stores its own active provider in the database. When it writes `config.toml`, any `[model_providers.*]` sections that were manually added by the user (legacy aliases, third-party providers, etc.) are silently dropped because they never existed in the DB-stored config.

This causes old Codex threads that reference legacy provider names like `proxy` or `kimi3` to break after a provider switch or failover.

**Root cause:** `write_codex_live_for_provider` → `plan_codex_live_write` reads config from the DB (which only contains the CC-Switch-managed `custom` block), then writes it back to `config.toml` via `write_codex_live_config_atomic` — non-managed sections are lost because they never existed in the DB input.

## Fix

Read the live `config.toml` before each write, identify `[model_providers.*]` sections that are NOT present in the new config (i.e., not managed by CC Switch), and merge them back in. This preserves user-added provider aliases across all write paths (`write_codex_live_atomic` and `write_codex_live_config_atomic`).

The merge function (`merge_live_non_managed_provider_sections`) is applied at the lowest-level write functions, so all code paths benefit automatically.

## Verification

```bash
# 1. Remove proxy/kimi3 from config (only 'custom' remains)
# 2. Add proxy/kimi3 back manually
# 3. Trigger provider switch via CC Switch
# 4. Confirm proxy/kimi3 still present after switch ✅
```

Deployed patched build on production — legacy sections survive provider switches and failovers.

### DIFF
--- a/src-tauri/src/codex_config.rs
+++ b/src-tauri/src/codex_config.rs
@@ -987,10 +987,15 @@ pub fn write_codex_live_atomic(
         None
     };
 
-    // 准备写入内容
-    let cfg_text = match config_text_opt {
+    // 准备写入内容，同时 merge live config 中非 CC-Switch 管理的 sections
+    let raw_cfg_text = match config_text_opt {
         Some(s) => s.to_string(),
         None => String::new(),
+    };
+    let cfg_text = if raw_cfg_text.trim().is_empty() {
+        raw_cfg_text
+    } else {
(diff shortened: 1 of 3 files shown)


farion1231:
@codex review

PR #6870, opened 2026-08-27
feat(provider): extend PPIO presets for Pi

### LINKED ISSUE #6868: Extend PPIO presets to Pi and Claude Code model discovery
## Self Checks

- [x] I have read the FAQ section in the README.
- [x] I searched existing issues, including closed issues.

## Related App

- Claude Code
- General

## Problem or Motivation

The completed PPIO integration from #6226 covers the existing provider surfaces, but Pi does not yet expose a PPIO preset. Claude Code also uses the native Anthropic endpoint, so automatic model discovery needs an explicit OpenAI-compatible models URL instead of relying on endpoint derivation.

## Proposed Solution

- Add a PPIO Pi provider preset with the OpenAI-compatible base URL and `deepseek/deepseek-v4-flash-0731` as its model.
- Preserve the documented text-only, reasoning, context-window, output-limit, and DeepSeek compatibility metadata.
- Add `https://api.ppio.com/openai/v1/models` as the Claude Code preset `modelsUrl`.
- Extend focused regression coverage without asserting preset or icon source positions.

## Additional Context

This change does not add user-facing text or modify Rust code. The models endpoint and configured model will be validated with a live authenticated request, and the complete frontend validation suite will be run before the pull request.

## Contribution

hu-miao:
## Summary

- Add a PPIO provider preset for Pi using the OpenAI-compatible endpoint.
- Configure the Claude Code PPIO preset with an explicit models endpoint.
- Extend focused PPIO regression coverage without source-order assertions.

## Related Issue

Fixes #6868

## Screenshots

Not applicable. This change only updates provider presets and tests.

## Checklist

- [x] `pnpm typecheck` passes.
- [x] `pnpm format:check` passes.
- [x] `pnpm test:unit` passes.
- [x] Rust checks are not applicable because no Rust code changed.
- [x] Localization updates are not applicable because no user-facing text changed.

## Validation

- Focused provider tests: 3 test files and 28 tests passed.
- Complete unit suite: 133 test files and 1,014 tests passed.
- Live `GET https://api.ppio.com/openai/v1/models`: HTTP 200, and the configured model is present.

### DIFF
--- a/src/config/claudeProviderPresets.ts
+++ b/src/config/claudeProviderPresets.ts
@@ -425,6 +425,7 @@ export const providerPresets: ProviderPreset[] = [
     isPartner: true,
     partnerPromotionKey: "ppio",
     endpointCandidates: ["https://api.ppio.com/anthropic"],
+    modelsUrl: "https://api.ppio.com/openai/v1/models",
     icon: "ppio",
     iconColor: "#2874FF",
   },
--- a/src/config/piProviderPresets.ts
+++ b/src/config/piProviderPresets.ts
@@ -466,6 +466,35 @@ const piProviderPresetDefinitions: PiProviderPreset[] = [
     icon: "teamorouter",
   },
   {
+    name: "PPIO",
+    providerKey: "cc-switch-ppio",
+    websiteUrl: "https://ppio.com",
+    apiKeyUrl: "https://ppio.com/activity/ccswitch",
+    settingsConfig: {
+      name: "PPIO",
+      baseUrl: "https://api.ppio.com/openai/v1",
+      api: "openai-completions",
+      apiKey: "",
+      models: [
+        {
+          ...piModel("deepseek/deepseek-v4-flash", {
+            id: "deepseek/deepseek-v4-flash-0731",
+            name: "Deepseek V4 Flash 0731",
+            contextWindow: 1_048_576,
+            maxTokens: 393_216,
+            thinkingProfile: "deepseekV4",
+          }),
+          compat: { ...DEEPSEEK_THINKING_COMPAT },
+        },
+      ],
+    },
+    category: "aggregator",
+    isPartner: true,
+    partnerPromotionKey: "ppio",
+    icon: "ppio",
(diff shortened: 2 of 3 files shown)


farion1231:
@codex review

farion1231 (approved):
Thanks for this — verified, and merging.

I confirmed the `modelsUrl` change fixes a real failure rather than a theoretical one. All three endpoints cc-switch derives automatically for the PPIO Claude preset return 404, so "Fetch models" could never succeed:

| URL | Status |
| --- | --- |
| `https://api.ppio.com/anthropic/v1/models` | 404 |
| `https://api.ppio.com/v1/models` (after stripping `/anthropic`) | 404 |
| `https://api.ppio.com/models` | 404 |
| `https://api.ppio.com/openai/v1/models` | **200** (114 models) |

The generic fallback in `build_models_url_candidates` only strips a known compatibility suffix and retries at the root, so it structurally cannot reach PPIO's `/openai/v1` subpath. Pinning `modelsUrl` on the preset is the right layer for this.

I also checked the Pi preset values against the live API for `deepseek/deepseek-v4-flash-0731`: `context_size` 1048576, `max_output_tokens` 393216, text-only input, reasoning supported — all match. For future reference, those overrides are deliberate and should not be "corrected" back to the catalog base entry (1000000/384000), which carries DeepSeek's own upstream figures rather than PPIO's platform figures.

One thing I fixed directly in a28d4f7e0 rather than sending back for another round: the new Pi block was appended at the end of the preset array. The preset selector preserves file order within the sponsor group, so PPIO rendered last among sponsors instead of in its slot. I moved the block between TeamoRouter and ClaudeCN to match the README sponsor table and the six other preset files — a pure move, no field changes.

A follow-up for me, not for this PR: Jiekou and Novita have exactly the same gap (both serve OpenAI under `/openai/v1`, and all three derived candidates 404), and `jiekouProviderPresets.test.ts` currently asserts the absence of `modelsUrl`, which locks the broken behaviour in.

PR #6873, opened 2026-08-27
fix(db): bound database growth from oversized error bodies, backups, and the dedup ledger (#6706)

### LINKED ISSUE #6706: c盘直接给我干爆了，这啥玩意 22g?????
### Self Checks / 自检

### Related App / 涉及应用

General / 通用

### Your Question / 你的问题

Image

### What Have You Tried / 你已经尝试过什么

_No response_

### Environment / 环境信息

_No response_

zayokami:
## Summary

Fixes #6706 (database folder growing to 22 GB). The reporter's folder held ten daily full-image backups of an already-bloated ~2.4 GB database (`db_backup_*.db` × 10 = 22.4 GB) plus an orphaned `.db-journal`. Three independent growth vectors plus backup hygiene, each now bounded:

### 1. Oversized `proxy_request_logs.error_message` (`0e16f0a3`)

Upstream error bodies (relay full-page HTML, long JSON) were persisted verbatim since error request logging was introduced — a chatty failing upstream could push the DB to multiple GB on its own.

- `UsageLogger` truncates `error_message` to `MAX_ERROR_MESSAGE_CHARS = 2000` on char boundaries before insert.
- Schema v17 → v18 migration truncates legacy oversized rows with SQL `substr()`, which is character-based and matches the logger-side cap. The step no-ops on layouts where the column does not exist yet.
- Space actually returns to disk via the existing incremental auto-vacuum path (`ensure_incremental_auto_vacuum` at init converts older DBs with a full VACUUM; startup/periodic maintenance issues `PRAGMA incremental_vacuum` after reclaimed-row maintenance).

### 2. Unbounded backups folder (`c13dc4b6`, `eb15c025`)

Retention only ran when a new backup was created, and only enforced a count.

- New `backup_max_total_mb` setting (default **2048 MB**, `0` = unlimited) with a selector in Settings → Database Backups; localized in en/zh/zh-TW/ja.
- `cleanup_db_backups` still keeps the newest N files first, then trims oldest-first until the folder fits the cap. The newest restore point always survives even when it alone exceeds the cap, and protected paths (just-created backup / restore source) are never deleted in either pass.
- `periodic_backup_if_needed` now runs count + size retention even when no new backup is due, so an oversized folder like the reporter's is bounded on first launch after upgrade instead of waiting up to a full backup interval.
- Deleting a backup now also removes its SQLite sidecar files (`-journal`/`-wal`/`-shm`), and orphan sidecars whose `.db` no longer exists are swept during cleanup — the reporter's folder contained exactly such residue. Sidecars of surviving backups stay untouched (they may belong to an in-progress restore).

### 3. Uncapped `session_usage_dedup` ledger (`c13dc4b6`)

The ledger added by #6463 has no timestamp column and is never pruned by design, so it grows forever. It is now trimmed to its newest 1M rows during periodic maintenance (rowid order = insertion order). Only the pi session importer consults this ledger, and per-file sync cursors remain the primary guard against re-counting, so evicting the oldest entries cannot resurrect historical usage.

### Test-only prerequisite (`9a679306`)

On Windows, the v3.10.3 legacy-database probe read `$HOME` unconditionally; Git Bash/MSYS injects a HOME pointing at the real user profile, redirecting explicitly test-isolated runs back to real user data. Tests now skip that probe when `CC_SWITCH_TEST_HOME` is set. Production behavior unchanged.

## Upgrading-user outcome

After upgrading with this PR, the first launch migrates v17→v18 (truncating oversized error bodies), converts pre-incremental-vacuum databases if needed, runs retention once immediately — the 22.4 GB backups folder drops to ≤ the size cap while keeping one restore point — and periodic maintenance reclaims freed pages via incremental vacuum.

## Test plan

- [x] New unit tests: logger char-boundary truncation + end-to-end oversized-error insert; v17→v18 migration (oversized/multibyte/short rows); size-cap cleanup incl. protected-path survival; startup retention without a due backup; dedup-ledger cap eviction order; sqlite sidecar removal + orphan sweep
- [x] `cargo fmt --check`, `cargo clippy --all-targets -- -D warnings`, full `cargo test` (2806 passed)
- [x] `pnpm typecheck`, `pnpm format:check`

### DIFF
--- a/src-tauri/src/config.rs
+++ b/src-tauri/src/config.rs
@@ -214,7 +214,9 @@ pub fn get_app_config_dir() -> PathBuf {
     #[cfg(windows)]
     {
         let default_db = default_dir.join("cc-switch.db");
-        if !default_db.exists() {
+        // 设置了 CC_SWITCH_TEST_HOME 时跳过遗留探测：Git Bash/MSYS 会注入
+        // 指向真实用户目录的 HOME，把显式隔离的测试重新指回真实数据。
+        if !default_db.exists() && std::env::var_os("CC_SWITCH_TEST_HOME").is_none() {
             if let Ok(home_env) = std::env::var("HOME") {
                 let trimmed = home_env.trim();
                 if !trimmed.is_empty() {
--- a/src-tauri/src/database/backup.rs
+++ b/src-tauri/src/database/backup.rs
@@ -444,6 +444,15 @@ impl Database {
             }
         }
 
+        // Retention (count + total-size cap) also runs when no new backup is
+        // due, so an oversized backups folder is bounded on the first launch
+        // after an upgrade instead of waiting up to a full backup interval.
+        let backup_dir = get_app_config_dir().join("backups");
+        if backup_dir.exists() {
+            let _backup_file_guard = lock_backup_file_operations()?;
+            Self::cleanup_db_backups(&backup_dir, &[])?;
+        }
+
         // Periodic maintenance is always enabled, regardless of auto-backup settings.
         let mut reclaimed_rows = 0u64;
         match self.cleanup_old_stream_check_logs(7) {
@@ -460,6 +469,14 @@ impl Database {
             }
             Err(e) => {
                 log::warn!("Periodic rollup_and_prune failed: {e}");
+            }
+        }
+        match self.cap_session_usage_dedup(super::dao::usage_rollup::SESSION_USAGE_DEDUP_MAX_ROWS) {
+            Ok(deleted) => {
+                reclaimed_rows += deleted;
+            }
+            Err(e) => {
+                log::warn!("Periodic session_usage_dedup cap failed: {e}");
             }
         }
         if reclaimed_rows > 0 {
@@ -598,10 +615,12 @@ impl Database {
         }
     }
 
-    /// 清理旧的数据库备份，保留最新的 N 个
+    /// 清理旧的数据库备份：先按份数保留最新的 N 个，再把目录总大小压到
+    /// `backup_max_total_mb` 上限内（至少保留 1 份）。protected_paths
+    /// （新建备份/恢复源）在两轮中都绝不删除。
     fn cleanup_db_backups(dir: &Path, protected_paths: &[&Path]) -> Result<(), AppError> {
         let retain = crate::settings::effective_backup_retain_count();
-        let entries = match fs::read_dir(dir) {
+        let mut entries = match fs::read_dir(dir) {
             Ok(iter) => iter
                 .filter_map(|entry| entry.ok())
                 .filter(|entry| {
@@ -615,31 +634,78 @@ impl Database {
             Err(_) => return Ok(()),
         };
 
-        if entries.len() <= retain {
+        let is_protected = |path: &Path| -> bool {
+            protected_paths
+                .iter()
+                .any(|protected| Self::same_existing_backup_path(path, protected))
+        };
+
+        entries.sort_by_key(|entry| entry.metadata().and_then(|m| m.modified()).ok());
+
+        if entries.len() > retain {
+            let remove_count = entries.len() - retain;
+            let mut removed = 0;
+            entries.retain(|entry| {
+                if removed >= remove_count {
+                    return true;
+                }
+                let path = entry.path();
+                if is_protected(&path) {
+                    return true;
+                }
+
+                match fs::remove_file(&path) {
+                    Ok(()) => {
+                        removed += 1;
+                        false
+                    }
+                    Err(err) => {
+                        log::warn!("删除旧数据库备份失败 {}: {}", path.display(), err);
+                        true
+                    }
+                }
+            });
+        }
+
+        let max_total = crate::settings::effective_backup_max_total_bytes();
+        if max_total == u64::MAX {
             return Ok(());
         }
 
-        let remove_count = entries.len().saturating_sub(retain);
-        let mut sorted = entries;
-        sorted.sort_by_key(|entry| entry.metadata().and_then(|m| m.modified()).ok());
-
-        let mut removed = 0;
-        for entry in sorted {
-            if removed >= remove_count {
-                break;
-            }
+        // Newest-first: keep backups until the accumulated size exceeds the
+        // cap. The newest non-protected backup is always kept even when it
+        // alone exceeds the cap, so a restore point always survives.
+        let mut kept = 0usize;
+        let mut total = 0u64;
+        for entry in entries.iter().rev() {
+            let Ok(metadata) = entry.metadata() else {
+                continue;
+            };
+            let size = metadata.len();
             let path = entry.path();
-            if protected_paths
-                .iter()
-                .any(|protected| Self::same_existing_backup_path(&path, protected))
-            {
+            if is_protected(&path) {
+                total = total.saturating_add(size);
                 continue;
             }
 
-            if let Err(err) = fs::remove_file(&path) {
-                log::warn!("删除旧数据库备份失败 {}: {}", path.display(), err);
-            } else {
-                removed += 1;
+            if total.saturating_add(size) <= max_total || kept == 0 {
+                total = total.saturating_add(size);
+                kept += 1;
+                continue;
+            }
+
+            match fs::remove_file(&path) {
+                Ok(()) => {
+                    log::info!(
+                        "删除超限数据库备份 {}（目录总大小超过上限）",
+                        path.display()
+                    );
+                }
+                Err(err) => {
+                    log::warn!("删除超限数据库备份失败 {}: {}", path.display(), err);
+                    total = total.saturating_add(size);
+                    kept += 1;
+                }
             }
         }
         Ok(())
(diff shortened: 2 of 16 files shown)


farion1231 (commented):
四层修复每层都对着 #6706 现场的一个实锤，测试也扎实——protected 存活、孤儿侧车清扫、多字节截断边界都有反向断言；恢复/导入路径重放迁移链，让旧备份里的超长行不会回流，这点做得很对。我在合并树上跑了全量测试（Rust + 前端）全绿。

有一个 P1 和一个 P2 希望合并前处理，另有几条小的可以顺手或 follow-up：

**[P1] 磁盘已满的用户升级后会卡死在启动，永远到不了清理**

保留策略要等 `Database::init` 成功返回后才会执行，但对 #6706 的目标用户（盘满）init 走不完：

- 迁移前备份失败只是 warn 继续（[mod.rs L127-L142](https://github.com/farion1231/cc-switch/blob/eb15c025fc022f2f37cd4155ca80e5baa1663409/src-tauri/src/database/mod.rs#L127-L142)），这步不阻塞；
- 但 v17→v18 的存量截断 UPDATE 需要写回滚 journal（约等于被修改页的原像，存量超长行多时是 GB 级），满盘时 `SQLITE_FULL`，`apply_schema_migrations()?` 硬失败（同上 L142）；
- init 失败后只有「重试/退出」对话框（[lib.rs L594-L599](https://github.com/farion1231/cc-switch/blob/eb15c025fc022f2f37cd4155ca80e5baa1663409/src-tauri/src/lib.rs#L594-L599)），重试再撞同一堵墙——清理代码永远不可达。

v17 在满盘下是能正常启动的（没有大 UPDATE），所以这对最需要本修复的人群反而是升级即不可用。修法便宜：清理只依赖 settings.json 和文件系统、删文件不需要可用磁盘空间，提到 init 最前即可：

```rust
// backup.rs
/// Init 打开数据库前先执行备份保留：满盘时（#6706 现场）
/// 靠这一步腾出后续迁移所需的 journal 空间。
pub(crate) fn preflight_backup_retention() {
    let backup_dir = get_app_config_dir().join("backups");
    if !backup_dir.exists() {
        return;
    }
    let Ok(_guard) = lock_backup_file_operations() else {
        return;
    };
    if let Err(e) = Self::cleanup_db_backups(&backup_dir, &[]) {
        log::warn!("Preflight backup retention failed: {e}");
    }
}

// mod.rs：Database::init() 第一行调用
Self::preflight_backup_retention();
```

**[P2] 进程崩溃残留的临时快照绕过大小上限**

临时快照 `.cc-switch-backup-*.tmp` 建在 backups 目录内（[backup.rs L549-L553](https://github.com/farion1231/cc-switch/blob/eb15c025fc022f2f37cd4155ca80e5baa1663409/src-tauri/src/database/backup.rs#L549-L553)）。正常发布和出错返回都有保障（发布重试消费 tmp 有测试锁着、Err 路径 `TempPath` 析构会删），但进程崩溃/强杀时析构不运行，残留文件接近整库大小，而保留策略只认 `.db` 扩展名——既不计入总量也永不清理，重复中断可以无限累积。建议在孤儿侧车清扫的同一个循环里顺带处理（持 `BACKUP_FILE_OPERATION_LOCK` 期间不存在进行中的备份，删除是安全的；若担心多实例并行可加文件年龄阈值）：

```rust
if name.starts_with(".cc-switch-backup-") && name.ends_with(".tmp") {
    let _ = fs::remove_file(&path);
    continue;
}
```

**[P3] size pass 的 protected 备份没有计入 kept**（[backup.rs L725-L732](https://github.com/farion1231/cc-switch/blob/eb15c025fc022f2f37cd4155ca80e5baa1663409/src-tauri/src/database/backup.rs#L725-L732)）

Codex 机器人在 L727 提过这条，最后一个 commit 没有回应。我核了实际影响是瞬态的——periodic 路径上创建备份后，同函数尾部那次无 protected 的清理会立即收敛；手动/迁移前备份最多多留一份到下次清理。但修法是安全的（protected 分支里补 `kept += 1` 即可；比 protected 新的普通备份在从新到旧的迭代里先被访问，不受影响），建议采纳或在线程里回应一下。

**[P3] 手动删除备份不清侧车**：`delete_backup`（[backup.rs L1261](https://github.com/farion1231/cc-switch/blob/eb15c025fc022f2f37cd4155ca80e5baa1663409/src-tauri/src/database/backup.rs#L1261)）只删 `.db` 本体；残留会被你新增的孤儿清扫在下次周期收走，但辅助函数已经有了，删除成功后补一行 `Self::remove_backup_sidecars(&backup_path);` 更完整。

**[P3] 账本淘汰的双算窗口**：裁掉最老条目后，「旧会话文件被重写 + 明细已被 30 天归并」同时发生时历史用量会重复入账——PR 描述里已经披露这个权衡，我同意 100 万行下窗口足够窄，不要求本 PR 改；以后若要收紧，方向是按文件维度淘汰而非全局行数。

**nit**：备份上限下拉框对非预设值（手改 settings.json 填如 3000）会显示空白，纯显示问题。

修掉 P1（强烈建议连 P2 一起，就几行的事）就可以合并；推了修复我会尽快复审。

zayokami:
Thanks for the thorough second pass — pushed `4ad4d344` addressing P1 and P2, plus both easy P3s:

**[P1]** Added `preflight_backup_retention()`, called as the first line of `Database::init()`, before the connection even opens. It's exactly your proposed shape: reads the backup dir straight off `get_app_config_dir()`, takes the file-operation lock, runs `cleanup_db_backups` with no protected paths, and warns-and-continues on failure. Retention now runs regardless of whether the migration below it can succeed, so a full-disk upgrade frees space before the v17→v18 journal write has a chance to hit `SQLITE_FULL`.

**[P2]** Folded into the same orphan-sweep loop that already walked the backups directory for dangling sidecars: any `.cc-switch-backup-*.tmp` is now removed unconditionally there too, so a crash-killed snapshot (destructor never ran) stops being invisible to both the size cap and cleanup.

**[P3] protected → kept**: added `kept += 1` to the protected branch in the size pass, per your note that the fix is safe (protected entries are already newest-first, so a later non-protected entry can't be reordered ahead of them).

**[P3] `delete_backup` sidecars**: added the `remove_backup_sidecars` call after the manual delete, matching the other two deletion paths now.

**[P3] ledger double-count**: left as-is per your comment, not required for this PR.

**nit**: left as a follow-up, purely cosmetic.

Tests added: `preflight_backup_retention_runs_before_database_connection_opens` (exercises the standalone call, no open connection), `cleanup_db_backups_removes_orphan_crash_snapshots`, `cleanup_db_backups_size_cap_counts_protected_toward_kept`, `delete_backup_removes_sqlite_sidecars`. Full `cargo fmt --check`, `cargo clippy --all-targets -- -D warnings`, and `cargo test` (2689 lib tests + all integration suites) are green locally.

farion1231 (commented):
Re-checked `4ad4d344` — all four items from the previous review are properly closed, each with a dedicated test: `preflight_backup_retention()` runs as the first line of `Database::init()` before the connection opens, crash-orphaned `.cc-switch-backup-*.tmp` snapshots are swept in the orphan loop, the protected branch now counts toward `kept` (with the reverse assertion locking the sentinel semantics), and `delete_backup` removes sidecars. The code itself is merge-ready. One remaining blocker, purely mechanical:

**[P1] Schema version collision with current `main` — rebase the migration to v18→v19**

`main` has since shipped its own v17→v18 (session-log byte-cursor columns, [schema.rs#L547-L555](https://github.com/farion1231/cc-switch/blob/273c9cc24fd091f8aa2f49d6d683d6c1a736e430/src-tauri/src/database/schema.rs#L547-L555)), while this PR uses the same slot for the error-message truncation ([schema.rs#L539-L542](https://github.com/farion1231/cc-switch/blob/4ad4d344943ce956293f7edfb7f2e050356fe57a/src-tauri/src/database/schema.rs#L539-L542)). Keeping either side verbatim would permanently skip the other migration. Rebase steps:

1. Keep `main`'s `17 =>` arm (byte cursor); add an `18 =>` arm for the truncation, logging v18→v19.
2. Rename the truncation `migrate_v17_to_v18` → `migrate_v18_to_v19`; bump `SCHEMA_VERSION` to 19 ([mod.rs#L56](https://github.com/farion1231/cc-switch/blob/4ad4d344943ce956293f7edfb7f2e050356fe57a/src-tauri/src/database/mod.rs#L56)).
3. Rename the migration test accordingly and seed it with `set_user_version(conn, 18)`.

⚠️ One silent trap: both branches changed `SCHEMA_VERSION` from 17 to 18 identically, so git auto-merges that line to 18 with **no textual conflict** — the visible conflicts are only the log line and the two same-named migration functions. The bump to 19 has to be done by hand, otherwise one of the two migrations never runs.

The two migrations are independent (`session_log_sync` vs `proxy_request_logs`), and `main`'s byte-cursor step is guarded by `table_exists`, so ordering is safe. I resolved the merge exactly this way locally against `main`=273c9cc2 and ran the full suites: `cargo fmt --check` / `clippy -D warnings` clean, all Rust tests green (2774 lib + integration suites), `tsc` clean, vitest 1071/1071 — so this rebase is safe to do mechanically.

**Non-blocking follow-up**: the size cap only counts each backup's own `metadata.len()` ([backup.rs#L746-L755](https://github.com/farion1231/cc-switch/blob/4ad4d344943ce956293f7edfb7f2e050356fe57a/src-tauri/src/database/backup.rs#L746-L755)); sidecars of a surviving `.db` are neither counted nor removed. I traced the write paths — a published backup is never opened by a writing connection (built under a `.tmp` name and closed before publish; restore opens it read-only with staging in the system temp dir) — so large live sidecars can't be produced today. Counting each backup plus its sidecars as one bundle would still be nice hygiene against pre-existing debris; fine as a follow-up.

Fix the v19 rebase and this is good to merge — happy to re-review quickly.

zayokami:
Rebased onto latest main and moved the truncation migration to the v19 slot as you described:

- Kept main's v17→v18 arm (session-log byte cursor) verbatim — `migrate_v17_to_v18` now refers only to that function.
- Truncation now runs as v18→v19: added the `18 =>` arm calling the renamed `migrate_v18_to_v19`, and bumped `SCHEMA_VERSION` to 19 in mod.rs.
- Renamed the migration test to `migrate_v18_to_v19_truncates_oversized_error_messages`, seeded with `set_user_version(conn, 18)`. The byte-cursor test stays seeded at 17.

Verified locally: `cargo fmt --check` clean, `cargo clippy -- -D warnings` clean, full `cargo test` 2906 passed / 0 failed — including both `migrate_v17_to_v18_adds_byte_cursor_to_existing_sync_table` and `migrate_v18_to_v19_truncates_oversized_error_messages`.

The branch also now includes the sidecar-removal commit from the earlier discussion, plus the #6895 proxy reproduction tests that were already on the branch.

PR #6884, opened 2026-08-27
fix(codex): fall back to create_new + copy when hard links are unsupported (#6679)

### LINKED ISSUE #6679: Codex routing fails on WSL UNC path because hard_link returns os error 50
### Self Checks / 自检

### CC Switch Version / 版本号

3.20

### Operating System / 操作系统

Windows

### Related App / 涉及应用

Codex

## Environment

- CC Switch: 3.20.0
- OS: Windows with WSL2
- Distribution: Ubuntu
- Codex home: `\\wsl.localhost\Ubuntu\home\lee\.codex`

## Problem

Enabling or repeatedly switching Codex routing fails with:

```text
切换路由状态失败: 安装 Codex auth 失败
(\\wsl.localhost\Ubuntu\home\lee\.codex\auth.json):
The request is not supported. (os error 50)

Switching back to the official configuration works, but enabling routing again eventually reproduces the error.


Suspected cause
In CodexAuthFileTransaction::install() within src-tauri/src/services/proxy.rs, the new auth file is installed using:
std::fs::hard_link(&temporary, &self.path)
Windows cannot create this hard link through the WSL UNC provider, so the operation returns ERROR_NOT_SUPPORTED (os error 50).
This appears separate from the previous ReplaceFileW fallback added for WSL paths.


Possible solution
For \\wsl.localhost\<distro>\... paths:
- Avoid relying on Windows hard-link creation.
- Use a rollback-safe rename/install strategy supported by the WSL UNC provider, or
- Perform the file operation inside the target distribution through wsl.exe.
A regression test covering Codex routing with CODEX_HOME on a WSL UNC path would also help.

### Expected Behavior / 期望行为

WSL环境下正常切换profile

### Actual Behavior / 实际行为

_No response_

### Additional Context / 补充信息

_No response_

zayokami:
## Summary

Fixes #6679. Split out of #6510 per review, and reworked per the P1 finding there.

WSL UNC paths (`\\wsl.localhost\...`) reject hard-link creation with ERROR_NOT_SUPPORTED (os error 50), which broke both the install step and its rollback in `CodexAuthFileTransaction`, leaving auth.json stranded in its quarantine file and routing switches failing.

### Protocol (reworked per review)

- `hard_link` remains the **primary** atomic install — a failed hard link never creates the destination, so `install`'s "destination exists ⇒ a concurrent Codex write happened" heuristic keeps its exact meaning.
- The `create_new + copy` fallback is now entered **only on `ErrorKind::Unsupported`**.
- If the copy or flush fails mid-way, the fallback **deletes the partially created destination before returning**, so the heuristic above can never misfire onto our own debris. Previously that combination silently destroyed the quarantined original credentials (P1).
- The capability probe in `begin()` accepts `Unsupported` (the fallback covers it) while still failing closed for other errors.

### Tests

- New test-only injection (`MidCopyFailureInjection`) forces the copy-fallback path on any filesystem and fails the copy after a configurable prefix:
  - helper level: failure leaves the destination **vacant** and the source intact;
  - transaction level: a failed install does **not** discard the quarantined generation via the exists() heuristic, and a subsequent rollback restores the original auth.json byte-for-byte.
- Existing vacant/occupied lifecycle tests unchanged; the ignored WSL2 contract test `codex_auth_transaction_runs_on_wsl_unc_home` runs this same lifecycle against a real `\\wsl.localhost` home in CI.

### CI

The second WSL2 test wiring is adapted to current main's structure (test binary discovered via `CC_SWITCH_TEST_EXE`, no cargo re-link involved). No `--list`/TEMP reorder is included — that concern is gone from main's job shape and stays out of this PR.

## Test plan

- [x] `cargo fmt --check`, `cargo clippy --all-targets -- -D warnings`
- [x] full `cargo test` (2715 across suites; local-only failures are the known `CC_SWITCH_TEST_HOME` machine-environment set tracked separately)
- [x] new injection tests: 10 passed / 1 ignored (WSL2-only) in the auth subset

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -237,16 +237,23 @@ jobs:
         run: |
           $env:TEMP = $env:CC_SWITCH_WSL_TEST_TEMP
           $env:TMP = $env:CC_SWITCH_WSL_TEST_TEMP
-          $testName = "config::tests::atomic_write_replaces_existing_wsl_unc_file"
+          $testNames = @(
+            "config::tests::atomic_write_replaces_existing_wsl_unc_file",
+            "services::proxy::tests::codex_auth_transaction_runs_on_wsl_unc_home"
+          )
           $testList = & $env:CC_SWITCH_TEST_EXE --list --ignored
           if ($LASTEXITCODE -ne 0) {
             exit $LASTEXITCODE
           }
-          $expected = "${testName}: test"
-          if ($testList -notcontains $expected) {
-            throw "Expected ignored test was not discovered: $testName"
+          foreach ($testName in $testNames) {
+            $expected = "${testName}: test"
+            if ($testList -notcontains $expected) {
+              throw "Expected ignored test was not discovered: $testName"
+            }
           }
-          & $env:CC_SWITCH_TEST_EXE $testName --ignored --exact --nocapture
-          if ($LASTEXITCODE -ne 0) {
-            exit $LASTEXITCODE
+          foreach ($testName in $testNames) {
+            & $env:CC_SWITCH_TEST_EXE $testName --ignored --exact --nocapture
+            if ($LASTEXITCODE -ne 0) {
+              exit $LASTEXITCODE
+            }
           }
--- a/src-tauri/src/services/proxy.rs
+++ b/src-tauri/src/services/proxy.rs
@@ -155,7 +155,7 @@ impl CodexAuthFileTransaction {
                     )
                 })?;
             }
-            Err(error) if error.kind() == std::io::ErrorKind::Unsupported => {
+            Err(error) if Self::hard_link_unsupported(&error) => {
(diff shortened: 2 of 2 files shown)


zayokami:
Root cause of the red WSL2 run: the probe never reached the fallback. std's Windows mapping leaves Win32 `ERROR_NOT_SUPPORTED` (os error 50 — what \\wsl.localhost returns for hard_link creation) **uncategorized** instead of `ErrorKind::Unsupported`, so neither the probe nor install_file_if_vacant's fallback arm fired and begin() failed closed (658960a5).

Hard-link support is now classified by a shared predicate accepting both `ErrorKind::Unsupported` and raw code 50; a unit test pins it against synthesized kinds plus the raw code. The same lifecycle now proceeds into the create_new + copy path on real UNC homes, which is exactly what #6679 needs.

PR #6908, opened 2026-08-27
fix(proxy): stop mask_url panicking on multi-byte UTF-8 near the truncation boundary

xu-kai-quan:
## Summary
- `mask_url`'s fallback branch (used when the proxy URL fails `Url::parse`, e.g. a garbled global-proxy setting) sliced the raw string at a fixed byte offset of 20. If a multi-byte UTF-8 character straddles that offset, `&url[..20]` panics instead of returning a masked string.
- Snap the truncation point to the nearest valid char boundary at or before 20 instead.

## Test plan
- [x] Added `test_mask_url_does_not_panic_on_multibyte_boundary` with a CJK string (`这是一个无效的代理地址不能解析`) that previously panicked with `byte index 20 is not a char boundary`
- [x] `cargo test --lib proxy::http_client::tests` passes (8/8)

### DIFF
--- a/src-tauri/src/proxy/http_client.rs
+++ b/src-tauri/src/proxy/http_client.rs
@@ -325,9 +325,14 @@ pub fn mask_url(url: &str) -> String {
             None => format!("{}://{}", parsed.scheme(), host),
         }
     } else {
-        // URL 解析失败，返回部分内容
+        // URL 解析失败，返回部分内容。截断点回退到最近的字符边界，
+        // 避免在多字节 UTF-8 字符中间切割导致 panic。
         if url.len() > 20 {
-            format!("{}...", &url[..20])
+            let cut = (0..=20)
+                .rev()
+                .find(|&i| url.is_char_boundary(i))
+                .unwrap_or(0);
+            format!("{}...", &url[..cut])
         } else {
             url.to_string()
         }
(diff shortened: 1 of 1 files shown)


farion1231:
@codex review

farion1231 (approved):
Thanks — nice catch. I reproduced it on main through the real call path, not only the unit: `build_client(Some("这是一个无效的代理地址不能解析"))` panics with `end byte index 20 is not a char boundary; it is inside '的' (bytes 18..21)`, because the `map_err` at the `Url::parse` site is the one place `mask_url` is evaluated unconditionally (the `log::debug!` call sites are lazy). With this patch the same input returns `Err("Invalid proxy URL '这是一个无效...': relative URL without a base")`, ASCII input still truncates at 20 as before, and a 4-byte emoji straddling the boundary correctly falls back to 19. I also swept `src-tauri/src` for other fixed-byte-offset slices of user or upstream text — this was the last unguarded one.

Two tiny optional tightenings for the test, fine as a follow-up or left as is:
- `ends_with("...")` also passes for a cut at 0/3/18; `assert_eq!(mask_url(bad), "这是一个无效...")` pins the actual behaviour.
- `assert!(build_client(Some(bad)).is_err())` would cover the path that actually crashed.

Merging. Thanks!

PR #6909, opened 2026-08-27
fix(tests): skip symlink-based tests gracefully without SeCreateSymbolicLinkPrivilege on Windows

xu-kai-quan:
## Summary
- Two tests unconditionally `.expect()`'d `std::os::windows::fs::symlink_dir`, which requires `SeCreateSymbolicLinkPrivilege` (admin or Developer Mode enabled).
- On a standard Windows dev account they panic with OS error 1314 (`ERROR_PRIVILEGE_NOT_HELD`) instead of skipping, unlike the graceful-skip idiom already used elsewhere in this codebase (`session_usage_codex.rs`'s `replay_real_codex_corpus`).

## Test plan
- [x] Reproduced the panic on a non-admin Windows account before the fix (`cargo test --lib`: both tests failed with `symlink: Os { code: 1314, ... }`)
- [x] After the fix, ran `cargo test --lib -- symlink --nocapture`: both tests print `[SKIP] no symlink privilege on this Windows account: ...` and pass

### DIFF
--- a/src-tauri/src/codex_config.rs
+++ b/src-tauri/src/codex_config.rs
@@ -7669,7 +7669,10 @@ model_catalog_json = "cc-switch-model-catalog.json"
         #[cfg(unix)]
         std::os::unix::fs::symlink(&outside_dir, base_dir.join("link")).expect("symlink");
         #[cfg(windows)]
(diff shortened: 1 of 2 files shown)


farion1231:
@codex review

PR #6917, opened 2026-08-27
fix(codex): sanitize xAI native Responses for Grok

### LINKED ISSUE #6815: [Bug] v3.20.0 xAI (Grok) OAuth + Codex: mcp__codex_app__automation_update 仍被拒绝 (root oneOf/anyOf with non-object branch)
### Self Checks / 自检

### CC Switch Version / 版本号

3.20.0

### Operating System / 操作系统

Windows

### Related App / 涉及应用

Codex

### Steps to Reproduce / 重现步骤

1. 使用 CC Switch v3.20.0
2. 在「设置 → OAuth 授权中心」完成 xAI (Grok OAuth) 设备码登录（已重新登录确认）
3. 在 Codex 标签页添加并启用「xAI (Grok) OAuth」托管供应商
4. 开启本地代理服务，并开启 Codex 路由
5. 启动 Codex，发起任意需要工具调用的对话（或正常对话触发内置工具注入）
6. 请求到达 /responses 端点时立即失败

### Expected Behavior / 期望行为

代理应在转发前对工具 schema 做规范化处理，使根节点满足 xAI 严格要求（type: "object"），请求正常到达上游并返回结果。

### Actual Behavior / 实际行为

CC Switch 本地代理在处理 Codex /responses 请求时失败，上游返回 HTTP 400。

完整错误信息：
Provider: xAI (Grok) OAuth; model: grok-4.6; upstream_status: HTTP 400;
cause: Failed to start sampling: [invalid_client_tool_schema] mcp__codex_app__automation_update: tool parameter root must be an object type (root schema is an anyOf/oneOf union with a non-object branch)

### Additional Context / 补充信息

1. 工具名特征：错误中明确出现的是 mcp__codex_app__automation_update（带 mcp__ 前缀），而不是普通的 codex_app__automation_update。

2. Schema 问题本质：该工具的 parameters 根节点是 oneOf / anyOf 联合类型，且包含非 object 分支（常见为 null）。xAI 的 Responses API 要求根必须是 "type": "object"，因此直接拒绝。

3. 已尝试但无效的操作：
   - 升级到最新 v3.20.0
   - 在 OAuth 授权中心重新登录 xAI 账号
   - 确认本地代理已启动 + Codex 路由已开启

4. 历史相关：
   - 类似问题在 DeepSeek、CLIProxyAPI、codex-router 等项目中多次出现
   - v3.18.0 曾对 Responses→Chat 桥接路径做过 schema 归一化（处理 type: null 和部分 oneOf），但当前「原生 Responses + xAI OAuth 托管路径」似乎仍未覆盖 mcp__ 前缀的这种形态

5. 影响：任何携带该工具定义的 Codex 请求都会在采样开始前被 xAI 拒绝，导致带工具调用的会话完全无法使用 Grok。

建议：在 xAI OAuth 的原生 Responses 转发路径中，对所有工具（尤其是名称匹配 *automation_update* 或根 schema 为 oneOf/anyOf 且含非 object 分支的工具）执行与 Responses→Chat 桥接相同的清洗逻辑，将根节点强制规范为 { "type": "object", ... }。

loocor:
## Summary / 概述

Keep Codex Desktop on native `api.x.ai/v1/responses` instead of a Chat/Anthropic bridge, and make Grok requests survive Codex-private payload shapes.

xAI request/response compatibility (isolated in `transform_codex_responses_xai_sanitize.rs`, thin gate in `forwarder`/`handlers`):

- Collapse root `oneOf`/`anyOf` function schemas that xAI rejects (#6815)
- Rewrite collaboration `agent_message` items into ordinary `message` items
- Remap unknown OpenAI role SKUs (`gpt-5.6-sol` / `luna` / `terra`) onto the provider's configured model; keep catalog slugs such as `grok-4.5`
- Rewrite whole-number JSON floats in completed function-call arguments (`92116.0` → `92116`)

macOS production packaging:

- Enable `custom-protocol` so release builds embed `dist` instead of loading Vite at `localhost:3000` (blank window)
- Keep Web Inspector off by default after the white-screen diagnosis

## Related Issue / 关联 Issue

Fixes #6815

## Screenshots / 截图

N/A

## Checklist / 检查清单

- [x] `cargo test --lib transform_codex_responses_xai_sanitize` passes

### DIFF
--- a/src-tauri/Cargo.toml
+++ b/src-tauri/Cargo.toml
@@ -28,7 +28,7 @@ serde_json = { version = "1.0", features = ["preserve_order"] }
 serde = { version = "1.0", features = ["derive"] }
 log = "0.4"
 chrono = { version = "0.4", features = ["serde"] }
-tauri = { version = "2.8.2", features = ["tray-icon", "protocol-asset", "image-png", "devtools"] }
+tauri = { version = "2.8.2", features = ["tray-icon", "protocol-asset", "image-png"] }
 tauri-plugin-log = "2"
 tauri-plugin-opener = "2"
 tauri-plugin-process = "2"
--- a/src-tauri/Cargo.toml
+++ b/src-tauri/Cargo.toml
@@ -18,6 +18,7 @@ doctest = false
 [features]
 default = []
 test-hooks = []
+custom-protocol = ["tauri/custom-protocol"]
 
 [build-dependencies]
 tauri-build = { version = "2.4.0", features = [] }
--- a/src-tauri/Cargo.toml
+++ b/src-tauri/Cargo.toml
@@ -27,7 +27,7 @@ serde_json = { version = "1.0", features = ["preserve_order"] }
 serde = { version = "1.0", features = ["derive"] }
 log = "0.4"
 chrono = { version = "0.4", features = ["serde"] }
-tauri = { version = "2.8.2", features = ["tray-icon", "protocol-asset", "image-png"] }
+tauri = { version = "2.8.2", features = ["tray-icon", "protocol-asset", "image-png", "devtools"] }
 tauri-plugin-log = "2"
 tauri-plugin-opener = "2"
 tauri-plugin-process = "2"
--- a/src-tauri/src/lib.rs
+++ b/src-tauri/src/lib.rs
@@ -389,6 +389,7 @@ pub fn run() {
             {
                 return;
             }
+            #[cfg(target_os = "macos")]
             tray::mark_main_page_ready();
             log::info!("主窗口 page load finished: {}", payload.url());
             if startup_page_handled.swap(true, Ordering::Relaxed) {
--- a/src-tauri/src/lib.rs
+++ b/src-tauri/src/lib.rs
@@ -394,7 +394,7 @@ pub fn run() {
             if startup_page_handled.swap(true, Ordering::Relaxed) {
                 return;
             }
-            if crate::settings::get_settings().silent_startup {
+            if crate::settings::get_settings().silent_startup && !cfg!(debug_assertions) {
                 log::info!("主页面加载完成（静默启动，保持隐藏）");
                 return;
             }
@@ -1332,7 +1332,7 @@ pub fn run() {
                 // 仅 Linux 生效：解决 Wayland 下系统窗口按钮不可用的问题
                 #[cfg(target_os = "linux")]
                 let _ = window.set_decorations(!settings.use_app_window_controls);
-                if settings.silent_startup {
+                if settings.silent_startup && !cfg!(debug_assertions) {
                     // 静默启动模式：保持窗口隐藏
                     let _ = window.hide();
                     #[cfg(target_os = "windows")]
--- a/src-tauri/src/lib.rs
+++ b/src-tauri/src/lib.rs
@@ -376,6 +376,8 @@ pub fn run() {
 
             // Show and focus window regardless
             if let Some(window) = app.get_webview_window("main") {
+                #[cfg(target_os = "macos")]
+                tray::apply_tray_policy(app, true);
                 let _ = window.unminimize();
                 let _ = window.show();
                 let _ = window.set_focus();
@@ -1779,10 +1781,10 @@ pub fn run() {
                         {
                             let _ = window.set_skip_taskbar(false);
                         }
+                        tray::apply_tray_policy(app_handle, true);
                         let _ = window.unminimize();
                         let _ = window.show();
                         let _ = window.set_focus();
-                        tray::apply_tray_policy(app_handle, true);
                     } else if crate::lightweight::is_lightweight_mode() {
                         if let Err(e) = crate::lightweight::exit_lightweight_mode(app_handle) {
                             log::error!("退出轻量模式重建窗口失败: {e}");
--- a/src-tauri/src/lib.rs
+++ b/src-tauri/src/lib.rs
@@ -390,6 +390,7 @@ pub fn run() {
                 return;
             }
             tray::mark_main_page_ready();
+            log::info!("主窗口 page load finished: {}", payload.url());
             if startup_page_handled.swap(true, Ordering::Relaxed) {
                 return;
             }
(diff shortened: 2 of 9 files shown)


laigawa-oss:
所以是解决了么？预计什么时候会发版呢？

farion1231 (commented):
Thanks for this — the four xAI compat fixes target real, reproducible failures on the same Codex Desktop × Grok path (the #6815 root-union rejection, `agent_message` items, subagent SKU model names, whole-float tool arguments), and the test coverage is solid, including idempotency locks. Widening the gate to API-key cards pointing at `api.x.ai` with `wire_api = "responses"` is also right — those hit the same strict-parser wall as the OAuth path. I merged this branch onto current `main` locally and the full Rust suite passes (2754 lib + 9 integration).

That said, a few things before this can merge:

**[P1] Please split the macOS window/packaging commits into a separate PR.**
`edd39bc16` (custom-protocol), `88fcec17c` + `e9656858d` (inspector on/off — these two cancel each other out), `bf943b280` (reload on reveal), and `852e269e6` (activation policy) touch `lib.rs`/`tray.rs`/`Cargo.toml` and are unrelated to the xAI fix. They also change user-visible startup behavior: macOS now defers showing the main window until the first page load, and [`silent_startup && !cfg!(debug_assertions)`](https://github.com/farion1231/cc-switch/blob/cfe4525b664a62fd935b604316528905b9ce537e/src-tauri/src/lib.rs#L1336) silently changes silent-startup semantics in dev builds on Windows too. That needs its own review and testing, not a ride-along.

On the packaging motivation specifically: the release pipeline builds via `pnpm tauri build` (`release.yml`), where the Tauri CLI enables `custom-protocol` automatically — shipped v3.20.x DMGs embed `dist` and don't load Vite. A blank window pointing at `localhost:3000` is what you get from a direct `cargo build --release` outside the CLI, so the `Cargo.toml` feature is a no-op for the official pipeline. The hidden-window WKWebView reload idea may well have value on its own — happy to look at it in a dedicated PR with repro notes.

**[P2] `required` should be intersected, not unioned, when collapsing union branches.**
In [`flatten_union_branches_to_object`](https://github.com/farion1231/cc-switch/blob/cfe4525b664a62fd935b604316528905b9ce537e/src-tauri/src/proxy/providers/transform_codex_responses_xai_sanitize.rs#L402-L445), merging the branches' `required` lists as a union turns "needs `id` OR needs `slug`" into "needs both", which tightens the schema and pushes the model into emitting invalid calls for tools with nullable unions. Keeping only the keys required by *every* object branch is the conservative direction:

```rust
// required: keep only keys present in every branch's `required`
let merged_required: Vec<Value> = first_required
    .iter()
    .filter(|item| rest.iter().all(|branch| branch_requires(branch, item)))
    .cloned()
    .collect();
```

(Exact shape up to you — the point is intersection, not union.) Please fix in this PR; it's a few lines next to code you're already touching.

**[P2, follow-up is fine] `automation_update` loses its whole parameter schema.**
The special case at [L545](https://github.com/farion1231/cc-switch/blob/cfe4525b664a62fd935b604316528905b9ce537e/src-tauri/src/proxy/providers/transform_codex_responses_xai_sanitize.rs#L545) replaces `parameters` with an empty `additionalProperties: true` object, so Grok can call the tool but gets no guidance on its fields. To be clear, this is still strictly better than the status quo (a hard 400 on every turn, or dropping the tool group entirely as the user-side shim in #6815 does) — and I understand why: the real schema's union branches are `$ref`s, which the generic collapse can't classify either. The proper fix is a shallow `$defs`/`$ref` resolve + lifting the object branch with its properties preserved (the Python shim linked in #6815 shows the approach). Fine as a follow-up issue if you'd rather keep this PR focused.

**[P3] Run the model rewrite before the model-conditional sanitize.**
[`apply_xai_native_responses_request_compat`](https://github.com/farion1231/cc-switch/blob/cfe4525b664a62fd935b604316528905b9ce537e/src-tauri/src/proxy/providers/transform_codex_responses_xai_sanitize.rs#L628-L650) sanitizes first and rewrites `model` last, but [`request_targets_grok_45`](https://github.com/farion1231/cc-switch/blob/cfe4525b664a62fd935b604316528905b9ce537e/src-tauri/src/proxy/providers/transform_codex_responses_xai_sanitize.rs#L126-L135) reads the pre-rewrite `model`, so a `gpt-5.6-sol` request rewritten to `grok-4.5` skips the grok-4.5 field strips. In practice Codex's Responses requests don't carry those Chat-only sampling fields, so this is theoretical today — but reordering the calls is ~3 lines and makes the pipeline "decide model, then sanitize for it".

**[P3] Consider letting unknown `grok-*` request models through.**
`rewrite_xai_unknown_request_model` forces anything outside the catalog onto the configured model. On an API-key card without a `modelCatalog`, a user who deliberately picks a real xAI model in Codex gets silently remapped with only a log line. Allowing request models with a `grok` prefix (case-insensitive) alongside the catalog would keep the SKU remap (`gpt-5.6-sol`/`luna`/`terra`) working without overriding intentional choices.

**[P3] Delete the superseded namespace-restore stream instead of `#[allow(dead_code)]`.**
`create_namespace_restore_sse_stream` / `restore_sse_block` in `transform_codex_responses_namespace.rs` are fully replaced by `create_xai_native_responses_sse_stream`; keeping them behind allow-attributes just leaves a second copy to drift.

**To merge:** split out the five macOS commits (P1) and fix the `required` intersection (first P2). The `$ref`-aware `automation_update` schema and the P3s are welcome here too but fine as follow-ups. The xAI side is close — happy to re-review quickly once it's split.

loocor:
Thanks for the careful review. The notes are right.

I don't have the bandwidth to keep following this PR. If any of it still has some reference value, that is already more than enough.

Thank you again for building CC Switch and sharing such a useful tool with everyone.

farion1231:
Thank you so much for all the work you've put into this — the fix chain you built is genuinely valuable. Totally understand about the maintenance bandwidth. I'll continue the remaining fixes on top of your branch, and your work will be credited in the release notes. 🙏

farion1231:
The fix chain from this PR has landed on main (d8065cc62...054673e0b): all six of your commits are merged as-is with your authorship preserved in the history (b7da894b3 through 527b56f8b), followed by three review follow-up commits. Since the changes reached main directly rather than through this PR's merge button, I'm closing it — but this is your fix, and the release notes for the next version will credit you accordingly.

Thanks again for the investigation and the clean fix chain. 🙏

PR #6929, opened 2026-08-28
refactor: use shared live configuration core

SaladDay:
## What

- Use `cc-switch-core` for the existing Claude live-settings metadata cleanup.
- Use `cc-switch-core` for the existing Codex `auth` and `config` snapshot preparation.
- Pin the same core revision used by cc-switch-lite.

The host keeps its current error translation and write orchestration.

## Boundary

This PR does not change `AppType`, file I/O, provider policy, proxy behavior, the database, or the UI.

Those exclusions are deliberate. The full app has a public `AppType::FromStr` error contract and Windows/Unix file behavior that do not match the current core APIs. Moving them in this PR would change behavior.

## Validation

- `cargo test --locked`
- `cargo clippy --locked --all-targets -- -D warnings`
- Two independent blind reviews after the compatibility rollback, followed by one final blind review of the shared revision pin

The final diff is limited to `Cargo.toml`, `Cargo.lock`, and `services/provider/live.rs`.

### DIFF
--- a/src-tauri/Cargo.lock
+++ b/src-tauri/Cargo.lock
@@ -768,6 +768,7 @@ dependencies = [
  "base64 0.22.1",
  "brotli 7.0.0",
  "bytes",
+ "cc-switch-core",
  "chrono",
  "dirs 5.0.1",
  "flate2",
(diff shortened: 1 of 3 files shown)


farion1231:
@codex review

PR #6955, opened 2026-08-29
fix(proxy): route Claude Copilot models by endpoint capability

### LINKED ISSUE #6954: bug(proxy): Claude Copilot routing ignores supported endpoints
## Bug description

Claude and Claude Desktop requests routed through a managed GitHub Copilot provider choose the upstream wire format from the model vendor. This fails for Copilot models whose vendor is not OpenAI but whose live model metadata declares the Responses API.

For example, the current Copilot model catalog includes a model with metadata equivalent to:

```json
{
  "id": "grok-4.6",
  "vendor": "xAI",
  "supported_endpoints": ["/responses"]
}
```

CC Switch 3.20.1 routes that model to `/chat/completions`, which Copilot rejects with:

```text
model "grok-4.6" is not accessible via the /chat/completions endpoint
```

## Steps to reproduce

1. Add a Claude Desktop GitHub Copilot provider using a managed account.
2. Configure one of the Claude Desktop route IDs to use a Copilot model that advertises `/responses` but has a non-OpenAI vendor.
3. Send a request through the Claude Desktop proxy gateway.
4. Observe that the upstream request uses `/chat/completions` and fails with `unsupported_api_for_model`.

## Expected behavior

The final mapped model's `supported_endpoints` metadata should take precedence over vendor heuristics. A model declaring `/responses` should use the Responses request conversion and `/v1/responses` route.

When older or cached model metadata has no `supported_endpoints`, the existing OpenAI-vendor fallback should remain unchanged.

## Root cause

`resolve_claude_api_format` currently calls a vendor-only predicate. The Copilot `/models` response is parsed without preserving `supported_endpoints`, so the Claude route cannot make a capability-based decision.

## Related work

- #1681 fixed the equivalent unsupported endpoint symptom for GPT-family models, but vendor routing still misses non-OpenAI Responses models.
- #5336 introduces capability-based Copilot routing for the Codex integration. This report is specifically about the existing Claude and Claude Desktop path.

## Environment

- CC Switch 3.20.1
- Claude Desktop proxy mode
- GitHub Copilot managed account provider

hueifeng:
## Summary / 概述

Use GitHub Copilot model endpoint capabilities when choosing the Claude/Claude Desktop upstream wire format.

Copilot's `/models` response can advertise a non-OpenAI model with `supported_endpoints: ["/responses"]`. The existing Claude path only checks `vendor == OpenAI`, so such a model is sent to `/chat/completions` and rejected with `unsupported_api_for_model`.

This PR:

- preserves `supported_endpoints` in `CopilotModel`;
- selects Responses when the final model explicitly supports `/responses`;
- keeps the existing OpenAI-vendor behavior only as a backward-compatible fallback when endpoint metadata is absent;
- keeps model selection and Claude Desktop route mapping configuration-driven—there is no model-specific production mapping;
- adds regression coverage for a non-OpenAI Responses-only model, legacy metadata, and an OpenAI chat-only model.

This intentionally limits the change to the existing Claude/Claude Desktop integration. #5336 contains broader capability-based routing work for the Codex integration.

## Related Issue / 关联 Issue

Fixes #6954

Related: #1681, #5336

## Screenshots / 截图

Not applicable; backend proxy routing only.

## Verification

- `pnpm typecheck`
- `pnpm format:check`
- `cargo fmt --check`
- `cargo test --lib proxy::providers::copilot_auth::tests::`
- `cargo test --lib proxy::providers::copilot_model_map::tests::`
- `cargo test --lib proxy::forwarder::tests::`
- Live Claude Desktop proxy request with a configured route targeting a non-OpenAI `/responses`-only model: HTTP 200 via `/v1/responses`
- Existing OpenAI Responses route: HTTP 200 via `/v1/responses`

`cargo clippy --all-targets --all-features -- -D warnings` was also attempted. It is currently blocked by a pre-existing `clippy::redundant_iter_cloned` warning in `src/database/backup.rs:2496`, outside this diff.

## Checklist / 检查清单

- [x] Updated i18n files if user-facing text changed / No user-facing text changed

### DIFF
--- a/src-tauri/src/proxy/forwarder.rs
+++ b/src-tauri/src/proxy/forwarder.rs
@@ -2664,10 +2664,7 @@ impl RequestForwarder {
 
         let model = body.get("model").and_then(|value| value.as_str());
         if let Some(model_id) = model {
-            if self
-                .is_copilot_openai_vendor_model(provider, model_id)
-                .await
-            {
+            if self.is_copilot_responses_model(provider, model_id).await {
                 return "openai_responses".to_string();
             }
         }
@@ -2718,7 +2715,7 @@ impl RequestForwarder {
         }
     }
 
-    async fn is_copilot_openai_vendor_model(&self, provider: &Provider, model_id: &str) -> bool {
+    async fn is_copilot_responses_model(&self, provider: &Provider, model_id: &str) -> bool {
         let Some(app_handle) = &self.app_handle else {
             log::debug!("[Copilot] AppHandle unavailable, fallback to chat/completions");
             return false;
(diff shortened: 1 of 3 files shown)


farion1231:
@codex review

PR #6977, opened 2026-08-29
fix(proxy): support versioned OpenAI-compatible base URLs

czhmartinez:
## Summary
- Claude-to-OpenAI Chat conversion no longer forces `/v1` when the configured base URL already ends in a numeric version segment such as Zhipu's `/api/paas/v4`.
- This prevents requests from being built as `/api/paas/v4/v1/chat/completions` and now targets `/api/paas/v4/chat/completions`.
- Non-versioned base URLs keep the existing `/v1/chat/completions` behavior.

## Testing
- `cargo test -p cc-switch rewrite_claude_transform_endpoint --lib`
- `cargo test -p cc-switch versioned_base_detection_matches_numeric_versions_only --lib`

Verified against Zhipu pay-as-you-go: `/api/paas/v4/chat/completions` returns HTTP 200 with `glm-5.3-flash`.

### DIFF
--- a/src-tauri/src/proxy/forwarder.rs
+++ b/src-tauri/src/proxy/forwarder.rs
@@ -1460,7 +1460,13 @@ impl RequestForwarder {
             let api_format = resolved_claude_api_format
                 .as_deref()
                 .unwrap_or_else(|| super::providers::get_claude_api_format(provider));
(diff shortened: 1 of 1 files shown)


farion1231:
@codex review

PR #6519, opened 2026-08-17
feat(cursor, shared-memory): manage Cursor app with shared skills & MCP, add shared memory panel

1716775457damn:
## Summary

Two features, split into two commits:

### 1. Cursor management (`db66e3a`)
- Manage the Cursor app alongside CC Switch, letting Cursor use the shared skills and MCP configurations.
- Adds Cursor CLI/config handling, settings, types, i18n entries, and tests.

### 2. Shared memory panel (`df11465`)
- Adds a cross-platform shared memory panel as a unified memory channel.
- Includes frontend panel, cloud fetch/push commands, vitest config fix for `realpath` environments, and tests.

## Verification
- `cargo check --lib` (MSVC toolchain) passes
- `tsc --noEmit` passes
- Related vitest suites pass (92/92)

## Notes
- Based on `4080a8e`; upstream commits `4080a8e..d4fefef` were compared and do not overlap with these changes.

PR #7001, opened 2026-08-31
feat(shared-memory): add shared memory panel as cross-platform memory channel

1716775457damn:
﻿## Summary

Adds a cross-platform shared memory panel as a unified memory channel.

- Frontend panel (`SharedMemoryPanel`) with settings, fetch, and push
- Backend commands (`shared_memory_*`), settings persistence, Cloudflare Worker protocol (GET/PUT `{url}/api`)
- i18n entries (en/ja/zh/zh-TW), vitest config fix for `realpath` environments
- Codex review fix: invoke arg renamed to `incoming` to match the Rust command parameter

## Verification
- `cargo check --lib` (MSVC toolchain) passes
- `cargo fmt --check` passes
- `tsc --noEmit` passes
- Prettier check passes on all changed frontend files

## Notes
- Split from #6519 into focused per-feature PRs per contribution policy
- Based on `origin/main` `d8065cc6`


farion1231:
@codex review

1716775457damn:
Thanks for the review — all three findings are valid, I'll fix them in this PR:

1. `shared_memory_save_settings` replaces the whole struct, so saving only `url`/`token` drops the backend-managed `enabled` / `lastSyncAt` / `lastSyncBytes`. I'll carry over the currently loaded values in the save payload so the panel keeps showing the real sync status after reconnecting.
2. Pull returns without persisting `updatedAt`/`bytes` while Push does, so a pull-only session shows a stale push timestamp (or "Never updated") after remount. I'll persist the fetched snapshot metadata before returning.
3. Concurrent pull + push can upload the pre-pull editor buffer and then get overwritten by the pending pull response. I'll disable both transfer actions while either request is in flight.

Pushing the fixes shortly.

1716775457damn:
CI note for run #4684: `Frontend Checks` is the only failing job, on `tests/config/codexReasoningLevelPresets.test.ts:121` — the xAI (Grok) preset now exposes an extra `xhigh` reasoning level while the test still asserts exactly `['low', 'medium', 'high']`. Nothing in the shared-memory panel touches the Codex preset config, and the ubuntu/windows/macos `Backend Checks` jobs are all green, so this looks like a pre-existing test-vs-config mismatch on `main` rather than a regression from these commits.

The identical failure also shows up on #7015, so it will probably block every PR against current `main` until one of the two sides is aligned. I can send a small standalone PR to reconcile the preset and the test if that's preferable to bundling the fix into the feature PRs.

1716775457damn:
Two follow-ups on this PR:

**1. The main-branch test breakage.** Since the same Frontend Checks failure on `tests/config/codexReasoningLevelPresets.test.ts:121` blocks every PR against current main, I'll send a small standalone PR that updates the expectation to `['low', 'medium', 'high', 'xhigh']`. Exposing `xhigh` reads as intended behavior on the preset side (it's in the shipped config, not something these PRs introduce), so aligning the test is the correct direction — and it keeps #7001 and #7015 scoped to their own features instead of each carrying an unrelated test fix.

**2. Concurrency handling for the shared-memory transfers.** The save-payload and pull-metadata fixes are straightforward and I'm including them here. For the pull/push overlap, disabling both actions while a request is in flight covers the single-window case, but not a second window or a second device pointed at the same Worker endpoint — in that scenario the stale-buffer overwrite still happens, just without a visible in-flight flag. If the Worker can support it, sending the last-known `updatedAt` with the PUT and rejecting writes older than the stored snapshot would close that properly. If there's no versioning on the Worker yet, I'd rather state that concurrent writers are unguarded in the panel than leave it looking protected when it isn't.

1716775457damn:
Checked all three Codex findings against `892b203` — they're fixed:

- `shared_memory_save_settings` carries `enabled`, `last_sync_at` and `last_sync_bytes` over from `existing` before writing, so a url/token-only save no longer resets the status line to "Never updated".
- `shared_memory_fetch` persists `updated_at`/`bytes` before returning (guarded by `is_some()`), so a pull-only session now survives a remount.
- Pull and Push are mutually exclusive — both buttons are `disabled={fetching || pushing || !settingsLoaded}`, which closes the single-window race.

Two things I'd still change:

1. The carry-over is unconditional: `incoming.enabled = existing.enabled` means no caller can ever flip `enabled` through this command. Nothing tries to today, so it's harmless — but it makes a field that looks writable in the payload silently read-only. Either take `enabled` from `incoming` when the caller supplies it, or narrow the command/type so the contract says so explicitly.

2. `handleFetch` does `setContent(data.content ?? "")` with no guard, so a Pull silently discards un-pushed edits in the editor. The local buffer and the cloud snapshot are different things and this throws away the local one without a prompt — the button-disabling fixes concurrent requests, not this. Suggest tracking a dirty flag on `content` and requiring an explicit confirm before a Pull overwrites a modified buffer.

On CI run #4684: same root cause as #4678 on #7015 — `codexReasoningLevelPresets.test.ts:121` asserting `['low','medium','high']` against a preset that now ships `xhigh`. All 7 checks are green on the force-pushed head; I'll send the expectation fix as a standalone PR so neither feature PR has to carry an unrelated test change.

Cross-window / cross-device stale writes remain unguarded, as noted in my previous comment — nothing here claims otherwise, and I'd rather have that stated than implied.

1716775457damn:
Writing down a concrete shape for the two open items so they're actionable instead of reminders:

1. `enabled` carry-over — the fix I'd go with is `enabled: incoming.enabled.unwrap_or(existing.enabled)` rather than always taking `existing`. That keeps the current behavior for callers that don't set it (nothing does today, so no behavior change), while a caller that explicitly passes the field can actually flip it. Paired with making `enabled` optional in the TypeScript payload type, the contract then says "omit to keep, pass to change" instead of a field that looks writable and isn't.

2. Pull over a dirty buffer — track `dirty = content !== lastPulledContent` and require an explicit confirm before a Pull replaces modified content. Cheapest safe version: if `dirty`, show a confirm dialog naming the discard ("Pull will replace N unsaved edits"). Slightly better: offer "Save as draft" in that dialog so nothing is lost by accident. Note this is independent of the button-disabling, which only addresses concurrent in-flight requests.

3. On the Worker concurrency question — if there's no `updatedAt` versioning on the PUT endpoint yet, I'd rather the panel state that concurrent writers are unguarded (tooltip or a small note next to Sync) than disable the buttons and imply protection that doesn't exist for a second window or device.

PR #7010, opened 2026-08-31
fix(grokbuild): point session_summary at the selected model profile

### LINKED ISSUE #7003: bug(grokbuild): imported providers omit the session title model
### Self Checks / 自检

- [x] Read the README FAQ and contribution guidelines.
- [x] Searched open and closed issues/PRs for `session_summary` and Grok title configuration.

### CC Switch Version / 版本号

3.20.1; source checked against `main` at `054673e0b8cca7d8bd8c67658f2c28e63b2855e0`.

### Operating System / 操作系统

macOS. Grok Build CLI 1.0.13. No Windows/Linux desktop UI claim.

### Related App / 涉及应用

Grok Build with a custom/BYOK provider.

### Steps to Reproduce / 重现步骤

1. Add/import a Grok Build custom provider with a non-default model ID, e.g. `custom-model`, a provider endpoint and its API key.
2. Inspect the generated TOML. Both `build_grokbuild_settings` (deep-link import) and `buildGrokBuildConfig` (form) write `[models].default`, but omit `[models].session_summary`.
3. Start Grok Build with that configuration and make a short request. The main conversation can use the custom model while the separate session-title request falls back to the client's default model (observed as `grok-4.6` in CLI 1.0.13).

The generated shape is currently:

```toml
[models]
default = "custom-model"

[model."custom-model"]
model = "custom-model"
base_url = "https://relay.example.invalid/v1"
name = "Custom"
api_key = "not-a-real-key"
api_backend = "responses"
context_window = 500000
```

### Expected Behavior / 期望行为

Newly generated custom-provider configurations should also point `models.session_summary` at the selected model profile, so generating a session title does not require an unrelated default model/provider. An explicitly configured, separate summary model should remain unchanged when editing a provider.

### Actual Behavior / 实际行为

Main conversation success does not imply that title generation works: the auxiliary request can use the fallback model instead. Where that model/provider is unavailable, only title generation may fail. This report is about client configuration, not a claim that all normal chat requests fail.

### Additional Context / 补充信息

Grok documents `models.session_summary` separately from `models.default`: https://docs.x.ai/build/settings/reference#models . The installed CLI configuration reference describes it as the model for session titles and summaries.

Proposed small fix: set the summary profile in the two existing configuration builders; preserve a separately selected summary profile, and update the reference if it followed a default profile being renamed. No new UI option, dependency, provider protocol changes, credential-resolution changes, or silent API retries are needed. I can attach regression tests and a focused PR.

ntdatt812:
Fixes #7003.

Grok Build resolves `[models].session_summary` separately from `[models].default`. Neither generator wrote it, so a configured custom provider answered the conversation while the session-title request went to the client's own fallback model — observed as `grok-4.6` on CLI 1.0.13. Where that fallback model or provider is not reachable, *only* title generation fails, which is hard to connect back to the provider that was just set up.

Both builders now set it:

- `build_grokbuild_settings` (deep-link import, `src-tauri/src/deeplink/provider.rs`) — a fresh config, so it simply emits the line
- `buildGrokBuildConfig` / `updateGrokBuildConfig` (form, `src/utils/grokBuildConfig.ts`) — has to respect what is already there

## The editing case

Two rules, both in `resolveSessionSummaryProfile`:

- **A summary profile the user pointed somewhere else is left alone.** Someone who deliberately routes titles to a cheaper model should keep that through an edit.
- **One that was tracking the default profile follows it through a rename.** Otherwise the rename strands `session_summary` on a `[model.*]` table that no longer exists, which is worse than never having set it.

## Tests

**Frontend** — `npx vitest run src/utils/grokBuildConfig.test.ts` → **8 passed** (5 existing, 3 new):

- a generated config points its summary at the selected profile
- an explicitly different summary profile survives an edit
- a summary that tracked the default follows a rename

Mutation-checked: dropping the `current === previousProfile` branch fails `follows the default profile through a rename when it was tracking it`, 7 passed / 1 failed.

**Rust** — a test asserting the deep-link import emits both `default` and `session_summary`, placed next to the existing `grokbuild_deeplink_never_carries_env_key_or_resolved_secret`.

One disclosure: I could type-check it but not run it here. `cargo check --lib --profile test` is clean, so the test compiles; linking the test binary on this machine needs a GNU toolchain workaround (`aws-lc-sys` wants a `nanosleep64` shim, and `build.rs` emits MSVC manifest flags that GNU `ld` rejects) that I did not want to carry anywhere near a commit. The `build.rs` I patched locally to get that far is restored and not in this diff. CI runs it properly.

### DIFF
--- a/src-tauri/src/deeplink/provider.rs
+++ b/src-tauri/src/deeplink/provider.rs
@@ -486,7 +486,10 @@ fn build_grokbuild_settings(request: &DeepLinkImportRequest) -> serde_json::Valu
 
     json!({
         "config": format!(
-            "[models]\ndefault = {model_value}\n\n[model.{model_value}]\nmodel = {model_value}\nbase_url = {endpoint_value}\nname = {name_value}\napi_key = {api_key_value}\napi_backend = \"{}\"\ncontext_window = {}\n",
+            // session_summary as well as default: Grok Build resolves the
+            // session-title model separately, so an import that set only
+            // `default` left titles going to the client's fallback model.
+            "[models]\ndefault = {model_value}\nsession_summary = {model_value}\n\n[model.{model_value}]\nmodel = {model_value}\nbase_url = {endpoint_value}\nname = {name_value}\napi_key = {api_key_value}\napi_backend = \"{}\"\ncontext_window = {}\n",
             crate::grok_config::DEFAULT_API_BACKEND,
             crate::grok_config::DEFAULT_CONTEXT_WINDOW,
         )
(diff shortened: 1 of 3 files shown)


farion1231:
@codex review

PR #7011, opened 2026-08-31
feat(presets): add Tencent Token Plan presets across six apps

zmq1121:
## Summary / 概述

Add Tencent Token Plan provider presets: 6 products × 6 apps (Claude, Claude Desktop, Codex, OpenCode, Hermes, OpenClaw) = 36 presets.

**Products** (domestic / intl for each):
- Token Plan (personal) — `api.lkeap.cloud.tencent.com/plan`, Auto id `tc-code-latest`
- Token Plan Enterprise Pro — `tokenhub.tencentmaas.com/plan`, 17 models
- Token Plan Enterprise Lite — `tokenhub.tencentmaas.com/plan`, Auto only
- Intl variants use `tokenhub-intl.tencentcloudmaas.com/plan` (Singapore), Auto id `auto`

Key facts baked into the presets:

- Subscription line ≠ TokenHub marketplace (Hunyuan preset): different keys, different endpoints, never mixed as candidates
- Keys are site-scoped: domestic and intl endpoints do not interoperate (verified by live calls — all cross-site requests fail 401)
- Auto routing id is asymmetric: domestic personal uses `tc-code-latest`, intl personal uses `auto`
- Codex presets use `openai_chat` local routing with `codexChatReasoning` (thinking + reasoning_effort both effective on /plan endpoints, verified with real keys): glm-5.3 → low/high/max strict enum (medium/xhigh → 400), kimi-k2.7-code → enabled only, minimax-m2.5/m2.7 + domestic auto ignore thinking-off, intl auto honors it
- `modelsUrl` set only on the domestic personal Claude preset (/plan/v3/models works there only; 404 on enterprise/intl endpoints)
- minimax-m2.5 removed from doc tables but still served → included where the plan covers it; kimi-k2.5 retired 8/31 → excluded

All endpoints, model lineups and thinking levels were verified with real API keys on 2026-08-31 (124 live calls across all four endpoints). Lineup sources: 1823/130060, 1823/130659, 1823/131173 + intl 1300/81315, 1300/81489, 1300/81490.

Tests: new `tokenPlanProviderPresets.test.ts` (43 tests) locks endpoints, lineups, key URLs and OpenClaw rebase across six apps; 11 reasoning-level rows registered in `codexReasoningLevelPresets.test.ts`. Full suite 1066/1066, `pnpm typecheck` and `pnpm format:check` pass.

## Checklist / 检查清单


farion1231 (commented):
Thanks for this — the research quality is well above the usual bar for a preset PR. I independently
re-checked all six endpoints against the official docs and every one matches, including the
asymmetric Auto id (`tc-code-latest` on the domestic personal plan vs `auto` everywhere else), and
dropping `kimi-k2.5` ahead of its 2026-08-31 retirement is the right call. The new test file locks
real invariants — lineups, cross-site isolation, per-model reasoning levels — rather than restating
the presets. Merge against `main` is clean, `typecheck` and `format:check` pass, and your 11 new
reasoning-level rows all pass.

Four things to fix before merge.

### [P1] The OpenClaw entries assume no official integration page exists — there are two

[The comment here](https://github.com/farion1231/cc-switch/blob/09cf86b376cc84a6b6094d5979dff8cc85db122e/src/config/openclawProviderPresets.ts#L1821-L1827)
says Tencent hasn't published an OpenClaw page for the subscription line, and derives the numbers
from the platform model-list page instead. Tencent does publish one, for both tiers:
[personal (1823/130062)](https://cloud.tencent.com/document/product/1823/130062) and
[enterprise (1300/81503)](https://intl.cloud.tencent.com/document/product/1300/81503).

This repo treats a vendor's official OpenClaw page as the authoritative shape and copies it verbatim
— see the Qianfan Token Plan entry
[here](https://github.com/farion1231/cc-switch/blob/09cf86b376cc84a6b6094d5979dff8cc85db122e/src/config/openclawProviderPresets.ts#L2171-L2179),
which carries all five fields (`reasoning`, `input`, `cost`, `contextWindow`, `maxTokens`). The
Tencent entries here carry only `id`/`name`/`contextWindow`.

The consequential omission is `maxTokens`. Per
[OpenClaw's model-providers doc](https://github.com/openclaw/openclaw/blob/main/docs/concepts/model-providers.md),
an omitted `maxTokens` defaults to **8192** (omitted `reasoning` defaults to `false`, omitted
`input` to `["text"]`). Tencent's page specifies 32768 for Auto, 384000 for the DeepSeek pair and
131072 for `hy3-preview` — so as written, every model's output is capped at 8192 and long
completions get truncated. The missing `reasoning` also mislabels
[`hy3-preview`](https://github.com/farion1231/cc-switch/blob/09cf86b376cc84a6b6094d5979dff8cc85db122e/src/config/openclawProviderPresets.ts#L1853),
which the official page marks `reasoning: true`.

Official values vs. what's here:

| model | official ctx | official maxTokens | official reasoning | this PR |
|---|---|---|---|---|
| `tc-code-latest` | 196608 | 32768 | false | 200000, both omitted |
| `glm-5` | 202752 | 16384 | false | 200000, both omitted |
| `deepseek-v4-flash-202605` / `-pro-202606` | 1000000 | 384000 | false | 1048576, both omitted |
| `hy3-preview` | 262144 | 131072 | **true** | 256000, both omitted |
| enterprise `auto` | 196608 | 32768 | false | 200000, both omitted |
| enterprise `glm-5.2` | 1048576 | 131072 | false | 1048576, maxTokens omitted |

Fix: copy the five fields verbatim for the models the official pages cover (including `cost` — both
pages state it explicitly as all-zero, which also means the
[`expect(model.cost).toBeUndefined()`](https://github.com/farion1231/cc-switch/blob/09cf86b376cc84a6b6094d5979dff8cc85db122e/tests/config/tokenPlanProviderPresets.test.ts#L315)
assertion needs updating). For the models you include beyond those pages
(`minimax-m2.7`, `glm-5.1`, `glm-5.2`, `hy3`), fill `maxTokens`/`reasoning` from the platform list
and note the boundary in the comment. Please also drop the "no official page" sentence — otherwise
the next maintainer will keep deriving numbers from the platform page.

One thing that page settles in your favour: it lists `minimax-m2.5`, which independently backs your
decision to ship it even though the plan table omits it.

### [P2] The six Auto entries fall back to a 128K context window

`tc-code-latest` and the five `auto` entries are the only catalog rows in the file without
`contextWindow` — e.g.
[the personal one](https://github.com/farion1231/cc-switch/blob/09cf86b376cc84a6b6094d5979dff8cc85db122e/src/config/codexProviderPresets.ts#L1434-L1437)
(the others are at L1535, L1606, L1747, L1846, L1878). The backend fills the gap with
[`spec.context_window.unwrap_or(default_context_window)`](https://github.com/farion1231/cc-switch/blob/09cf86b376cc84a6b6094d5979dff8cc85db122e/src-tauri/src/codex_config.rs#L1402),
and `default_context_window` comes from `model_context_window` in config.toml, defaulting to
[128_000](https://github.com/farion1231/cc-switch/blob/09cf86b376cc84a6b6094d5979dff8cc85db122e/src-tauri/src/codex_config.rs#L2113-L2114)
— which `generateThirdPartyConfig` doesn't emit. So the default model of all six presets ships with
a 128K window.

Every one of the 52 catalog rows on `main` declares `contextWindow`, routing aliases included
(`ark-code-latest` = 256000, `qianfan-code-latest` = 131072). Fix: add
`contextWindow: 196608` to the six Auto rows — that's the value both official OpenClaw pages give
for Auto, so it also lines up with the P1 fix.

### [P2] `glm-5.3` ends up defaulting to `max` reasoning effort

Declaring `["low", "high", "max"]`
([here](https://github.com/farion1231/cc-switch/blob/09cf86b376cc84a6b6094d5979dff8cc85db122e/src/config/codexProviderPresets.ts#L1611-L1615)
and L1752) omits `medium`, which is the ProxyChat template's `default_reasoning_level`
(`src-tauri/src/resources/gpt5_5_template.json`). The resolution order is explicit value → template
default if still supported →
[`canonical.last()`](https://github.com/farion1231/cc-switch/blob/09cf86b376cc84a6b6094d5979dff8cc85db122e/src-tauri/src/codex_config.rs#L1377-L1382),
i.e. the highest declared level. So `medium` drops out and GLM-5.3 lands on `max` by default — the
slowest and most quota-hungry setting, on a plan whose whole point is a fixed token budget.

If `max` really is the vendor default, keep it and say so in the comment. Otherwise add
`defaultReasoningLevel: "high"` to both `glm-5.3` rows (the Kimi Code entries use the same explicit
override for exactly this reason).

### [P2] Domestic-site customers on the Singapore region have no usable preset

[1823/130659](https://cloud.tencent.com/document/product/1823/130659) and
[1823/131173](https://cloud.tencent.com/document/product/1823/131173) both list two regions for the
enterprise plans: Guangzhou `tokenhub.tencentmaas.com` and Singapore
**`tokenhub-intl.tencentmaas.com`** — note that is a different domain from the international site's
`tokenhub-intl.tencentcloudmaas.com`. Someone on the domestic site who opened the Singapore region
can use neither preset: the Guangzhou endpoint is the wrong region, the intl one is the wrong site,
and the docs state cross-region and cross-site calls aren't supported.

The comment reasons this away with "keys are per-site and don't interoperate", which is true but
only covers cross-*site* — it doesn't cover same-site cross-*region*. Either add the Singapore URL as
a second `endpointCandidates` entry on the two domestic enterprise presets
(`claudeProviderPresets.ts` L994/L1039, `codexProviderPresets.ts` L1596/L1842, plus the Desktop
file), or just say in the comment that this region needs the base URL edited by hand. I'd lean
towards the candidate entry, with a note that it requires a Singapore-region subscription.

### nit

The OpenClaw enterprise `auto` rows use `contextWindow: 200000` while `glm-5.3`/`minimax-m3` in the
same preset are 1M; same root cause as above, aligning on the official 196608 covers it. The PR also
carries a one-line comment change in the Hunyuan preset (doc ref `1823/130988` → `136956`) — no
runtime effect, just flagging it as out of scope for the title.

### Not blocking

Several of the new ids (dated DeepSeek variants, `glm-5.3`, `glm-5-turbo`, the Kimi and MiniMax
entries) aren't in `is_confirmed_text_only_model`, so they're advertised as image-capable. That's
pre-existing behaviour rather than something this PR introduces — `main` already has 16 rows in the
same shape, and the registry is deliberately fail-open so a future `-vision` variant isn't blocked
locally. Better fixed by extending the registry in a separate PR than by adding declarations here.
(`auto`/`tc-code-latest` should stay fail-open regardless — they're routing aliases and Tencent does
serve `-vision-exp` models.)

### CI

CI hasn't run yet — it needs a maintainer to approve workflows for a first-time contributor. When it
does, expect two failures in `tests/config/codexReasoningLevelPresets.test.ts` on the xAI `grok-4.5`
rows: those are pre-existing on `main`, unrelated to this PR, and being fixed separately.

Fix the P1 and the three P2s and this is good to merge — happy to re-review quickly once they're in.

zmq1121:
Thanks for the exceptionally thorough review — all four items are fixed in 94309a61 (1069/1069 tests pass locally, +3 new).

**P1** — Rewrote all six OpenClaw entries against the two official pages (1823/130062 personal, 1300/81503 enterprise): the five fields are now copied verbatim, including all-zero cost and the official contextWindow/maxTokens figures (Auto 196608/32768, glm-5 202752/16384, deepseek 1000000/384000 on the personal page vs 1048576/393216 on the enterprise page, hy3-preview 262144/131072 with reasoning:true). Models beyond those pages (minimax-m2.7, glm-5.1, glm-5.2, hy3, kimi-k2.6, glm-5.3, glm-5-turbo, the deepseek GA ids) take maxTokens from the platform list (1300/78934), with the boundary marked per-entry in comments. The "no official page" sentence is gone, and the cost assertion in the test now locks the all-zero shape plus maxTokens/reasoning/input on every row.

**P2 (Auto window)** — All six Auto rows in the Codex catalogs now declare contextWindow: 196608, with a comment citing the OpenClaw pages as the only official source (the platform list has no Auto row). A new test locks it across all six presets.

**P2 (glm-5.3 default)** — Both glm-5.3 rows declare defaultReasoningLevel: "high". For the record: live testing didn't surface any vendor default either way (we only probed explicit values), so "high" is the deliberate pick per your reasoning about quota on a fixed-budget plan, not a documented vendor default. Both rows are registered in the reasoning-level whitelist with the explicit default.

**P2 (Singapore region)** — Added the second endpointCandidates entry on both domestic enterprise presets (Guangzhou primary + tokenhub-intl.tencentmaas.com), with a note that the Singapore entry requires a Singapore-region subscription. One thing worth flagging: 1300/81489 and 81490 list the same two-region structure for the intl enterprise plans (Guangzhou on tokenhub.tencentcloudmaas.com), so I applied the symmetric fix there too — intl enterprise presets now carry Singapore primary + Guangzhou candidate. Happy to revert that half if you'd rather keep the diff scoped to the domestic side; the isolation test was updated to distinguish site domains (tencentmaas vs tencentcloudmaas) from the tokenhub-intl prefix, since the same-site Singapore candidate now matches that prefix.

The out-of-scope Hunyuan comment change (130988 → 136956) is kept — 130988 turned out to be the multimodal nav page with no model list, while 136956 is the actual image-understanding list the comment cites. Separately, glad the enterprise page independently backs minimax-m2.5.

farion1231 (approved):

PR #7015, opened 2026-08-31
feat(cursor): manage Cursor app with shared skills & MCP

1716775457damn:
﻿## Summary

Manage the Cursor app alongside CC Switch, letting Cursor use the shared skills and MCP configurations.

- Adds Cursor CLI/config handling, settings, types, i18n entries, and tests
- Registers Cursor in the unified Skills/MCP panels (enabled_cursor persistence, schema v18 migration)
- Codex review fixes: skills DAO cursor read/write, agents panel import filtering, settings cleanup

## Verification
- `cargo check --lib` (MSVC toolchain) passes
- `cargo fmt --check` passes
- `tsc --noEmit` passes
- Prettier check passes on all changed frontend files

## Notes
- Split from #6519 into focused per-feature PRs per contribution policy
- Based on `origin/main` `d8065cc6`


farion1231:
@codex review

1716775457damn:
CI note on run #4678: the only failing job is `Frontend Checks`, and it is unrelated to the Cursor changes here. `tests/config/codexReasoningLevelPresets.test.ts:121` asserts the xAI (Grok) preset exposes `['low', 'medium', 'high']`, but the preset now also ships an `xhigh` level, so the deep-equal fails. All three matrix `Backend Checks` jobs (ubuntu/windows/macos) passed.

So this looks like a stale frontend test expectation rather than a regression from this PR. Happy to either update the test to expect `['low', 'medium', 'high', 'xhigh']` or drop `xhigh` from the preset, depending on which behavior is intended — let me know which one you'd prefer.

1716775457damn:
Thanks for the Codex review — both findings are valid, I'll fix them:

1. Schema bump: I'll introduce a dedicated v19 migration that adds the `enabled_cursor` column to *both* the MCP and skills tables, instead of relying on `CREATE TABLE IF NOT EXISTS`. You're right that an existing install already at `user_version = 18` would silently skip it, and a v17 DB would only get the skills column via `migrate_v17_to_v18`.
2. `SkillService::import_from_apps`: I'll call `sync_to_app_dir` for each enabled selection right after `db.save_skill`, so a skill that only lives in `~/.agents/skills` is actually written to the selected app's skills directory instead of being marked managed and waiting for an unrelated global sync.

On CI run #4678: the only failing job is Frontend Checks, and it's a stale assertion, not a regression — `tests/config/codexReasoningLevelPresets.test.ts:121` still expects the xAI (Grok) preset to expose `['low', 'medium', 'high']`, but the preset now also ships `xhigh`. All three Backend Checks matrix jobs (ubuntu/windows/macos) pass. My preference is to update the test to `['low', 'medium', 'high', 'xhigh']` since exposing `xhigh` is intended, but tell me if you'd rather drop `xhigh` from the preset instead.

1716775457damn:
Status update on the two open threads here.

**CI run #4678 is resolved.** The failure was confined to Frontend Checks and was a stale expectation, not a regression: `tests/config/codexReasoningLevelPresets.test.ts:121` asserted `['low','medium','high']` for the xAI (Grok) presets while the preset also ships `xhigh`. The expectation now reads `['low','medium','high','xhigh']` for both the API-key and OAuth variants, and all 7 checks — including the three backend matrix jobs — are green on `8fea2f3`. Worth checking whether `main` still carries the stale assertion, otherwise every PR touching that file will hit the same red.

**On the Codex schema finding: I said I'd land a v19 migration, and checking the current head — it isn't there yet.** The migration chain still terminates at v18 (`match` arm `17 => migrate_v17_to_v18`, `set_user_version(18)`), and `mcp_servers.enabled_cursor` is only ever added in `migrate_v0_to_v1`. So a database already at `user_version = 18` never receives that column: `create_tables` won't backfill it, the v17→v18 arm only touches `session_log_sync` and `skills.enabled_cursor`, and no migration re-runs. On such an install the MCP DAO's `enabled_cursor` reads/writes fail with *no such column*, which is exactly what Codex flagged.

Fix I'll push next: bump `SCHEMA_VERSION` to 19 and add `migrate_v18_to_v19` calling `add_column_if_missing(conn, "mcp_servers", "enabled_cursor", "BOOLEAN NOT NULL DEFAULT 0")`, guarded by `table_exists` like the v18 arm does so databases missing the table fall through to the fresh DDL instead of erroring. Existing rows keep the `0` default, which is the correct "not enabled for Cursor" semantics for pre-Cursor data.

I'll verify the second finding (`SkillService::import_from_apps` calling `sync_to_app_dir` per enabled selection after `db.save_skill`) in the same push.

1716775457damn:
Confirming the schema gap is real, not theoretical: with the chain terminating at v18, an install already at `user_version = 18` never receives `mcp_servers.enabled_cursor`, and the MCP DAO fails with `no such column` on that path. Two additions worth folding into the v19 migration:

- Backfill `enabled_cursor` for existing rows (defaulting from `enabled`) rather than relying on a column default, so a migrated install behaves the same as a fresh `CREATE TABLE` one. Otherwise the two paths silently diverge and the bug only reproduces on upgraded databases.
- Make the migration idempotent via a `PRAGMA table_info(mcp_servers)` check before `ALTER TABLE`, so a re-run or partially-applied DB doesn't abort mid-transaction.

On the test: `['low', 'medium', 'high', 'xhigh']` is the correct expectation assuming `xhigh` is intended for Grok. Agreed it should land against main separately — otherwise every PR touching that preset file keeps hitting the same red.

PR #7044, opened 2026-09-01
fix(proxy): resolve conversation truncation and close dangling content blocks in Responses API

sugu6:
Fixes three critical proxy streaming issues, plus a CI cleanup:

### 1. Close dangling content blocks before stream_truncated

When an upstream Responses API stream is truncated mid-block (web search limit exceeded or stream ended before a terminal event), the proxy now emits synthetic `content_block_stop` events for any open content blocks before the `stream_truncated` error. This prevents downstream clients from seeing "Content block not found" on the next turn.

Applied at three code paths in `streaming_responses.rs`:
- web search limit-exceeded (non-citations path)
- web search limit-exceeded (`preserve_web_search_citations` path)
- generic stream truncation (missing terminal event)

### 2. Resolve conversation truncation in Responses API proxy

- `handlers.rs`: merge the `output` array from `response.completed` with items collected from `output_item.done` events, so output items received before a stream interruption are not lost. Deduplicate by item id and sort by `output_index`.
- `transform_codex_anthropic.rs`: tolerate a missing `index` field in `content_block_delta`/`content_block_stop` events emitted by third-party gateways, falling back to `output_index` then 0 (with a debug log).
- `handlers.rs`: harden `body_looks_like_sse` so prose error bodies like `: Bad Gateway` / `: connection refused` are no longer misclassified as SSE (SSE comment identifiers must be uppercase).

### 3. Claude Desktop: inject cache_control breakpoints

`forwarder.rs`: for Claude Desktop proxy mode with native Anthropic-format upstreams (non-Bedrock), inject `cache_control` breakpoints into the request body so Prompt Caching works in proxy mode.

### 4. Remove redundant `test-proxy-fix.yml`

This workflow duplicated the backend tests already covered by `ci.yml` and failed at startup due to missing `permissions` (0 jobs created). Deleted.

## Test plan

- `ci.yml` passed on all 3 platforms (Ubuntu 22.04, Windows, macOS): fmt, clippy `-D warnings`, and the full `cargo test` suite
- WSL2 filesystem contract test passed on the Windows runner
- The updated `test_streaming_web_search_limit_does_not_finalize_parallel_function` now asserts the dangling function block receives a `content_block_stop` before the `stream_truncated` error and that no `message_stop` is emitted


farion1231:
@codex review

Issue #1876, opened 2026-04-03
[Feature]: Allow syncing configs without skills — respect Git-managed skill directories


XIYBHK:
## Problem

CC Switch's WebDAV sync currently backs up **everything** together — provider configs/database **and** custom skills (`~/.claude/skills/`). There is no way to sync configs without also syncing skills.

This causes problems for users who manage their skills folder separately (e.g., as a **Git repository**).

## What Happened

My `~/.claude/skills/` directory is a GitHub repo. I committed and pushed changes on my work machine, then pulled on my home machine — the git state was fully in sync.

However, CC Switch **also synced the skills folder** via WebDAV, **overwriting files in the working directory** with whatever state existed on the source machine at sync time. This included **uncommitted working directory changes** from the work machine, resulting in unexpected `git diff` on the home machine despite the git repo being clean.

In short: CC Switch's file-level sync conflicts with git's version control, because it operates on filesystem state rather than respecting repository boundaries.

## Expected Behavior

Users should be able to choose **what** gets synced. At minimum:

- **Option A**: Sync configs only (provider settings, database) — **without** syncing skills
- **Option B**: Sync everything (current behavior)

## Suggested Implementation

Add a configuration option, for example:

```json
{
  "sync": {
    "configs": true,
    "skills": false
  }
}
```

Or provide a `.ccswitchignore` / exclude list so users can opt out specific directories like `skills/`.

## Why This Matters

- Users who version-control their skills via Git will have their working directory silently modified
- This can cause phantom diffs, merge conflicts, or loss of uncommitted work
- The current all-or-nothing approach forces users to choose between CC Switch config sync and Git-based skill management — they should be able to use both

PR #7082, opened 2026-09-03
feat(settings): add provider-only management scope

hehuaiyu:
## Summary

- Add independent management switches for MCP, Skills, and Sessions, plus a Manage providers only preset.
- Keep all switches enabled by default for backward compatibility.
- Hide disabled resource sections and reject their scans, imports, deep links, background tasks, and mutation commands.
- Preserve unmanaged MCP and Skills data during provider rewrites and profile autosave.
- Require every resource manager to be enabled for atomic full-data sync, import, and restore because protocol v2 stores db.sql and skills.zip as one snapshot.

## Related Issue

Related to #1876, which describes the risk of full-data sync overwriting independently managed Skills. This PR prevents unmanaged resources from participating in the atomic snapshot instead of introducing a partial-sync protocol.

## Screenshots

Not included. The settings control and hidden resource states are covered by component tests.

## Testing

- pnpm typecheck
- pnpm format:check
- pnpm build:renderer
- pnpm exec vitest run tests/components/ManagementScopeSettings.test.tsx tests/components/UsageDashboard.test.tsx (7 passed)
- Full frontend suite: 1074 passed; 2 pre-existing timeouts remain in tests/integration/App.test.tsx
- cargo fmt --check
- cargo clippy --all-targets
- Focused Rust management-scope and profile-preservation tests passed before the final Pi command guard; the post-guard all-target Clippy build passed. A later Cargo test retry was stopped because Cargo linked all Windows integration-test targets despite the name filter.

## Checklist

- [x] pnpm typecheck passes
- [x] pnpm format:check passes
- [x] cargo clippy passes
- [x] Updated i18n files for user-facing text

### DIFF
--- a/src-tauri/src/codex_config.rs
+++ b/src-tauri/src/codex_config.rs
@@ -972,6 +972,15 @@ pub fn delete_codex_provider_config(
     Ok(())
 }
 
+fn preserve_live_mcp_when_unmanaged(incoming: &str) -> Result<String, AppError> {
+    if crate::settings::mcp_management_enabled() {
+        return Ok(incoming.to_string());
+    }
+
+    let current = read_codex_config_text()?;
+    crate::mcp::replace_toml_mcp_sections(&current, incoming)
+}
+
 /// 原子写 Codex 的 `auth.json` 与 `config.toml`，在第二步失败时回滚第一步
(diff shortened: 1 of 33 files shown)


farion1231:
@codex review

PR #7097, opened 2026-09-03
fix(terminal): respect default profile and portable mode in Windows Terminal launch (#5322)

### LINKED ISSUE #5322: [Bug] v3.16.5–v3.17.0: Windows Terminal ignores the default profile and hardcodes cmd
> **Update — reproduced on v3.17.0 (2026-07-14):** The v3.17.0 release tag still contains the same two `wt cmd /K` launch paths. The issue therefore affects both v3.16.5 and v3.17.0. A separate `0x80070002` launch failure observed on v3.17.0 is documented in the latest comment below without assuming that v3.17.0 introduced it.
### Self Checks / 自检

- [x] I searched existing open and closed issues.
- [x] I verified the behavior against the v3.16.5 tag and current `main` source.

### CC Switch Version / 版本号

v3.16.5 and v3.17.0 and v3.17.0

### Operating System / 操作系统

Windows 11, build 26200.8655

### Related App / 涉及应用

Claude Code

### Summary / 问题摘要

The problem reported in #2830 still exists in the latest release.

When **Preferred Terminal** is set to **Windows Terminal**, CC Switch always opens a Windows Terminal tab running `cmd.exe`, even when the user's Windows Terminal default profile is PowerShell 7.

在 CC Switch 中选择 Windows Terminal 后，程序仍然固定启动 CMD，而不是使用 Windows Terminal 中设置的默认 Profile（本机为 PowerShell 7）。

### Steps to Reproduce / 复现步骤

1. Install CC Switch v3.16.5 on Windows.
2. Set the Windows Terminal default profile to PowerShell 7.
3. In CC Switch, set **Preferred Terminal** to **Windows Terminal**.
4. Click **Open in Terminal** on a Claude provider card.
5. Exit Claude Code or inspect the process tree/shell.

### Expected Behavior / 预期行为

Windows Terminal should start its configured default profile.

With PowerShell 7 configured as the default profile, the process chain should be equivalent to:

```text
WindowsTerminal.exe -> pwsh.exe -> claude
```

If CC Switch itself is elevated, the separate Windows Terminal window and its PowerShell/Claude child processes should inherit that elevated token. If CC Switch is not elevated, they should remain unelevated.

### Actual Behavior / 实际行为

The process chain is effectively:

```text
WindowsTerminal.exe -> cmd.exe -> claude
```

The Windows Terminal default profile setting is ignored.

### Source-level Cause / 源码原因

The v3.16.5 tag, v3.17.0 tag, and current `main` still explicitly pass `cmd /K` to `wt`:

```rust
"wt" => run_windows_start_command(
    &["wt", "cmd", "/K", &bat_path],
    "Windows Terminal",
),
```

The same behavior appears in two launch paths:

- https://github.com/farion1231/cc-switch/blob/v3.17.0/src-tauri/src/commands/misc.rs#L3238
- https://github.com/farion1231/cc-switch/blob/v3.17.0/src-tauri/src/commands/misc.rs#L3475

Because `cmd` is explicitly supplied as the Windows Terminal commandline, `defaultProfile` cannot select the user's configured shell.

### Acceptance Detail / 验收注意事项

Simply changing the command to:

```text
wt new-tab -- claude --settings ...
```

is not fully equivalent to launching the default profile shell. Passing `claude` as the tab commandline replaces the profile's shell executable, so PowerShell profile initialization and the actual `pwsh.exe -> claude` process relationship are lost.

The fix should either:

1. genuinely start the selected/default Windows Terminal profile shell and execute Claude within it; or
2. add explicit Windows Terminal profile selection, such as the approach proposed in #1599.

### Related / 相关记录

- #2830 was closed as completed, but the code remains present in v3.16.5 and current `main`.
- #2831 contains an unmerged proposed fix.
- #1599 proposes explicit Windows Terminal profile selection and is also still unmerged.

eicky:
## Summary

When "Windows Terminal" was selected as the preferred terminal on Windows, CC Switch hardcoded `["wt", "cmd", "/K", &bat_path]`, which overrode the user's configured `defaultProfile` (e.g. PowerShell 7) and always forced a `cmd.exe` tab (#5322, #2830).

This PR fixes the issue by:
1. Parsing `defaultProfile` from Windows Terminal's `settings.json` (using `json5` to support JSONC comments and trailing commas).
2. Classifying the shell family (`Pwsh`, `PowerShell`, or `Cmd`) by inspecting commandlines, dynamic generator source metadata (`Windows.Terminal.PowershellCore`), well-known GUIDs, and profile names.
3. Supporting Portable Mode and Scoop persistent directory discovery (`wt.exe` / `wt.shim` sibling settings and `%USERPROFILE%\scoop\persist\windows-terminal...`).
4. Sorting candidate settings files by modification time (`mtime` descending) so the active, most recently edited configuration takes precedence over stale leftover files.
5. Safely falling back to `cmd /K` if the default profile is a non-Windows batch shell (e.g. WSL, Git Bash) or if configuration discovery fails.
6. Escaping single quotes in PowerShell launch commands to protect paths containing apostrophes.

## Related Issue

Fixes #5322

## Screenshots

| Before | After |
|---|---|
| Always opened `cmd.exe` tab regardless of default profile | Launches configured default profile (e.g. `pwsh.exe`) |

## Checklist

- [x] `pnpm typecheck` passes
- [x] `cargo clippy` passes
- [x] Updated i18n files if user-facing text changed (N/A - backend only)

### DIFF
--- a/src-tauri/src/commands/misc.rs
+++ b/src-tauri/src/commands/misc.rs
@@ -4445,7 +4445,7 @@ del \"%~f0\" >nul 2>&1
     std::fs::write(&bat_file, &content).map_err(|e| format!("写入批处理文件失败: {e}"))?;
 
     let bat_path = bat_file.to_string_lossy();
-    let ps_cmd = format!("& '{}'", bat_path);
+    let ps_cmd = format!("& '{}'", bat_path.replace('\'', "''"));
 
     // Try the preferred terminal first
     let result = match terminal {
@@ -4453,7 +4453,7 @@ del \"%~f0\" >nul 2>&1
             &["powershell", "-NoExit", "-Command", &ps_cmd],
             "PowerShell",
         ),
-        "wt" => run_windows_start_command(&["wt", "cmd", "/K", &bat_path], "Windows Terminal"),
+        "wt" => launch_wt_terminal(&bat_path, &ps_cmd),
         _ => run_windows_start_command(&["cmd", "/K", &bat_path], "cmd"), // "cmd" or default
     };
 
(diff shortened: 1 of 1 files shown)


farion1231:
@codex review

PR #7108, opened 2026-09-04
fix(providers): preserve "apply common config" checkbox when editing suppliers

mo2g:
## Problem

When suppliers A and B both have "Apply common config" checked, clicking Edit on supplier A can open an edit form where the checkbox appears unchecked; exiting and re-entering may show it checked again. This can happen for Claude, Codex, and Gemini suppliers.

The root cause is a mismatch between the frontend and backend models for "apply common config":

- Since the runtime-overlay change (`6d078e7f`), the backend strips the common config snippet from stored provider snapshots on save and keeps `meta.commonConfigEnabled` as the single source of truth. The snippet is merged dynamically only when writing live.
- The three frontend hooks still infer the checkbox from whether the currently edited JSON/TOML/env contains the snippet. That inference races with async snippet loading, the Codex TOML merge, and `EditProviderDialog`'s live-settings refresh, so an initialized `true` can be overwritten by an intermediate `false`.

## Fix

- `useCommonConfigSnippet.ts` / `useCodexCommonConfig.ts` / `useGeminiCommonConfig.ts`: in edit mode with an explicit `meta.commonConfigEnabled`, skip the content-derived sync effect and treat that flag as authoritative.
- `useCodexCommonConfig.ts`: during edit initialization, set the checkbox state synchronously from the flag before asynchronously merging the snippet for the editor preview; a failed or stale merge only reports an error and no longer unchecks the box.
- Legacy providers without the `commonConfigEnabled` field keep the previous content-inference behavior for compatibility.

## Test Coverage

- New `tests/hooks/useCommonConfigEditState.test.tsx` covering:
  - Claude: stays checked when the snapshot does not contain the snippet and after an external snapshot refresh.
  - Codex: an external `config.toml` rewrite during the in-flight merge does not uncheck the box.
  - Gemini: stays checked when the snapshot env does not contain the snippet and after an external env refresh.
  - All three scenarios also verify the user can still uncheck the box manually.
- Regression verification: all 3 new cases fail on the old code and pass after the fix; the full `pnpm test:unit` suite passes 136 files / 1077 tests.

## Files Changed

| File | Delta |
|------|-------|
| `src/components/providers/forms/hooks/useCommonConfigSnippet.ts` | +21, -2 |
| `src/components/providers/forms/hooks/useCodexCommonConfig.ts` | +32, -11 |
| `src/components/providers/forms/hooks/useGeminiCommonConfig.ts` | +14, -1 |
| `tests/hooks/useCommonConfigEditState.test.tsx` | +184 |

### DIFF
--- a/src/components/providers/EditProviderDialog.tsx
+++ b/src/components/providers/EditProviderDialog.tsx
@@ -17,6 +17,7 @@ import {
   type ManagedAuthProvider,
 } from "@/lib/api";
 import { extractCodexExperimentalBearerToken } from "@/utils/providerConfigUtils";
+import { useInitialDataCommonConfig } from "@/components/providers/forms/hooks/useInitialDataCommonConfig";
 
 interface EditProviderDialogProps {
   open: boolean;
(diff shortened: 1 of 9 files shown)


farion1231:
@codex review

PR #7139, opened 2026-09-05
fix(hermes): default api_mode to chat_completions for UI-created providers

### LINKED ISSUE #4011: Bug: Missing api_mode field when creating custom Hermes provider via UI

paopao-amns:
## Summary / 概述

UI-created custom Hermes providers can be saved without `api_mode`: the API
Mode dropdown only serializes the field once touched. Hermes then falls back
to its URL heuristics, which only recognize a handful of official endpoints —
third-party gateways silently get the wrong protocol.

The deeplink import path already writes `api_mode` explicitly for exactly
this reason (`build_hermes_settings`); this PR makes the UI path match.

- inject `api_mode: chat_completions` in `set_provider` **after** the on-disk
  forward-compat merge (both update and insert branches), so an `api_mode`
  set outside the UI (Hermes Web UI / hand-edited YAML) always wins over the
  default
- ship `api_mode` in `HERMES_DEFAULT_CONFIG_OBJ` so the add-mode form
  serializes the field into the DB record from the start
- 3 new roundtrip tests (default injected on add / explicit value preserved /
  on-disk value kept when an older payload omits the field) + 1 frontend test

Verified with a patched dev build on Windows + Hermes 0.21: provider add
writes `api_mode` correctly and an explicit protocol selection survives the
on-disk merge.

## Related Issue / 关联 Issue

Fixes #4011

Note: #6029 addresses the same issue — this PR implements the backend-first
approach proposed there, plus the frontend default and the on-disk
preservation test. Happy to coordinate if that PR gets revived.

## Screenshots / 截图

N/A (no UI change)

## Checklist / 检查清单

- [x] Updated i18n files if user-facing text changed / 无用户可见文本改动

### DIFF
--- a/src-tauri/src/hermes_config.rs
+++ b/src-tauri/src/hermes_config.rs
@@ -776,6 +776,28 @@ pub fn get_provider(name: &str) -> Result<Option<serde_json::Value>, AppError> {
     Ok(get_providers()?.get(name).cloned())
 }
 
+/// Forward-compat default for `api_mode` (#4011): the API Mode dropdown only
+/// serializes the field once touched, so a UI-created provider can omit it
+/// entirely. Hermes then falls back to its URL heuristics, which only
+/// recognize a handful of official endpoints — third-party gateways silently
+/// get the wrong protocol. The deeplink import path already writes
+/// `api_mode` explicitly for exactly this reason; the UI path now matches.
+///
+/// MUST run after the on-disk forward-compat merge, not before it — otherwise
+/// the injected default would occupy the key and an `api_mode` the user set
+/// outside the UI (Hermes Web UI / hand-edited YAML) would be reset.
+fn ensure_api_mode(provider: &mut serde_yaml::Value) {
+    const DEFAULT_API_MODE: &str = "chat_completions";
+    if provider.get("api_mode").is_none() {
+        if let serde_yaml::Value::Mapping(m) = provider {
+            m.insert(
+                serde_yaml::Value::String("api_mode".to_string()),
+                serde_yaml::Value::String(DEFAULT_API_MODE.to_string()),
+            );
+        }
+    }
+}
+
 /// Set (upsert) a custom provider by name.
 ///
 /// Upserts into the `custom_providers:` YAML sequence (matched by `name`).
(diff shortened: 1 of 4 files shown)


farion1231:
@codex review

PR #1556, opened 2026-03-18
fix(opencode): separate model properties and advanced properties sect…

qqzhangyanhua:
…ions

问题描述： 在使用的时候 
f8c4f5a8f9fcf1a5acaf7884847d3745
如图，我需要配置 "modalities": {
                        "input": ["text", "image"],
                        "output": ["text"]
                    } 才能在opencode被识别为多模态的模型，但是模型属性都是往options里面去push值，在不改动原来的流程下增加了个属性 高级属性和options同级别如 `   "claude-sonnet-4-5-20250929": {
      "modalities": {
        "input": [
          "text",
          "image"
        ],
        "output": [
          "text"
        ]
      },
      "name": "claude-sonnet-4-5-20250929",
      "options": {
        "thinking": {
          "type": "enabled",
          "budgetTokens": 8192
        }
      }
    }`
    结果 
f0f96b799ef366c67ed3419a5a09d514

Fix the issue where editing "模型属性" (Model Properties) would incorrectly populate "高级属性" (Advanced Properties) fields on blur.

Changes:
- Restore original logic for "模型属性" section to operate on model.options (nested properties)
- Add new "高级属性" section to operate on top-level fields (limit, modalities, variants, etc.) that are siblings to options
- Update getModelAdvancedFields() to return all non-reserved top-level fields (excluding name, limit, options)
- Add proper JSDoc comments explaining the three-tier field classification
- Remove unused imports (getModelExtraFields, isKnownModelKey)

The two sections are now completely isolated:
- Model Properties: edits model.options.* (nested)
- Advanced Properties: edits model.* top-level fields (siblings to options)

Fixes the cross-contamination bug where both sections were operating on overlapping field sets.

PR #7141, opened 2026-09-05
fix(opencode): expose top-level model modalities

starprince1234:
## Summary / 概述

OpenCode model editing now separates nested SDK options from top-level advanced properties. This makes it possible to configure model-level `modalities` (including image input) directly in CC Switch and preserve the value when syncing to `opencode.json`.

## Changes

- Keep “Model Properties” bound to `model.options.*`.
- Add “Advanced Properties” bound to top-level model fields alongside `options`.
- Support adding/editing/removing `modalities`, `variants`, `cost`, and other preserved top-level fields.
- Add localized labels and placeholders for English, Simplified Chinese, Traditional Chinese, and Japanese.
- Add a regression test covering top-level `modalities` editing.

Fixes #1556

## Verification

- `pnpm typecheck`
- `pnpm format:check`
- `pnpm exec vitest run tests/components/OpenCodeFormFields.test.tsx`

The focused test passes. The full unit suite was also run; 134 files passed and 2 pre-existing integration tests failed due to timeout/duplicate test DOM state in `tests/integration/App.test.tsx`.

### DIFF
--- a/src/components/providers/forms/OpenCodeFormFields.tsx
+++ b/src/components/providers/forms/OpenCodeFormFields.tsx
@@ -22,7 +22,7 @@ import {
 import { opencodeNpmPackages } from "@/config/opencodeProviderPresets";
 import { cn } from "@/lib/utils";
 import {
-  getModelExtraFields,
+  getModelAdvancedFields,
   isKnownModelKey,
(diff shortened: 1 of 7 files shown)


farion1231:
@codex review

PR #7159, opened 2026-09-06
feat(pi): add Tencent TokenHub and Token Plan presets

2691176649-cloud:
## Summary / 概述

Add 8 Tencent Pi presets covering TokenHub pay-as-you-go and Token Plan subscription lines, align the Pi model catalog entries those presets need, and extend the Token Plan preset tests with Pi-specific assertions plus TokenHub pay-as-you-go coverage.

## Related Issue / 关联 Issue

N/A — maintenance preset sync for an existing provider line.

## Validation / 验证

- Real-key connectivity checks completed in Pi: corresponding providers/models are discoverable and usable.
- Local checks for this change set:
  - `pnpm typecheck`
  - `pnpm format:check`
  - `pnpm test:unit`

## Notes / 备注

- No Rust files changed.
- Follows the existing Tencent split between TokenHub `/v1` and Token Plan `/plan/v3`.
- Reference product pages:
  - https://cloud.tencent.com/product/tokenhub
  - https://www.tencentcloud.com/products/tokenhub


farion1231:
@codex review

farion1231 (commented):
Thanks for this — the data fidelity here is unusually good, and I verified it rather than taking it on trust.

I pulled Tencent's own tables and diffed them against the PR: the domestic list (1823/130051) matches **24/24** model IDs in the same order, and the international list (intl 1300/78934) matches **26/26**. Both endpoints are verbatim (`tokenhub.tencentmaas.com/v1` from 1823/130078, `tokenhub-intl.tencentcloudmaas.com/v1` from intl 1300/78939 — the latter is the first time this repo has used that host, so it mattered). Your two exclusions are also correct: `hy3-preview` has zero hits in either TokenHub list, and `qwen3.5-flash`/`qwen3.5-plus` are both tagged （2026-09-08下线）. And at 625 additions / 0 deletions, the regression surface for existing users is genuinely nil.

The gap is in one place the model tables don't cover: **thinking control**. The new Tencent block is 313 lines and contains `thinkingProfile` zero times and `compat` zero times, and three of the four issues below come from that.

---

### [P2] DeepSeek models can't have thinking turned off

Tencent's DeepSeek guide ([1823/132248](https://cloud.tencent.com/document/product/1823/132248)) documents the parameter as:

| 字段 | 类型 | 默认值 | 取值范围 |
|---|---|---|---|
| `type` | string | **`"enabled"`** | `"enabled"` / `"disabled"` |

So thinking is on unless the client explicitly sends `"thinking": {"type": "disabled"}`. A request capture against Pi 0.85.1 shows that when a user selects `off`, no `thinking` field is sent at all — meaning the user turns it off in the UI and keeps paying for reasoning tokens.

All 31 DeepSeek entries in this PR ([first one at piProviderPresets.ts#L1603](https://github.com/farion1231/cc-switch/blob/22d86f6cd2af34434540bfdf8233d81db385905f/src/config/piProviderPresets.ts#L1603-L1610)) carry neither piece. Every other DeepSeek V4 entry in the Pi catalog does: the DeepSeek preset sets the level map ([#L898-L906](https://github.com/farion1231/cc-switch/blob/22d86f6cd2af34434540bfdf8233d81db385905f/src/config/piProviderPresets.ts#L898-L906)), and OpenCode Go — the closest analogue, a third party reselling DeepSeek over an OpenAI-compatible endpoint — sets both ([#L1309-L1323](https://github.com/farion1231/cc-switch/blob/22d86f6cd2af34434540bfdf8233d81db385905f/src/config/piProviderPresets.ts#L1309-L1323)).

Suggested fix, following the OpenCode Go shape:

```ts
{
  ...piModel("deepseek/deepseek-v4-flash", {
    id: "deepseek-v4-flash-202605",
    name: "DeepSeek V4 Flash Official",
    thinkingProfile: "deepseekV4",
  }),
  compat: { ...TENCENT_DEEPSEEK_THINKING_COMPAT },
}
```

One caveat on that compat: I'd suggest **not** reusing `DEEPSEEK_THINKING_COMPAT` ([#L56-L60](https://github.com/farion1231/cc-switch/blob/22d86f6cd2af34434540bfdf8233d81db385905f/src/config/piProviderPresets.ts#L56-L60)) wholesale. It also sets `requiresReasoningContentOnAssistantMessages: true`, while the same Tencent doc states twice that `reasoning_content` need *not* be echoed back on multi-turn requests. "Not needed" isn't "rejected", so it may well work — but replaying reasoning text every turn is real input cost on a token-budget product line. A Tencent-specific constant may be the better fit:

```ts
// 1823/132248: thinking.type defaults to "enabled"; disabling requires an explicit
// {"type":"disabled"}. The same doc says reasoning_content need not be echoed back
// on multi-turn, so we don't reuse DEEPSEEK_THINKING_COMPAT wholesale.
const TENCENT_DEEPSEEK_THINKING_COMPAT = {
  ...OPENAI_COMPLETIONS_COMPAT,
  thinkingFormat: "deepseek",
} as const;
```

I couldn't settle that half from the repo alone — it depends on what Pi actually emits with that flag set. Since you already have a request capture set up, comparing the assistant-message size on a multi-turn request with and without it would decide it in one run.

### [P2] Kimi K2.7 Code HighSpeed is declared text-only, so images get dropped

[piModelCatalog.ts#L286](https://github.com/farion1231/cc-switch/blob/22d86f6cd2af34434540bfdf8233d81db385905f/src/config/piModelCatalog.ts#L282-L290) sets `input: ["text"]`.

Tencent's Kimi guide ([intl 1300/80635](https://intl.cloud.tencent.com/zh/document/product/1300/80635)) lists `kimi-k2.7-code-highSpeed` in its 视觉能力 section, and its parameter template says of `kimi-k2.7-code / kimi-k2.7-code-highspeed`: 「图片仅支持 Base64」— a transport restriction that presupposes image support. The domestic model list (1823/130051) independently marks that row 「图片、视频理解」. Its twin `moonshotai/kimi-k2.7-code` is already `["text", "image"]` in the same catalog.

This isn't only metadata: Pi replaces tool-returned images with a "model does not support images" placeholder, which breaks screenshot analysis on this model.

```ts
input: ["text", "image"],
```

(The Base64-only restriction is Pi-side transport behaviour, nothing a preset field can express — no action needed there.)

### [P2] The K2.7 Code entries offer an `off` level that can't work

Same doc, same row: 「仅支持思考」, and more explicitly earlier on the page: 「`kimi-k2.7-code` 思考模式不可关闭」.

This PR adds 8 K2.7 Code entries across four presets (`kimi-k2.7-code` ×4 + `kimi-k2.7-code-highspeed` ×4, [first pair at #L1669-L1672](https://github.com/farion1231/cc-switch/blob/22d86f6cd2af34434540bfdf8233d81db385905f/src/config/piProviderPresets.ts#L1669-L1672)) with no level map, so Pi will surface `off` for a model that can't honour it.

The repo already has the profile for exactly this — the Kimi preset uses it ([#L96-L101](https://github.com/farion1231/cc-switch/blob/22d86f6cd2af34434540bfdf8233d81db385905f/src/config/piProviderPresets.ts#L96-L101)):

```ts
thinkingProfile: "offUnsupported",
```

### [P2] Dropping `minimax-m2.5` from Pi only — which way should it go?

Pi's Token Plan presets omit it ([personal preset at #L1611](https://github.com/farion1231/cc-switch/blob/22d86f6cd2af34434540bfdf8233d81db385905f/src/config/piProviderPresets.ts#L1611)) while Claude, Claude Desktop, Codex, OpenCode, OpenClaw and Hermes all still carry it for the same product.

I confirmed your premise — `minimax-m2.5` has zero hits in both [1823/130060](https://cloud.tencent.com/document/product/1823/130060) and [1823/130659](https://cloud.tencent.com/document/product/1823/130659). But that's the same state #7011 was in when it added the model: its note reads `// minimax-m2.5 型号列表已除名但真 Key 实测可用`. So "removed from the doc table" was already weighed and overridden by a live call.

Whichever way it goes, it should be all seven apps, not one:

- **If it now fails against `/plan/v3`** — drop it everywhere in one change and put that evidence in the comment, replacing the #7011 note.
- **If it only left the doc but still answers** — add it back to Pi; the catalog entry already exists in this PR ([piModelCatalog.ts#L228](https://github.com/farion1231/cc-switch/blob/22d86f6cd2af34434540bfdf8233d81db385905f/src/config/piModelCatalog.ts#L228-L236)), so it's two lines.

Either way the `"piModels" in product ? …` branch at [tokenPlanProviderPresets.test.ts#L411](https://github.com/farion1231/cc-switch/blob/22d86f6cd2af34434540bfdf8233d81db385905f/tests/config/tokenPlanProviderPresets.test.ts#L411) should go — right now it asserts two different model lists for the same product in one file.

### nit

The four `tencent/*` catalog keys are appended after `zai/*`, which breaks the alphabetical grouping. Purely cosmetic — I checked, `cn_official` presets are name-sorted in the selector, so file position doesn't affect the UI.

### Not for this PR

The bot's note about a strict GLM-5.3 level map is factually right (`/plan/v3` only accepts `low`/`high`/`max`), but it's a repo-wide gap rather than something this PR introduced — 50 of the 102 reasoning-capable models already in the Pi presets ship without a level map. Same for the `reasoning: false` on the Auto alias, which follows the OpenClaw preset and Tencent's own OpenClaw integration page. Please don't feel you need to take those on here; I'll open a follow-up to sweep them together.

---

Fixing the first three (all mechanical) plus an answer on `minimax-m2.5` gets this merged. Ping me when it's pushed and I'll re-review quickly.

2691176649-cloud:
Addressed both Codex review comments (P2):

1. `tencent/tokenplan-auto` is now reasoning-capable, with plan-specific thinking maps — `none/high` for personal and international presets, `high`-only for domestic enterprise presets.
2. `glm-5.3` in both enterprise Pro presets now carries a strict `low/high/max` thinking map (medium/xhigh disabled, matching the `/plan/v3` real-key results in `codexProviderPresets.ts`).

Local checks pass: `pnpm typecheck`, `pnpm format:check`, `pnpm test:unit` (135 files / 1083 tests).

2691176649-cloud:
Thanks for the detailed review — all four points addressed in `019d9e31`.

1. **DeepSeek thinking off** — added `thinkingProfile: "deepseekV4"` plus a Tencent-specific `TENCENT_DEEPSEEK_THINKING_COMPAT` (no `requiresReasoningContentOnAssistantMessages`, per 1823/132248 stating `reasoning_content` need not be echoed) to all 31 Tencent DeepSeek entries.
2. **Kimi K2.7 Code HighSpeed images** — catalog entry now `input: ["text", "image"]`.
3. **K2.7 Code off level** — all 8 entries now use `thinkingProfile: "offUnsupported"`.
4. **minimax-m2.5** — decided to drop it from **all apps**: beyond the doc removal, our platform is actively retiring the model, so keeping it only in Pi would have been inconsistent. Removed from Codex/OpenCode/OpenClaw/Hermes/Pi presets and the Pi catalog; the `piModels` test branch is gone.

Also reverted the earlier GLM-5.3 level map and `reasoning:false` on the Auto alias — agreed those are repo-wide and belong in your follow-up sweep, not this PR.

`pnpm typecheck`, `pnpm format:check`, and `pnpm test:unit` (135 files / 1083 tests) all pass locally. Ready for re-review — ping me if anything else comes up.

farion1231 (approved):
Re-reviewed `019d9e31` — all four points from the last round are closed, and I checked each one against the built presets rather than the diff:

- 31/31 Tencent DeepSeek entries carry `thinkingProfile: "deepseekV4"` plus the new `TENCENT_DEEPSEEK_THINKING_COMPAT`
- 8/8 K2.7 Code entries declare `off` unavailable
- `kimi-k2.7-code-highspeed` is image-capable
- `minimax-m2.5` is gone from every app, with a new regression test guarding it

Merge tree against `main` is clean: typecheck, `format:check`, 1087 unit tests and CI all green. Thanks for also reverting the GLM-5.3 and Auto edits once we agreed those were repo-wide — this stayed exactly in scope across both rounds.

Merging.

farion1231:
Merged as `1b34d322` — thank you for this one.

Two things made it easy to take. Every model ID, ordering and endpoint traced back to a Tencent page I could open and diff myself, and when I said the GLM-5.3 level map and the Auto `reasoning` flag were repo-wide gaps, you *reverted* your earlier edits instead of folding them in. Two fix rounds in a day, each with the doc citation attached, and zero scope creep in either — that is rare, and it is why this took one re-read rather than three.

I verified the fixes against the built presets rather than the diff (dumping every model of all eight Tencent presets with its level map, compat and input): 31/31 DeepSeek entries, 8/8 K2.7 Code entries, image capability on highspeed, and no Tencent `minimax-m2.5` left anywhere. Adding the regression test for that removal was a good instinct.

Three notes for the record — none of them held up the merge:

**1. One I owe you.** I missed a case last round that is the same defect I asked you to fix. The four new `minimax-m2.7` entries ship an empty map, so Pi offers an `off` that Tencent silently ignores — and the evidence was already in the repo: `codexProviderPresets.ts` gives it `reasoningLevels: ["high"]` with the note 「minimax-m2.5/m2.7 关不掉（参数被静默忽略，非报错），只列 high 防"选 none 却照样思考"的假档」. That is my miss, and it goes into the follow-up sweep I already promised, alongside GLM-5.3 and the Auto flag. Please don't spend another round on it.

**2. One question only you can settle.** `deepseekV4` has no `off` key, and per `docs/pi-thinking-level-map-requirements-zh.md` a *missing* key means that level falls back to Pi's default — which is exactly the state the original capture showed. So the fix really rests on `thinkingFormat: "deepseek"` making Pi's default `off` emit `{"type":"disabled"}`. That matches the existing precedent (OpenCode Go and the DeepSeek official preset ship the same shape), so I merged on it — but nobody has actually captured it. If your interception setup can confirm it in one run, that would settle it for all three presets at once. If Pi *doesn't* emit disabled, the fix belongs in the profile and I'll take it repo-wide.

**3. A comment wording nit.** The intl TokenHub note reads 「国际站仍列 minimax-m2.5」 and then 「minimax-m2.5 官方已除名」 in the same paragraph. I re-checked intl 1300/78934 (2026-08-28) and it is indeed still listed with full capabilities, so for the international line "已除名" isn't quite right — the real reason is the retirement you mentioned. Worth a one-line reword next time you touch the file, plus the retirement date if you know it, so nobody later reads the comment against the live table and thinks something was missed.

Same category: `deepseek-v3.2` gets the profile literally named `deepseekV4`, asserting `high`/`max` are accepted. The context numbers check out against the intl table (128k/32k), but if v3.2 wasn't part of the real-key pass, `{}` would be the honest value there. Happy to fold that into the sweep too.

Thanks again — the TokenHub line and Pi's Token Plan gap were both real holes, and the thinking-control work you added on top is what makes the presets actually correct rather than merely present.

PR #7162, opened 2026-09-06
feat(pricing): add GPT-6 Astra pricing

yovinchen:
## Summary / 概述

- add the built-in GPT-6 Astra pricing row
- use USD 10/M input, USD 50/M output, USD 1/M cache read, and USD 12.5/M cache write

## Validation / 验证

- cargo fmt --check
- cargo test model_pricing_seed_includes_gpt_6_astra


farion1231 (approved):
Verified against the vendor pricing page; seed row and its unit test both look right. Thanks!

farion1231:
Thanks @yovinchen — merged as 291946d51.

Nice timing on this one: a pricing audit I had run locally against the vendor page landed on exactly the same four numbers (10 / 50 input-output, 1 cached read, 12.5 cache write), so your row is an independent confirmation rather than something I had to re-derive. The cache-write figure is the part worth calling out — it is 1.25x input, which is the convention that started with the GPT-5.6 family, not the older GPT behaviour where writes were free.

Seeding only the base row is also the right call: `strip_reasoning_effort_suffix` in `usage_stats.rs` falls `-low` / `-medium` / `-high` / `-xhigh` back to `gpt-6-astra`, so the effort variants price correctly without extra rows.

For users: the row lands via INSERT OR IGNORE on next launch, and the startup cost backfill re-prices any historical `gpt-6-astra` requests that were recorded at $0. Shipping in the next release.

PR #7163, opened 2026-09-06
feat(pricing): add GLM-5.3 Flash pricing

yovinchen:
## Summary / 概述

- add the built-in GLM-5.3-Flash pricing row
- record the regular list price instead of the two-week 50% promotional price
- use USD 0.15/M input, USD 0.50/M output, USD 0.03/M cache read, and no cache-write charge

## Validation / 验证

- cargo fmt --check
- cargo test model_pricing_seed_includes_glm_5_3_flash


farion1231 (approved):
Verified against the vendor pricing page; rebased onto main to resolve the test-file conflict with #7162 (both tests kept, diff otherwise unchanged). Thanks!

farion1231:
Thanks @yovinchen — merged as 66ae64461.

This one had a mechanical conflict with #7162: both PRs appended their new test right after `model_pricing_seed_includes_claude_5_1_and_standard_sonnet_5_prices`, so git interleaved the two functions line by line. I rebased your branch onto main and kept both tests intact — the diff is otherwise byte-identical to what you submitted.

The numbers match the official Z.ai pricing page (GLM-5.3-Flash: $0.15 input / $0.50 output / $0.03 cached input per MTok), and a zero cache-write cost is correct for the GLM family, consistent with the neighbouring rows.

For users: the row lands via INSERT OR IGNORE on next launch, and the startup cost backfill re-prices any historical `glm-5.3-flash` requests that were recorded at $0. Shipping in the next release, alongside your GPT-6 Astra row.

PR #7164, opened 2026-09-06
feat(pricing): add Gemini 3.8 Flash pricing

yovinchen:
## Summary / 概述

- add the built-in Gemini 3.8 Flash pricing row
- use USD 0.75/M input, USD 3.75/M output, USD 0.075/M cache read, and no cache-write charge

## Validation / 验证

- cargo fmt --check
- cargo test model_pricing_seed_includes_gemini_3_8_flash


farion1231:
@codex review

farion1231 (approved):
Verified and rebased onto `main` (the tests.rs collision with #7162/#7163 was purely positional — all three tests coexist; the rebased diff is byte-identical to your original).

Checks I ran on the merge tree:
- All four numbers cross-checked against three sources: Google's own pricing page ($0.75 / $3.75 / $0.075 through 2026-12-31), models.dev, and a local audit run — all agree.
- `cache_creation = "0"` is right: Google's caching storage charge is per-token-hour, which this table can't express, and every other Gemini row uses "0".
- Seed-only is right for a new model id: `INSERT OR IGNORE` just adds the row to existing databases, and past $0-cost records self-heal through the usage cost backfill.
- `cargo fmt --check` clean, `cargo clippy --all-targets` clean, full suite 2922 passed / 0 failed.

Thanks for the three-part September pricing series.

farion1231:
Merged as 9692ff5e2 — thanks! I rebased the branch for you first; the conflict was only that #7162, #7163 and this one all inserted their test at the same anchor, so keeping all three resolved it.

That's the September pricing set complete (GPT-6 Astra, GLM-5.3 Flash, Gemini 3.8 Flash). All three will be credited to you in the next release notes.

One small thing for next time, purely cosmetic: the neighbouring `gemini-3.7-flash` and `gemini-3.6-flash` rows carry a comment recording when their introductory price expires and what it reverts to, because those rows are deliberately kept out of the pricing-audit ignore list so the audit re-flags them on expiry. I've added the matching note for 3.8 in a follow-up commit, so nothing for you to do here.

PR #7165, opened 2026-09-06
feat(proxy): cross-provider subagent routing in takeover mode

xinzengxx:
### Motivation

A common cost-optimization setup is to run the main conversation (planning, review) on a flagship provider A (e.g. official Anthropic or a high-quality relay) while delegating Task/subagent work to a cheap provider B (e.g. a GLM flash-tier model). Today cc-switch's proxy takeover only supports model mapping *within the current provider* — there is no way to send subagent traffic to a different provider, so this combo is impossible without manually switching providers.

### Approach

- **Routing decision (`proxy::subagent_route`)**: when proxy takeover is active for Claude and a subagent route rule is configured, requests whose model matches the subagent baseline are rerouted by prepending the target provider to the failover candidate list — `[B, A]` — instead of the usual `[A]`. Everything downstream is reused as-is: automatic failover, circuit breaking, and model mapping. If B degrades, existing failover already falls back to A and back, with per-provider usage attribution intact.
- **Persistence**: new `subagent_route` field (`providerId` + optional `model`) on `AppProxyConfig`, stored in `proxy_config` via a v19 schema migration, with DAO read/write, corrupt-JSON fallback to "disabled", and backward-compatible deserialization.
- **Takeover live env injection**: when takeover activates, `CLAUDE_CODE_SUBAGENT_MODEL` is injected into the live Claude env so Claude Code itself tags subagent requests with the routed model. A user-explicit `CLAUDE_CODE_SUBAGENT_MODEL` on the provider always wins over the injected value. Since Claude Code v2.1.251, the per-invocation model the main model passes when dispatching a Task overrides that env var — so the injection also sets `CLAUDE_CODE_SUBAGENT_MODEL_FORCE=1` (official switch, Claude Code ≥ v2.1.257) to keep the routing intent intact; a user-explicit FORCE value (including `"0"`, i.e. opt-out) is always preserved, and the FORCE key is cleaned from the live env when the route is disabled.
- **Frontend**: routing is configured from the Claude provider form itself — the model mapping section's Subagent row becomes a routing row: a target-provider dropdown (existing Claude providers, excluding the one being edited), plus a model input whose fetch button pulls the *target* provider's real model list (preset `modelsUrl` overrides and custom User-Agent included) and supports the `[1M]` context marker in both target modes. Live takeover env is re-synced after saving an active rule. i18n updated for en / zh / zh-TW / ja.

The subagent baseline for matching is the provider's effective `CLAUDE_CODE_SUBAGENT_MODEL` (explicit env if set, otherwise the rule's model) — the same source the live injection uses. Model matching normalizes the `[1M]` suffix, and the target is deduplicated if it already appears in the failover chain.

### Scope

- Claude app only (`AppType::Claude` short-circuits the decision for codex/gemini).
- Zero behavior change when proxy takeover is off, or when any rule precondition is unmet: rule disabled, no baseline, request model mismatch, target equals current provider, or target provider no longer exists — in every case the request path is untouched.

### Testing

Unit tests:

- Decision function (`src-tauri/src/proxy/subagent_route.rs`): reroute on matching request, baseline fallback to route model, explicit env wins as baseline, non-matching request unchanged, `[1M]` suffix normalization, disabled route, non-claude app short-circuit, target == current noop, missing target falls back to current, dedup of target already in the failover chain, failover queue head as baseline source, empty provider list, effective-model precedence.
- Persistence (`src-tauri/src/database/dao/proxy.rs`, `schema.rs`): `subagent_route` roundtrip, backward compatibility when the field is absent, v18→v19 migration adds the column, disabled/corrupt-JSON fallback (with warning log).
- Takeover injection (`src-tauri/src/services/proxy.rs`): injection appended when rule active and env unset, skipped when the provider sets `CLAUDE_CODE_SUBAGENT_MODEL` explicitly, noop when the rule has no model, FORCE=1 appended/suppressed/preserved per the rules above, stale FORCE and model keys removed when the route is disabled.
- Live re-sync (`src-tauri/src/commands/proxy.rs`): saving the config while takeover is active re-syncs the injected env; skipped when takeover is off.
- Frontend (`tests/components/`): provider-form routing row (target dropdown excludes the edited provider, save writes `subagentRoute` without touching the env field, self-target clears the rule, fetch targets B's endpoint, `[1M]` toggle in both modes); failover panel save preserves an existing `subagentRoute`.

Manual acceptance (done against real providers + a live Claude Code session):

1. Providers A (flagship) and B (cheap) configured under Claude, takeover + subagent routing enabled with B and a model name.
2. Main conversation and Task subagents run simultaneously: proxy logs / usage attribution show main requests on A, subagent requests on B.
3. B degraded (invalid key): subagent requests fail over to A without interrupting the session; restoring B routes back to it.
4. Switching the main provider keeps the subagent route pointing at B.
5. Takeover off: plain single-provider behavior, no residue.
6. Explicit `CLAUDE_CODE_SUBAGENT_MODEL` on A: kept as-is, injection does not override it.

### Operational notes

- Requires proxy takeover mode for Claude and Claude Code ≥ v2.1.257 for the FORCE switch; on older Claude Code versions the routing degrades to "subagents that don't get an explicit per-call model".
- Editing the routing rule while takeover is active re-syncs the Claude live config automatically, but a running Claude Code session must be restarted to pick up the new subagent model (env is read at process start).
- Forks and skills running in a subagent with `model: inherit` still run on the main conversation's model (Claude Code-defined exceptions).

### Design doc

- Spec: `docs/superpowers/specs/2026-09-06-subagent-cross-provider-routing-design.md`
- Plan: `docs/superpowers/plans/2026-09-06-subagent-cross-provider-routing.md`


farion1231:
@codex review

PR #7170, opened 2026-09-06
fix(env): preserve colons in Unix environment source paths

### LINKED ISSUE #7169: Unix environment variable cleanup fails when shell config path contains a colon
### Self Checks / 自检

### CC Switch Version / 版本号

3.20.1

### Operating System / 操作系统

macOS

### Related App / 涉及应用

Codex

### Steps to Reproduce / 重现步骤

1. Create a shell configuration file under a directory whose name contains a colon:

   ```shell
   mkdir -p '/tmp/foo:bar'
   printf 'export OPENAI_API_KEY=test-key\n' > '/tmp/foo:bar/.zshrc'
   ```

2. Let the environment conflict scanner produce an `EnvConflict` whose `sourcePath` is:

   ```text
   /tmp/foo:bar/.zshrc:1
   ```

3. Attempt to delete or restore the environment variable through the Unix environment cleanup logic.

The scanner stores file conflicts in the `path:line` format:

```rust
source_path: format!("{}:{}", file_path, line_num + 1)
```

However, the cleanup code parses the value using `split(':')` and uses the first segment as the file path.

### Expected Behavior / 期望行为

CC Switch should parse the line number from the final colon and operate on the complete shell configuration path:

```text
/tmp/foo:bar/.zshrc
```

Both deletion and restoration should modify the intended file.

### Actual Behavior / 实际行为

The path is split at every colon, so CC Switch treats the file path as:

```text
/tmp/foo
```

It then attempts to read or write an unrelated, usually nonexistent path.

Unix and macOS permit colons in file and directory names, so the generated `sourcePath` representation is ambiguous when parsed from the beginning.

### Additional Context / 补充信息

The value can be parsed safely without changing the serialized data structure:

```rust
let (file_path, line_num) = source_path.rsplit_once(':')?;
line_num.parse::<usize>()?;
```

The right-hand side is validated as the line number, while the complete left-hand side remains the path.

A similar delimiter issue is already handled in the OpenCode SQLite session parser. It searches from the end because the database path itself may contain colons.

Linux is affected as well.

cuishuang:
## Summary / 概述

Fix Unix environment variable deletion and restoration when the shell configuration path contains a colon.

Environment conflicts encode file locations as `path:line`, but the cleanup logic previously used `split(':')` and treated the first segment as the file path. For a valid path such as:

```text
/tmp/foo:bar/.zshrc:1
```
this incorrectly resolved the path as /tmp/foo.

## Related Issue / 关联 Issue

Fixes https://github.com/farion1231/cc-switch/issues/7169

## Screenshots / 截图

Not applicable. This is a backend-only path parsing fix.

## Checklist / 检查清单

### DIFF
--- a/src-tauri/src/services/env_manager.rs
+++ b/src-tauri/src/services/env_manager.rs
@@ -102,16 +102,28 @@ fn delete_single_env(conflict: &EnvConflict) -> Result<(), String> {
 }
 
 #[cfg(not(target_os = "windows"))]
+fn parse_file_source_path(source_path: &str) -> Result<&str, String> {
+    let (file_path, line_num) = source_path
+        .rsplit_once(':')
+        .ok_or_else(|| "无效的文件路径格式".to_string())?;
+
+    if file_path.is_empty() {
+        return Err("无效的文件路径格式".to_string());
+    }
+
+    line_num
+        .parse::<usize>()
+        .map_err(|_| "无效的文件路径格式".to_string())?;
+
+    Ok(file_path)
+}
+
+#[cfg(not(target_os = "windows"))]
 fn delete_single_env(conflict: &EnvConflict) -> Result<(), String> {
     match conflict.source_type.as_str() {
         "file" => {
             // Parse file path and line number from source_path (format: "path:line")
-            let parts: Vec<&str> = conflict.source_path.split(':').collect();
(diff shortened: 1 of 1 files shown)


farion1231:
@codex review

PR #7172, opened 2026-09-07
fix(pi): allow provider key rename after removal from live config

ramonmi:
## Summary / 概述

Provider key was locked by record-existence (isEdit) rather than live-config membership, so removing a provider from models.json left its id permanently uneditable. Align Pi with opencode/openclaw/hermes by gating the key lock on piCurrentState.enabledProviderIds, and allow the backend to rename once the key is no longer in models.json.

## Checklist / 检查清单


farion1231:
@codex review

ramonmi:
The issues with the review and test errors have been resolved

PR #7194, opened 2026-09-07
fix(prompts): refresh active prompts after external file edits

allenxu09:
## Summary / 概述

Editing the global `CLAUDE.md` or `AGENTS.md` outside CC Switch left the Prompts list showing the old database content. Loading the list now reads the live file back into the currently enabled prompt, and returning focus to the open Prompts panel refreshes the list. Refreshes remain queued while an edit, confirmation, or write is in progress.

Inactive templates and prompt metadata are preserved. Empty, whitespace-only, missing, and unreadable files leave the saved prompt intact. Unchanged content does not cause another database write. Pi retains its existing native-file activation behavior. Backfill uses the shared sync/restore mutex with `try_lock`; while it is occupied, list reads return persisted prompts without reading stale live content back into the database.

## Validation / 验证

Latest review fixes: updated empty/whitespace preservation coverage and parameterized edit-overlay queueing for both focus and prompt-imported events. Formatting and `git diff --check` completed. Tests were not run for this revision, as requested. The results below are from earlier revisions.

- Reproduced stale content and missing focus refresh with regression tests before implementing the fix.
- Added a Rust integration test covering external edits to both Claude and Codex files, persisted content, empty/missing files, unchanged timestamps, and inactive templates.
- Added focus-refresh and listener-cleanup coverage, and verified deferred refresh while editing.
- `pnpm test:unit`: 135 files, 1,088 tests passed.

- `cargo fmt --check` and `cargo test`: 2,946 tests passed, 5 ignored.

## Related Issue / 关联 Issue

Reported during local use; no GitHub issue filed.

## Checklist / 检查清单

- [x] `pnpm typecheck`
- [x] `pnpm format:check`
- [x] `cargo clippy --manifest-path src-tauri/Cargo.toml -- -D warnings`
- [x] No user-facing text changes; i18n updates not needed.

### DIFF
--- a/src-tauri/src/services/prompt.rs
+++ b/src-tauri/src/services/prompt.rs
@@ -59,7 +59,19 @@ impl PromptService {
         if matches!(app, AppType::Pi) {
             return get_pi_prompts(state);
         }
-        state.db.get_prompts(app.as_str())
+        let mut prompts = state.db.get_prompts(app.as_str())?;
+        // External editors change the live file without updating the saved
+        // selection. Refresh only that selection; inactive templates are separate.
+        if let Some(prompt) = prompts.values_mut().find(|prompt| prompt.enabled) {
+            if let Some(content) = Self::get_current_file_content(app.clone())? {
+                if prompt.content != content {
+                    prompt.content = content;
+                    prompt.updated_at = Some(get_unix_timestamp()?);
+                    state.db.save_prompt(app.as_str(), prompt)?;
+                }
+            }
+        }
+        Ok(prompts)
     }
 
     pub fn upsert_prompt(
--- a/src-tauri/src/services/prompt.rs
+++ b/src-tauri/src/services/prompt.rs
@@ -63,12 +63,17 @@ impl PromptService {
         // External editors change the live file without updating the saved
         // selection. Refresh only that selection; inactive templates are separate.
         if let Some(prompt) = prompts.values_mut().find(|prompt| prompt.enabled) {
-            if let Some(content) = Self::get_current_file_content(app.clone())? {
-                if prompt.content != content {
+            match Self::get_current_file_content(app.clone()) {
+                Ok(Some(content)) if prompt.content != content => {
                     prompt.content = content;
                     prompt.updated_at = Some(get_unix_timestamp()?);
                     state.db.save_prompt(app.as_str(), prompt)?;
                 }
+                Ok(_) => {}
+                Err(error) => log::warn!(
+                    "Failed to refresh {} prompt from live file; keeping saved prompts: {error}",
+                    app.as_str()
+                ),
             }
         }
         Ok(prompts)
--- a/src-tauri/tests/prompt_live_sync.rs
+++ b/src-tauri/tests/prompt_live_sync.rs
@@ -50,12 +50,33 @@ fn prompt_list_refreshes_external_edits_without_changing_inactive_templates() {
         }
 
         let mut saved = state.db.get_prompts(app.as_str()).unwrap()["active"].clone();
+        saved.content = "saved instructions".into();
+        fs::write(&path, &saved.content).unwrap();
         saved.updated_at = Some(42);
         state.db.save_prompt(app.as_str(), &saved).unwrap();
         assert_eq!(
             PromptService::get_prompts(&state, app.clone()).unwrap()["active"].updated_at,
             Some(42)
         );
+
+        // An editor may save UTF-16: the live read fails, but saved templates
+        // must remain available so the user can repair the file from the UI.
+        let utf16 = [0xff, 0xfe, b'A', 0];
+        fs::write(&path, utf16).unwrap();
+        let prompts = PromptService::get_prompts(&state, app.clone()).unwrap();
+        assert_eq!(prompts["active"].content, saved.content);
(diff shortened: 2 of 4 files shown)


farion1231:
@codex review

allenxu09:
@farion1231 ptal

farion1231 (commented):
Thanks — this closes a real gap: an external edit to `~/.claude/CLAUDE.md` / `~/.codex/AGENTS.md` now shows up in the Prompts list without a restart, and the refresh reuses the panel's existing write/overlay queueing so it doesn't collide with an in-progress edit. Test coverage is thorough (both apps, missing file, unreadable file, inactive templates untouched), and the second commit's "keep saved templates on read error" change is the right call.

Two P2s before merge, plus a couple of nits.

**[P2] An empty/whitespace-only live file wipes the saved template.**
`get_prompts` writes any differing content back, including `""` — the new test asserts this on purpose ([prompt_live_sync.rs#L36](https://github.com/farion1231/cc-switch/blob/36a7016c6cf68fd370ab90f726fe45c54c0fac13/src-tauri/tests/prompt_live_sync.rs#L36)). But that means a user who temporarily empties `CLAUDE.md` to "turn instructions off" loses the template's content permanently, with no undo. It also contradicts the guard `enable_prompt` already has for exactly this reason ([prompt.rs#L126 on main](https://github.com/farion1231/cc-switch/blob/d45eebac9/src-tauri/src/services/prompt.rs#L126), from #225). Suggested fix at [prompt.rs#L67](https://github.com/farion1231/cc-switch/blob/36a7016c6cf68fd370ab90f726fe45c54c0fac13/src-tauri/src/services/prompt.rs#L67):

```rust
Ok(Some(content)) if !content.trim().is_empty() && prompt.content != content => {
```

and flip the `""` case in the test to assert the saved content is kept.

**[P2] The backfill doesn't take the sync/restore lock, so a list load during a restore can overwrite the restored prompt with the stale local file.**
File import, backup restore, and WebDAV/S3 download all do "swap DB → `run_post_import_sync` projects DB to live" under `sync_mutex()` ([import_export.rs#L19](https://github.com/farion1231/cc-switch/blob/d45eebac9/src-tauri/src/commands/import_export.rs#L19), [webdav_sync.rs#L144](https://github.com/farion1231/cc-switch/blob/d45eebac9/src-tauri/src/commands/webdav_sync.rs#L144)). `get_prompts` doesn't participate, and nothing blocks navigating to the Prompts page while a restore is running. If the list load lands between the DB swap and the prompt projection, it reads the freshly restored DB + the *old* live file, writes the old content into the DB, and the projection then writes it back to disk — the restore silently didn't happen, and the DB write also queues an auto-sync upload of the stale content.

I reproduced this with a small integration test (snapshot with content R → local drifts to L → `import_sql_string` → `get_prompts` → `sync_all_to_live`): on this branch the live file ends as L; on `main` it ends as R. Happy to paste the test if useful.

The window is narrow (ms-scale, between the swap and the projection), but it's a data-loss path on restore and the fix is small. Suggested: guard in the command layer at [commands/prompt.rs#L16](https://github.com/farion1231/cc-switch/blob/d45eebac9/src-tauri/src/commands/prompt.rs#L16) — `try_lock` on `sync_protocol::sync_mutex()`; if a restore holds it, skip the backfill and return the DB list as before. (`sync_mutex` is `pub(crate)`, so it's reachable from the command.)

**nit** — `PromptPanel.test.tsx` switched the "queues external reloads while editing" case from the `prompt-imported` event to `focus`, so the `prompt-imported` queueing path no longer has a test. Worth keeping both triggers.

**nit** — every window focus runs a full `reload`, which sets `loading` and swaps the list for the loading placeholder for the round-trip, and `reloadLockRef` makes `handleEdit`/`handleDelete` drop a click that lands during it. Probably a one-frame flash in practice; a silent refresh (or comparing `getCurrentFileContent` before reloading) would avoid it. Fine as a follow-up.

With the two P2s addressed this is good to merge — ping me and I'll re-check quickly.

farion1231:
Thanks @allenxu09 — re-checked `bb35ef67a` and both P2s are closed, so I'm merging this.

- The empty/whitespace guard now matches what `enable_prompt` already does, and the test pins both `""` and whitespace-only files.
- Putting the `try_lock` in `PromptService::get_prompts` rather than the command was the better call — it also covers the non-UI callers. I verified it on a merge with current `main`: with the sync lock held, a list load leaves the restored content alone and the projection writes it to disk; with the guard removed, the same two tests fail. All four restore paths (file import, backup restore, WebDAV/S3 download) hold that same lock, so the window is closed.
- Thanks for restoring the `prompt-imported` queueing case via `it.each` as well.

Optional follow-ups, none blocking — happy to take a small PR for any of them:

1. The guard has no regression test in the repo. `sync_mutex` is `pub(crate)`, so it needs an in-crate unit test rather than `tests/prompt_live_sync.rs`; I have the two I used (~20 lines) and can paste them.
2. The skip branch is silent. Uploads (manual and auto-sync) hold the same lock for the network round-trip, and a backfill write itself queues an auto-sync upload, so a refocus in those few seconds skips the backfill until the next one. A `log::debug!` there would make that visible.
3. The focus refresh still flips `loading` and swaps the list for the placeholder for a frame; a silent refresh would avoid it.

Much appreciated — this makes the Prompts panel match what's actually on disk.

PR #6972, opened 2026-08-29
fix(proxy): restore standalone route on startup

allenxu09:
## Summary
- preserve the global proxy enabled state when CC Switch exits normally
- restore a standalone local proxy on startup even when no CLI takeover is enabled
- keep manual stop behavior unchanged so explicitly disabling the route still persists

## Verification
- `cargo test --manifest-path src-tauri/Cargo.toml startup_restores_standalone_proxy_enabled_at_shutdown`
- `cargo test --manifest-path src-tauri/Cargo.toml startup_restore_includes_enabled_grokbuild_route`
- `cargo fmt --manifest-path src-tauri/Cargo.toml --all -- --check`
- `cargo clippy --manifest-path src-tauri/Cargo.toml --all-targets -- -D warnings`

Closes #6961

PR #7210, opened 2026-09-08
fix(proxy): preserve per-app settings on shutdown and port allocation

### LINKED ISSUE #7204: [Bug] Codex failover settings are overwritten by Claude settings after restart
### Environment
- CC Switch: v3.20.2
- OS: macOS
- App: Codex local proxy / failover routing

### Reproduction
1. Set Claude failover Max Retries = 6.
2. Set Codex failover Max Retries = 0, with other app-specific values different.
3. Confirm Codex values are saved.
4. Exit CC Switch normally and start it again.
5. Open Codex failover settings.

### Actual behavior
Codex settings become the Claude values. In my case Max Retries changed from 0 to 6.

When Claude and Codex use identical values, restart appears to preserve them, which masks the problem.

### Likely cause
In v3.20.2, stop_with_restore_keep_state() reads the legacy global proxy config via get_proxy_config(), which reads the Claude row. update_proxy_config() then updates shared fields without a WHERE app_type clause, so Claude values are written to all application rows.

Relevant files:
- src-tauri/src/services/proxy.rs
- src-tauri/src/database/dao/proxy.rs

### Expected behavior
Claude, Codex, Gemini, and Grok Build failover settings should remain independent across exit/restart.

Komikawayi:
## Summary / 概述

修复正常退出 CC Switch 后，Codex 等应用的重试次数和超时配置被 Claude 配置覆盖的问题，同时修复动态监听端口保存路径上的相同问题。

- `stop_with_restore_keep_state()` 为已废弃的 `live_takeover_active` 字段调用旧的全局配置写入接口。该接口从 Claude 行读出配置，再把重试及超时字段更新到所有应用行。移除此无效回写，保留 Live 配置恢复、备份清理和健康状态重置。
- `persist_ephemeral_listen_port_if_needed()` 改用现有全局配置接口，仅更新全局字段，避免回写各应用独立的重试和超时配置。
- 保持旧配置接口及主动更新配置的现有语义，不涉及数据库迁移或前端改动。

## Related Issue / 关联 Issue

Fixes #7204

## Verification / 验证

基于上游 `main` / v3.20.2 (`f3b18df`)；Windows 本地验证。

新增两个回归测试，分别覆盖退出和动态端口保存。四个应用使用不同重试次数（Claude/Codex/Gemini/Grok Build：6/0/2/3）与超时配置，并比较全部应用级配置字段。另检查退出时配置备份仍能恢复并清理、动态端口正确持久化，以及固定端口不被改写。

- **修复前**：两个测试均失败，Codex 的重试次数从 0 被改成 6，三个超时字段也被覆盖。
- **修复后**：两个测试通过。
- `cargo test --manifest-path src-tauri/Cargo.toml --lib services::proxy::tests -- --test-threads=1`：96 passed。
- `cargo test --manifest-path src-tauri/Cargo.toml --lib database::dao::proxy::tests -- --test-threads=1`：3 passed。
- `cargo fmt --manifest-path src-tauri/Cargo.toml -- --check`：通过。
- `git diff --check`：通过。

测试使用内存数据库和隔离测试目录；未执行已安装应用的完整退出/重启 UI 测试。

## Screenshots / 截图

不适用：仅后端逻辑与测试变更。

## Checklist / 检查清单

- [ ] `pnpm typecheck` passes — 未运行，无 TypeScript 变更。
- [ ] `pnpm format:check` passes — 未运行；Rust 格式检查已通过。
- [ ] `cargo clippy` passes — 未运行。
- [x] 无需更新 i18n：未修改用户界面文案。

### DIFF
--- a/src-tauri/src/services/proxy.rs
+++ b/src-tauri/src/services/proxy.rs
@@ -999,10 +999,15 @@ impl ProxyService {
             return Ok(());
         }
 
-        let mut resolved_config = config.clone();
+        // 端口是全局字段，不能通过旧接口回写各应用独立的重试和超时配置。
+        let mut resolved_config = self
+            .db
+            .get_global_proxy_config()
+            .await
+            .map_err(|e| format!("获取全局代理配置失败: {e}"))?;
         resolved_config.listen_port = actual_port;
         self.db
-            .update_proxy_config(resolved_config)
+            .update_global_proxy_config(resolved_config)
             .await
             .map_err(|e| format!("保存动态代理端口失败: {e}"))
     }
@@ -1803,20 +1808,16 @@ impl ProxyService {
         // 2. 恢复原始 Live 配置
         self.restore_live_configs().await?;
 
-        // 3. 更新 proxy_config 表中的 live_takeover_active 标志（兼容旧版）
-        //    注意：保留 proxy_config.enabled 状态，下次启动时自动恢复
-        if let Ok(mut config) = self.db.get_proxy_config().await {
-            config.live_takeover_active = false;
-            let _ = self.db.update_proxy_config(config).await;
-        }
+        // 保留各应用的 enabled 和故障转移配置，下次启动时自动恢复。
+        // live_takeover_active 已废弃，无需通过旧接口回写配置。
 
-        // 4. 删除备份（Live 配置已恢复，备份不再需要）
+        // 3. 删除备份（Live 配置已恢复，备份不再需要）
         self.db
             .delete_all_live_backups()
             .await
             .map_err(|e| format!("删除备份失败: {e}"))?;
 
-        // 5. 重置健康状态
+        // 4. 重置健康状态
         self.db
             .clear_all_provider_health()
             .await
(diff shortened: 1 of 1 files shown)


farion1231:
@codex review

farion1231:
Thanks @Komikawayi — merging this.

Nice find on the root cause. What made it worth a closer look: the block being removed from `stop_with_restore_keep_state` never did what its comment claimed. The legacy `update_proxy_config` UPDATE doesn't even list `live_takeover_active` in its SET clause, so the only observable effect that step ever had was fanning the Claude row's retries/timeouts out to every app row. Deleting it outright is the right fix, not a workaround.

What I checked on my side:
- Merge tree against current `main` (2f3c026): `cargo test` 2820/0 in lib plus integration suites, `cargo fmt --check` clean, `cargo clippy --all-targets` with zero warnings (covering the clippy run you noted as skipped).
- Reverting only the two fix hunks while keeping your tests makes both fail with exactly the #7204 shape — codex `maxRetries` 0 → 6 and the three timeouts picking up Claude's values — so the regression coverage is real.
- The shutdown test's `#[serial]` + `TempHome` + `reload_settings()` setup matches the existing pattern in that module; the real `~/.claude/settings.json` was untouched during the run.
- Runtime never reads `max_retries` or the timeouts from the legacy `ProxyConfig` — `handler_context`/`forwarder` take them from `AppProxyConfig` — and the frontend no longer invokes the legacy `update_proxy_config`/`get_proxy_config` commands, so the two call sites you changed were the only live ones.
- Dry-run merges with #6972 and #7120, which touch the same shutdown path, are conflict-free in either order.

Two non-blocking things, in case you want a follow-up:
1. The legacy `Database::update_proxy_config` still has no `WHERE` and still writes `max_retries` + the three timeouts table-wide. After this PR its only remaining caller is `ProxyService::update_config` behind the `update_proxy_config` Tauri command, which nothing in the UI invokes anymore — but anyone wiring that command back up would reintroduce the bug. Dropping those four per-app columns from that UPDATE (or retiring the legacy trio entirely) would close the door.
2. The `use_ephemeral_proxy_port` test helper still goes through the legacy interface and resets per-app fields to defaults as a side effect; `update_global_proxy_config` would be the cleaner shape now.

Release note will mention that already-overwritten per-app values need to be re-entered once.

PR #7219, opened 2026-09-08
fix(codex): detect growing rollouts with a persisted byte cursor

### LINKED ISSUE #6060: Codex usage sync misses active Windows rollouts when mtime stays unchanged

woniuxiaoshu:
## Summary / 概述

Fix Codex usage imports getting stuck when Windows keeps an active rollout's modification time unchanged while its contents grow.

The unchanged-file check now compares mtime with the persisted consumed byte offset as well as file size. This uses the existing `session_log_sync.last_byte_offset` column: legacy Codex cursors with a NULL byte offset are re-read once, and their existing line cursor prevents duplicate imports. Cursor updates remain in the same transaction as the final insert batch.

The parser leaves incomplete trailing JSON records for the next scan, while continuing to accept a complete final JSON record without a newline. Token accounting, request IDs, pricing and fork replay rules are unchanged.

## Related Issue / 关联 Issue

Fixes #6060. Related: #6023, #6080, #6027, #6246.

This is a one-file alternative against current `main` (v3.20.2) that reuses the byte-cursor column already present in schema v18. It does not add a schema migration or reset existing usage/cursors.

## Screenshots / 截图

No UI changes. The byte-cursor fix was verified in a locally built Windows application: previously stuck logs were imported, subsequent scans detected new usage despite unchanged mtime, and restarting did not duplicate the unchanged sessions. Input/output/cache counters were reconciled with the consumed log prefix. Private session logs and database snapshots are intentionally not attached.

## Validation / 验证

- [x] Codex session usage tests: 51 passed, 0 failed, 1 existing real-corpus test ignored.
- [x] Regression coverage: append with unchanged mtime; legacy line-only cursor catch-up; incomplete final record; complete final record without a newline followed by further appends; repeated scans after cursor reload.
- [x] `cargo fmt -- --check`
- [x] `cargo clippy --locked --lib`
- [x] `pnpm typecheck` (same `tsc --noEmit` command invoked directly)
- [x] `pnpm format:check` (same Prettier command invoked directly)

Windows release-build/live-application verification covered the byte-cursor fix before the final EOF-without-newline compatibility adjustment. The final adjustment is covered by the Rust regression tests above. The full repository test suite was not rerun.

## Checklist / 检查清单

- [x] TypeScript type checking passes.
- [x] Code formatting passes.
- [x] Clippy passes for the Rust library.
- [x] No user-facing text changed; i18n updates are not applicable.

Developed with AI assistance; the patch and regression cases were reviewed and validated locally.

### DIFF
--- a/src-tauri/src/services/session_usage_codex.rs
+++ b/src-tauri/src/services/session_usage_codex.rs
@@ -217,6 +217,7 @@ struct ParsedCodexFile {
     parent: ParentResolution,
     token_events: Vec<ParsedTokenEvent>,
     line_offset: i64,
+    byte_offset: i64,
     has_billable_tokens: bool,
 }
 
@@ -493,6 +494,7 @@ fn token_snapshot_source(payload: &serde_json::Value) -> Option<String> {
 ///   仍用旧价，下一个同步 pass 生效。
 struct CodexSyncPass {
     cursors: HashMap<String, (i64, i64)>,
+    byte_offsets: HashMap<String, Option<i64>>,
     pricing: HashMap<String, Option<ModelPricing>>,
 }
 
@@ -511,8 +513,15 @@ impl CodexSyncPass {
             })
             .and_then(|rows| rows.collect::<Result<HashMap<_, _>, _>>())
             .map_err(|e| AppError::Database(format!("预载同步游标失败: {e}")))?;
+        let mut stmt = conn.prepare("SELECT file_path, last_byte_offset FROM session_log_sync")?;
+        let byte_offsets = stmt
+            .query_map([], |row| {
+                Ok((row.get::<_, String>(0)?, row.get::<_, Option<i64>>(1)?))
+            })?
+            .collect::<Result<HashMap<_, _>, _>>()?;
         Ok(Self {
             cursors,
+            byte_offsets,
             pricing: HashMap::new(),
         })
     }
@@ -783,7 +792,7 @@ fn parse_codex_file(
 ) -> Result<ParsedCodexFile, AppError> {
     let file =
         fs::File::open(file_path).map_err(|e| AppError::Config(format!("无法打开文件: {e}")))?;
-    let reader = BufReader::new(file);
+    let mut reader = BufReader::new(file);
     let mut root_meta_seen = false;
     let mut root_timestamp = None;
     let mut meta_thread_id = None;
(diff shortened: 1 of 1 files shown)


farion1231:
@codex review

farion1231:
Merged, thank you @woniuxiaoshu! This fixes a long-standing Windows gap: Codex keeps the rollout handle open for the whole session, so mtime never moved while the file grew and usage silently stopped being imported. Persisting the observed length alongside the existing cursor without a schema migration was the right call.

One follow-up landed on main right after the merge: the parser now counts an incomplete trailing record toward the stored length, so a crashed or never-resumed rollout that ends mid-record (or in whitespace) is skipped on later passes instead of being fully reparsed every minute. Existing installs will reparse each Codex rollout once after upgrading to seed the new cursor; nothing is double-counted.

PR #7228, opened 2026-09-08
fix(omo-slim): import agents from the active preset

### LINKED ISSUE #6992: OMO Slim "Import from local" does not read per-agent models from presets.<active> - agent rows empty with recommended-model placeholders
### Summary

The "Import from local" action in the Oh-My-OpenCode-Slim section does not populate per-agent model fields when the source configuration uses the standard preset-based layout generated by the OMO-S installer. The agent model fields import empty, and the UI then displays built-in recommended models (e.g. `claude-opus-5` for Orchestrator) as placeholder text, which can be mistaken for imported values.

### Environment

- CC Switch 3.20.1, Windows 11
- oh-my-opencode-slim 2.2.17 (installed via its official installer)
- Config source: `~/.config/opencode/oh-my-opencode-slim.json`

### Steps to reproduce

1. Install oh-my-opencode-slim with its official installer. The generated config uses the recommended preset structure - models are defined under `presets.<preset>.<agent>.model`, with no top-level `agents` block:

```json
{
  "preset": "opencode-go",
  "presets": {
    "opencode-go": {
      "orchestrator": { "model": "opencode-go/deepseek-v4-flash", "variant": "high" },
      "oracle": { "model": "opencode-go/deepseek-v4-pro", "variant": "max" }
    }
  }
}
```

2. Customize per-agent models in the active preset.
3. In CC Switch, open the Oh-My-OpenCode-Slim section and click "Import from local", pointing at the file above.

### Expected behavior

The agent rows reflect the models defined in the active preset, e.g. `orchestrator: opencode-go/deepseek-v4-flash (variant: high)`.

### Actual behavior

- All per-agent model fields are empty.
- Each empty combobox shows its built-in recommendation as placeholder text (e.g. `claude-opus-5` for Orchestrator).
- The complete `presets` block is preserved only inside the "Other Fields" raw JSON and remains invisible to the per-agent rows.

### Root cause

- `src-tauri/src/services/omo.rs` (`import_from_local`) extracts only the top-level `agents` key from the source file. There is no handling of the `presets` key anywhere in the OMO Slim import path.
- `src/types/omo.ts` (`OMO_SLIM_BUILTIN_AGENTS`) renders `recommended` values as placeholder text in `OmoFormFields.tsx` when the imported model is empty.

Because the OMO-S installer always generates the preset-based structure, every default installation is affected. The top-level `agents` layout is an override mechanism, not the default.

### Suggested fix

On import, read the top-level `preset` key to determine the active preset, then flatten `presets.<active>.<agent>.model` / `variant` into the per-agent fields. Top-level `agents` overrides should keep precedence, consistent with OMO-S merge semantics. Alternatively, display a hint that the models originate from the active preset.

### Workaround

Keep the config file as the source of truth and use CC Switch as a profile switcher only, or re-enter agent models manually after import. Note that manual re-entry creates a flat top-level `agents` block, which overrides `presets` at OMO-S startup.

happys2333:
## Summary / 概述

OMO Slim's local import only reads the top-level `agents` object, so configs generated with `preset` and `presets` leave the agent form empty.

This change:

- resolves only the active preset for OMO Slim;
- deep-merges it with top-level `agents`, with top-level values winning, matching OMO Slim 2.2.17;
- preserves the complete `preset` and `presets` data, custom agents, and unknown fields;
- rejects malformed or missing active preset data instead of choosing another preset.

Saving the imported profile writes the effective merged agents to the top-level `agents` object. This keeps the imported values stable, but changing `preset` later will not replace those top-level overrides automatically.

## Related Issue / 关联 Issue

Fixes #6992

## Screenshots / 截图

Not applicable; this is a backend-only import fix.

## Checklist / 检查清单

- [ ] `pnpm typecheck` passes / 通过 TypeScript 类型检查 — not run; no frontend code changed
- [ ] `pnpm format:check` passes / 通过代码格式检查 — not run; no frontend code changed
- [ ] `cargo clippy` passes (if Rust code changed) / 通过 Clippy 检查（如修改了 Rust 代码） — dependency resolution was blocked because the local environment could not reach crates.io
- [x] Updated i18n files if user-facing text changed / 如修改了用户可见文本，已更新国际化文件 — not applicable; no user-facing text changed

Additional verification:

- [x] `cargo fmt --check -- src/services/omo.rs`
- [x] Added regression tests for active preset selection, deep root overrides, field preservation, invalid preset structures, and standard OMO isolation
- [ ] `cargo test services::omo::tests --lib` — dependency resolution was blocked before compilation because `async-stream` was not cached and crates.io was unreachable

This PR is opened as a draft because the project-level Rust tests could not be completed locally.

### DIFF
--- a/src-tauri/src/services/omo.rs
+++ b/src-tauri/src/services/omo.rs
@@ -843,6 +843,71 @@ impl OmoService {
         }
     }
 
+    fn merge_objects(
+        base: &Map<String, Value>,
+        override_: &Map<String, Value>,
+    ) -> Map<String, Value> {
+        let mut merged = base.clone();
+        for (key, override_value) in override_ {
+            let value = match (merged.get(key), override_value) {
+                (Some(Value::Object(base)), Value::Object(override_)) => {
+                    Value::Object(Self::merge_objects(base, override_))
+                }
+                _ => override_value.clone(),
+            };
+            merged.insert(key.clone(), value);
+        }
+        merged
+    }
+
+    fn imported_agents(
+        v: &OmoVariant,
+        obj: &Map<String, Value>,
+    ) -> Result<Option<Value>, AppError> {
+        if v.category != SLIM.category {
+            return Ok(obj.get("agents").cloned());
+        }
+
+        let root_agents = match obj.get("agents") {
+            Some(Value::Object(agents)) => Some(agents),
+            Some(_) => {
+                return Err(AppError::Config(
+                    "OMO Slim agents must be an object".to_string(),
+                ));
+            }
+            None => None,
+        };
+        let Some(preset) = obj.get("preset") else {
+            return Ok(root_agents.cloned().map(Value::Object));
+        };
+        let preset = preset
+            .as_str()
+            .ok_or_else(|| AppError::Config("OMO Slim preset must be a string".to_string()))?;
+        if preset.is_empty() {
+            return Ok(root_agents.cloned().map(Value::Object));
+        }
+
+        let presets = obj
+            .get("presets")
+            .and_then(Value::as_object)
+            .ok_or_else(|| {
+                AppError::Config(
+                    "OMO Slim presets must be an object when preset is set".to_string(),
+                )
+            })?;
+        let active = presets.get(preset).ok_or_else(|| {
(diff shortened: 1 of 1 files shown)


farion1231:
@codex review

PR #7240, opened 2026-09-09
fix(windows): prevent WebView2 startup blank screen

kyleforai:
## Summary

- Fix the blank/white window issue on affected Windows hosts.
- Defer Windows WebView2 window creation until application setup completes.
- Add a COM-aware WebView2 callback wait implementation.
- Add the Windows `--single-process` WebView2 workaround.
- Vendor `webview2-com` and pin the Wry fix so clean builds reproduce the fix.
- Fix pnpm build permission configuration.

## Testing

- `pnpm run build:renderer` passed.
- `cargo fmt --manifest-path src-tauri/Cargo.toml -- --check` passed.
- `cargo check --manifest-path src-tauri/Cargo.toml --lib -j1` passed.
- `pnpm tauri build --no-bundle --ci` passed.
- Release executable launched successfully and displayed the CC Switch UI instead of a blank page.
- 2793 Rust tests passed; 3 environment-dependent tests failed because of Windows symlink privileges and missing `LOCALAPPDATA`.

## Related Issue

Not applicable.


farion1231:
@codex review

farion1231:
Thanks for digging into the WebView2 blank-window problem. Closing this PR in its current shape, though — not because of the fix itself, but because of how it is delivered:

- `[patch.crates-io] wry = { git = "https://github.com/npiesco/wry", rev = "efbd299" }` pins our release builds to a commit on a personal fork. Signed/notarized release artifacts cannot depend on a third-party fork we do not control.
- Vendoring the whole `webview2-com` crate into `src-tauri/vendor/` (~2.5k lines) turns an upstream dependency into code we have to maintain forever.
- There is no linked issue and no reproduction beyond "affected Windows hosts", so we cannot tell which Windows / WebView2 runtime versions this addresses.

The part that is genuinely reviewable is the `lib.rs` change that defers Windows window creation until setup completes. If that change alone fixes the blank screen on your machine, a minimal PR with just that (plus an issue describing the Windows and WebView2 runtime versions and how to reproduce) would be very welcome. For the wry side, the right path is to wait for the fix to land in an upstream wry release and bump normally.

PR #7244, opened 2026-09-09
feat(dsh): add DeepSeek Harness as an AppType target

mackenzietsai:
## Summary / 概述

Add **DeepSeek Harness (DSH)** as a first-class application target, alongside the existing apps. DSH is a local agent harness whose provider configuration lives in `~/.dsh/settings.yaml` (`llm-pi-ai.providers.<id>` entries with `apiKeyEnv` secret references, plus an `agent-default-model` section).

### Backend (`src-tauri/src/dsh_config.rs`)

- Path resolution: explicit settings override → `DSH_HOME` env → `~/.dsh`.
- Read: `get_providers` / `get_provider` / `get_default_model`.
- Write: block-level YAML splice for `set_provider` (replaces only the target provider block, preserving comments and formatting elsewhere; covers all four structural cases: existing block replace / insert within section / `providers: {}` expansion / append without `llm-pi-ai` section), `remove_provider` (folds empty providers back to `{}`), `set_default_model` (rewrites only that top-level section).
- Safety on every write: backup to `~/.cc-switch/backups/dsh/` (rotating), atomic write, in-process write lock, parse + read-back validation before commit, `baseURL` required, secrets always stored as `apiKeyEnv` references (usage credential resolution returns an empty key for DSH — no key material ever crosses the boundary).
- App wiring: `AppType::Dsh` (`as_str` = `dsh`, aliases `dsh` / `deepseek-harness`), additive mode, `VisibleApps.dsh` (hidden by default), startup live import (same pattern as Hermes: new ids imported, existing ids updated), `read_live_settings` / `write_live_snapshot`, DSH prompt target `~/.dsh/AGENTS.md` (auto-import disabled), deeplink builder.
- MCP/Skill file sync intentionally **not** enabled for DSH: DSH manages MCP through its profile composition and skills through its own package manager, so the existing sync surfaces don't apply.

### Frontend

- DSH in the app switcher, app visibility settings, and provider views, using the existing DeepSeek icon (icon-only, consistent with the other apps; tooltip and overflow menu still show the name).
- i18n keys added in all four languages (zh / zh-TW / en / ja): `apps.dsh`, `settings.dshConfigDir`, `browsePlaceholderDsh`.

### Tests / 測試

- 10 new `dsh_config` unit tests: upsert preserves siblings/comments, block replace, append without `llm-pi-ai` section, inline `{}` expansion, missing `baseURL` rejected, remove preserves siblings, fold-to-`{}`, remove idempotent, default-model section isolation, temp-file e2e roundtrip with backup verification.
- Opt-in production-shape round-trip test (`DSH_ROUNDTRIP_SETTINGS` env): copies a real settings.yaml to a temp dir (the original is never written) and verifies the full cycle — upsert with backup, sibling survival, default-model immutability, re-upsert replacement, remove with field-level restoration.

Validated: `cargo test` 2,955 passed / 0 failed; `pnpm test:unit` 135 files / 1,087 tests; `pnpm typecheck` clean; `pnpm format:check` clean; `cargo fmt --check` clean; `cargo clippy` clean on the new module.

## Related Issue / 关联 Issue

None — new feature.

## Screenshots / 截图

DSH appears in the app switcher with the DeepSeek icon (icon-only, like the other apps) and in Settings → General → Homepage Display once enabled. The DSH provider view lists providers imported from `~/.dsh/settings.yaml`.

## Checklist / 检查清单

### DIFF
--- a/src-tauri/src/app_config.rs
+++ b/src-tauri/src/app_config.rs
@@ -32,7 +32,8 @@ impl McpApps {
             AppType::OpenCode => self.opencode,
             AppType::OpenClaw => false, // OpenClaw doesn't support MCP
             AppType::Hermes => self.hermes,
-            AppType::Pi => false, // Pi core has no native MCP registry.
+            AppType::Dsh => false, // DSH mounts MCP via profile cordis patches; no native config-file sync yet.
+            AppType::Pi => false,  // Pi core has no native MCP registry.
             AppType::ClaudeDesktop => false,
         }
     }
@@ -47,7 +48,8 @@ impl McpApps {
             AppType::OpenCode => self.opencode = enabled,
             AppType::OpenClaw => {} // OpenClaw doesn't support MCP, ignore
             AppType::Hermes => self.hermes = enabled,
-            AppType::Pi => {}            // Pi core has no native MCP registry.
+            AppType::Dsh => {} // DSH mounts MCP via profile cordis patches; no native sync yet.
+            AppType::Pi => {}  // Pi core has no native MCP registry.
             AppType::ClaudeDesktop => {} // Claude Desktop 3P provider config doesn't support MCP here
         }
     }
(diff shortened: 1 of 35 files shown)


farion1231:
@codex review

PR #7167, opened 2026-09-06
fix(proxy): add opt-in Claude Code steer rectifier

SinCircle:
## Summary

Add a disabled-by-default rectifier for Claude Code steer messages on Chat Completions routes. Matching messages are converted from `system` to `user`, preserving content and position.

## Related Issue

Refs #7166

## Validation

Frontend checks and rectifier tests passed. Full backend checks passed on Windows and Linux in [CI](https://github.com/SinCircle/cc-switch/actions/runs/34040344209).

PR #7253, opened 2026-09-09
feat(proxy): add opt-in conversation system conversion

### LINKED ISSUE #7252: Cache misses from Claude Code conversation system messages
## Environment

CC Switch 3.20.2, Claude Code 2.1.266, Windows. Chat Completions route through a gateway to DeepSeek.

## Reproduction

Run a conversation with several tool calls, then send a message while Claude Code is working. Compare cached input before and after the steer message.

## Expected

The unchanged conversation prefix remains reusable.

## Observed

I still see cache misses after #6941. In a controlled test, cached input dropped from 98.96% to 0% on steer. Changing only the injected message's role to `user` gave 97.35%. Token reminders, todo reminders and background notifications showed the same pattern.

Outbound comparisons confirmed that existing messages, tools and the initial system message stayed unchanged. These results are specific to the tested gateway route.

Related: #7167.

SinCircle:
## Summary

Adds opt-in conversion from `system` to `user` for Claude Code steer messages, token/todo reminders and background notifications, addressing cache misses reproduced on a Chat Completions gateway.

Groups these controls and a convert-all option under Response system conversion. All default off; conversion preserves content, order and initial system instructions, including replayed history.

## Related

Fixes #7252.

Follows up on #7167 and the remaining cache misses after #6941. Rebased on v3.20.2, with the steer handling included here.

## Validation

Frontend checks and a Windows build passed. Full backend tests and Clippy passed on Windows and Linux in [CI](https://github.com/SinCircle/cc-switch/actions/runs/34333598265). Live comparisons verified the installed fix.

### DIFF
--- a/src-tauri/src/proxy/forwarder.rs
+++ b/src-tauri/src/proxy/forwarder.rs
@@ -1428,6 +1428,17 @@ impl RequestForwarder {
         };
         if adapter.name() == "Claude" {
             if let Some(api_format) = resolved_claude_api_format.as_deref() {
+                if matches!(app_type, AppType::Claude) {
+                    let converted =
+                        super::response_system_rectifier::rectify_response_system_messages(
+                            &mut mapped_body,
+                            &self.rectifier_config,
+                            api_format,
+                        );
+                    if converted > 0 {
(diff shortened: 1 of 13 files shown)


farion1231:
@codex review

PR #7263, opened 2026-09-10
fix(codex): honor proxy URL when model_provider is omitted

Komikawayi:
## Summary / 概述

未指定 `model_provider` 时，Codex 默认使用内置 `openai`，但 CC Switch 改写地址时写入了 Codex 不使用的顶层 `base_url`。代理配置看似包含本地地址，Responses 请求却仍可能发往官方。

将缺省供应商按内置 `openai` 处理，写入有效的 `openai_base_url`。第三方接管随后复用已有规范化流程，生成自定义供应商并注入 `PROXY_MANAGED`。已有供应商选择不变。

## Related Issue / 关联 Issue

Related to #7217：修复缺失供应商选择时的路由分支。不覆盖客户端显式以 `modelProvider=openai` 覆盖自定义供应商的情况；登录状态修复见 #7262。

## Verification / 验证

- 新回归测试修复前失败、修复后通过，覆盖空配置、遗留顶层地址、供应商名称冲突及重复接管。
- `codex_config::tests`：117 passed；排除依赖符号链接权限的既有测试 `resolve_catalog_rejects_symlink_escaping_config_dir`。
- `services::proxy::tests`：95 passed。
- 将测试生成的三组配置交给官方 Codex app-server v0.153.4：均选择自定义供应商，携带占位 bearer 请求本地 `/v1/responses`，未尝试连接 `api.openai.com`。模拟上游返回预设 400，仅验证路由和认证头，不验证模型生成。
- `cargo fmt --check`、`git diff --check`、`cargo clippy --lib -- -D warnings` 通过。

## Checklist / 检查清单

- [ ] `pnpm typecheck`（无前端改动，未运行）
- [ ] `pnpm format:check`（无前端改动，未运行）
- [x] `cargo clippy` passes
- [x] 无用户界面文案变更

### DIFF
--- a/src-tauri/src/codex_config.rs
+++ b/src-tauri/src/codex_config.rs
@@ -4101,9 +4101,9 @@ pub fn restore_codex_settings_for_backfill(
 ///
 /// Supported fields:
 /// - `"base_url"`: writes to `[model_providers.<current>].base_url` if `model_provider` exists,
-///   otherwise falls back to top-level `base_url`.
+///   otherwise uses `openai_base_url` for Codex's default built-in provider.
 /// - `"wire_api"`: writes to `[model_providers.<current>].wire_api` if `model_provider` exists,
-///   otherwise falls back to top-level `wire_api`.
+///   otherwise leaves the built-in provider's Responses protocol unchanged.
 /// - `"model"` / `"model_catalog_json"`: writes to top-level field.
 ///
 /// Empty value removes the field.
@@ -4119,7 +4119,9 @@ pub fn update_codex_toml_field(toml_str: &str, field: &str, value: &str) -> Resu
             let model_provider = doc
                 .get("model_provider")
                 .and_then(|item| item.as_str())
-                .map(str::to_string);
+                .map(str::to_string)
+                // Codex defaults to openai when the selector is absent.
+                .or_else(|| (!doc.contains_key("model_provider")).then(|| "openai".to_string()));
 
             if let Some(provider_key) = model_provider {
                 // validate_reserved_model_provider_ids（0.148 起）对配置里出现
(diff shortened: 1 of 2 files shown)


Komikawayi:
@codex

farion1231:
Thanks, Komi — merging this as-is.

This closes a gap that has been sitting open since #6256: a provider card whose TOML has no `model_provider` used to get the takeover address written as a top-level `base_url`, which Codex ignores, so the built-in `openai` provider silently went straight to api.openai.com. Routing the implicit selector through the same `openai_base_url` branch the explicit `model_provider = "openai"` case already used, and letting the existing reroute migrator turn it into a `cc-switch` table, is exactly the right size of fix.

What I checked on my side:
- Merge tree against current main: `cargo test --lib` 2818 passed, `cargo clippy --all-targets -D warnings` clean; no conflict with your companion #7262 either.
- Reverting only the `codex_config.rs` hunk makes the new takeover test fail at the "explicit provider" assertion, so the regression coverage is real.
- Codex 0.153 source: `openai_base_url` feeds `create_openai_provider`, and there is no top-level `base_url` in `ConfigToml` — matches your reading.

On #7217: I agree with your scoping. Preset cards carry `model_provider = "custom"`, so that reporter's bypass is more likely the Desktop app overriding the provider when signed in; I'll follow up on the issue separately rather than claim this PR fixes it.

Two small things, neither blocking, in case you feel like a follow-up:
1. The placeholder-cleanup fallback (`cleanup_codex_takeover_placeholders_in_live`, only reached when the SSOT write-back is skipped) strips `base_url` and the bearer from the migrated `cc-switch` table but leaves the table and the selector behind, so Codex would route to an address-less custom provider instead of falling back to the built-in one as it did before. Dropping the whole cc-switch-owned table (and `model_provider`) when its `base_url` matched the local proxy would close that.
2. `model_provider = ""` counts as "present" for `contains_key` but as "absent" for `active_codex_model_provider_id` and the migrator; worth aligning the two (pre-existing edge, not introduced here).

The `println!("takeover_fixture=…")` in the new test can go too — I assume it was for feeding the app-server run.

Thanks again.

PR #7280, opened 2026-09-10
fix(codex): coalesce adjacent commentary with pending tool calls

### LINKED ISSUE #6529: Codex Responses→Chat splits commentary and tool calls into consecutive assistant messages

fszcd:
## Summary

Codex Responses -> Chat Completions conversion emitted two consecutive `assistant` messages when one Responses model turn contained a commentary `message` directly followed by `function_call` items. Chat models can imitate the first text-only assistant message as a complete turn and return `finish_reason=stop` before emitting the tool call that Codex expects, so long agent tasks end after a progress update.

中文概述：同一 Responses turn 中的 commentary `message` 与紧随其后的 `function_call` 现在会合并为同一条 Chat assistant 消息，避免 Chat 模型把纯进度播报当作完整回合而在工具调用前提前 `stop`。

Fixes #6529 (supersedes #6530, including the review feedback there).

## What changed

- merge pending tool calls into a directly adjacent assistant message that has no existing `tool_calls`; preserve its commentary content
- keep the new-assistant-message path for every other boundary shape (user/tool boundaries, prior tool-call batches, media flushes)
- attach pending reasoning per segment, so a turn whose commentary carries reasoning `A` and whose calls carry `A` and `B` serializes as `A\n\nB` instead of duplicating `A` (addresses the review comment on #6530)
- keep `backfill_tool_call_reasoning_placeholders` behavior for coalesced calls without any reasoning (thinking upstreams still get a non-empty `reasoning_content`)

## Edge cases covered by regressions

- coalescing with reasoning on the commentary, on the call, or as a separate `reasoning` item between them
- parallel tool calls with repeated and distinct reasoning segments (per-segment dedup)
- user boundary between a prior assistant message and a new tool-call turn (must stay unmerged)
- commentary without any reasoning (placeholder backfill still applies)
- `custom_tool_call` coalescing
- media from a preceding tool result (synthetic user message ordering is unchanged, coalescing still applies to the following commentary)
- multi-round history: no assistant message except the final answer is text-only

Also verified compatibility with `codex_chat_history` restore: restored `function_call` items are inserted directly before their outputs, so they flow through the same coalescing path. The production request path has a single conversion entry point in `forwarder.rs`.

## Real-world validation

Replayed 124 `response_item` entries extracted from an actual Codex agent session (the exact history shape that triggered premature stops with a Chat upstream):

- with the fix: 77 messages, 38 tool messages, 13 commentary+tool-call turns coalesced, and zero occurrences of a text-only assistant message followed by a tool-call assistant message
- with the merge temporarily disabled: the same replay fails the invariant check (RED), confirming the regression actually exercises the bug

## Standard validation

- `cargo test --lib --manifest-path src-tauri/Cargo.toml proxy::providers::transform_codex_chat` (106 passed, including the 8 new regressions) on the pinned Rust 1.95 toolchain
- `cargo check --tests --manifest-path src-tauri/Cargo.toml`

### DIFF
--- a/src-tauri/src/proxy/providers/transform_codex_chat.rs
+++ b/src-tauri/src/proxy/providers/transform_codex_chat.rs
@@ -4195,6 +4195,214 @@ mod tests {
     }
 
     #[test]
+    fn responses_request_to_chat_coalesces_when_reasoning_sits_between_commentary_and_call() {
+        let mut call = test_function_call("call_1");
+        call["reasoning_content"] = json!("need to update the file");
+
+        let result = convert_test_input(vec![
+            json!({
+                "type": "message",
+                "role": "assistant",
+                "content": [
+                    {"type": "output_text", "text": "Part 1 written. Appending sections 4-5."}
+                ]
+            }),
+            json!({
+                "type": "reasoning",
+                "summary": [
+                    {"type": "summary_text", "text": "need to update the file"}
+                ]
+            }),
+            call,
+            test_function_output("call_1", json!("Success")),
+        ]);
+        let messages = result_messages(&result);
+
+        assert_eq!(message_roles(&result), vec!["assistant", "tool"]);
+        assert_eq!(
+            messages[0]["content"],
+            "Part 1 written. Appending sections 4-5."
+        );
+        assert_eq!(messages[0]["tool_calls"][0]["id"], "call_1");
+        assert_eq!(messages[0]["reasoning_content"], "need to update the file");
+    }
+
+    #[test]
+    fn responses_request_to_chat_backfills_reasoning_placeholder_for_coalesced_call() {
+        let result = convert_test_input(vec![
+            json!({
+                "type": "message",
+                "role": "assistant",
+                "content": [
+                    {"type": "output_text", "text": "Running the build now."}
+                ]
+            }),
+            test_function_call("call_build"),
+            test_function_output("call_build", json!("Build succeeded")),
+        ]);
+        let messages = result_messages(&result);
+
+        assert_eq!(message_roles(&result), vec!["assistant", "tool"]);
+        assert_eq!(messages[0]["content"], "Running the build now.");
+        assert_eq!(messages[0]["tool_calls"][0]["id"], "call_build");
+        assert_eq!(messages[0]["reasoning_content"], "tool call");
+    }
+
+    #[test]
+    fn responses_request_to_chat_coalesces_custom_tool_call_with_commentary() {
+        let input = json!({
+            "model": "kimi-k3",
+            "tools": [{
+                "type": "custom",
+                "name": "apply_patch",
+                "description": "Apply a patch to files."
+            }],
+            "input": [
+                json!({
+                    "type": "message",
+                    "role": "assistant",
+                    "content": [
+                        {"type": "output_text", "text": "Applying the next patch."}
+                    ]
+                }),
+                json!({
+                    "type": "custom_tool_call",
+                    "id": "ctc_1",
+                    "call_id": "call_patch",
+                    "name": "apply_patch",
+                    "input": "*** Begin Patch\n*** End Patch"
+                }),
+                json!({
+                    "type": "custom_tool_call_output",
+                    "call_id": "call_patch",
+                    "output": {"text": "Success"}
+                })
+            ]
+        });
+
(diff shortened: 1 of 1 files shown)


farion1231:
@codex review

farion1231:
Merged, thank you @fszcd! Nice first contribution: the fix stays exactly as narrow as the bug (only a directly adjacent assistant message without `tool_calls` is coalesced), and the per-segment reasoning dedup closes the parallel-call gap that the earlier attempt left open.

What I checked on my side:
- Merge tree against current main: `cargo fmt --check` and clippy clean, `transform_codex_chat` 106 passed, `codex_chat_history` 8 passed, the whole `proxy::` suite 1472 passed.
- Disabling only the `merge_pending_tool_calls_into_adjacent_assistant` gate makes 7 of the 8 new tests fail (the user-boundary test is a guard and passes both ways), so the regressions lock the actual bug.
- Traced the three code paths that push an assistant message: each flushes pending tool calls and media first, so the coalescing gate can only fire within one model turn, never across a user boundary. The `codex_chat_history` restore inserts recovered `function_call` items right before their outputs, so it flows through the same path as you described.
- The root cause is our own round trip: the response side splits one Chat `content + tool_calls` message into a `message` and `function_call` item, and the request side used to project those back as two assistant messages. After this fix the Chat history matches what the upstream model originally produced.

Credit also to @BigStrongSun, whose #6530 first landed on this approach; the release note will name both of you.

Thanks again.

fszcd:
Thanks for the thorough review and merge! Really appreciate you verifying the gate behavior and tracing the restore path in \codex_chat_history\ — glad the fix closes the loop on the round trip. Credit to @BigStrongSun for the original direction in #6530. Happy to help with any follow-ups in the next release.

PR #7287, opened 2026-09-10
fix(proxy): clamp sub-floor max_tokens to the Responses API minimum

### LINKED ISSUE #7103: [Bug] Claude Desktop probe fails when Responses upstream requires max_output_tokens >= 16
### Self Checks / 自检

### CC Switch Version / 版本号

3.20.1

### Operating System / 操作系统

macOS 26.6.2

### Related App / 涉及应用

Other: Claude Desktop 1.44121.4

For comparison, Claude Code CLI 2.1.259 works with the same upstream model through CC Switch.

### Steps to Reproduce / 重现步骤

1. Configure a Claude Desktop provider in local routing mode.
2. Set the upstream API format to `openai_responses`.
3. Set the upstream base URL to `https://opencode.ai/zen/v1`.
4. Map a Claude Desktop-safe route such as `claude-haiku-4-5` or `claude-fable-5` to `muse-spark-1.3-contributor-free`.
5. Enable the CC Switch local router and select/check this provider in Claude Desktop.

### Expected Behavior / 期望行为

Claude Desktop's configured-model availability check should succeed, allowing the model to be selected and used through the local gateway.

### Actual Behavior / 实际行为

Claude Desktop reports `Configured model not available`. The gateway returns HTTP 400:

```text
Gateway rejected model "claude-haiku-4-5" (HTTP 400)
requestUrl: http://127.0.0.1:15721/claude-desktop/v1/messages
responseBody: {
  "model": "muse-spark-1.3-contributor-free",
  "error": {
    "param": "max_output_tokens",
    "type": "invalid_request_error",
    "message": "Upstream request failed: `max_output_tokens` The number must be `>= 16`."
  }
}
```

The same Muse model succeeds through Claude Code CLI because normal CLI requests use an output-token limit greater than 16.

### Root Cause / 根因分析

Claude Desktop sends a very small model-availability probe with `max_tokens=1`. CC Switch's Anthropic-to-Responses conversion currently copies that value directly:

```rust
if let Some(v) = body.get("max_tokens") {
    result["max_output_tokens"] = v.clone();
}
```

OpenCode Zen's Muse Responses backend requires `max_output_tokens >= 16`, so the probe fails before Claude Desktop allows the configured model to be used. This is a probe compatibility issue rather than an invalid API key, missing model access, or an incorrect model route.

Relevant source:
`src-tauri/src/proxy/providers/transform_responses.rs`

### Suggested Fix / 修复建议

Please consider supporting a provider-configurable minimum for `max_output_tokens`, or applying a narrowly scoped compatibility clamp for Claude Desktop availability probes / upstreams that declare a minimum. For this provider, the effective conversion would be:

```text
max_output_tokens = max(requested_value, 16)
```

A configurable or narrowly scoped solution would avoid changing the semantics of unrelated Responses providers that legitimately accept one-token requests.

No API keys or account credentials are included in this report.

SailingLoong:
## Summary / 概述

Responses API 拒绝 `max_output_tokens` 小于 16（"The number must be `>= 16`"）。Claude Desktop 的模型可用性探测发 `max_tokens=1`，而 Anthropic→Responses 转换此前原样拷贝该值，严格上游（如 OpenCode Zen Go）直接 400，Claude Desktop 便判定「配置的模型不可用」，本地路由下映射的模型无法选用。

转换时把 1..=15 的预算抬到 16；16 及以上、0、非整数保持原有透传语义。正常会话请求（Claude Code CLI 的输出上限远大于 16）转换字节不变，不影响 prompt cache；Codex OAuth 路径随后照旧删除 `max_output_tokens`，不受影响。

## Related Issue / 关联 Issue

Fixes #7103

## Verification / 验证

- 新增回归测试 `test_anthropic_to_responses_clamps_sub_floor_max_tokens`：1/8/15 → 16；16/1024 原样；0/字符串/浮点原样透传。修复前失败、修复后通过。
- `cargo test --lib`：2824 passed / 0 failed。
- `cargo fmt --check`、`cargo clippy --lib --all-targets`（0 warning）、`git diff --check` 通过。

## Checklist / 检查清单

- [ ] `pnpm typecheck`（无前端改动，未运行）
- [ ] `pnpm format:check`（无前端改动，未运行）
- [x] `cargo clippy` passes
- [x] 无用户界面文案变更

### DIFF
--- a/src-tauri/src/proxy/providers/transform_responses.rs
+++ b/src-tauri/src/proxy/providers/transform_responses.rs
@@ -24,6 +24,11 @@ use super::reasoning_bridge::{
 
 pub(crate) const TOOL_RESULT_ERROR_MARKER: &str = "[cc-switch:tool-result-error]";
 
+/// OpenAI Responses API 拒绝 `max_output_tokens` 小于 16。Anthropic 侧的合法
+/// 小预算探测请求（如 Claude Desktop 的模型可用性探测发 `max_tokens=1`）
+/// 会被严格网关整体拒绝，转换时把低于下限的预算抬到最小值而不是让请求失败。
+pub(crate) const RESPONSES_MIN_MAX_OUTPUT_TOKENS: u64 = 16;
+
 fn has_http_url_scheme(value: &str) -> bool {
     value
         .get(.."http://".len())
@@ -1805,9 +1810,20 @@ pub fn anthropic_to_responses(
         result["input"] = json!(input);
     }
 
-    // max_tokens → max_output_tokens (Responses API uses max_output_tokens for all models)
+    // max_tokens → max_output_tokens (Responses API uses max_output_tokens for
+    // all models). The Responses API rejects values below 16 ("The number must
+    // be `>= 16`"), while Anthropic clients legitimately send tiny probe
+    // budgets (Claude Desktop's model-availability probe uses max_tokens=1),
+    // so sub-floor budgets are clamped up to the minimum instead of failing
+    // the whole request. 0 and non-integer values keep their existing
+    // pass-through semantics.
     if let Some(v) = body.get("max_tokens") {
-        result["max_output_tokens"] = v.clone();
+        result["max_output_tokens"] = match v.as_u64() {
+            Some(n) if (1..RESPONSES_MIN_MAX_OUTPUT_TOKENS).contains(&n) => {
+                json!(RESPONSES_MIN_MAX_OUTPUT_TOKENS)
+            }
+            _ => v.clone(),
+        };
     }
 
     // 直接透传的参数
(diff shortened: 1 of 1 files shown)


farion1231:
@codex review

farion1231:
Merged, thank you @SailingLoong! Three PRs went after #7103; yours is the one that fixes it at exactly the right size — Responses is the only format with a 16-token floor, so a clamp scoped to that one converter is the whole fix, and the 0 / non-integer pass-through keeps invalid budgets as an upstream error instead of silently rewriting them.

What I checked on my side:
- Merge tree against current `main` (b78192e): `cargo fmt --check` clean, `cargo test --lib` 2830 passed / 0 failed.
- Reverting only the clamp hunk while keeping your test makes `test_anthropic_to_responses_clamps_sub_floor_max_tokens` fail with `left: 1, right: 16`, so the regression coverage locks the actual bug.
- Ordering vs. the Codex OAuth path holds: the clamp runs before the OAuth branch strips `max_output_tokens`, so ChatGPT-backed providers are unaffected.
- Nothing inside cc-switch itself sends a sub-16 budget; the only callers this touches are client-side probes (Claude Desktop's availability check, Claude Code's startup quota probe), and a probe that now gets up to 16 tokens instead of 1 costs nothing measurable.

One nit I left as-is: `RESPONSES_MIN_MAX_OUTPUT_TOKENS` is `pub(crate)` but only used in that file; it can go private whenever someone next touches the module.

PR #7305, opened 2026-09-11
feat(proxy): strip Claude Code fingerprint headers on non-Anthropic transform paths

cuiyf5516:
## Motivation and use case

When CC Switch's local proxy converts Claude Code requests to OpenAI Chat, OpenAI Responses, or Gemini format, the forwarded requests can still carry Claude Code and Stainless SDK identification headers. Although the request format has changed, an upstream gateway can still use these headers to identify the client.

This PR adds a per-provider, opt-in setting to remove `x-app`, `claude-code-session-id`, `x-stainless-*`, and `x-claude-code-*` on these conversion paths. This reduces the client and session identification metadata forwarded upstream and provides a compatibility option for gateways that treat these headers specially.

For example, when using Claude Code with an OpenAI-compatible gateway that applies special handling to Claude Code-specific headers, users can enable this setting for that provider so converted requests no longer carry those headers.

To use it, enable **Strip Claude Code fingerprint headers** in the Claude provider's advanced settings and route requests through the local proxy using one of the supported API format conversions. The setting is disabled by default. Anthropic passthrough, Codex-to-Anthropic impersonation, and Copilot retain their existing behavior.

This option only removes the specified headers. `User-Agent` remains controlled by the existing custom User-Agent setting, so this does not fully conceal client identity or guarantee that a gateway will accept a request.

## Summary
- add an opt-in Claude provider setting to strip Claude Code client-fingerprint headers
- remove x-app, x-stainless-*, and x-claude-code-* headers on Claude-to-OpenAI/Gemini transform paths
- preserve headers for Anthropic passthrough, Codex-to-Anthropic impersonation, and Copilot
- add localized UI copy and regression tests

## Tests
- pnpm exec vitest run tests/components/ClaudeFormFields.test.tsx
- pnpm typecheck
- cargo test --manifest-path src-tauri/Cargo.toml --lib strip_claude_code_fingerprint_defaults_to_disabled
- cargo test --manifest-path src-tauri/Cargo.toml --lib claude_code_fingerprint_headers_are_dropped_for_non_anthropic_upstreams
- `cargo fmt --check --manifest-path src-tauri/Cargo.toml` — passed locally with Rust 1.95

### DIFF
--- a/src-tauri/src/provider.rs
+++ b/src-tauri/src/provider.rs
@@ -544,6 +544,14 @@ pub struct ProviderMeta {
     /// Custom User-Agent for local proxy routing.
     #[serde(rename = "customUserAgent", skip_serializing_if = "Option::is_none")]
     pub custom_user_agent: Option<String>,
+    /// Claude→非 Anthropic 转换路径是否剥离 Claude Code 客户端指纹头
+    /// （x-app / x-stainless-* / x-claude-code-*）。缺省（None）不剥离；
+    /// 显式 true 开启——用于向指纹敏感的网关隐藏 Claude Code 客户端特征。
+    #[serde(
+        rename = "stripClaudeCodeFingerprint",
+        skip_serializing_if = "Option::is_none"
+    )]
+    pub strip_claude_code_fingerprint: Option<bool>,
     /// Local proxy request overrides applied to the transformed upstream request.
     #[serde(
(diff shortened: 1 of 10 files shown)


farion1231:
@codex review

cuiyf5516:
@farion1231 Fixed the formatting issue. Could you please take another look?

PR #7327, opened 2026-09-12
fix(sessions): preserve text selection while scrolling

### LINKED ISSUE #7325: Text selection is lost when scrolling through session messages
## Bug description

In the session manager, selecting text across several messages and scrolling down can make the selection shrink or disappear without clicking anywhere else.

This happens in v3.20.3 on Windows. I reproduced it in a session with 432 messages: after selecting text across the first few messages, scrolling down gradually removed the selected content until nothing remained selected.

## Cause

The message list is virtualized. When a selected message moves outside the rendered range, its DOM nodes are unmounted, so the browser loses part or all of the text selection.

## Expected behavior

Selected text should remain selected when scrolling through the conversation. Dragging and scrolling should also allow the selection to continue across messages.

MOLAaaaaaaa:
## Summary / 概述

Selecting text across session messages and scrolling could make the selection shrink or disappear. The virtualized list unmounted the selected DOM nodes once they moved outside the rendered range.

This keeps the selected messages and the range between them and the viewport mounted while a native text selection exists. Clearing the selection restores the normal virtualized range.

修复会话正文滚动时选中文字逐渐减少或消失的问题。保留选区覆盖的消息及其与可视区域之间的消息，清除选区后恢复正常虚拟化。

Closes #7325

## Tests

- `pnpm typecheck`
- 16 focused tests passed, including reverse selection, element-node endpoints, clearing the selection and switching sessions.
- Chromium verification with 432 fixture messages: selection remained unchanged after scrolling 7,800 pixels; mounted rows returned from 52 to 15 after clearing it.

Selecting across a large range keeps more messages mounted for the lifetime of that selection.

### DIFF
--- a/src/components/sessions/SessionManagerPage.tsx
+++ b/src/components/sessions/SessionManagerPage.tsx
@@ -1,5 +1,6 @@
 import { useCallback, useEffect, useMemo, useRef, useState } from "react";
 import { useSessionSearch } from "@/hooks/useSessionSearch";
+import { useMessageSelectionRange } from "@/hooks/useMessageSelectionRange";
 import { useTranslation } from "react-i18next";
 import { useVirtualizer } from "@tanstack/react-virtual";
 import { toast } from "sonner";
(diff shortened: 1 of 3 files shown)


farion1231:
@codex review

PR #7328, opened 2026-09-12
feat(sessions): select and copy multiple messages

### LINKED ISSUE #7326: Select and copy multiple messages in the session manager
## Feature request

Sometimes I want to copy several messages from a session, for example a question and its replies, rather than copy each message separately.

The session manager already has a conversation table of contents. It would be useful to select a turn there and copy the selected messages together.

## Expected behavior

- Select individual messages, or select a question and its following replies from the table of contents.
- Select all messages or clear the selection.
- Copy the selected messages in their original order, including the role and complete message content, even when a long message is collapsed in the detail pane.

MOLAaaaaaaa:
## Summary / 概述

Copying several messages from a session currently requires copying each message separately or dragging a text selection across the conversation.

This adds message selection to the detail pane and turn selection to the existing table of contents. A turn includes a user question and the messages before the next question. Selected messages are copied in their original order with role labels and complete content, including collapsed long replies.

会话详情支持逐条选择消息，也可以在现有目录中选择整轮提问和回复，一次复制所选消息。复制保留原始顺序和完整正文，支持全选和清空选择。

Closes #7326

## Tests

- `pnpm typecheck`
- 17 focused tests passed, including complete long-message copying, selection reset on session change, partial turn selection and navigation.
- Chromium verification of multi-turn copying, select all/clear, individual selection and the narrow-window table of contents.
- Updated en, ja, zh and zh-TW translations.

### DIFF
--- a/src/components/sessions/SessionManagerPage.tsx
+++ b/src/components/sessions/SessionManagerPage.tsx
@@ -69,6 +69,7 @@ import {
   getBaseName,
   getProviderIconName,
   getProviderLabel,
+  getRoleLabel,
   getSessionDirectoryGroupKey,
   getSessionKey,
   groupSessionsByProviderAndDirectory,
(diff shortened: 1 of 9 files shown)


farion1231:
@codex review

PR #7348, opened 2026-09-12
fix(misc): ignore shell startup output when probing tool versions in WSL

### LINKED ISSUE #7347: [Bug] WSL: 本地环境检查的「当前版本」显示为 Ubuntu 版本号（MOTD 混入版本探测输出） / WSL tool version shows the Ubuntu release number
### CC Switch Version / 版本号

v3.20.3

### Operating System / 操作系统

Windows 11 + WSL2 (Ubuntu 24.04)

### Related App / 涉及应用

Claude Code (any tool detected through WSL)

## 现象 / What happens

设置 → 本地环境检查，WSL 里的 Claude Code 卡片显示：

```
WSL · Ubuntu-24.04
当前版本   24.04.4     ← Ubuntu 的发行版号，不是 Claude Code 的版本
最新版本   2.1.270
```

`24.04.4` 来自 `/etc/os-release` 的 `PRETTY_NAME="Ubuntu 24.04.4 LTS"`。实际装的 Claude Code 是 2.1.270。

The WSL Claude Code card reports the Ubuntu release (`24.04.4`) as the current version. The installed Claude Code is 2.1.270.

**It is intermittent**: the same probe returns the right version later the same day, which makes it look unreproducible. See "Why it is intermittent" below.

## 根因 / Root cause

`try_get_version_wsl` (no forced shell) runs, via `wsl.exe -d <distro> -- sh -c`:

```sh
"${SHELL:-sh}" -lic 'claude --version' 2>/dev/null || "${SHELL:-sh}" -lc '…' 2>/dev/null || "${SHELL:-sh}" -c '…'
```

and passes the whole stdout to `extract_version`, which returns the **first** `\d+\.\d+\.\d+` match.

`-lic` starts a login + interactive shell. On Ubuntu under WSL, `/etc/profile.d/update-motd.sh` prints the MOTD in interactive shells:

```sh
# /etc/profile.d/update-motd.sh (Ubuntu 24.04)
if echo "$-" | grep -qs "i"; then
  if [ ! -e "$HOME/.hushlogin" ] && [ -z "$MOTD_SHOWN" ] && ! find $stamp -newermt 'today 0:00' …; then
    update-motd --show-only          # prints to stdout
    touch $stamp                     # stamp="$HOME/.motd_shown"
```

`update-motd --show-only` starts with:

```
Welcome to Ubuntu 24.04.4 LTS (GNU/Linux 6.6.87.2-microsoft-standard-WSL2 x86_64)
```

That line precedes `2.1.270 (Claude Code)` on stdout, so the first match is `24.04.4`.

Reproduced end to end through `wsl.exe` with a throwaway `HOME` (so no stamp exists): the probe's stdout begins with the sudo hint from `/etc/bash.bashrc`, then the MOTD above, then the tool's version line. Feeding MOTD + `2.1.270 (Claude Code)` through the same regex yields `24.04.4`.

## Why it is intermittent / 为什么时有时无

The MOTD is shown **once per day** per `$HOME/.motd_shown`. Whichever interactive login shell runs first that day — often CC Switch's own probe — prints it and touches the stamp. Every later probe that day gets clean output. So the wrong version appears after the first probe of the day and disappears on refresh or the next check.

## Relation to existing PRs / 与现有 PR 的关系

- #5817 isolates tool output from rc files for `try_get_version`, but explicitly keeps `try_get_version_wsl` sharing stdout with the shell's startup files (its comment: 「仍与 rc 共用 stdout 的 `try_get_version_wsl` 保留」). On that path it only strips OSC escape payloads, which does not cover plain-text MOTD lines.
- #3512 changes the flag for the Linux build running inside WSL (`try_get_version`), not the Windows → `wsl.exe` probe.
- #2824 / #3937 are the same class of symptom on macOS (rc output winning the first semver match).

## 建议修复 / Proposed fix

Keep the login shell (it provides PATH), but make the payload print a sentinel before running the tool, and parse only what follows the last sentinel:

```sh
"${SHELL:-sh}" -lic 'echo __CCSWITCH_VERSION__; claude --version'
```

Anything the startup files print — MOTD, the sudo hint, `Using Node vX.Y.Z`, OSC sequences glued to a line — lands before the marker. Taking the text after the **last** occurrence also handles the `||` fallback chain concatenating several attempts. If the marker is absent (the shell failed before running the payload), the current behaviour is kept. I'm happy to send a PR.

## 临时绕过 / Workaround

Inside the distro: `touch ~/.hushlogin` stops the MOTD, and the version is detected correctly.

lyfuci:
## Summary / 概述

On Windows, the local environment check for a tool installed in WSL can show the **distro's release number** as the tool's current version — e.g. `24.04.4` for Claude Code 2.1.270 on Ubuntu 24.04.

`try_get_version_wsl` runs `"${SHELL:-sh}" -lic '<tool> --version'` through `wsl.exe` and passes the whole stdout to `extract_version`, which returns the first `x.y.z` it finds. A login + interactive shell runs the distro's startup files first, and their output lands on the same stdout. On Ubuntu, `/etc/profile.d/update-motd.sh` prints the MOTD in interactive shells once a day:

```
Welcome to Ubuntu 24.04.4 LTS (GNU/Linux 6.6.87.2-microsoft-standard-WSL2 x86_64)
...
2.1.270 (Claude Code)
```

The once-a-day stamp (`$HOME/.motd_shown`) is why this looks unreproducible: the wrong version appears after the first probe of the day and disappears on the next refresh.

WSL 下工具版本探测走 login+交互 shell，发行版启动文件（Ubuntu 的 update-motd 每天打印一次欢迎语）和 `--version` 输出共用 stdout，`extract_version` 取到的第一个版本号是 `24.04.4`。

**Fix.** The payload now prints a sentinel before running the tool — `echo __CCSWITCH_VERSION__; <tool> --version` — and parsing looks only at what follows its **last** occurrence. Whatever the startup files print (MOTD, the sudo hint from `/etc/bash.bashrc`, `Using Node vX.Y.Z`, OSC sequences) lands before it.

- **Last occurrence**: the `-lic || -lc || -c` fallback chain can print the sentinel several times.
- **Substring, not whole line**: an OSC sequence without a trailing newline can be glued in front of it.
- **Sentinel absent** (the shell failed before running the payload): the output is used as before, so behaviour is unchanged in that case.
- The existing stderr fallback for tools that print their version there is kept; with the sentinel, "empty" now means empty after it.
- The failure branch also uses the post-sentinel stdout, so startup-file text can't be mistaken for a "not found" diagnostic.

The helpers are plain functions next to `extract_version`, not `cfg(windows)`, so they are unit-tested on every platform; they carry `cfg_attr(not(windows), allow(dead_code))` because only the Windows probe calls them.

## Related Issue / 关联 Issue

Fixes #7347

**Relation to #5817.** #5817 isolates tool output from rc files for `try_get_version`, but keeps `try_get_version_wsl` sharing stdout with the shell and only strips OSC payloads there — which doesn't cover plain-text MOTD lines. Both PRs touch the output-handling lines in `try_get_version_wsl`, so whichever lands second needs a small rebase. To keep that overlap small, this PR does **not** change `extract_version` or its return type (#5817 turns it into `Option`), and the new test doesn't depend on it.

## Verification / 验证

**End to end through the real `wsl.exe`** (Windows 11 → WSL2 Ubuntu 24.04, invoked the same way as `Command::new("wsl.exe").args(["-d", distro, "--", "sh", "-c", cmd])`):

| Case | Old parsing (first match in whole stdout) | New parsing (after last sentinel) |
|---|---|---|
| Normal HOME, MOTD already shown today | `2.1.270` | `2.1.270` |
| Throwaway HOME, so the MOTD fires | **`24.04.4`** | `2.1.270` |

The first row confirms the sentinel survives the `wsl.exe → sh -c → $SHELL -lic` quoting chain and the normal case is unaffected. The second reproduces the reported card exactly and shows the fix. (The throwaway HOME is test scaffolding to trigger the MOTD without touching the real stamp.)

**Unit test** `version_probe_sentinel_drops_shell_startup_output`: MOTD before the sentinel, repeated sentinels from the fallback chain, an OSC sequence glued to the sentinel, an empty payload (stderr fallback), and no sentinel at all.

**What I could not do:** compile the `cfg(windows)` body of `try_get_version_wsl` locally. The crate needs a C toolchain for the Windows target (`ring`, `aws-lc-sys`, bundled `libsqlite3-sys`) and I don't have one here. The change inside that function is limited to building the payload and slicing stdout with the tested helper. The `windows-latest` backend job will compile and lint it — note that workflows on this PR need maintainer approval to run.

## Screenshots / 截图

| WSL Claude Code card: 当前版本 `24.04.4`, 最新版本 `2.1.270` (after the first probe of the day) | 当前版本 `2.1.270` — verified via the end-to-end table above; no Windows build available here for a UI screenshot |

## Checklist / 检查清单

- [ ] `pnpm typecheck` passes — not run: no TypeScript changes
- [ ] `pnpm format:check` passes — not run: no frontend changes
- [x] `cargo clippy` passes (if Rust code changed) — `cargo clippy --manifest-path src-tauri/Cargo.toml -- -D warnings` on Linux
- [x] Updated i18n files if user-facing text changed — N/A, no user-facing text changed

Also run: `cargo fmt --check`; `cargo test --lib -- version_probe_sentinel test_extract_version` (2 passed).

### DIFF
--- a/src-tauri/src/commands/misc.rs
+++ b/src-tauri/src/commands/misc.rs
@@ -1067,6 +1067,32 @@ async fn fetch_pypi_latest_version(client: &reqwest::Client, package: &str) -> O
 static VERSION_RE: Lazy<Regex> =
     Lazy::new(|| Regex::new(r"\d+\.\d+\.\d+(-[\w.]+)?").expect("Invalid version regex"));
 
+/// WSL 版本探测载荷先打印的哨兵。用户 shell 的启动文件（Ubuntu 的 update-motd、
+/// `/etc/bash.bashrc` 的 sudo 提示、nvm 的 `Using Node vX.Y.Z` 等）输出都在它前面，
+/// 解析时只看它之后的内容（#7347）。
+#[cfg_attr(not(windows), allow(dead_code))]
+const VERSION_PROBE_SENTINEL: &str = "__CCSWITCH_VERSION__";
+
+/// 版本探测交给用户 shell 执行的命令：先打哨兵，再跑 `--version`。
+/// 不含单引号，可以直接嵌进外层 `'...'`。
+#[cfg_attr(not(windows), allow(dead_code))]
+fn version_probe_payload(tool: &str) -> String {
+    format!("echo {VERSION_PROBE_SENTINEL}; {tool} --version")
+}
+
+/// 取哨兵**最后一次**出现之后的输出。
+///
+/// 取最后一次：兜底链 `-lic || -lc || -c` 失败重试时会多次打印哨兵。
+/// 按子串而非整行查找：OSC 序列可能没有换行、直接粘在哨兵前面。
+/// 找不到哨兵（shell 在执行载荷之前就失败）时原样返回，保持原有行为。
+#[cfg_attr(not(windows), allow(dead_code))]
+fn after_version_sentinel(output: &str) -> &str {
+    match output.rfind(VERSION_PROBE_SENTINEL) {
+        Some(i) => output[i + VERSION_PROBE_SENTINEL.len()..].trim(),
+        None => output,
+    }
+}
+
 /// 从版本输出中提取纯版本号
 fn extract_version(raw: &str) -> String {
     VERSION_RE
@@ -1322,17 +1348,18 @@ fn try_get_version_wsl(
             default_flag_for_shell(shell)
         };
 
-        (shell.to_string(), flag, format!("{tool} --version"))
+        (shell.to_string(), flag, version_probe_payload(tool))
     } else {
+        let payload = version_probe_payload(tool);
         let cmd = if let Some(flag) = force_shell_flag {
             if !is_valid_shell_flag(flag) {
                 return ShellProbe::NotFound(format!("[WSL:{distro}] invalid shell flag: {flag}"));
             }
-            format!("\"${{SHELL:-sh}}\" {flag} '{tool} --version'")
+            format!("\"${{SHELL:-sh}}\" {flag} '{payload}'")
         } else {
             // 兜底：自动尝试 -lic, -lc, -c
             format!(
-                "\"${{SHELL:-sh}}\" -lic '{tool} --version' 2>/dev/null || \"${{SHELL:-sh}}\" -lc '{tool} --version' 2>/dev/null || \"${{SHELL:-sh}}\" -c '{tool} --version'"
+                "\"${{SHELL:-sh}}\" -lic '{payload}' 2>/dev/null || \"${{SHELL:-sh}}\" -lc '{payload}' 2>/dev/null || \"${{SHELL:-sh}}\" -c '{payload}'"
             )
         };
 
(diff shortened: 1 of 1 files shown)


farion1231:
@codex review

farion1231 (approved):
感谢 @lyfuci 的修复和详细的复现说明！通过标记隔离 WSL shell 启动输出，解决了 Ubuntu 欢迎信息中的版本号被误识别的问题，也考虑到了 shell 多次回退和 OSC 控制序列等边界情况。已完成审查，未发现阻塞问题，相关 CI 均已通过。感谢贡献！

PR #7370, opened 2026-09-13
fix(proxy): clamp xhigh reasoning effort to high for o-series models

SailingLoong:
Follow-up to the second suggestion in #7318's review: `xhigh` can reach o-series models (o1 / o3 / o4-mini), which only accept low / medium / high.

## Problem / 根因

Since #7318 passes `xhigh` through verbatim (and `max` / `thinking: adaptive` already mapped to it before), an o-series target receives `reasoning_effort: "xhigh"` — a value that model family does not accept. The max/adaptive paths were pre-existing; explicit `/effort xhigh` became reachable with #7318.

## Fix

Add `clamp_reasoning_effort(model, effort)` and apply it at both injection sites (Chat Completions `reasoning_effort`, Responses `reasoning.effort`): when the target is an o-series model, `xhigh` is downgraded to `high` — the closest supported intent, mirroring how grok-4.5 documents a silent xhigh→high fallback. Every other effort value and every non-o-series model is byte-identical to before.

## Tests

- Unit: `clamp_reasoning_effort` downgrades for `o3-mini` / `o1` (incl. mixed case) and passes through for `gpt-5.4` / `grok-4.6` and non-xhigh efforts.
- Chat path: explicit `xhigh` and `thinking: adaptive` against `o3-mini` both emit `high`.
- Responses path: `xhigh` against `o3-mini` emits `high`; against `gpt-5.4` stays `xhigh`.
- `cargo test --lib` 2861 passed / 0 failed; `cargo fmt --check` clean.

Happy to adjust.


farion1231:
@codex review

PR #7378, opened 2026-09-14
fix(proxy): omit missing tool descriptions in the Codex responses-to-chat converter

SailingLoong:
Follow-up invited in #7319's review: the Codex Responses→Chat converter has one remaining `unwrap_or(Value::Null)` for tool descriptions.

## Problem / 根因

`responses_function_tool_to_chat_tool` (transform_codex_chat.rs) still serializes a missing tool `description` as `"description": null`. Hosted tools and undescribed custom tools reach this branch without the field, and strict OpenAI-compatible upstreams reject the whole request on a null value (400 "expected string, received null") — the same failure mode #7319 fixed for the anthropic→chat and anthropic→responses converters.

## Fix

Build the function object with `name` first, insert `description` only when present and non-null, then `parameters` — the same shape as the merged #7319 converters. Insertion order preserves the original `name→description→parameters` key order, so tools that do carry a description serialize byte-identically to before (prompt-cache prefix safe). Explicit `null` in the source is also omitted, matching #7319.

## Tests

- `responses_request_to_chat_omits_missing_tool_description`: two tools in one request — the one without `description` has no `description` key in the outbound chat tool; the one with it keeps it.
- `cargo test --lib` 2858 passed / 0 failed; `cargo fmt --check` clean; `cargo clippy --lib` clean.

Happy to adjust.


allenxu09 (approved):
LGTM

farion1231:
@codex review

farion1231 (approved):
Thanks @SailingLoong for the focused follow-up and regression test! Reviewed the change and verified that existing descriptions retain their serialized field order, while missing and null descriptions are omitted as intended. Looks good to merge.

PR #7398, opened 2026-09-14
fix(proxy): keep reasoning effort for muse-spark models

### LINKED ISSUE #7397: fix(proxy): reasoning effort silently dropped for muse-spark models
## Self Checks

- [x] This is a bug report (not a question / feature request)
- [x] I searched existing issues (found #7314 / #7318 for the same class of bug on grok models)

## Version

Source: main @ 42ac174d. Behavior verified against the desktop release-channel build (proxy log module paths match `cc_switch_lib::proxy::providers::transform_responses`).

## OS

Windows 11 (proxy) + WSL2 Ubuntu (Claude Code client)

## Related App

Claude Code

## Steps to Reproduce

1. Use a Responses-compatible channel with model `muse-spark-1.3-contributor`.
2. Set `/effort xhigh` (or `low`) in Claude Code.
3. Observe the upstream request: no `reasoning.effort` is sent. Verified by probing:
   - Direct to upstream: `reasoning.effort=max` -> HTTP 400, `xhigh` -> 200, omitted -> defaults to `high`.
   - Through the proxy: `output_config.effort=max` -> 200 (should be 400 if passed through), `output_config.effort=low` on a reasoning-heavy prompt burns the whole output budget as reasoning tokens with empty visible output (i.e. it ran at default `high`, not `low`).

## Expected Behavior

`output_config.effort` is forwarded as `reasoning.effort` (with the existing `max` -> `xhigh` clamp, since Contributor tiers reject `max`).

## Actual Behavior

Effort is silently dropped for all `muse-spark-*` models, so upstream always runs at its default level regardless of `/effort`.

## Root Cause

`supports_reasoning_effort()` in `src-tauri/src/proxy/providers/transform.rs:71-83` whitelists only o-series / gpt-5+ / grok families. The Responses gate at `transform_responses.rs:1841-1847` therefore never injects `reasoning` for `muse-spark-*` models. Same class as #7314 (fixed by #7318 for grok-4.6).

## Suggested Fix

Add `muse-spark` to the whitelist (one line + tests), keeping the existing `max` -> `xhigh` clamp so Contributor tiers never receive the rejected `max` value. PR follows.

jasonrale:
## Summary

`muse-spark-*` models were missing from `supports_reasoning_effort()`, so `output_config.effort` was silently dropped in the Anthropic -> Responses transform and upstream always ran at its default level regardless of `/effort`.

Second commit refines the first: Muse Spark effort passes through **verbatim, including `max`** (Standard tier defines `max` above `xhigh`; clamping it would silently downgrade users who asked for `max`). If a Contributor tier rejects `max`, the upstream 400 surfaces instead of a silent clamp. All other families keep the existing `max` -> `xhigh` mapping — zero behavior change for them.

Same class as #7314, same fix shape as #7318.

## Related issue

Fixes #7397

## Behavior change (scoped)

- Before: `muse-spark-*` + any `/effort` -> upstream default (`high`).
- After: all levels pass through verbatim on `muse-spark-*` (`max` -> `max`).
- GPT / Grok / o-series: untouched.

## Checklist

- [x] Rust tests added/updated (whitelist asserts, `max`-verbatim unit tests for standard + contributor, 3 end-to-end `anthropic_to_responses` tests; pre-existing `resolve` tests now pass an explicit model, expectations unchanged)
- [x] `rustfmt --check` passes on touched files
- [ ] Full `cargo test` / `cargo clippy` — could not run locally (Tauri system deps missing in this environment); logic verified with an isolated harness executing identical assertions against verbatim copies of the changed functions (RED before each change, GREEN after). Please rely on CI.
- [x] No user-facing text changed (no i18n impact)


farion1231:
@codex review

PR #7412, opened 2026-09-15
fix(ui): prevent provider card text overflow

### LINKED ISSUE #7411: fix(ui): prevent provider card text wrapping and usage overflow
## Problem

When a custom provider has a long name or fallback icon initials, the provider card can become difficult to read at narrower widths:

- fallback initials in the left icon can wrap or be squeezed;
- the provider title and status badges can wrap unexpectedly;
- long usage-query values can overflow into the provider identity area and overlap the left-side content.

## Expected behavior

Provider icon initials should remain on one line, provider names should truncate instead of wrapping, and status labels should remain readable. Inline usage data should stay within its allocated area, truncate long fields, and expose the full value through a tooltip/title.

## Scope

This issue tracks a focused UI layout fix for provider cards. Existing related work such as #7035 addresses whitespace parsing in fallback initials, but not the card flex layout or usage overflow described here.

huyikai:
## Summary

Fixes the provider card layout issues described in #7411:

- keep provider title and status badges on a single line, truncating the title when space is limited;
- constrain the inline usage footer to the available card width so long usage values cannot overlap provider identity content;
- preserve the existing full-value `title` tooltip for truncated usage metadata.

## Verification

- `git diff --check`
- Targeted Vitest could not be run because dependencies are not installed in the checkout (`vitest: command not found`).

Closes #7411


farion1231:
@codex review

huyikai:
Added mocked boundary coverage in commit `249ed3e2`. The tests run without a Tauri runtime and cover long custom usage metadata plus long Token Plan labels at the component level. Both ProviderCard layout and UsageFooter boundary tests pass (5 tests total), typecheck and renderer build also pass. The earlier Renderer `invoke` error is expected when running outside Tauri; the mock test setup avoids that limitation.

PR #7436, opened 2026-09-16
chore(dev): change dev server port from 3000 to 3210

iSunker:
## Summary

Moves the Vite dev server port from `3000` to `3210`.

## Scope

Two files, two lines. The port is declared in both, and they must stay in lockstep:

| File | Change |
| --- | --- |
| `vite.config.ts` | `server.port`: `3000` → `3210` |
| `src-tauri/tauri.conf.json` | `build.devUrl`: `http://localhost:3000` → `http://localhost:3210` |

No runtime behaviour, database schema, or user data is affected. The runtime proxy port (`15721`) is untouched.

## Key decisions

**Why two files.** `vite.config.ts` decides where Vite listens; `tauri.conf.json`'s `devUrl` tells Tauri which URL to attach the webview to. They are independent literals with no shared source of truth, so changing one alone leaves `tauri dev` waiting on a dead URL.

**Why `strictPort: true` stays.** It is deliberate and correct. Without it Vite would silently drift to the next free port while Tauri kept attaching to `devUrl`, producing a blank window that is harder to diagnose than a hard failure. The trade-off is that any process already holding the port becomes a fatal startup error.

**Why `3210`.** It sits outside the commonly-contested dev range (`3000`, `5173`, `8080`) and is not registered to a well-known service, so collisions with other local tooling are unlikely.

## Motivation

`3000` is heavily contested on developer machines. When another process already holds it the failure is opaque:

```
Error: Port 3000 is already in use
beforeDevCommand exited with a non-zero status code
```

The process tree is `pnpm → tauri CLI → vite`. Interrupting `pnpm tauri dev` with Ctrl+C terminates only the `pnpm` parent, leaving the Vite child alive and still bound to the port. The next launch then fails even though nothing appears to be running, and the error points at the port rather than at the orphaned process.

Moving the default to `3210` reduces how often that collision occurs.

## Verification

- `pnpm tauri dev` completes; the native window opens and the renderer loads.
- `curl http://localhost:3210/` returns `HTTP 200` with `<title>Claude Code 供应商切换器</title>`.
- The Vite HMR client is injected and hot reload works.
- Backend sync runs normally after launch (session sync and Codex sync both complete).
- `pnpm typecheck` passes (`tsc --noEmit`, exit 0).


farion1231:
@codex review

farion1231:
@codex review

iSunker:
就是说主要端口从3000改成一个不常用的，比如3210

iSunker:
另外我想问一下，这个context机制是怎么走的？cli面板上看不到是不是满了，既怕清理没了，也怕爆缓存

PR #7437, opened 2026-09-16
fix(codex): stop takeover from stripping the provider's model_catalog_json pointer

### LINKED ISSUE #7430: [Bug] Codex 代理接管重写 config.toml 时丢弃 model / model_catalog_json，自定义模型列表清空
### Self Checks / 自检

相关但不同：#6017（异常退出恢复路径不合并通用配置）、#6183（热切换丢 `[windows]`/`[projects]` 段）、#7235（插件/外观设置丢失）。本 issue 是接管重写丢弃**顶层 `model` 与 `model_catalog_json`**，且这两个键**本就存在于供应商模板里**。

### CC Switch Version / 版本号

3.20.2

### Operating System / 操作系统

Windows

### Related App / 涉及应用

Codex

### Steps to Reproduce / 重现步骤

1. Codex 供应商配置中包含顶层 `model = "gpt-6-astra"` 和 `model_catalog_json = 'C:\Users\…\cc-switch-model-catalog.json'`（该 catalog 由 CC Switch 自己生成，Codex 靠它渲染自定义模型列表）
2. 开启本地路由（代理接管），`base_url` 指向 `http://127.0.0.1:15721/v1`
3. 触发一次重新接管（Codex Desktop 更新后 CC Switch 重新接管即可复现）
4. 打开 `~/.codex/config.toml`：`model` 与 `model_catalog_json` 两行消失；Codex 模型选择器中自定义模型列表变空

### Expected Behavior / 期望行为

接管重写保留与供应商无关的顶层键，至少保留 `model` / `model_catalog_json`（或按 #6017 的思路在接管重写时合并通用配置片段）。

### Actual Behavior / 实际行为

两行被丢弃。接管前备份（`proxy_live_backup`）里这两行**存在**，接管后的 live 配置里**不存在**——差异 100% 由接管写入引入：

```diff
# 接管前（proxy_live_backup, 2026-09-15 19:41 本地时间）
model_provider = "custom"
model = "gpt-6-astra"
model_catalog_json = 'C:\Users\ezra\.codex\cc-switch-model-catalog.json'
model_reasoning_effort = "high"
disable_response_storage = true

# 接管后（live config.toml）
model_provider = "custom"
model_reasoning_effort = "high"
disable_response_storage = true
```

### Additional Context / 补充信息

- 这两个键在供应商模板（数据库 `settings_config`）里**就存在**，接管输出仍然丢失，说明接管写入路径与正常切换路径的模板处理不一致。
- 该 catalog 指向的文件本身也是 CC Switch 维护的（本例中 31 个模型），接管丢弃指针后 Codex 端表现为「自定义模型列表每次 Codex 更新后消失」。

YUZHEthefool:
## Summary / 概述

A proxy re-takeover rebuilt Live from the provider row and, for a provider with **no inline `modelCatalog`**, fabricated an empty one (`{ "models": [] }`) before writing. That flipped `prepare_codex_live_config_text_with_optional_catalog`'s presence gate on for exactly the providers that have no mapping table, and catalog projection then rewrote or removed the top-level `model_catalog_json` pointer — so a pointer the provider's own `config.toml` text carries was silently stripped, and Codex's model picker went empty on every re-takeover.

This is the loss the restore path already guards against: the gate exists on purpose (it is why a snapshot backup keeps its raw pointer), and `codex_restore_from_backup_preserves_model_catalog_pointer` documents the failure mode. The takeover attach now follows the same rule as the restore path:

- the provider declares a `modelCatalog` → copy it, projection runs as before;
- the provider declares none → leave the config text alone, so a pointer it carries survives.

Emptying the mapping table still clears the live pointer. The write layer keys on the *presence* of the key, and the form used to omit it whenever the table was empty — which would have made "user cleared the table" indistinguishable from "provider never had a mapping". The form now persists an explicit `{ "models": [] }` when the provider had mappings before, so clearing behaves exactly as before while an absent key now means "never configured" (keep the provider's own pointer).

## Related Issue / 关联 Issue

Fixes #7430

## Verification / 验证

Reproduced first: with a provider whose `config.toml` text carries `model_catalog_json` and no inline `modelCatalog`, the takeover output dropped exactly that line and nothing else — matching the report's diff. The new test fails without the fix (`takeover must not strip the provider's own model_catalog_json pointer, got: …` — the output has no `model_catalog_json` line) and passes with it.

- `codex_takeover_keeps_provider_model_catalog_pointer_without_inline_catalog` — pointer **and** the top-level `model` survive, and the route is still rewritten to the local proxy.
- `codex_takeover_drops_pointer_for_explicitly_empty_catalog` — the flip side, so "what the table shows is what Codex lists" keeps holding.
- `resolveCodexCatalogForSave` unit tests — filled table persisted as-is, emptied table persisted as explicit empty, never-configured provider declares nothing.

Full suites are unchanged against a baseline run of the same tree (`vitest`: same 45 pre-existing failures; `cargo test --lib`: same 12 pre-existing environment-dependent failures), with only the new tests added. `cargo fmt --check`, `cargo clippy -D warnings`, `pnpm typecheck` and `prettier --check` all pass.

## A note on the `model` line in the report

The takeover output is rebuilt from the provider row (only `mcp_servers` is carried over from the existing Live config), so a top-level `model` the row declares does survive — the new test asserts exactly that. I could not reproduce a `model` drop from a row that carries it; if that line still disappears for the reporter, then the row's stored `config` text does not contain it (Codex Desktop's model picker writes `model` into Live only, and the DB row is the SSOT on the next write). Happy to follow up with a full pre/post dump of `settings_config.config` from the affected database.

## Checklist / 检查清单

- [x] Updated i18n files if user-facing text changed / 如修改了用户可见文本，已更新国际化文件 — no user-facing text changed / 未修改用户可见文本


farion1231:
@codex review

farion1231:
@codex review

cyjin-yl:
any updates?

PR #7445, opened 2026-09-16
fix(proxy): send model before large canonical request payloads

### LINKED ISSUE #7442: Large request bodies: canonical sorting places model after messages and delays streaming gateway routing
### Summary

The proxy canonicalizes all outgoing JSON keys alphabetically, placing `messages` before `model`. This causes very large requests to perform poorly with gateways that scan the request stream for the top-level `model` before selecting an upstream. I reproduced this with Codex → CC Switch 3.20.3 → OpenCode Go and a synthetic image payload of about 21 MB.

### Evidence and reproduction

1. Send a streaming Responses request through CC Switch's `openai_chat` adapter, using a synthetic large image payload and `deepseek-flash`.
2. The resulting Chat Completions body starts with keys such as `max_tokens`, `messages`, then `model`. The model name follows approximately 21 MB of message content.
3. Replay that converted body over HTTP/1.1. Keeping the model after messages produced HTTP 500/503 or a timeout in these tests.
4. Move **only the top-level `model` key** to the beginning, keeping JSON values, images, `reasoning_effort=max`, and `stream_options.include_usage=true` unchanged. That replay completed with HTTP 200 in 11.6 seconds.
5. Applying the same model-first ordering in a temporary HTTP/1.1 relay allowed the complete Codex/CC Switch path to finish a ~21 MB synthetic request in 20.6 seconds. Three subsequent existing-thread requests also completed successfully.

These timings are individual observations, not a controlled performance benchmark or a claim that all upstream 5xx errors have this cause. HTTP/2 upload problems were also observed independently; the field-order comparison above used HTTP/1.1 throughout to isolate the JSON-order change.

OpenCode's public gateway implementation scans until it finds the root `model`, buffering preceding chunks before it can return a forwardable stream:
https://github.com/anomalyco/opencode/blob/dev/packages/console/app/src/routes/zen/util/requestBody.ts

CC Switch's relevant path (also present on current main, 06082e1):
- `src-tauri/src/proxy/forwarder.rs`: `prepare_upstream_request_body`
- `src-tauri/src/proxy/json_canonical.rs`: `canonicalize_value`

Body overrides are followed by another call to `prepare_upstream_request_body`, so manually ordering provider overrides does not solve this.

### Proposed fix

Keep deterministic canonicalization and private-field filtering, but put a present top-level `model` first when preparing the outgoing request body. Preserve the canonical order of remaining keys and nested objects, all field values and array order, and the existing canonical hash helpers. Add tests for a large body, missing model, deterministic ordering, and the override preparation path.

No HTTP transport/default changes are proposed in this fix. This is an AI-assisted investigation/contribution, locally reproduced with synthetic inputs; no credentials or conversation contents are included.

pepsile:
## Summary / 概述

Large image-heavy requests currently serialize `messages` before `model` because the proxy sorts every JSON object alphabetically. Streaming gateways such as OpenCode Go must buffer those messages before discovering which model/upstream to use.

Put the top-level `model` first in the finalized upstream body. Keep private-field filtering, canonical ordering for the remaining keys and nested objects, JSON values, array order, and canonical hash helpers unchanged. The same preparation runs after local body overrides, so overriding a model does not regress the ordering.

The investigation in #7442 includes a synthetic ~21 MB request: a same-transport HTTP/1.1 replay succeeded after changing only the root field order. Those measurements validate the diagnosis/workaround, not a packaged build of this branch. This PR does not change HTTP transport defaults or add a service/configuration option.

## Related Issue / 关联 Issue

Fixes #7442

## Validation / 验证

- Regression tests cover early model visibility before a large message, preparation after body overrides, deterministic remaining/nested ordering, value/array preservation, and stable canonical hashes. The existing test covers bodies without a model and private-field filtering.
- Red/green check: the three new tests fail against the original implementation; all four body-preparation tests pass after the fix.
- `cargo +stable test --offline`: **3,018 passed, 0 failed, 9 ignored**.
- `cargo +stable fmt --check`: passed.
- `pnpm test:unit`: **1,120 passed**; TypeScript and Prettier checks passed.
- Toolchain: macOS arm64, Rust 1.97.1 (`+stable`, matching CI's stable-channel policy; the repository's local toolchain file pins 1.95).
- Strict `cargo +stable clippy --offline -- -D warnings` reports two lints in unchanged files: `clippy::for_kv_map` at `src/gemini_mcp.rs:52` and `clippy::useless_borrows_in_formatting` at `src/tray.rs:829`. Those files are identical to base `06082e1`; this focused PR does not edit them.
- No UI changes; screenshots and i18n updates are not applicable.

## Checklist / 检查清单

- [x] `pnpm typecheck` passes
- [x] `cargo +stable clippy --offline` passes with the two existing warnings documented above
- [x] i18n: not applicable; no user-facing text changed

AI-assisted contribution; the changed code and tests were reviewed locally. The fork branch `fix/model-first-upstream` is temporary and intended for deletion after merge; upstream `main` is retained.


farion1231:
@codex review

farion1231:
@codex review

pepsile:
Hi, thanks for requesting the reviews. Neither `@codex review` invocation seems to have received a bot response yet. All CI checks have passed. Could you check whether the review integration is working, or review this manually when convenient? Happy to address any feedback. Thanks!

PR #7454, opened 2026-09-16
fix(proxy): lift additional_tools input carriers on the Codex chat conversion

### LINKED ISSUE #7451: 本地代理无法处理 OpenAI Responses 请求的 role:"developer",导致 Codex 0.154 全部 400
**现象**
Codex 0.154 通过 CC Switch 本地代理(127.0.0.1:15721)请求时,任何输入(中文"你好"、英文"hello"等)全部 400:

> CC Switch local proxy failed while handling Codex endpoint /responses ... upstream_status: HTTP 400; cause: BadRequestError ... Invalid JSON data: Failed to deserialize the JSON body into the target type: messages[1]: missing field `content`

**根因(已复现)**
Codex 0.154 起把系统提示词改用 `role: "developer"` 发送。我把 Codex 真实发出的 16KB 请求体(7 条 input,其中 5 条 `role: developer`,含 1 条 `additional_tools`)原样经代理发给上游,得到上述 400;把同一体中所有 `role: "developer"` 改成 `role: "system"`、其余字段一字不动,再走代理 → 正常返回 `status: completed`。

说明代理在把 OpenAI `/responses` 请求转成上游 chat 格式时,**没有处理 `developer` 角色那条 message,其 `content` 字段丢失**,上游因此报 `missing field "content"`。

**修法建议**
代理在 responses→chat 转换时,把 `role: "developer"` 映射成 `system`(OpenAI 规范里 `developer` 即系统提示词的新写法)。这样 Codex 0.154+ 就能正常经代理使用。

**环境**
- Codex 0.154.0(codex-tui)
- Windows 10
- CC Switch 本地代理 127.0.0.1:15721

复现用的完整请求体可提供(16KB,JSON)。

SailingLoong:
## Summary / 概述

Hi! This picks up #7451 — Codex 0.154 requests failing with `400 messages[1]: missing field "content"` through the local proxy. Credit to @zhanghuihubei for the full 16KB capture and the A/B isolation, which made it possible to pin this down offline.

- **Root cause**: it's one layer below the role mapping. `Codex 0.154+` ships extra tools in an `additional_tools` **input carrier** (a Responses private extension — `{"type":"additional_tools","role":"developer","tools":[…]}`, note: a role but **no `content`**). In the Responses→Chat item walk, the generic "item with a role becomes a message" fallback converted that carrier into `{"role":"system","content":null}` — landing exactly at `messages[1]` — which strict chat gateways reject (gateways that normalize null fields away surface it as `missing field "content"`). The developer→system role mapping itself has been in place since #6252; replaying the issue's body through current `main` reproduces the same degenerate output with *either* role, so the breakage follows the carrier shape, not the role. On top of that, the tools carried inside it (the `functions`/`collaboration` namespaces) were dropped entirely on the Chat and Anthropic conversions — the xAI Responses passthrough already promotes the same carriers (`promote_additional_tools`).
- **Fix**, two small pieces in `transform_codex_chat.rs`:
  - an explicit skip arm for `additional_tools` carriers in `append_responses_item_as_chat_message` — a carrier declares tools, it is not a message;
  - carried tools are lifted through the existing `CodexToolContext` registry (`collect_input_declared_tools`, formerly the `tool_search_output`-only collector) — reusing its dedup, namespace flattening, and custom-tool conversion, and also feeding the Anthropic conversion, which builds its context via the same entry point.
- **Known boundary (out of scope)**: `custom`-typed children *inside* a namespace tool (e.g. the grammar-format `functions.exec`) still don't convert — `add_namespace_tool` lifts only `function` children today, for top-level declarations as well, and lifting them properly needs its own response-side name round-trip. The primary 400 from this issue is fixed without touching that.

## Related Issue / 关联 Issue

Fixes #7451

## Screenshots / 截图

N/A — backend-only change.

## Checklist / 检查清单

- Backend-only diff (one `.rs` file) — no TS/format/i18n surface touched, the changed-areas job skips the frontend checks.

## Validation

- `cargo fmt --check` clean; `cargo test --lib` 2885 passed, 1 failed — `update_current_claude_desktop_provider_syncs_profile_when_proxy_takeover_is_active` binds a fixed port that a locally running app holds on this machine; environment-only, unrelated to this change (CI runs it fine).
- `cargo clippy --all-targets -- -D warnings` flags one pre-existing `op_ref` in an untouched test in `transform_codex_chat.rs:4466` (`url == &data_url`) — present on clean `upstream/main` with local clippy 1.95; not introduced here.
- Red-green: reverting only the two production hunks (keeping the tests) makes `additional_tools_carrier_is_lifted_not_converted_to_a_message` fail at the content assertion while the carrier-free regression tests stay green.
- Four new tests: carrier emits no message and its function/namespace tools are lifted (`functions__exec_command`, `wait`); carrier tools dedup against top-level `tools`; plain developer messages convert unchanged with no fabricated `tools` array.

Happy to adjust — e.g. if you'd rather keep the collector name as-is, or lift the carrier in the forwarder instead of the transform.


farion1231:
@codex review

farion1231:
@codex review

farion1231 (approved):
Reviewed against main 37d047609: merge tree green (cargo test 2901 passed / 0 failed, fmt clean), and the full request capture from #7451 replayed through both the Chat and Anthropic conversions — the null system message is gone and the 7 carried function tools now reach the upstream. Approving to merge.

farion1231:
Merged as a35e5000b — thank you @SailingLoong. The carrier-vs-role diagnosis was exactly right: replaying @zhanghuihubei's 16KB capture against the merge tree, the `content: null` system message is gone and the 7 function tools from `functions`/`collaboration` now reach both Chat and Anthropic upstreams (on `main` before this, the Anthropic conversion silently dropped all of them too).

Three optional follow-ups, none blocking, if you'd like to pick any up:

1. An Anthropic-side regression test for the lifted carrier tools (`responses_request_to_anthropic`). The replay passes today but nothing locks it.
2. The boundary you called out — `custom` children inside a namespace (`functions.exec`). Worth its own issue so it stays tracked; Codex's developer prompt does steer the model toward `functions.exec`.
3. The collector still recurses into every nested value (inherited from the `tool_search_output` collector). #6159 scanned only direct `input[]` items, which is stricter against a structured tool output that happens to contain that shape. Cheap to tighten if you agree.

Credit also to @BigStrongSun, whose #6159 reached the same fix in August.

Issue #5510, opened 2026-07-18
Preserve custom Codex config fields in provider deeplink imports


aofee:
## Self Checks / 自检

- [x] I have read the FAQ and contribution guide.
- [x] I searched open and closed issues for existing Codex deeplink config preservation requests.

## Related App / 涉及应用

Codex

## Problem or Motivation / 问题或动机

A Codex provider deeplink can carry an inline `config` payload and the import preview can show settings from that payload. However, after confirmation, `build_codex_settings` reconstructs a fixed TOML template and drops non-provider Codex settings such as:

```toml
web_search = "live"

[features]
standalone_web_search = true

[tools.web_search]
context_size = "high"
```

This makes the preview and persisted provider configuration diverge. It also prevents provider distributors from using the documented `config` field for supported Codex preferences.

Example inputs can use `https://api.example.com/v1` and `sk-test-placeholder`; no real credentials are needed to reproduce this.

## Proposed Solution / 建议方案

Preserve the safe, non-provider-specific portion of the inline Codex config when building the stored provider settings, using the existing Codex common-config sanitization rules.

URL parameters should remain authoritative for provider-specific values:

- `endpoint` overrides `base_url`
- `model` overrides the configured model
- `apiKey` remains the stored credential
- provider routing/auth tables are regenerated by CC Switch
- credentials, model catalog paths, extra provider tables, and CC Switch-managed MCP fields are not copied from the inline config

Deeplinks without `config` should remain byte-for-byte compatible with the existing generated template.

## Contribution / 参与贡献

- [x] I am interested in contributing a focused implementation and regression tests.

PR #7455, opened 2026-09-16
fix(deeplink): preserve the supplied inline Codex config on import

### LINKED ISSUE #7438: [Bug] Codex deeplink import rewrites auth mode and drops required provider headers/bearer token (3.20.0)
## Environment

- CC Switch 3.20.0 (installed macOS app)
- Target: Codex
- Related: #3751 (closed), #5510 (open). This report specifically concerns required provider authentication fields, not only general Codex preferences.

## Reproduction

Use only a dummy credential; no live model request is necessary. Generate a link with this Python script, open the printed link, and confirm import **without enabling** the provider:

```python
import base64, json, urllib.parse
key = "fixture-not-a-real-key"
config = '''model_provider = "custom"
model = "gpt-5-codex"
model_reasoning_effort = "high"
disable_response_storage = true

[model_providers.custom]
name = "Codex auth import repro"
base_url = "https://api.example.com/v1"
wire_api = "responses"
requires_openai_auth = false
http_headers = { "x-openai-actor-authorization" = "1" }
supports_websockets = true
experimental_bearer_token = "fixture-not-a-real-key"
'''
payload = {"auth": {"OPENAI_API_KEY": key}, "config": config}
params = {
    "resource": "provider", "app": "codex",
    "name": "Codex auth import repro", "model": "gpt-5-codex",
    "homepage": "https://api.example.com",
    "endpoint": "https://api.example.com/v1", "apiKey": key,
    "enabled": "false", "configFormat": "json",
    "config": base64.b64encode(json.dumps(payload).encode()).decode(),
}
print("ccswitch://v1/import?" + urllib.parse.urlencode(params))
```

After importing, edit the provider and inspect its saved TOML.

## Expected

The explicitly supplied authentication mode and required provider fields survive import:

- `requires_openai_auth = false`
- `http_headers = { "x-openai-actor-authorization" = "1" }`
- `experimental_bearer_token` with the explicitly supplied dummy credential
- `supports_websockets = true`

If these fields cannot be safely supported, the importer should reject or clearly warn instead of silently substituting a different authentication configuration. No environment-variable secret resolution is requested.

## Actual / verified evidence

A real desktop import with a dummy key and loopback endpoint was confirmed locally. Both the provider editor and a read-only query of the test provider's saved `settings_config` showed:

- `requires_openai_auth` changed to **true**
- `http_headers` missing
- `experimental_bearer_token` missing
- `supports_websockets` missing
- dummy key retained only in `auth.OPENAI_API_KEY`
- test provider remained inactive (`is_current = false`)

The example.com script above is a sanitized equivalent of that reproduction, not a live service or credential.

## Suspected cause

In v3.20.0 `src-tauri/src/deeplink/provider.rs`:

1. `merge_codex_config` extracts only the key, endpoint and model from the inline config.
2. `build_codex_settings` reconstructs a fixed TOML template with `requires_openai_auth = true` and discards the supplied provider fields.

Source: https://github.com/farion1231/cc-switch/blob/v3.20.0/src-tauri/src/deeplink/provider.rs

## Impact

A valid distributor-generated config becomes a different authentication configuration after confirmation. This blocks one-click onboarding for endpoints that require these fields. We are holding the distributor-side change until the persisted import can satisfy this contract.

Please add a regression test that parses the deeplink, imports/builds the provider, and asserts the final saved configuration, rather than only checking the decoded input/preview.

SailingLoong:
## Summary / 概述

Hi! This picks up #7438 — deeplink import rewriting a distributor-supplied Codex config into a different authentication configuration. Credit to @linhay for the repro script, the exact source pointers, and the request for an end-to-end regression test — this PR follows all three.

- **Root cause**: `merge_codex_config` extracts only the API key / endpoint / model from an inline payload, and `build_codex_settings` then rebuilt its fixed TOML template with `requires_openai_auth = true` — silently discarding the explicitly supplied provider table (`http_headers`, `experimental_bearer_token`, `supports_websockets`, the declared auth mode, and anything else the distributor configured).
- **Fix**: when the deeplink carries a usable inline config (`configFormat: "json"` whose `config` string parses as TOML), use it **verbatim** as the provider's settings config:
  - the merged request key stays authoritative for `auth.OPENAI_API_KEY` (URL param wins over payload auth, same priority the merge step already applies), and any *extra* auth fields the distributor supplied are kept;
  - an unusable payload (bad base64 / bad JSON / bad TOML) falls back to the fixed template instead of failing the import — no behavior change for template-style links;
  - mirrors the Claude env helper's documented re-decode pattern (`decode cost is negligible on this one-shot import path`), so no signature threading through the build call graph.
- Deliberately unchanged: `configFormat: "toml"` payloads still take the template path (the JSON `{auth, config}` envelope is what the codex merge step consumes today).

## Related Issue / 关联 Issue

Fixes #7438

## Screenshots / 截图

N/A — backend-only change.

## Checklist / 检查清单

- Backend-only diff (one `.rs` file) — no TS/format/i18n surface touched, the changed-areas job skips the frontend checks.

## Validation

- `cargo fmt --check` clean; `cargo test --lib` 2884 passed, 1 failed — `update_current_claude_desktop_provider_syncs_profile_when_proxy_takeover_is_active` binds a fixed port that a locally running app holds on this machine; environment-only, unrelated to this change (CI runs it fine).
- `cargo clippy --all-targets -- -D warnings` flags one pre-existing `op_ref` in an untouched test in `transform_codex_chat.rs:4466` — present on clean `upstream/main` with local clippy 1.95; not introduced here.
- Red-green: reverting only the early-return in `build_codex_settings` makes `build_codex_settings_preserves_supplied_inline_config` fail (the template's `requires_openai_auth = true` comes back) while the fallback test stays green.
- The regression test asserts the final saved configuration, as requested in the issue: all four fields from the repro (`requires_openai_auth = false`, `http_headers`, `experimental_bearer_token`, `supports_websockets`) survive into the built provider settings, plus the merged auth key.

Happy to adjust — e.g. if you'd rather see a warning/confirmation in the import preview when a supplied config diverges from the template, that UI surface is better designed separately.


farion1231:
@codex review

farion1231:
@codex review

farion1231 (commented):
Thanks for picking this up so quickly — the direction is right. Adopting the distributor's inline TOML verbatim is the smallest fix that actually satisfies #7438 (the "regenerate the template and merge safe fields back" alternative would still drop the provider-table fields the reporter needs), the bad-payload fallback is the correct failure mode, and the full Rust suite is green on a merge tree against current `main`.

Verbatim adoption does open three gaps that the template path never had, though. I reproduced each one end-to-end (parse link → import → switch → read `~/.codex/config.toml`) on the merge tree; two of them need to land before this can merge.

### [P1] Inline config can reference the user's environment credentials (`env_key` / `env_http_headers`)

A link only has to carry *some* `apiKey` to pass validation, and the inline table goes into the stored config untouched — including `env_key`. On switch, the live `config.toml` keeps `env_key = "OPENAI_API_KEY"` verbatim and no `experimental_bearer_token` is injected, because [`codex_provider_table_declares_auth`](https://github.com/farion1231/cc-switch/blob/71b9e05027ca8d69cf0931f7269c68637139d521/src-tauri/src/codex_config.rs#L2796-L2807) treats a table with `env_key` as "already has its own auth" and skips injection. Codex then reads the real `$OPENAI_API_KEY` from the user's environment and sends it as the bearer to whatever `base_url` the link chose. `env_http_headers` is the same leak via headers.

Two things make this worse than it looks: the confirmation dialog masks the value of any key containing `KEY` ([`isSensitiveConfigKey`](https://github.com/farion1231/cc-switch/blob/71b9e05027ca8d69cf0931f7269c68637139d521/src/utils/deeplinkRisk.ts#L10-L29) via [`sanitizeTomlForPreview`](https://github.com/farion1231/cc-switch/blob/71b9e05027ca8d69cf0931f7269c68637139d521/src/utils/deepLinkConfigPreview.ts#L38-L41)), so the user sees `env_key = "OPEN************"` and cannot tell which variable is being referenced. And the Grok Build merge path in this same file already rejects exactly this class of link for exactly this reason ([provider.rs#L925](https://github.com/farion1231/cc-switch/blob/71b9e05027ca8d69cf0931f7269c68637139d521/src-tauri/src/deeplink/provider.rs#L925)).

Minimal fix — reject rather than strip (stripping silently changes the distributor's config), in [`merge_codex_config`](https://github.com/farion1231/cc-switch/blob/71b9e05027ca8d69cf0931f7269c68637139d521/src-tauri/src/deeplink/provider.rs#L791) right after the TOML parses, so the preview surfaces the error the same way the Grok Build path does:

```rust
if let Some(providers) = toml_value.get("model_providers").and_then(|v| v.as_table()) {
    if providers.values().any(|t| t.get("env_key").is_some() || t.get("env_http_headers").is_some()) {
        return Err(AppError::InvalidInput(
            "This link's Codex config references environment credentials (`env_key` / \
             `env_http_headers`), which cannot be imported from an untrusted link. Add the \
             provider manually and enter the key yourself."
                .to_string(),
        ));
    }
}
```

### [P1] URL `endpoint` / `model` are silently dropped once an inline config is present

The merge step and the confirmation dialog still treat URL params as authoritative ([`parse_and_merge_config`](https://github.com/farion1231/cc-switch/blob/71b9e05027ca8d69cf0931f7269c68637139d521/src-tauri/src/deeplink/provider.rs#L650): "URL params > inline config"), but the early return at [provider.rs#L443](https://github.com/farion1231/cc-switch/blob/71b9e05027ca8d69cf0931f7269c68637139d521/src-tauri/src/deeplink/provider.rs#L443) stores whatever `base_url` / `model` the TOML declares. Reproduced: URL `endpoint=https://url.example/v1&apiKey=sk-url` + inline `base_url = "https://config.example/v1"` → both the DB row and the live `config.toml` route to `config.example`, and the injected `experimental_bearer_token = "sk-url"` goes there too. The dialog showed that key next to `url.example`. The Claude path avoids this by lifting the `env` block and then overwriting the URL-driven fields.

The other shape of the same gap: an inline config with preferences only and no `[model_providers]` table (e.g. `web_search` + `[features]`, the #5510 shape) now produces a provider with **no routing at all**, where `main` would have built the template pointing at the URL endpoint.

Minimal fix in `supplied_codex_settings`: only adopt configs that declare their own routing, and stamp the merged endpoint/model back onto the active table (`update_codex_toml_field` is already the syntax-preserving helper for exactly these two fields):

```rust
let doc: toml::Value = toml::from_str(config_toml).ok()?;
// No active custom provider table → the template is the right shape (it carries the URL endpoint/model).
let active = doc.get("model_provider").and_then(|v| v.as_str())?;
doc.get("model_providers")?.get(active)?.get("base_url")?;

// URL params stay authoritative, same contract as the Claude env path.
let mut config_text = config_toml.to_string();
let endpoint = get_primary_endpoint(request).trim().trim_end_matches('/').to_string();
if !endpoint.is_empty() {
    config_text = crate::codex_config::update_codex_toml_field(&config_text, "base_url", &endpoint).ok()?;
}
if let Some(model) = request.model.as_deref() {
    config_text = crate::codex_config::update_codex_toml_field(&config_text, "model", model).ok()?;
}
```

The #7438 fixture still passes with this (it declares `model_provider = "custom"` and its `base_url` matches the URL endpoint), and a prefs-only or empty payload falls back to the template instead of persisting a card that cannot route.

### [P2] An empty `config` string counts as a usable config

`toml::from_str::<toml::Value>("")` is `Ok`, so `{"config": ""}` skips the template at [provider.rs#L418](https://github.com/farion1231/cc-switch/blob/71b9e05027ca8d69cf0931f7269c68637139d521/src-tauri/src/deeplink/provider.rs#L418) and stores `config = ""`. When that card is the first Codex provider (so it becomes current on add), the **import itself fails** with `provider.codex.config.missing` ([codex_config.rs#L3352](https://github.com/farion1231/cc-switch/blob/71b9e05027ca8d69cf0931f7269c68637139d521/src-tauri/src/codex_config.rs#L3352)); with another card already current, import succeeds and the first switch fails with the same error. On `main` an empty string is equivalent to no config and the template works. The `model_provider` requirement above closes this for free.

### [P2] (follow-up is fine) other execution-bearing top-level keys pass through

`notify` (a command Codex runs), `[projects."…"] trust_level`, `approval_policy`, `sandbox_mode` all pass `build_codex_settings` and the add-time validator unchanged. Before this PR none of them could enter through a link; the Claude path deliberately lifts only `env`. `[mcp_servers]` is fine — the post-switch MCP projection replaces that table wholesale. If you are already adding a denylist for the P1 above, stripping `notify` and `projects` there is ~10 more lines; otherwise a follow-up issue is fine.

### P3 / nit

- `auth`: keeping arbitrary extra fields from the payload has no legitimate use today and changes how Codex resolves the auth mode (`personal_access_token` / `bedrock_api_key` / `auth_mode` outrank `OPENAI_API_KEY`). Carrying only `OPENAI_API_KEY` is simpler and safer.
- Tests: the issue asked for "parse the deeplink → import → assert the saved configuration"; the new tests call `build_codex_settings` directly, which is why the empty-config failure at import time was not caught. [`tests/deeplink_import.rs`](https://github.com/farion1231/cc-switch/blob/71b9e05027ca8d69cf0931f7269c68637139d521/src-tauri/tests/deeplink_import.rs#L46) already has the isolated-home harness — adding the #7438 fixture there (and a `ProviderService::switch` + read of `get_codex_config_path()` for the env_key rejection) would lock the actual contract.
- nit: when the URL `apiKey` differs from the payload's `experimental_bearer_token`, the stored TOML keeps the payload's token while `auth.OPENAI_API_KEY` gets the URL key. The switch-time injection overwrites the token, so live is correct; only the stored copy is stale.

Fix the two P1s and the empty-config guard (the second snippet covers both P1 #2 and the P2) and this is good to merge — happy to re-review quickly.

PR #2481, opened 2026-04-30
feat: Liquid Glass UI + Security Hardening + Bug Fixes

xjwm5685-ui:
### UI Overhaul — Liquid Glass Design System
- Translucent frosted glass surfaces with specular highlights across all components
- Dynamic Island / macOS Dock app switcher with spring physics and magnification
- Stacked bar chart usage dashboard (blue-purple gradient) replacing area chart
- Model breakdown list with color dots and percentage bars (inspired by Claude Code Desktop)
- Spring animations throughout: cards, dialogs, buttons, view transitions, staggered entrances
- Responsive hover lift, press squish, and elastic micro-interactions
- Two app switcher modes: Island (morphing pill) and Dock (3D magnification)

### Bug Fixes
- **fix(#2471)**: Strip empty `pages` field from Responses API → Anthropic tool_use conversion
- **fix(#2465)**: `effortLevel: "max"` not persisting — use `deepFill` (only add missing fields) instead of `deepMerge` (overwrite) when re-applying common config snippet
- Remove dead `update_proxy` function
- Fix body_filter log to show all removed keys across entire request

### Security Hardening
- **AES-256-GCM encryption** for API keys in SQLite (`crypto.rs`) — key derived from machine fingerprint
- **WebDAV password encryption** in settings.json
- **Deep-link SSRF protection** — block localhost, private IPs, CGN, link-local addresses
- **Input validation** on deep-link name (≤256), API key (≤1024), repo field (whitelist chars)
- **SwitchLock TOCTOU fix** — double-check pattern eliminates race window
- **cargo-audit CI step** for dependency security scanning
- **CSP tightened** — removed generic `http:` from connect-src

### Refactoring
- `lib.rs` setup extracted into named functions (`init_database`, `import_initial_data`, `register_deeplink_handler`, `setup_tray`)
- Forwarder status updates extracted into `record_success` / `record_failure` helpers
- Named proxy config constants replacing magic numbers in schema.rs

### Build
- Release profile optimized for speed: `codegen-units=16, lto=false, opt-level=2` (~5-8x faster builds, slightly larger binary)

## Test plan
- [ ] Verify liquid glass renders correctly in both light and dark themes
- [ ] Test Island/Dock mode toggle and persistence
- [ ] Verify #2471 fix: Responses API Read tool with `pages: ""` no longer errors
- [ ] Verify #2465 fix: `effortLevel: "max"` persists after save/reopen
[...]

PR #7477, opened 2026-09-18
fix(ui): prevent flicker during view transitions

wc7086:
## Summary / 概述

Page navigation can flash as nested view, provider-list, and toolbar transitions wait for outgoing content, while cached session details appear in a second render. Mount the destination immediately using the same 0.3s fade-in and 10px upward slide as the settings and statistics pages. Remove the redundant list/toolbar transitions, keep the fixed header background solid, and derive the first visible session during render.

Render the session table-of-contents trigger in a body portal so the animated page's transform cannot change its fixed-position reference. This preserves the page animation while eliminating the trigger's jump when the transform finishes; Radix dialog interaction and focus restoration remain intact.

Full-screen panels had two further defects, both addressed here: the opaque surface faded in from `opacity: 0`, and opening or closing a panel never replayed the page entrance animation.

### Fixed elements inside the animated view container

`position: fixed` descendants of the view container are positioned against the container while its transform is applied, not against the viewport. framer-motion resets the transform to `none` once the animation settles (the element ends with `style="opacity: 1; transform: none;"`), so this is confined to the ~300ms entrance window rather than persisting — but an element visible throughout that window still jumps. The session table-of-contents trigger is rendered unconditionally, which is why it jumped on every entry into the sessions view.

Two further fixed elements live inside the container: `ProviderList`'s search popover (`top-[6.5rem]`) and `UnifiedSkillsPanel`'s import backdrop (`inset-0`). Both are rendered only on demand (`isSearchOpen`, `importDialogOpen`), so neither is mounted while the container animates and neither was observed to misbehave. They are portaled anyway so the rule holds unconditionally, rather than depending on when they happen to be mounted. Portaling is safe here: the popover's keyboard handling is a global listener and its focus effect goes through a ref, so nothing depends on DOM containment.

Standing rule for `App.tsx`: anything `position: fixed` rendered inside the view container must be portaled to `document.body`. All seven fixed elements in the tree now satisfy it — two sit outside the container, four are portaled, and `FullScreenPanel` portals itself.

### The opaque panel surface no longer fades

`FullScreenPanel` mounted its full-screen opaque layer at `opacity: 0` and faded it in. For the whole transition the user sees that layer cross-faded with the page behind it — two screens blended together, on both open and close. The surface is now opaque from the first frame, and the entry animation (same 0.3s fade-in and 10px slide as the pages) lives on the panel's content. The `slide-from-right` preset is unchanged: a slide never blends, because each pixel is either panel or page.

### Why "add provider" behaved differently

"Add provider" is not a page. It is a `FullScreenPanel` — a `createPortal` to `document.body`, rendered as a sibling of `renderContent()` and outside `<main>` (`App.tsx:1740` vs `:1742`). Opening or closing it does not change `currentView`, so the keyed container is never remounted and its entrance animation never replays; views get that for free.

It has been an overlay since the initial commit (as `AddProviderModal`), re-skinned twice (Radix `Dialog`, then `FullScreenPanel`) and never reconsidered as a view. The flat `type View` union and the flat `renderContent()` switch cannot express the panel nesting it needs anyway — `AddProviderDialog` opens two more full-screen panels of its own (`UniversalProviderFormModal`, `AuthSettingsPanel`).

So rather than change the navigation model, `FullScreenPanel` now exposes `subscribeToFullScreenPanelDismiss`, driven by its existing body-scroll-lock refcount. It fires when that count reaches zero — the last panel closing, not a nested panel returning to its parent. `App.tsx` subscribes and replays the view's entrance animation **on the existing element**: it removes the animation class, forces a reflow, and re-adds it. The container is deliberately not remounted, so the scroll position and any view-local state survive; an earlier revision of this branch bumped a counter in the container key instead, which reset both on every dismissal.

### The animation moved to CSS

The entrance animation is now a keyframe in `index.css` (`@keyframes view-enter`, `.view-enter`), carrying the same 0.3s fade-in and 10px slide.

Driving it with framer-motion was not viable. `useAnimationControls()` does not reach a `motion.div` that a `key` change has remounted — the element stays at its `initial` state, i.e. invisible. Measured in the test environment:

    AFTER MOUNT:    opacity: 1; transform: none;                (controls reach it)
    AFTER REMOUNT:  opacity: 0; transform: translateY(10px);    (controls do not)

A CSS animation runs on a fresh element by itself and restarts on demand through the class toggle, so one mechanism covers both the mount case and the replay case. `App.tsx` no longer imports framer-motion at all.

## Related Issue / 关联 Issue

N/A.

## Screenshots / 截图

Browser frame sampling used the real session components with six cached user messages and transition wrappers extracted from the application. At a 1000 × 700 viewport, the page fades from opacity 0 to 1 and slides from y=10 to y=0; the floating TOC trigger remains at y=580 throughout (0px jump, previously approximately 16px). Light and dark modes were checked. At 1400px width, the floating trigger remains hidden as intended.

## Validation / 验证

- Regression test first failed with the trigger inside the transformed page, then passed after the portal change.
- 24 tests passed across `SessionToc.test.tsx`, `SessionManagerPage.test.tsx`, and `App.test.tsx`, including dialog selection, Escape, focus restoration, and portal cleanup.
- `pnpm exec tsc --noEmit` and `pnpm format:check` passed.
- Browser checks confirmed the restored slide/fade, stable trigger position, TOC opening/selection, and focus restoration.

Added with the panel and replay work:

- `tests/components/FullScreenPanel.test.tsx` gains three assertions: the opaque surface must not mount at `opacity: 0`, the first-frame contract, and the dismiss notification firing when the last panel closes but not for a nested one.
- `tests/integration/App.test.tsx` gains an assertion that the view container is the **same DOM node** after a dismissal, so the replay cannot silently regress into a remount that resets scroll. It also asserts that the dismissal actually fired, so the test cannot pass vacuously.
- Every new behavioural assertion was confirmed to fail before its fix (RED), including the replay one — reverting to the keyed remount produces an `Object.is` inequality on the container element.
- 67 tests pass across `FullScreenPanel`, `ProviderList`, `UnifiedSkillsPanel`, `AddProviderDialog`, `EditProviderDialog` and the `App` integration suite.
- `pnpm typecheck` and `pnpm format:check` pass.
- The full suite has pre-existing flaky timeouts in `tests/integration/App.test.tsx` and `tests/components/PiProviderForm.test.tsx` under parallel load. They reproduce on the unmodified base commit and pass when those files run alone.

## Diff size / 差异规模

`App.tsx`, `FullScreenPanel.tsx`, `ProviderList.tsx` and `UnifiedSkillsPanel.tsx` show re-indentation alongside the behavioural change — removing an `AnimatePresence` wrapper and adding a `createPortal` wrapper each shift a nesting level. Reading hunk by hunk rather than by line count is recommended.

## Overlap / 重叠

#2481 (Liquid Glass UI) touches the same two spots in `App.tsx`: the `key={currentView}` container (`mode="wait"` → `popLayout`) and the header className (`backdrop-blur-md` → `glass-header`). Same regions, different approach — it keeps `AnimatePresence` and changes the mode, while this branch drops it in favour of the settings page's mount-only animation. Flagging it so a conflict there is expected.

## Checklist / 检查清单

- [x] TypeScript type checking passes.
- [x] `pnpm format:check` passes.
- [x] Rust/Clippy: not applicable; no Rust code changed.
- [x] i18n: trigger accessibility label reuses the existing `sessionManager.tocTitle` translation.
- [x] Regression tests cover the new `FullScreenPanel` behaviour and the in-place replay.


wc7086:
@farion1231 麻烦看一下，这个bug会导致很多页面闪烁，非常响体验

farion1231:
@codex review

farion1231:
@codex review

farion1231:
## GPT-6 Astra 自动代码审查



审查提交：`09b1151ac6638a2c874146d10ee88eae6d5ea90c`
Base：`06082e189d65e6d6dbadc35dacdac1ce6c79d89a`

由 Codex 定时任务使用 GPT-6 Astra 子 agent 审查，并经主 agent 复核。

### [P2] 为页面重放动画保留减少动态效果设置

位置：[src/index.css:314](https://github.com/farion1231/cc-switch/blob/09b1151ac6638a2c874146d10ee88eae6d5ea90c/src/index.css#L314-L316)

系统启用 prefers-reduced-motion: reduce 时，关闭添加/编辑供应商等最后一个全屏面板，现在仍会让底层页面从透明状态向上移动 10px，持续 300ms。FullScreenPanel 自身继续通过 useReducedMotion 禁用动画，但新增的 dismiss 订阅会无条件移除并重新添加 view-enter，而这里的 CSS 没有减少动态效果分支，绕过了原有无动画关闭行为。这使已明确关闭动态效果的用户在每次返回页面时仍被强制播放大面积位移动画。

证据：src/App.tsx:1016-1021 的 dismiss 回调无条件重放 view-enter；src/components/common/FullScreenPanel.tsx:64 在最后一个面板关闭时通知，84-90 行则仍尊重 prefersReducedMotion。src/index.css:300-308 定义 opacity 0→1 和 translateY(10px)→none，314-316 行无条件启用 0.3s 动画；现有 204-218 行的 reduced-motion 媒体查询不覆盖此类。基线 FullScreenPanel 在减少动态效果模式下 duration 为 0，且基线 App 没有关闭面板后的页面重放。临时 PostCSS 局部检查 ViewMotion.review.test.tsx 在原始 head 上失败：期望 none，实际 view-enter 0.3s ease-out；仅将规则置于 no-preference 媒体查询后通过。该检查验证 CSS 规则，不是浏览器逐帧采样。

建议：将 .view-enter 的 animation 声明放入 @media (prefers-reduced-motion: no-preference)，或在 reduce 分支明确设置 animation: none；为减少动态效果模式下关闭最后一个面板补充回归覆盖。

### 验证与范围

已覆盖 PR 全部 10 个变更文件及相关调用方和测试，确认 1 项新增 P2 无障碍回归：面板关闭后的页面动画忽略系统减少动态效果偏好。67 项相关既有测试通过；未确认其他可修复问题。

- 读取完整 prs/7477.json，确认 base/head；reviews 与 inline 均为空，3 条历史评论没有实质问题报告。工作树 HEAD 与指定 head 一致。
- 检索工作树及适用祖先目录，未发现 AGENTS.md。以 git diff base...head 审查全部 10 个变更文件，并用 -w 区分行为变更与缩进调整；检查相关面板调用方、portal、会话选择和测试。
- git diff --check 06082e189d65e6d6dbadc35dacdac1ce6c79d89a...09b1151ac6638a2c874146d10ee88eae6d5ea90c 通过。
- 在 validation-7477 临时副本运行 FullScreenPanel、SessionToc、SessionManagerPage、ProviderList、UnifiedSkillsPanel 和 App integration 六个测试文件，67/67 通过；出现 Tauri window metadata mock 和 Radix Dialog 描述警告，但没有测试失败。初次运行因临时副本遗漏 tsconfig.node.json 无法收集测试，补齐后重跑通过。
- 临时 ViewMotion.review.test.tsx 对 head 的 CSS 进行 PostCSS 规则检查，确认 reduce 模式仍适用 view-enter 0.3s ease-out；在验证副本添加 no-preference 媒体查询后同一检查通过。
- 审查工作树 git status --short 为空；测试副本、缓存和验证性修改均位于 validation-7477，未修改主仓库或审查工作树源文件。

未验证或受限范围：
- 未运行整个项目构建、完整测试套件、全量 TypeScript 检查或 Rust 检查。
- 未启动真实 Tauri 应用，未对 macOS/Windows/Linux WebView 的逐帧动画、拖拽区域或所有窗口尺寸做视觉验证；CSS 回归证据为静态规则检查及调用链审查。
- 局部测试复用本机已安装依赖，未按该 PR lockfile 重新安装；Vitest 实际版本为 2.1.9，framer-motion 为 12.23.25。测试运行在隔离源码副本并单独设置缓存目录。

本评论仅针对以上提交的代码审查结果，不构成合并批准。

wc7086:
> ### [P2] 为页面重放动画保留减少动态效果设置
> 位置：[src/index.css:314](https://github.com/farion1231/cc-switch/blob/09b1151ac6638a2c874146d10ee88eae6d5ea90c/src/index.css#L314-L316)


这个是拉了旧commit，第二个个commit就是修这个的

